PR #283, opened 2017-09-29
Shortcuts Documentation

tommoor:
This also fixes the blue outline on modals and the scrollbar behind

closes https://github.com/jorilallo/atlas/issues/224


tommoor:
wtf.

PR #414, opened 2017-11-14
Public api doc

jorilallo:
Moving to SSR React page.

screen shot 2017-11-19 at 6 45 16 pm
screen shot 2017-11-19 at 6 45 25 pm
screen shot 2017-11-19 at 6 45 35 pm


tommoor:
Looks like a great first pass! We should definitely get that big list of methods in a sticky sidebar as an improvement later - I guess we need some visual design in general for the static pages navigation and how they hang together 😄

jorilallo:
yeah, decided just to stick the MVP for now :)

PR #424, opened 2017-11-19
Separate Layout component

tommoor:
Starting to put some basic mobile styles in place, I wanted to make sure this refactor got done first to make things easier.

Basically split the sidebar logic out of the Layout component which was getting way too big for it's boots. There's definitely more tidying that can be done here but baby steps…


jorilallo:
@tommoor specs failing

tommoor:
totally unrelated db specs failing 💁

tommoor:
Seems like some migrations did return promises and others didn't so fixed that up. We also had a bit of a mixture of styles as the codebase has evolved - brought this all inline and consistent.

PR #434, opened 2017-11-20
Collection Home

### LINKED ISSUE #237: Collections: dashboard with recently changed

### LINKED ISSUE #251: Collection Homepage
- Sticky documents
- Recently created documents
- Recently changed documents?

### LINKED ISSUE #397: Deleting the first document in a collection results in error
Seems like it tries to navigate to the document that was just deleted. I'd imagine creating a simple collection home would resolve this at the same time but filing for completeness 😄

tommoor:
image

closes #397
closes #251 
closes #237


server/api/documents.js
```diff
@@ -15,14 +15,17 @@ const authDocumentForUser = (ctx, document) => {
 const router = new Router();
 
 router.post('documents.list', auth(), pagination(), async ctx => {
-  let { sort = 'updatedAt', direction } = ctx.body;
+  let { sort = 'updatedAt', direction, collectionId } = ctx.body;
```
jorilallo: so far I just left out `id`s from the fields in the API, so you pass the collection ID with `collection` field. Once this is changed, could you also append to public API docs?

server/presenters/document.js
```diff
@@ -36,6 +36,7 @@ async function present(ctx: Object, document: Document, options: ?Options) {
     team: document.teamId,
     collaborators: [],
     starred: !!(document.starred && document.starred.length),
+    collectionId: document.atlasId,
```
jorilallo: why can't we just pull this from `collection`?

server/api/documents.js
tommoor: Oh yea, I'm really unused to that so I forgot.

server/presenters/document.js
tommoor: We could, but it's not reliably available.

jorilallo: https://github.com/outline/outline/blob/master/server/presenters/document.js#L44

I don't really see why it wouldn't exist, especially on the backend?

tommoor: On the backend it should be there yep, I guess it's just usually good practice to read from the model rather than the relationship where you can. I'll change it up though.

jorilallo: Yeah, I would like to keep the API clean and concise as it will also be exposed publicly

Issue #459, opened 2017-12-01
Friendly error message when .env not set


anchepiece:
yarn start returns this:
```
Running Outline in development mode with React hot reloading. To run Outline in production mode, use `yarn start`
buffer.js:183
    throw new TypeError(kFromErrorMsg);
    ^
....
```

Tests fail like this:
```
TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.
      
      at Function.Buffer.from (buffer.js:183:11)
      at new Buffer (buffer.js:158:17)
      at new EncryptedField (node_modules/sequelize-encrypted/index.js:9:16)
      at EncryptedField (node_modules/sequelize-encrypted/index.js:5:16)
      at Object.<anonymous> (server/sequelize.js:8:82)
      at Object.<anonymous> (server/test/helper.js:9:17)

```

I suggest duplicating the .env.sample file as a default .env as you are already checking for a default secret key.  

Maybe a good idea is to build a getting started/installer process for development?

anchepiece:
A second place to check is in sequelize.js, and maybe for testing a default could be allowed.

tommoor:
@anchepiece thanks for filing! I think a friendly error message in this scenario would be a great step

anchepiece:
@tommoor is there a reason you don't want to leave a default .env in the repo? I'll create a PR for that scenario but if you disagree with the idea just close it :smile:

tommoor:
Yea, It's good practice to have the `.env` in `.gitignore` ([like we have](https://github.com/outline/outline/blob/master/.gitignore#L3)) to ensure that secret keys are never accidentally committed to the repo.

anchepiece:
Ok I'll skip that PR. Thanks for the quick replies!

PR #460, opened 2017-12-01
Address failing startup without .env #459

anchepiece:
Basically check for an undefined SECRET_KEY in index before starting the server. An undefined entry is the result of a missing or misconfigured .env file or environment variables.

The error message could be more specific but as it is you are at least pointed at the right direction to get your install config going.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,18 +12,19 @@ if (process.env.NODE_ENV === 'production') {
   );
 }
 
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks @anchepiece solution looks good - this is the first third-party contribution we've had, I'm not sure why CI isn't working so need to look into that before I can merge.

PR #471, opened 2017-12-04
More detailed error boundary for Document

jorilallo:
screen shot 2017-12-03 at 7 16 32 pm

Wrapped the Document scene into an error boundary so that the navigation etc will work as I'm expecting that most our bugs will be in this view. Also made the error message little friendlier.

### DIFF
--- a/app/components/ErrorBoundary/ErrorBoundary.js
+++ b/app/components/ErrorBoundary/ErrorBoundary.js
@@ -9,6 +9,14 @@ import PageTitle from 'components/PageTitle';
 class ErrorBoundary extends Component {
   @observable error: boolean = false;
 
+  componentWillReceiveProps(nextProps: Object) {
+    if (
+      (this.props.location || nextProps.location) &&
+      this.props.location.pathname !== nextProps.location.pathname
+    )
+      this.error = false;
+  }
+
   componentDidCatch(error: Error, info: Object) {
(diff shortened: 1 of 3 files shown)


app/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -9,6 +9,14 @@ import PageTitle from 'components/PageTitle';
 class ErrorBoundary extends Component {
   @observable error: boolean = false;
 
+  componentWillReceiveProps(nextProps: Object) {
```
jorilallo: This needed to be added as moving between documents doesn't remove the error. Location prop is optional for the component

app/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -2,13 +2,29 @@
...
+  location?: Location,
+  children?: ?React.Element<any>,
+};
+
 @observer
 class ErrorBoundary extends Component {
+  props: Props;
   @observable error: boolean = false;
 
+  componentWillReceiveProps(nextProps: Object) {
```
tommoor: Could we lose this logic and add a `key={this.props.location}` to ErrorBoundary instead?

jorilallo: derp, thanks

PR #476, opened 2017-12-05
support local development with docker

sasso:
start local development with one command! (`make up`)

commands: 
* `make up` start all required services and run migrations
* `make build` rebuild docker image
* `make destroy` stop and remove all containers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM node:latest
+
+ENV APP_PATH /opt/outline
+RUN mkdir -p $APP_PATH
+
+WORKDIR $APP_PATH
+COPY . $APP_PATH
+RUN yarn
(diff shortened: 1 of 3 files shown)


tommoor:
Hey @sasso, thanks for the PR! Going to checkout this out tonight properly but from a glance - looks great, this is something I've been meaning to setup for a while 😄

sasso:
sounds good! I haven't implemented fakes3 yet but could be usefull to remove that dependency for local dev. It would be also nice to modify `.env.sample` to match hosts with docker-compose (postgres, redis) would make it even faster to bootstrap dev 😄

tommoor:
@sasso I'm thinking I'll modify this to include as many env variables in the docker-compose file rather than in .env as possible for development, that way as you say getting going will be easier…

sasso:
👍

PR #487, opened 2017-12-18
Unfurling of Slack links

tommoor:
- If there is not an authentication in the db for slack then we prompt them to install the app.
- Any documents that the user tries to unfurl but doesn't have access to will be silently skipped.
- Otherwise a successful unfurl looks something like:

image

(the color comes from the collection color 😄 )


app/scenes/Settings/Slack.js
```diff
@@ -17,11 +17,11 @@ class Slack extends Component {
         <h1>Slack</h1>
         <HelpText>
           Connect Outline to your Slack team to instantly search for documents
-          using the <Code>/outline</Code> command.
+          using the <Code>/outline</Code> command and unfurl Outline links.
```
jorilallo: I think "preview Outline links" would be more descriptive than "unfurl"

server/api/hooks.js
```diff
@@ -1,10 +1,56 @@
...
+  let unfurls = {};
+  for (let link of event.links) {
+    const id = link.url.substr(link.url.lastIndexOf('/') + 1);
+    const doc = await Document.findById(id);
+    if (!doc || doc.teamId !== user.teamId) continue;
+
+    unfurls[link.url] = {
+      title: doc.title,
+      text: doc.getSummary(),
+      color: doc.collection.color || undefined,
```
jorilallo: Is there a difference between null and undefined here?

server/models/Event.js
```diff
@@ -9,30 +9,6 @@ const Event = sequelize.define('event', {
   },
   name: DataTypes.STRING,
   data: DataTypes.JSONB,
-
```
jorilallo: ?

tommoor: You don't need to define these, the `belongsTo` call takes care of it

server/api/hooks.js
tommoor: Not sure, probably not - was just being cautious

app/scenes/Settings/Slack.js
tommoor: agree

tommoor:
I've added `SLACK_APP_ID` added to production env

.env.sample
```diff
@@ -7,19 +9,26 @@ DATABASE_URL_TEST=postgres://user:pass@localhost:5432/outline-test
...
 PORT=3000
 REDIS_URL=redis://redis:6379
-SLACK_KEY=71315967491.XXXXXXXXXX
-SLACK_SECRET=d2dc414f9953226bad0a356c794XXXXX
 URL=http://localhost:3000
 DEPLOYMENT=hosted
 ENABLE_UPDATES=true
+
+# Third party credentials (optional)
+SLACK_KEY=71315967491.XXXXXXXXXX
```
jorilallo: KEY and SECRET are actually mandatory right now for auth, but all others optional, right?

PR #491, opened 2017-12-19
Added task queue for emails

jorilallo:


server/mailer.js
```diff
@@ -87,6 +88,30 @@ class Mailer {
...
 
 const mailer = new Mailer();
+const mailerQueue = new Queue('email', process.env.REDIS_URL);
+
+mailerQueue.process(async function(job) {
+  // $FlowIssue flow doesn't like dynamic values
+  await mailer[job.data.type](job.data.opts);
+});
+
+const sendEmail = (type: string, to: string, options?: Object = {}) => {
```
tommoor: I think this would be neater if you just combined the second and third parameters, it would also make the function signature more similar to the individual email methods rather than passing an object in one place and a string in the other?

server/mailer.js
```diff
@@ -87,6 +88,30 @@ class Mailer {
 }
 
 const mailer = new Mailer();
+const mailerQueue = new Queue('email', process.env.REDIS_URL);
+
+mailerQueue.process(async function(job) {
+  // $FlowIssue flow doesn't like dynamic values
+  await mailer[job.data.type](job.data.opts);
```
tommoor: You should type the strings that `job.data.type` can have (the individual email methods)... eg something like:

```
type Job = {
  data: {
    type: "welcome" | "password_reset"
  }
}

PR #492, opened 2017-12-19
Fixes exception visiting doc unauthenticated

### LINKED ISSUE #480: Visiting document pages unauthenticated leads to an exception
Was looking through exceptions and visiting a document URL while being logged out leads to an exception. Redirect to marketing page still works tho.

tommoor:
Also pulled the `Auth` component out into it's own file for legibility, that index.js is getting big.

Closes #480

### DIFF
--- a/app/components/Auth.js
+++ b/app/components/Auth.js
@@ -1,0 +1,58 @@
+// @flow
+import React from 'react';
+import { Provider } from 'mobx-react';
+import stores from 'stores';
+import SettingsStore from 'stores/SettingsStore';
+import DocumentsStore from 'stores/DocumentsStore';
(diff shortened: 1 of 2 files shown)


app/components/Auth.js
```diff
@@ -0,0 +1,58 @@
...
+      stores.auth.fetch();
+      authenticatedStores.collections.fetchAll();
+    }
+
+    return <Provider {...authenticatedStores}>{children}</Provider>;
+  }
+
+  stores.auth.logout();
+  window.location.href = BASE_URL;
+  return null;
```
tommoor: This is the actual fix. Even though we redirect the component has to return a valid React node.

Issue #493, opened 2017-12-19
CMD+S in a new document exits edit mode


tommoor:
It should however publish and remain in edit mode the same as it does normally.

Issue #40, opened 2017-05-02
Drafts


jorilallo:
This would be killer but needs scoping. Probably should be in user's global state over individual collections. I imagine we could make this work very similarly to Medium's Save/Publish

tommoor:
I think this would be a good quick win. The main thing it enables is us to have automatic saving which we can't really do right now because saving for the first time is also publishing.

jorilallo:
I like the idea. What are you thinking for the implementation, a new backend model with simple APIs only scoped to the user? Feels pretty simple in my mind, and useful

tommoor:
I was thinking we'd add a `publishedAt` column to documents, migrate all the current documents so that `publishedAt` = `createdAt` and add a default scope to all queries so they only return published documents etc some new endpoints for drafts.

jorilallo:
Sounds good

PR #495, opened 2017-12-20
CMD+S in a new document should not exit edit mode

zacharyfmarion:
Hey guys! Jori especially ;). This project looks awesome! I wanted to get involved if possible so I set it up today. This is a tiny PR to address issue #493. It's a little weird to have the text be Publish when in fact you are just saving, so I'm not sure if this fix is sufficient. Generally there is a publish/save distinction if you can have content that is visible to you in draft form but is not ready for the group to see yet. Also it doesn't look like there are component level tests so I didn't write one, hopefully that's okay!

### DIFF
--- a/app/scenes/Document/Document.js
+++ b/app/scenes/Document/Document.js
@@ -168,7 +168,7 @@ class DocumentScene extends Component {
     document = await document.save();
     this.isSaving = false;
 
-    if (redirect || this.props.newDocument) {
+    if (redirect) {
       this.props.history.push(document.url);
       this.props.ui.setActiveDocument(document);
     }
--- a/app/scenes/Document/Document.js
+++ b/app/scenes/Document/Document.js
@@ -171,6 +171,9 @@ class DocumentScene extends Component {
     if (redirect) {
       this.props.history.push(document.url);
       this.props.ui.setActiveDocument(document);
+    } else if (this.props.newDocument) {
+      this.props.history.push(documentEditUrl(document));
+      this.props.ui.setActiveDocument(document);
     }
   };
 


tommoor:
Hey @zacharyfmarion ! 

You're right about the publish text – the idea here is definitely to introduce the ability to save without publishing but we'll also need to have a sidebar item to view those 'drafts', there's an issue for that here: https://github.com/outline/outline/issues/40

As far as the PR goes - makes sense, it's so obvious I wonder why it was ever like this 🧐

zacharyfmarion:
Okay I see so expected behavior is that is should still redirect to the new page, the editor should just be focused one the redirect occurs?

jorilallo:
A wild @zacharyfmarion appears! Good to see you again buddy :)

zacharyfmarion:
Hey @jorilallo good to see you too! I'm gonna be pretty bored over winter break so feel free to assign some issues to me if you think I'd be a good fit.

tommoor:
>  Also just to clarify, clicking the publish button should exit edit mode, correct?

Yep! I'll check out this PR soon, got a feeling it's going to cause the cursor to jump but that's actually also the case just hitting cmd+s in general right now (another issue I REALLY need to look at!)

jorilallo:
@zacharyfmarion If you want to pick up something, there's a few "good first issue" bugs, upgrading webpack (we're still on 1.x) and also improving performance. Happy to add you to our Slack if you want, just hit me up with your email!

zacharyfmarion:
@jorilallo Okay I'll take a look at some more. I'd love to be added to slack! It's zacharyfmarion@gmail.com

tommoor:
Just an FYI, I've been looking at improving this further to try and remove the flash of loading screen when you cmd+s and also fix the cursor losing focus, almost there…

PR #496, opened 2017-12-27
Backend support for team admins

jorilallo:
I'll implement the UI in a separate PR as this grew in size:

- Updated `sequelize` so that ops match current documentations. No change in Jest but did it anyway
- Started using dormant `User#isAdmin` as each user can only belong to one team at the time. Seemed to be the easiest for now but `Team` has functions to abstract admin promotion/demotion if we want to swap the implementation later
- Moved `auth()` middlewares to utilize `router.use()` where it makes sense (ie. we expect the same authentication for the API section)
- `auth` middleware now takes `adminOnly` option
- Added unit tests for `auth()` middleware (finally)
- New endpoints to fetch, promote and demote admin users
- User email is now only exposed to the user themselves and admins. There's no real reason for this but I don't see a reason to expose it for all right now so siding with privacy.

### DIFF
--- a/server/api/hooks.js
+++ b/server/api/hooks.js
@@ -59,6 +59,8 @@ router.post('hooks.slack', async ctx => {
   });
 
   if (!user) throw httpErrors.BadRequest('Invalid user');
+  if (!user.isAdmin)
+    throw httpErrors.BadRequest('Only admins can add integrations');
 
   const documents = await Document.searchForUser(user, text, {
     limit: 5,
--- a/server/api/team.js
+++ b/server/api/team.js
@@ -10,9 +10,8 @@ import pagination from './middlewares/pagination';
 import { presentUser } from '../presenters';
 
 const router = new Router();
-router.use(auth({ adminOnly: true }));
 
-router.post('team.users', pagination(), async ctx => {
+router.post('team.users', auth(), pagination(), async ctx => {
   const user = ctx.state.user;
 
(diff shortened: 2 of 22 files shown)


server/api/team.js
```diff
@@ -0,0 +1,70 @@
...
+import Team from '../models/Team';
+
+import auth from './middlewares/authentication';
+import pagination from './middlewares/pagination';
+import { presentUser } from '../presenters';
+
+const router = new Router();
+router.use(auth({ adminOnly: true }));
+
+router.post('team.users', pagination(), async ctx => {
```
tommoor: This endpoint definitely seems like it should be open to more than admins. We'll need to list the folks in a team for everyone - perhaps exposing the details should just be admin only?

server/api/hooks.js
```diff
@@ -59,6 +59,8 @@ router.post('hooks.slack', async ctx => {
   });
 
   if (!user) throw httpErrors.BadRequest('Invalid user');
+  if (!user.isAdmin)
```
tommoor: We should keep in the back of our mind using cancan like 'policies' or some similar concept in the future to disconnect the users role from hardcoded all over the place, this is totally the right approach for the short term though.

jorilallo: I like that. Maybe we'll end up moving `.isAdmin` to be something more granular

Issue #361, opened 2017-10-23
Removing of collections doesn't destroy documents and they show up on the dashboard


jorilallo:

tommoor:
Do you remember, is this a server side issue or a client thing?

jorilallo:
I think this is server-side; it was the recent documents on home view. I should actually look into this now

PR #499, opened 2017-12-28
Webhook / Integration Event bus

tommoor:
@jorilallo lmk what you think of something like this…


server/models/Document.js
```diff
@@ -204,6 +205,15 @@ Document.searchForUser = async (
   return _.sortBy(documents, doc => ids.indexOf(doc.id));
 };
 
+// Hooks
+
+Document.addHook('afterCreate', 'event', (document, options) => {
```
tommoor: Example implementation, obviously we'd add these in all of the key places.

server/services/slack/index.js
```diff
@@ -1,7 +1,13 @@
 // @flow
+import type { Event } from '../../events';
+
 const Slack = {
   id: 'slack',
   name: 'Slack',
+  on: (event: Event) => {
+    console.log(`Slack service received ${event.name}`);
```
tommoor: Example implementation, this would likely end up a switch statement or an `if` filter depending on the complexity of the integration and how many types of events it's looking to handle.

services/slack/index.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import type { Event } from '../../server/events';
+
+export default {
```
jorilallo: Not sure what we need here but maybe these should be classes that would inherit from a base that would house unified logging/etc. Not sure what else we need to expose but logging feels like an obvious one that we can abstract from 3rd party ones later

jorilallo:
@tommoor big fan, lgtm!

jorilallo:
Was thinking, if we have such global message bus, it would be nice to plug other, non-service, hooks to it. For example #361 could rely on this bus. While thinking about this, I realized the shortcoming of the current implementation: each hook will get called during the single `.process`. If any of the hooks fail, it will lead to a reprocessing of the whole event. Maybe easier to take this over Slack, just random thoughts at this point, not sure how to architect this...

tommoor:
I agree that this can be used for more than services, although I don't think #361 is a great candidate - it should probably be taken care of as part of deleting the collection using sequelize hooks imo. If part of that fails then that can bubble to a collection delete failure.

The point you bring up is pretty valid though, each hook should run on it's own job - I'll make that adjustment.

tommoor:
Alright, this should be mergeable now. It doesn't DO anything other than log out when a fictional Slack integration would have taken some action on an event 😄

PR #504, opened 2017-12-30
Added support for block quotes

jorilallo:
screen shot 2017-12-30 at 1 46 01 pm
screen shot 2018-01-01 at 11 19 37 am
screen shot 2018-01-01 at 11 19 30 am


tommoor:
Do you think we need the toolbar item? If not could remove that and it would make this immediately mergeable. If you think it's needed i'd still say lets remove it and get the fixes in this PR merged and create a new issue to add that item and icon 😄

jorilallo:
@tommoor added icon

PR #505, opened 2018-01-01
Toolbar select tweaks

jorilallo:
- When selecting with a mouse, only show the inline toolbar on `mouseup` event
- Delay showing the toolbar a tiny bit, feels more natural


tommoor:
Thanks! It's weird, this behavior changed with the Slate 30 update - there must have been a change in how the selection data was propagated but I also noticed that this had become less 'natural' feeling. 

Will checkout and give a proper 👍 shortly.

jorilallo:
Ah that makes sense. Wasn't sure if there was a way to get the device selection status from the editor so added it as regular event listeners. Let me know if there's a more natural way

PR #506, opened 2018-01-01
Update README.md

jorilallo:


README.md
```diff
@@ -2,7 +2,7 @@
   <img src="https://user-images.githubusercontent.com/31465/34380645-bd67f474-eb0b-11e7-8d03-0151c1730654.png" height="29" />
 </p>
 <p align="center">
-  <i>An open, extensible, knowledge base for your team built using React and Node.js.<br/>Try out Outline using our hosted version at <a href="https://www.getoutline.com">www.getoutline.com</a>.</i>
+  <i>An open, extensible, wiki for your team built using React and Node.js.<br/>Try out Outline using our hosted version at <a href="https://www.getoutline.com">www.getoutline.com</a>.</i>
```
tommoor: Seems like we're batting backwards and forwards a bit on what to call it - could we include both? Would be better from an SEO pov to be using both terms…

jorilallo: I think we should just call it a wiki for now, much easier for people to understand and probably search as well

PR #507, opened 2018-01-04
Webpack 3 + dynamic Editor loading

### LINKED ISSUE #428: Upgrade webpack
We're still on v1, pretty far back.

jorilallo:
Wanted to see how much we could save in the main bundle, and this leads to 50% size reduction by splitting the editor into it's own dynamically loaded bundle. Also updated to Webpack 3 in the process

Fixes #428


app/components/Editor/components/Toolbar/Toolbar.js
```diff
@@ -6,7 +6,6 @@ import { Portal } from 'react-portal';
 import { Editor, findDOMNode } from 'slate-react';
 import { Node, Value } from 'slate';
 import styled from 'styled-components';
-import _ from 'lodash';
```
jorilallo: random linting errors exposed by updated dependencies or our CI is not working right

app/scenes/Document/Document.js
```diff
@@ -128,6 +129,11 @@ class DocumentScene extends Component {
     }
   };
 
+  loadEditor = async () => {
+    const EditorImport = await import('components/Editor');
```
jorilallo: I actually liked `react-loadable` but was easier as we want to share the loading state with document loading indicator

webpack.config.js
```diff
@@ -45,22 +45,33 @@ module.exports = {
...
-    root: path.resolve(__dirname, 'app'),
-    extensions: ['', '.js', '.json'],
-    alias: { shared: path.resolve(__dirname, 'shared') }
+    modules: [
+      path.resolve(__dirname, 'app'),
+      'node_modules'
+    ],
+    alias: {
+      shared: path.resolve(__dirname, 'shared'),
+      'boundless-arrow-key-navigation': 'boundless-arrow-key-navigation/build',
```
tommoor: Any insight into why these modules need this now?

tommoor:
Code looks 👌 . I'll give it a run this eve before we merge to prod.

webpack.config.js
jorilallo: I got just syntax errors, seems that it pulls the source by default and I excluded transpiling node modules. If this should not be the case, let me know

jorilallo: no idea what was going on before, I doubt we were transpiling everything... or maybe I'm missing something

tommoor: Okay, looked into this - it seems that the `boundless` modules have a `module` key in the package json that points to uncompiled code and webpack3 is reading that. The correct way to fix for now is to add:

`mainFields: ["browser",  "main"]` within the resolve config alongside `alias`, you can then remove these items. I've tested this locally.

jorilallo: thanks!

PR #514, opened 2018-01-18
Team members frontend

jorilallo:
screen shot 2018-01-17 at 8 47 12 pm

Left this pretty bare as I didn't want to waste time on invites or admin stuff as it doesn't matter for launch. Would need an icon tho


tommoor:
Looks good! There's an appropriate icon already in the Sketch file - just needs exporting to a component, https://share.goabstract.com/26fbd51a-8a42-4901-b898-169325eece5b

PR #515, opened 2018-01-18
Added more metatags + screenshot

jorilallo:


server/pages/components/Layout.js
```diff
@@ -6,6 +6,11 @@ import Analytics from '../../../shared/components/Analytics';
 import globalStyles from '../../../shared/styles/globals';
 import { color } from '../../../shared/styles/constants';
 
+export const title = 'Outline';
+export const description =
+  'Your team’s knowledge base - Team wiki, documentation, playbooks, onboarding & more…';
+export const screenshotUrl = 'https://www.getoutline.com/screenshot.png';
```
tommoor: There's a env var for URL which is set to `https://www.getoutline.com` in prod, we should use that instead of hardcoding here. Otherwise 👍

jorilallo: We should probably take these out all together for non-SaaS installation

jorilallo: Changed. I added robots exclusion for non-hosted variant

PR #519, opened 2018-01-22
Added link=“prefetch” tags

jorilallo:
More prefetch tags and nicer reading webpack confif

### DIFF
--- a/server/utils/prefetchTags.js
+++ b/server/utils/prefetchTags.js
@@ -1,5 +1,7 @@
 // @flow
 import React from 'react';
+import fs from 'fs';
+import path from 'path';
 import webpackConfig from '../../webpack.config';
 
 const PUBLIC_PATH = webpackConfig.output.publicPath;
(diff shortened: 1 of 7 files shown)


server/utils/prefetchTags.js
```diff
@@ -0,0 +1,33 @@
...
+
+const PUBLIC_PATH = webpackConfig.output.publicPath;
+
+const prefetchTags = [
+  <link rel="dns-prefetch" href={process.env.AWS_S3_UPLOAD_BUCKET_URL} />,
+];
+
+if (process.env.NODE_ENV === 'production') {
+  try {
+    const manifest = require('../../dist/manifest.json');
```
tommoor: How about using the node API's to load this file rather than require - we can just have it output empty if the file isn't found and it'll avoid the env check. That way we'll be able to test it too in the future if we want by mocking a manifest.

jorilallo: I googled this a bit; apparently Node doesn't cache the reads so it will always get it from disk. This seemed a faster way to implement the call.

tommoor: True node doesn't but it'd be easy enough to keep it in a local const?

jorilallo: true, I'll swap

tommoor:
This is pretty cool, any trick to testing in development?

jorilallo:
@tommoor I tested locally just by running the prod build and inspecting HTML so this should be fine

PR #521, opened 2018-01-23
About page

tommoor:
Mostly fixing grammar, some small design tweaks. I also wanted to remove the explicit mention of it being a side project which I think is a negative and instead emphasis open source - which is positive 😄 

**NOTE**: base is your branch to make diffing easier.

### DIFF
--- a/server/pages/About.js
+++ b/server/pages/About.js
@@ -40,7 +40,7 @@ export default function About() {
       </Helmet>
       <Header>
         <h1>About Us</h1>
-        <p>Who builds this anyway?</p>
+        <p>The team behind Outline</p>
       </Header>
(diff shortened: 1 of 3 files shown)


server/pages/About.js
```diff
@@ -36,30 +36,29 @@ export default function About() {
   return (
     <Grid>
       <Helmet>
-        <title>About</title>
+        <title>About Us</title>
       </Helmet>
       <Header>
         <h1>About Us</h1>
-        <p>Lets get to know each others</p>
+        <p>Who builds this anyway?</p>
```
jorilallo: Not particularly fan of this, sounds a bit negative

PR #524, opened 2018-01-24
LinkToolbar changes

jorilallo:
Bug: When selecting large section of text with a link it it, LinkToolbar would be opened and focus drawn to it automatically.

With these changes, toolbar is only shown when user has the focus within the link bounds. I also took out the autofocus as I found it feeling wrong as it captures user's focus. Also adjusted the toolbar size to give more space for links

### DIFF
--- a/app/components/Editor/components/Toolbar/components/LinkToolbar.js
+++ b/app/components/Editor/components/Toolbar/components/LinkToolbar.js
@@ -163,6 +163,7 @@ class LinkToolbar extends Component {
             placeholder="Search or paste a link…"
             onKeyDown={this.onKeyDown}
             onChange={this.onChange}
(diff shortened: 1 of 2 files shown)


app/components/Editor/components/Toolbar/components/LinkToolbar.js
```diff
@@ -163,7 +163,6 @@ class LinkToolbar extends Component {
             placeholder="Search or paste a link…"
             onKeyDown={this.onKeyDown}
             onChange={this.onChange}
-            autoFocus
```
tommoor: I think we should retain `autoFocus` for when it's a new link, `href === ""` ?

jorilallo:
Good point, I'll change that. Thanks

PR #526, opened 2018-01-27
Map keyboard shortcuts on windows

### LINKED ISSUE #525: (Some) Shortcuts not working for Windows
The shortcuts that use the command (Mac) key does not seem to have an alternate in Windows. I tried using Ctrl, Win, Alt and Shift key as a replacement for ⌘, but didn't work. Eg: ⌘+b for bold, ⌘+/ for shortcuts guide.
As an alternative, I suggest the use of the Ctrl key for Windows. Ctrl is used for all formatting shortcuts in Windows.

My setup: Windows 10 (64 bit), Chrome v63.

jorilallo:
`event.metaKey` maps to Windows key and not the Ctrl so keyboard shortcuts weren't working on Windows.

Fixes #525

### DIFF
--- a/app/scenes/KeyboardShortcuts/KeyboardShortcuts.js
+++ b/app/scenes/KeyboardShortcuts/KeyboardShortcuts.js
@@ -38,7 +38,7 @@ function KeyboardShortcuts() {
         <Label>Jump to dashboard</Label>
 
         <Keys>
-          <Key>⌘</Key> + <Key>/</Key>
+          <Key>?</Key>
         </Keys>
         <Label>Open this guide</Label>
       </List>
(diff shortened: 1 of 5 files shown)


jorilallo:
Looks like this is not the only place, refactoring so no need to review yet

app/scenes/KeyboardShortcuts/KeyboardShortcuts.js
```diff
@@ -38,7 +38,7 @@ function KeyboardShortcuts() {
         <Label>Jump to dashboard</Label>
 
         <Keys>
-          <Key>⌘</Key> + <Key>/</Key>
+          <Key>?</Key>
```
tommoor: 💯

PR #532, opened 2018-01-29
Fixed collection deletion

jorilallo:
Delete collection documents with it. Also fixed content on dashboard.

### DIFF
--- a/app/stores/DocumentsStore.js
+++ b/app/stores/DocumentsStore.js
@@ -29,6 +29,7 @@ type Options = {
 
 class DocumentsStore extends BaseStore {
   @observable recentlyViewedIds: Array<string> = [];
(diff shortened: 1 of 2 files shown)


app/stores/DocumentsStore.js
```diff
@@ -41,17 +42,22 @@ class DocumentsStore extends BaseStore {
...
   @computed
   get recentlyViewed(): Array<Document> {
-    return _.take(
-      _.filter(this.data.values(), ({ id }) =>
-        this.recentlyViewedIds.includes(id)
-      ),
-      5
-    );
+    const docs = [];
+    this.recentlyViewedIds.forEach(id => {
```
jorilallo: Flow is sad with plain `map`

tommoor: Tom is sad with flow

PR #535, opened 2018-01-29
Added social links

jorilallo:
Promote that Twitter profile

New email footer: 

screen shot 2018-01-28 at 7 48 06 pm


tommoor:
Hah, I didn't approve because the build failed or something - anyway, looks good 💯

Issue #537, opened 2018-01-29
TypeError "cannot read property push of undefined"


tommoor:
## Error in Outline

**TypeError** in **/collections/ffeaa0ce-fb0d-4d12-9140-3589ad961d73**
Uncaught TypeError: Cannot read property 'push' of undefined

[View on Bugsnag](https://app.bugsnag.com/atlas-1/outline/errors/5a6cf784a7e8c60018dfc1c8?event_id=5a6f207568b330001c92764f&i=gh&m=ci)

## Stacktrace

    webpack:///app/menus/CollectionMenu.js:26 - r.onNewDocument

[View full stacktrace](https://app.bugsnag.com/atlas-1/outline/errors/5a6cf784a7e8c60018dfc1c8?event_id=5a6f207568b330001c92764f&i=gh&m=ci)

*Created by Tom Moor via Bugsnag*

tommoor:
I found the cause of this, taking care of it. PR later on this evening.

jorilallo:
__drumroll__

PR #541, opened 2018-01-30
Fixes Collection ... menu not working

tommoor:
The menu in the top right of the collection dashboard was not really working at all. I think we kinda just forgot to finish it off 😬 .

As it needed the logic to import files I refactored that out into it's own lib as it's used in several different places - which also means we can stop passing refs to the dropzone around.

closes https://github.com/outline/outline/issues/537

### DIFF
--- a/app/components/Sidebar/components/Collections.js
+++ b/app/components/Sidebar/components/Collections.js
@@ -76,13 +76,8 @@ type CollectionLinkProps = {
 @observer
 class CollectionLink extends Component {
   props: CollectionLinkProps;
-  dropzoneRef;
 
   @observable menuOpen = false;
-
-  handleImport = () => {
-    this.dropzoneRef.open();
-  };
 
   renderDocuments() {
     const {
@@ -119,7 +114,6 @@ class CollectionLink extends Component {
         collectionId={collection.id}
         activeClassName="activeDropZone"
         menuOpen={this.menuOpen}
-        dropzoneRef={ref => (this.dropzoneRef = ref)}
       >
         <SidebarLink
           key={collection.id}
@@ -139,8 +133,6 @@ class CollectionLink extends Component {
                 collection={collection}
                 onOpen={() => (this.menuOpen = true)}
                 onClose={() => (this.menuOpen = false)}
-                onImport={this.handleImport}
-                open={this.menuOpen}
               />
             </CollectionAction>
           </CollectionName>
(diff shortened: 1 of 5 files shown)


app/components/Sidebar/components/Collections.js
```diff
@@ -132,19 +126,16 @@ class CollectionLink extends Component {
...
-                collection={collection}
-                onOpen={() => (this.menuOpen = true)}
-                onClose={() => (this.menuOpen = false)}
-                onImport={this.handleImport}
-                open={this.menuOpen}
-              />
-            </CollectionAction>
           </CollectionName>
         </SidebarLink>
+        <CollectionAction>
```
tommoor: Had to move this out as the file input doesn't work inside the `SidebarLink` anchor, not a big deal though.

app/menus/CollectionMenu.js
```diff
@@ -39,32 +68,47 @@ class CollectionMenu extends Component {
...
-              Import document
-            </DropdownMenuItem>
-            <DropdownMenuItem onClick={this.onEdit}>Edit…</DropdownMenuItem>
-          </Flex>
-        )}
-        {allowDelete && (
-          <DropdownMenuItem onClick={this.onDelete}>Delete…</DropdownMenuItem>
-        )}
-      </DropdownMenu>
+      <span>
```
tommoor: Ideally we'd use a React 16 array response here, but we need to update the flow definitions for that - I started doing this but it's actually quite a big task to update Flow which we should tackle separately.

app/scenes/Collection/Collection.js
```diff
@@ -133,7 +133,10 @@ class CollectionScene extends Component {
             />
             <Actions align="center" justify="flex-end">
               <Action>
-                <CollectionMenu collection={this.collection} />
+                <CollectionMenu
+                  history={this.props.history}
```
tommoor: This is where we weren't passing `history` before.

app/utils/importFile.js
```diff
@@ -0,0 +1,36 @@
...
+type Options = {
+  file: File,
+  documents: DocumentsStore,
+  collectionId: string,
+  documentId?: string,
+};
+
+const importFile = async (
+  { documents, file, documentId, collectionId }: Options,
+  callback: Document => *
```
jorilallo: Why not return a promise? We dont use callbacks anywhere

tommoor: Because I'm a lazy bum. Will update 😄

PR #554, opened 2018-02-03
Fixes code blocks

tommoor:
There were several different bugs layered on top of each other with code blocks that made this something of a nightmare to figure out:

- `slate-edit-code` plugin depended on only having `code-line` blocks as children however it didn't automatically make this conversion. I [submitted a PR](https://github.com/GitbookIO/slate-edit-code/pull/34) and published the fixes that are relevant to us to my scoped npm.
- `slate-prism` also has a bug that causes the slate edit code normalization not to work on first load if there is a language class set on the block. I've gone ahead and disabled that for now, the default highlighting is still pretty good.
- The combination of the first two made any code blocks after the first appear empty.

### DIFF
--- a/app/components/Editor/components/Code.js
+++ b/app/components/Editor/components/Code.js
@@ -5,37 +5,44 @@ import type { SlateNodeProps } from '../types';
 import CopyButton from './CopyButton';
 import { color } from 'shared/styles/constants';
 
+function getCopyText(node) {
+  return node.nodes.reduce((memo, line) => `${memo}${line.text}\n`, '');
+}
+
 export default function Code({
   children,
   node,
   readOnly,
   attributes,
 }: SlateNodeProps) {
-  const language = node.data.get('language') || 'javascript';
+  // TODO: There is a currently a big in slate-prism that prevents code elements
+  // with a language class name from formatting correctly on first load.
+  // const language = node.data.get('language') || 'javascript';
 
   return (
     <Container {...attributes} spellCheck={false}>
-      {readOnly && <CopyButton text={node.text} />}
-      <Pre className={`language-${language}`}>
-        <code className={`language-${language}`}>{children}</code>
-      </Pre>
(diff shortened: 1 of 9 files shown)


app/components/Editor/components/Code.js
```diff
@@ -5,37 +5,44 @@ import type { SlateNodeProps } from '../types';
 import CopyButton from './CopyButton';
 import { color } from 'shared/styles/constants';
 
+function getCopyText(node) {
```
tommoor: Because `code` now contains individual `code-line` blocks we need to recombine them into a string with line breaks otherwise the copied string will all be on one line

app/components/Editor/components/Code.js
```diff
@@ -5,37 +5,44 @@ import type { SlateNodeProps } from '../types';
...
 }: SlateNodeProps) {
-  const language = node.data.get('language') || 'javascript';
+  // TODO: There is a currently a big in slate-prism that prevents code elements
+  // with a language class name from formatting correctly on first load.
+  // const language = node.data.get('language') || 'javascript';
 
   return (
     <Container {...attributes} spellCheck={false}>
-      {readOnly && <CopyButton text={node.text} />}
-      <Pre className={`language-${language}`}>
```
tommoor: Removed the extra container here

app/components/Editor/components/Code.js
```diff
@@ -5,37 +5,44 @@ import type { SlateNodeProps } from '../types';
...
+}
+
 export default function Code({
   children,
   node,
   readOnly,
   attributes,
 }: SlateNodeProps) {
-  const language = node.data.get('language') || 'javascript';
+  // TODO: There is a currently a big in slate-prism that prevents code elements
```
jorilallo: nit: `big` -> `bug`

tommoor: big bug?

Issue #560, opened 2018-02-04
Increase collections pagination limit


tommoor:
Right now the sidebar will only load a maximum of 15 collections due to the pagination limits, we should ensure it loads all collections.

PR #561, opened 2018-02-04
Increase collections pagination limit

tommoor:
I think this is the pragmatic solution for now but feel free to push back if you don't agree @jorilallo 😄 
Also renamed `fetchAll`, to the more accurate `fetchPage`.

closes https://github.com/outline/outline/issues/560

### DIFF
--- a/app/components/Auth.js
+++ b/app/components/Auth.js
@@ -44,7 +44,7 @@ const Auth = ({ children }: Props) => {
       }
 
       stores.auth.fetch();
-      authenticatedStores.collections.fetchAll();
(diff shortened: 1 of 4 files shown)


jorilallo:
I'm ok for now but ideally we would implement `fetchAll` so it will make calls until there's no more but might have been that the pagination didn't give enough info right now to do that nicely

tommoor:
Yea, it's not IDEAL as it's setup right now - it would be nice to have a total count and / or perhaps the nextPage url should be unset when there are no more pages?

PR #567, opened 2018-02-05
Check for parent document existance

jorilallo:


tommoor:
I saw the related error to this - any idea at all how it could happen?

jorilallo:
@tommoor my guess would be someone testing us. If we see it again, will need to investigate more

Issue #559, opened 2018-02-04
Fix table of contents


tommoor:
We have one, but it's not showing for some reason - I think it's related to when the lazy editor loading was implemented.

PR #570, opened 2018-02-07
Fix TOC

jorilallo:
TOC wasn't rendering on initially because `this.editor` was `null` at render time, and only set later but it won't trigger a re-render.

closes https://github.com/outline/outline/issues/559


enzolutions:
I created documents with h1 and h2, but I don't get a TOC how I could render that?

PR #573, opened 2018-02-07
Renamed Document#atlasId finally

### LINKED ISSUE #530: Rename `atlasId` etc
We still use the legacy "atlas" name for some database fields which should use "collection" instead. For a rainy day cleanup.

jorilallo:
Finally renamed this legacy field name

fixes #530

### DIFF
--- a/server/api/documents.js
+++ b/server/api/documents.js
@@ -20,7 +20,7 @@ router.post('documents.list', auth(), pagination(), async ctx => {
 
   const user = ctx.state.user;
   let where = { teamId: user.teamId };
-  if (collection) where = { ...where, atlasId: collection };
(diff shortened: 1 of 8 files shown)


server/api/documents.js
```diff
@@ -20,7 +20,7 @@ router.post('documents.list', auth(), pagination(), async ctx => {
 
   const user = ctx.state.user;
   let where = { teamId: user.teamId };
-  if (collection) where = { ...where, atlasId: collection };
+  if (collection) where = { ...where, collectionId: collection };
```
tommoor: Now we just need to change the api parameters to match :P

jorilallo: I actually designed the API to omit `Id` postfix, that's why it wasn't there :)

PR #577, opened 2018-02-07
Revert "Renamed Document#atlasId finally"

jorilallo:
Reverts outline/outline#573


tommoor:
@jorilallo ummm, do you remember why we reverted this? 🙈

PR #578, opened 2018-02-08
Fixes: Possible to create blocks inside document title

tommoor:
Two fixes - ensure that Slate gets rid of anything other than text nodes in headings, and make sure that the markdown shortcuts functionality doesn't attempt to even put anything there.

Works beautifully.


jorilallo:
Sweet, thanks!

Issue #558, opened 2018-02-04
Analytics should differentiate between homepage / app


tommoor:
Right now in the analytics visits to the homepage look identical to page loads inside the app.

We should be able to tell the difference between the two - perhaps send anonymized slugs?

PR #582, opened 2018-02-10
Analytics

tommoor:
- Upgrade `React` for Fragment support 🎉 
- Include GA autotrack so we can actually get some analytics on what folks are doing in the hosted app

closes https://github.com/outline/outline/issues/558


app/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import React from 'react';
+import * as React from 'react';
```
tommoor: This is the latest recommended way to import React, I'll probably change them all when I upgrade flow.

app/index.js
```diff
@@ -49,7 +49,7 @@ const RedirectDocument = ({ match }: { match: Object }) => (
 globalStyles();
 
 render(
-  <div style={{ display: 'flex', flex: 1, height: '100%' }}>
+  <React.Fragment>
```
tommoor: 😍

app/index.js
```diff
@@ -120,6 +120,20 @@ render(
...
+  </React.Fragment>,
   document.getElementById('root')
 );
+
+window.addEventListener('load', async () => {
+  // installation does not use Google Analytics, or tracking is blocked on client
+  // no point loading the rest of the analytics bundles
+  if (!process.env.GOOGLE_ANALYTICS_ID || !window.ga) return;
+
+  await import('autotrack/lib/plugins/outbound-link-tracker');
```
tommoor: Additional analytics code is async loaded, only if the browser hasn't blocked GA. So that's pretty nifty.

PR #593, opened 2018-02-13
Remove delete restriction

### LINKED ISSUE #592: Cannot delete only document in a collection
Looks like we accidentally left this restriction in from before there was a collection homepage.

tommoor:
Seems like this was just kinda totally unneeded - we can re-add a working version of this as other permissions / policies get built later.

closes #592


jorilallo:
Agree, it actually confused me once as well.. no need to keep around IMO

PR #595, opened 2018-02-13
Fix logout redirect loop

tommoor:
In my testing this fixes the redirect loop. I'm not a huge fan of putting the redirect in the model but it's kind of the only way to handle it when the clearing of local storage needs to be awaited – it's also a vital part of completing the logout.


jorilallo:
I think this is perfectly fine for the required purpose, thanks!

PR #602, opened 2018-02-18
Authorization Improvements

tommoor:
Moving to a more consistent authorization methodology, fixes some minor holes in the process and add regression specs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "boundless-popover": "^1.0.4",
     "bugsnag": "^1.7.0",
     "bull": "^3.3.7",
+    "cancan": "3.1.0",
     "copy-to-clipboard": "^3.0.6",
     "css-loader": "^0.28.7",
     "debug": "2.2.0",
--- a/server/api/__snapshots__/team.test.js.snap
+++ b/server/api/__snapshots__/team.test.js.snap
@@ -17,9 +17,10 @@ Object {
 
 exports[`#team.addAdmin should require admin 1`] = `
 Object {
-  "error": "authorization_error",
-  "message": "Authorization Required",
+  "error": "admin_required",
+  "message": "An admin role is required to access this resource",
   "ok": false,
+  "status": 403,
 }
 `;
 
@@ -40,15 +41,16 @@ Object {
 
 exports[`#team.removeAdmin should require admin 1`] = `
 Object {
-  "error": "authorization_error",
-  "message": "Authorization Required",
+  "error": "admin_required",
+  "message": "An admin role is required to access this resource",
   "ok": false,
+  "status": 403,
 }
 `;
 
 exports[`#team.removeAdmin shouldn't demote admins if only one available  1`] = `
 Object {
-  "error": "at_least_one_admin_is_required",
+  "error": "validation_error",
   "message": "At least one admin is required",
   "ok": false,
   "status": 400,
--- a/server/api/__snapshots__/team.test.js.snap
+++ b/server/api/__snapshots__/team.test.js.snap
@@ -2,29 +2,37 @@
 
 exports[`#team.addAdmin should promote a new admin 1`] = `
 Object {
-  "avatarUrl": "http://example.com/avatar.png",
-  "email": "user1@example.com",
-  "id": "46fde1d4-0050-428f-9f0b-0bf77f4bdf61",
-  "isAdmin": true,
-  "name": "User 1",
+  "data": Object {
+    "avatarUrl": "http://example.com/avatar.png",
+    "email": "user1@example.com",
+    "id": "46fde1d4-0050-428f-9f0b-0bf77f4bdf61",
+    "isAdmin": true,
+    "name": "User 1",
+    "username": "user1",
+  },
   "ok": true,
   "status": 200,
-  "username": "user1",
 }
 `;
 
 exports[`#team.addAdmin should require admin 1`] = `
 Object {
-  "error": "only_available_for_admins",
-  "message": "Only available for admins",
+  "error": "authorization_error",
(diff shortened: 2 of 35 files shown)


server/api/collections.js
```diff
@@ -32,13 +34,27 @@ router.post('collections.create', auth(), async ctx => {
...
+});
+
 router.post('collections.update', auth(), async ctx => {
   const { id, name, color } = ctx.body;
   ctx.assertPresent(name, 'name is required');
   if (color)
     ctx.assertHexColor(color, 'Invalid hex value (please use format #FFFFFF)');
 
   const collection = await Collection.findById(id);
+  authorize(ctx.state.user, 'update', collection);
```
tommoor: 🍖 Ensure that the user is authorized to update the collection

server/api/index.js
```diff
@@ -38,6 +38,16 @@ api.use(async (ctx, next) => {
       }
     }
 
+    if (message.match('Authorization error')) {
```
tommoor: This catches the error that's thrown by `authorize()` if the user does not have correct permissions.

I'm trying to match the pattern we had before here, which is return a `404` if attempting to read a document you don't have access to and a `401` if any other action.

server/api/middlewares/authentication.js
```diff
@@ -96,10 +95,6 @@ export default function auth(options: AuthOptions = {}) {
         }
       }
 
-      if (options.adminOnly && !user.isAdmin) {
```
tommoor: Moved this logic to `policies/`

server/api/team.js
```diff
@@ -31,41 +32,41 @@ router.post('team.users', auth(), pagination(), async ctx => {
...
-  if (!promotedUser) throw httpErrors.NotFound();
+  const user = await User.findById(userId);
+  authorize(ctx.state.user, 'promote', user);
 
-  await team.addAdmin(promotedUser);
+  const team = await Team.findById(teamId);
+  await team.addAdmin(user);
 
-  ctx.body = presentUser(ctx, promotedUser, { includeDetails: true });
+  ctx.body = {
```
tommoor: Updated the response here, this was missing in the user data with other parts of the response. These endpoints seem to be the only place where a `data` property wasn't used.

server/api/views.js
```diff
@@ -1,24 +1,26 @@
...
+const { authorize } = policy;
 const router = new Router();
 
 router.post('views.list', auth(), async ctx => {
   const { id } = ctx.body;
   ctx.assertPresent(id, 'id is required');
 
+  const user = ctx.state.user;
+  const document = await Document.findById(id);
+  authorize(user, 'read', document);
```
tommoor: Previously there was no check here - you could record a "view" against any document in the db if you know the `id`. `views.list` would then happily return that document.

server/api/views.js
```diff
@@ -42,11 +44,13 @@ router.post('views.create', auth(), async ctx => {
...
   const user = ctx.state.user;
   const document = await Document.findById(id);
-
-  if (!document || document.teamId !== user.teamId)
-    throw httpErrors.BadRequest();
+  authorize(user, 'read', document);
 
   await View.increment({ documentId: document.id, userId: user.id });
+
+  ctx.body = {
```
tommoor: Just trying to be more consistent with our API responses.

server/models/Team.js
```diff
@@ -42,13 +42,13 @@ Team.prototype.createFirstCollection = async function(userId) {
...
 Team.prototype.addAdmin = async function(user: User) {
-  return await user.update({ isAdmin: true });
+  return user.update({ isAdmin: true });
 };
 
 Team.prototype.removeAdmin = async function(user: User) {
   const res = await User.findAndCountAll({
     where: {
-      teamId: user.teamId,
+      teamId: this.id,
```
tommoor: The controller was checking the authenticated users team, but then the model was using the promoted users team – these are the same right now, but makes sense to make these match.

package.json
```diff
@@ -87,6 +87,7 @@
     "boundless-popover": "^1.0.4",
     "bugsnag": "^1.7.0",
     "bull": "^3.3.7",
+    "cancan": "3.1.0",
```
tommoor: I checked out the source code of this, it's nice and simple with a minimal API so we can move it in house without too many changes if it turns out that it's restrictive at all.

jorilallo:
Comment on return codes:

- `401` when authentication token is invalid
- `404` when user tries to access invalid resource (or they are trying to access a resource they should not see)
- `403` when user tries to perform an action they aren't allowed to perform (e.g. non-admin tries to promote an admin)
- `400` invalid request, e.g. wrong params etc

tommoor:
@jorilallo changes as discussed 🙏 . If you're down with this approach once this is merged there are a lot of places where we're calling `httpErrors.BadRequest` directly – we should strip all of those out in favor of specific error classes like `ValidationError` as each one is getting a unique error id based on the passed string.

server/api/__snapshots__/team.test.js.snap
```diff
@@ -40,15 +41,16 @@ Object {
...
+  "message": "An admin role is required to access this resource",
   "ok": false,
+  "status": 403,
 }
 `;
 
 exports[`#team.removeAdmin shouldn't demote admins if only one available  1`] = `
 Object {
-  "error": "at_least_one_admin_is_required",
+  "error": "validation_error",
```
tommoor: This is an example if we replace `httpErrors.BadRequest` out in all the spots it's used - the error responses are more useable. I'll leave that for a separate PR though.

jorilallo:
Looks great, thanks for this

Issue #63, opened 2017-05-18
Push updates to Slack


jorilallo:
- Probably when you "publish"
- Where, global settings or in collection settings? But per collection notifications

tommoor:
We decided I'll take this on next, some details here:
https://www.getoutline.com/doc/slack-WnjhTM1h85

PR #603, opened 2018-02-19
Post to Slack

tommoor:
Ongoing work to enable connect and post collection changes to Slack.

### Todo:

- [x] Add new Integration model and endpoints
- [x] Integrate new authorization work from #602 
- [x] Make sure that `documents.update` isn't triggered every time the user hits CMD+S
- [ ] Stretch: Remove need to redirect user through SPA and redirect directly to API

closes https://github.com/outline/outline/issues/63


tommoor:
I've moved back to work on #518 as it became clear that this won't be useful unless we have a concept of 'publishing' and separating the different types of saves – otherwise the notifications will be constant / annoying.

tommoor:
@jorilallo this is ready for some 👀 – I'm going to create a separate ticket for refactoring the OAuth flow.

server/presenters/slackAttachment.js
```diff
@@ -0,0 +1,15 @@
+// @flow
```
tommoor: Pretty happy with this refactor now it's used in two places 🙌

server/services/slack/index.js
```diff
@@ -0,0 +1,43 @@
+// @flow
```
tommoor: Slack posting magic in here 🍖

server/models/User.test.js
```diff
@@ -1,10 +1,11 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
-import { flushdb, seed } from '../test/support';
+import { flushdb } from '../test/support';
+import { buildUser } from '../test/factories';
 
 beforeEach(flushdb);
 
 it('should set JWT secret and password digest', async () => {
-  const { user } = await seed();
+  const user = await buildUser({ password: 'test123!' });
```
tommoor: We should gradually start replacing the `seed` calls with individual factory calls to make the tests run faster.

server/models/Document.js
```diff
@@ -223,23 +228,51 @@ Document.searchForUser = async (
...
-Document.addHook('afterUpdate', model => {
-  if (!model.previous('publishedAt') && model.publishedAt) {
-    events.add({ name: 'documents.publish', model });
-  }
-  events.add({ name: 'documents.update', model });
-});
-
 // Instance methods
 
+Document.prototype.publish = async function() {
```
tommoor: Pretty happy with this approach, although it took a while to get it working how I wanted. Basically you can now just deal with document models and add / update properties at will – the hooks will take care of ensuring that the associated collection is updated correctly.

server/api/documents.js
```diff
@@ -332,7 +331,7 @@ router.post('documents.create', auth(), async ctx => {
 });
 
 router.post('documents.update', auth(), async ctx => {
-  const { id, title, text, publish, lastRevision } = ctx.body;
+  const { id, title, text, publish, done, lastRevision } = ctx.body;
```
tommoor: `done` is passed if the user has finished editing the document, which triggers the `documents.update` event. Open to improved naming on this one.

tommoor:
Sitting for 3 weeks 😢 – should I deploy without review?

server/models/Document.js
```diff
@@ -223,23 +228,51 @@ Document.searchForUser = async (
...
-  events.add({ name: 'documents.create', model })
-);
+Document.addHook('beforeSave', async model => {
+  if (!model.publishedAt) return;
+
+  const collection = await Collection.findById(model.atlasId);
+  if (collection.type !== 'atlas') return;
+
+  await collection.updateDocument(model);
+  model.collection = collection;
```
jorilallo: 👏

server/models/Document.js
jorilallo: this is really nice

server/presenters/integration.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import { Integration } from '../models';
+
+function present(ctx: Object, integration: Integration) {
+  return integration.toJSON();
```
jorilallo: In general I would prefer at least pulling the keys out so that we don't accidentally expose data in the API. I been bitten by these before :(

PR #606, opened 2018-02-25
Validation Improvements

tommoor:
Final removal of direct `BadRequest` calls so that all our errors return a valid error id.

Also closes https://github.com/outline/outline/issues/592 (again) – apparently we left in a server side restriction.

### DIFF
--- a/server/api/documents.js
+++ b/server/api/documents.js
@@ -1,11 +1,11 @@
 // @flow
 import Router from 'koa-router';
-import httpErrors from 'http-errors';
 
 import auth from './middlewares/authentication';
 import pagination from './middlewares/pagination';
 import { presentDocument, presentRevision } from '../presenters';
 import { Document, Collection, Star, View, Revision } from '../models';
+import { ValidationError, InvalidRequestError } from '../errors';
 import policy from '../policies';
 
 const { authorize } = policy;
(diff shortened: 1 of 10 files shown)


server/api/middlewares/authentication.js
```diff
@@ -1,20 +1,10 @@
 // @flow
-import httpErrors from 'http-errors';
 import JWT from 'jsonwebtoken';
 import { type Context } from 'koa';
-
 import { User, ApiKey } from '../../models';
+import { AuthenticationError } from '../../errors';
 
-type AuthOptions = {
-  require?: boolean,
```
tommoor: This was unused, if it's not required just don't add the middleware to that route I guess? 😄

server/api/documents.js
```diff
@@ -327,27 +322,11 @@ router.post('documents.delete', auth(), async ctx => {
...
   if (collection.type === 'atlas') {
-    // Don't allow deletion of root docs
-    if (collection.documentStructure.length === 1) {
-      throw httpErrors.BadRequest(
-        "Unable to delete collection's only document"
-      );
-    }
-
     // Delete document and all of its children
-    try {
```
tommoor: These caches aren't needed, the error will bubble correctly.

server/api/documents.js
```diff
@@ -327,27 +322,11 @@ router.post('documents.delete', auth(), async ctx => {
 
   const collection = document.collection;
   if (collection.type === 'atlas') {
-    // Don't allow deletion of root docs
```
tommoor: This was the old logic for preventing deletion of the last doc.

Issue #598, opened 2018-02-14
Pinned Documents


tommoor:
It would be nice to be able to "pin" documents in a collection, which would make them always appear at the top of the collection dashboard.

Could be used for a bunch of different usecases.

tommoor:
Ongoing work here: https://github.com/outline/outline/tree/pinned-documents

- [x] Add pin / unpin API endpoints
- [x] Add pinned boolean to document presenter
- [ ] Add ability to pin / unpin from document menu
- [ ] Add ability to unpin from DocumentPreview (icon needed)
- [ ] Ensure listing documents by collection returns pinned first
- [ ] Ensure displaying documents on collection home shows pinned first

PR #608, opened 2018-02-27
Pinned documents

tommoor:
Complete other than any visual treatment changes we want to make

closes https://github.com/outline/outline/issues/598


app/scenes/Collection/Collection.js
```diff
@@ -55,15 +55,17 @@ class CollectionScene extends Component {
...
 
     if (collection) {
       this.props.ui.setActiveCollection(collection);
       this.collection = collection;
       await this.props.documents.fetchRecentlyModified({
         limit: 10,
-        collection: collection.id,
+        collection: id,
+      });
+      await this.props.documents.fetchPinned({
```
jorilallo: can we `await Promise.all` these two so that the scene will load faster

app/scenes/Collection/Collection.js
```diff
@@ -145,14 +155,13 @@ class CollectionScene extends Component {
               />{' '}
               {this.collection.name}
             </Heading>
+
+            {hasPinnedDocuments && <DocumentList documents={pinnedDocuments} />}
+
             <Subheading>Recently edited</Subheading>
-            <DocumentList
-              documents={this.props.documents.recentlyEditedIn(
```
tommoor: @jorilallo maybe you remember - was there some reason we did it this way? Rather than what I've changed it to... it just seems kinda dumb in hindsight, but I might be missing something

jorilallo: I don't remember at all

tommoor:
Updates made :smile:

image

PR #617, opened 2018-03-07
Remove moment.js

tommoor:
Looks like there was a critical security vulnerability in our version of moment.

Instead of upgrading I took this opportunity to remove this bulky dependency in favor of [date-fns](https://date-fns.org) – this should reduce the bundle size nicely as only individual functions are imported. I prefer the functional approach of this lib. 

lmk your thoughts.


jorilallo:
this is great, always hated `moment` and its bloat

PR #619, opened 2018-03-10
Improve scrolling behavior

tommoor:
Prevent scrolling in the sidebar/ content area scrolling the other section when the boundary of a scrolling area is reached.

https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior


jorilallo:
TIL

PR #620, opened 2018-03-11
Improved paragraph behavior

tommoor:
@jorilallo give this a go, let me know how it feels. The only thing that isn't working yet is multiple paragraphs still collapse down to one when you save.

### DIFF
--- a/app/components/Editor/Editor.js
+++ b/app/components/Editor/Editor.js
@@ -291,8 +291,7 @@ const StyledEditor = styled(Editor)`
 
   p {
     position: relative;
-    margin-top: 1.2em;
-    margin-bottom: 1.2em;
+    margin: 0;
   }
 
   a:hover {
--- a/app/components/Editor/components/BlockInsert.js
+++ b/app/components/Editor/components/BlockInsert.js
@@ -71,7 +71,11 @@ export default class BlockInsert extends Component {
 
       // do not show block menu on title heading or editor
       const firstNode = this.props.editor.value.document.nodes.first();
-      if (result.node === firstNode || result.node.type === 'block-toolbar') {
+      if (
+        result.node === firstNode ||
+        result.node.type === 'block-toolbar' ||
+        !!result.node.text.trim()
+      ) {
         this.left = -1000;
       } else {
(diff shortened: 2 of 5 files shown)


jorilallo:
Tested, seems to work ok. Few things I noticed:

- Extra newline is inserted after a list block which leads to two newlines instead of one
- I tried to compare exported output to Paper's and noticed how headings differ a tiny bit; we both render them equally, e.g. if there's no forced newlines before or after, it still looks good. On the MD export Paper renders the heading the same way, even if you have a extra newline below it; you have one before and after and that's it. Our serializer seem to add the newline so user ends up getting two. Not sure what is ideal, but it makes sense in our case as we store the document in MD and they don't.

Overall, this feels really good and more predictable than existing implementation. And kudos for the toolbar fix!!

tommoor:
Re extra paragraphs, are you sure this wasn’t just existing documents? I
meant to mention that some existing docs will get extra paras by default
because the format previously always collapsed multiple to a single new
line and it doesn’t any more.

Give it a go with fresh content


On Mar 18, 2018 at 14:28, Jori Lallo <notifications@github.com> wrote:


Tested, seems to work ok. Few things I noticed:

   - Extra newline is inserted after a list block which leads to two
   newlines instead of one
   - I tried to compare exported output to Paper's and noticed how headings
   differ a tiny bit; we both render them equally, e.g. if there's no forced
   newlines before or after, it still looks good. On the MD export Paper
   renders the heading the same way, even if you have a extra newline below
   it; you have one before and after and that's it. Our serializer seem to add
   the newline so user ends up getting two. Not sure what is ideal, but it
   makes sense in our case as we store the document in MD and they don't.

Overall, this feels really good and more predictable than existing
implementation. And kudos for the toolbar fix!!

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/outline/outline/pull/620#issuecomment-374049736>, or mute
the thread

tommoor:
@jorilallo added a migration to update existing docs for new serializer changes. I'd like to get this deployed asap.

PR #634, opened 2018-03-30
[WIP] Billing

jorilallo:
- [ ] Billing icon
- [ ] Billing UI
- [x] UI element to notify users of suspended team requiring payment information
- [x] Block document creation and editing on API level is team is suspended
- [ ] Show "unable to save document" error message in the UI
- [ ] Way to communicate to admins that they are close to paid limit
- [ ] Error handling for billing page
- [ ] Show to the user who to contact for upgrading
- [ ] Million other things after I read the code more thoroughly
- [ ] Test that none of the billing related functionality leaks into self-hosted

Open questions: 

- How should below billing level be handled? should we allow the user to subscribe or not?
- How will subscription be handled when team reaches free limit?
- Should we bother on moving Stripe code outside the repo or not


tommoor:
Merged master and moving work here: https://github.com/outline/outline/pull/678

Issue #612, opened 2018-03-02
Autosave


tommoor:
Now that the drafts work is landed we're in a good position to add autosaving of documents.

PR #644, opened 2018-05-07
Autosave Documents

tommoor:
closes https://github.com/outline/outline/issues/612

### Todo

- [x] Reloading the page the first time does not show latest version (I think this is a wider issue not related to autosave, but became obvious while testing)
- [x] A network request for `collections.info` is happening after every autosave.
- [x] Updated API documentation

### DIFF
--- a/app/models/Collection.js
+++ b/app/models/Collection.js
@@ -156,6 +156,9 @@ class Collection extends BaseModel {
         }
       }
     );
+    this.on('documents.publish', (data: { collectionId: string }) => {
+      if (data.collectionId === this.id) this.fetch();
+    });
     this.on('documents.move', (data: { collectionId: string }) => {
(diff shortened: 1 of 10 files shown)


app/models/Collection.js
```diff
@@ -133,11 +149,11 @@ class Collection extends BaseModel {
       if (data.collectionId === this.id) this.fetch();
     });
     this.on(
-      'collections.update',
-      (data: { id: string, collection: Collection }) => {
-        // FIXME: calling this.updateData won't update the
```
tommoor: Addressed by using `withRouter` around `SidebarLink` to ensure re-render happens. 

https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md

Issue #626, opened 2018-03-24
Public share links


tommoor:
This would be pretty legit and a good growth tactic. Needs some design love and a little discussion before we start implementation…

- Where do you share from (presume a new icon in top right)?
- How do you revoke share links?
- Can there be multiple share links to the same document?
- Any restrictions on who can share / can it be turned off at team level?

jorilallo:
@tommoor How about having a organization level option to revoke shared documents. This way the team admins will also have access and visibility to shared links. Not sure if document needs multiple, could keep the v1 KISS

tommoor:
In which case a simple share `key` stored on the doc would suffice to bypass auth for that document. We can make it read-only too.

The more complicated thing about this feature is making the rest of the app work without being signed in – you wouldn't have access to see the rest of the collection structure etc so it could look similar to how editing mode does currently for public viewing.

PR #651, opened 2018-05-23
Public share links

tommoor:
Ongoing work to allow public sharing of documents. Still to do:

- [x] List share links in settings
- [x] Allow revocation in UI
- [x] Filter shares returned from list endpoint
- [x] Return existing share link for create if one exists for the user
- [x] Remove header and other controls from public doc
- [x] Filter protected data from `documents.info` when requested with a share token
- [x] shares.create `id` -> `documentId`
- [x] Close share dialog on copy link
- [x] Cleanup – convert other lists in admin to use `List/Item`
- [x] Error handling, not found shareId backend and frontend

closes https://github.com/outline/outline/issues/626


tommoor:
Ready to go. Will release this weekend 😄

PR #678, opened 2018-06-10
WIP: Billing

tommoor:
- [x] Billing icon
- [x] UI element to notify users of suspended team requiring payment information
- [x] Block document creation and editing on API level is team is suspended
- [x] Refactor subscription logic
- [ ] Lock down edit functionality when read only
- [ ] Buy seats in batches
- [ ] Allow auto-adding seats
- [ ] Paid seats
- [ ] Billing UI
- [ ] Error handling for billing page
- [ ] Show to the user who to contact for upgrading
- [ ] Test that none of the billing related functionality leaks into self-hosted

Open questions: 

- How should below billing level be handled? should we allow the user to subscribe or not?
- How will subscription be handled when team reaches free limit?


enzolutions:
The billing section would be hidden for self-hosting mode, or do you plan to have a running mode for that?

tommoor:
@enzolutions the functionality will be hidden when self hosted.

PR #680, opened 2018-06-14
WIP: Editor embeds

tommoor:
### TODO

- [x] Make deleting an embed easier
- [x] Add document level ability to disable embeds
- [x] Add team level ability to disable embeds
- [x] GitHub
- [x] Numeracy
- [x] Mode Analytics
- [x] Figma
- [x] Airtable
- [x] Vimeo
- [x] RealtimeBoard
- [x] Loom
- [x] Lucidcharts
- [x] Framer
- [x] InVision
- [x] Typeform
- [x] Marvel
- [x] Spotify
- [x] Codepen
- [x] Trello
- [ ] Speakerdeck
- [ ] Prezi
- [ ] Leanboard
- [ ] Twitter
- [ ] Google Maps
- [ ] Looker


app/embeds/youtube.js
```diff
@@ -0,0 +1,32 @@
+// @flow
```
tommoor: A lot still in the air, but I'm super happy with this - embeds can be added to Outline by defining a single React component like this.

Configuration is defined as statics.

jhult:
@tommoor, would this work for including content from another page in Outline?

tommoor:
@jhult it's not the primary usecase, although I think it's a good idea. I'd be a bit worried about confusion that could arise because of the visual similarities but nothing that couldn't be overcome

Issue #681, opened 2018-06-14
Allowed domains parameter for Google Auth in environment


satyadeepk:
We love the simplicity of Outline and looking to host it for internal wiki in our company. 

Having another optional parameter in env to allow only specific domains in google auth will be useful for security. Something like below:
```
GOOGLE_CLIENT_ID=XXXX
GOOGLE_CLIENT_SECRET=XXXXXX
GOOGLE_ALLOWED_DOMAINS=example.com,getoutline.com
```
I can take a shot at implementing it, if you feel like it's a good idea.

satyadeepk:
Thanks. 
Google has an optional `hd` parameter to be passed in the OAuth access request, https://developers.google.com/identity/protocols/OpenIDConnect#hd-param
But they don't recommend it as a way to validate the domain:
> Don't rely on this UI optimization to control who can access your app, as client-side requests can be modified. Be sure to validate that the returned ID token has an hd claim value that matches what you expect (e.g. mycolledge.edu). Unlike the request parameter, the ID token claim is contained within a security token from Google, so the value can be trusted.

Also the above `hd` parameter doesn't seem to support multiple domains.

So I am planning to validate it after the OAuth callback response.

PR #682, opened 2018-06-16
Allowed domains env variable for Google Auth

satyadeepk:
Implements #681.

* Added an optional `GOOGLE_ALLOWED_DOMAINS` parameter in the environment which takes a comma separated list of domains to allow in Google auth.
* If it is not set, all Google apps domains are allowed by default.

@tommoor Please let me know if any changes have to be made. Open for feedback.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -20,7 +20,7 @@ SLACK_SECRET=d2dc414f9953226bad0a356cXXXXYYYY
 
 GOOGLE_CLIENT_ID=
 GOOGLE_CLIENT_SECRET=
-# Comma seperated list of domains to be allowed (optional)
+# Comma separated list of domains to be allowed (optional)
 # If not set, all Google apps domains are allowed by default
 GOOGLE_ALLOWED_DOMAINS=
 
--- a/.env.sample
+++ b/.env.sample
@@ -20,6 +20,9 @@ SLACK_SECRET=d2dc414f9953226bad0a356cXXXXYYYY
 
 GOOGLE_CLIENT_ID=
 GOOGLE_CLIENT_SECRET=
+# Comma seperated list of domains to be allowed (optional)
+# If not set, all Google apps domains are allowed by default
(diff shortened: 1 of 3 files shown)


.env.sample
```diff
@@ -20,6 +20,9 @@ SLACK_SECRET=d2dc414f9953226bad0a356cXXXXYYYY
 
 GOOGLE_CLIENT_ID=
 GOOGLE_CLIENT_SECRET=
+# Comma seperated list of domains to be allowed (optional)
```
tommoor: Small spelling mistake:seperated -> separated

server/auth/google.js
```diff
@@ -43,6 +44,13 @@ router.get('google.callback', async ctx => {
     return;
   }
 
+  // allow all domains by default if the env is not set
+  const allowedDomains = allowedDomainsEnv && allowedDomainsEnv.split(',');
+  if (allowedDomains && allowedDomains.indexOf(profile.data.hd) < 0) {
```
tommoor: You can use `allowedDomains.includes(profile.data.hd)` as this is server side.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

server/pages/Home.js
```diff
@@ -38,6 +38,12 @@ function Home(props: Props) {
               try signing in with your company Google account.
             </Notice>
           )}
+          {props.notice === 'hd-not-allowed' && (
+            <Notice>
+              Sorry, your Google apps domain is not allowed. Please try again
+              with an allowed company domain.
```
tommoor: What do you think about printing the allowed domains list in the error response here?

satyadeepk: Yeah, that might be revealing too much info to an outsider. A real member can contact their admin in any case.

satyadeepk:
@tommoor addressed your comments.

PR #747, opened 2018-08-10
Dashboard Tabs

tommoor:
Adds three tabs to the dashboard, stepping into increased information density on the home page and a new view to be able to see documents created by yourself.

---

image


satyadeepk:
Cool stuff! Will each view be paginated like an infinite scroll?

tommoor:
I'll get some pagination in before merging, yea. Although probably not the most efficient way to browse your wiki 😆

PR #762, opened 2018-08-27
`showCollection` option in Dashboard documents list

satyadeepk:
It was added in #718, but seems to have gone after the new tabs are added in the dashboard.


tommoor:
Sorry, about that – appreciate the repeat PR!

PR #770, opened 2018-09-06
Implement a local authentication provider (for development only)

japsu:
This implements a local single-user authentication provider that makes it easier to try out and develop Outline without GitHub or Slack integration.

Further improvements:

- [ ] Add profile pictures for team and user

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -14,9 +14,12 @@ DEPLOYMENT=self
 ENABLE_UPDATES=true
 DEBUG=sql,cache,presenters,events
 
+# Enable single-user development authentication (leave empty in production)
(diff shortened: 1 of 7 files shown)


tommoor:
Hey @japsu – thanks for taking the time to put this together.

Right now I'm not comfortable landing an authentication method that is only used for local development, it's very easy to register a Slack or Google application for the purpose and I'm concerned that this will lead to edge cases that aren't caught when not using OAuth.

Definitely appreciate the time you've put in – In future OSS contributions I'd highly recommend creating an issue and getting input from project maintainers before starting writing code.

Thanks again

PR #785, opened 2018-10-16
Make isHistory regex slightly more specific

JonathanUsername:
I believe this fixes a bug that means a user cannot edit a document if there is the word "history" in the user-generated part of the slug (ie. the title).

To reproduce the bug - make a document called "Transaction History", then try to edit it. The sidebar will keep opening and the url state will keep getting changed to history rather than edit.


tommoor:
Thanks for the PR, great find!

PR #788, opened 2018-10-20
server: fix content-length-range s3 uplaod policy

defunctzombie:
The max size should be an integer rather than a string (environment variables
are strings). More strict s3 implementations reject the upload because they do
not consider the policy valid if a string is used rather than an integer.


defunctzombie:
This was discovered while deploying outline to Google Cloud and using their S3 compatibility layer.

tommoor:
Great, thanks for the PR!

Issue #834, opened 2018-12-31
Add more collection document sort options


tommoor:
Separate, but similar to [chronological collections](https://github.com/outline/outline/issues/821). 

As well as "recently edited" documents it would be great to have sort / filters, for the following which would make _managing_ the knowledgebase easier – for example removing old and outdated info.

- [x] Alphabetical sort
- [x] Least recently updated
- [x] Most recently published
- [ ] Popularity (most viewed)

tommoor:
Counting as complete for now

PR #843, opened 2019-01-08
Add new document sort options to collections

tommoor:
Adds 3 additional tabs as sort options when viewing a collection and infinite pagination.

image

ref https://github.com/outline/outline/issues/834

### DIFF
--- a/app/components/DocumentPreview/components/PublishingInfo.js
+++ b/app/components/DocumentPreview/components/PublishingInfo.js
@@ -19,11 +19,12 @@ const Modified = styled.span`
 
 type Props = {
   collection?: Collection,
+  showPublished?: boolean,
   document: Document,
   views?: number,
 };
 
-function PublishingInfo({ collection, document }: Props) {
+function PublishingInfo({ collection, showPublished, document }: Props) {
   const {
     modifiedSinceViewed,
     updatedAt,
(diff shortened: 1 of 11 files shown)


app/components/DocumentPreview/components/PublishingInfo.js
```diff
@@ -48,7 +49,7 @@ function PublishingInfo({ collection, document }: Props) {
             </span>
           ) : (
             <Modified highlight={modifiedSinceViewed}>
-              &nbsp;modified <Time dateTime={updatedAt} /> ago
+              &nbsp;updated <Time dateTime={updatedAt} /> ago
```
tommoor: Bringing this inline with language used elsewhere

app/components/Tab.js
```diff
@@ -10,14 +10,19 @@ const NavItem = styled(NavLink)`
   text-transform: uppercase;
   color: ${props => props.theme.slate};
   letter-spacing: 0.04em;
-  margin-right: 20px;
+  margin-right: 24px;
   padding-bottom: 8px;
+
+  &:hover {
```
tommoor: Make these tabs more clickable feeling

app/scenes/Collection.js
```diff
@@ -207,8 +204,62 @@ class CollectionScene extends React.Component<Props> {
                   </React.Fragment>
                 )}
 
-                <Subheading>Recently edited</Subheading>
-                <DocumentList documents={recentDocuments} limit={10} />
+                <Tabs>
```
tommoor: Using the same pattern here as the `/dashboard` routes that have worked pretty well

PR #850, opened 2019-01-10
run "yarn build" when making the production docker image

defunctzombie:
Building the app assets during container build allows for a faster app start
and keeps the built assets consistent regardless of individual environment settings.

---

## Motivation

When altering the Dockerfile to perform a `yarn build` during the image creation phase (a habit which I have found leads to faster app restart), I was bit by the injection of environment variables by the webpack DefinePlugin. The plugin injected `undefined` for `BASE_URL` and this resulted in a logout loop on our deployment. I traced this back to my alteration of the Dockerfile.

## Proposed Implementation

In this changeset I propose an alternative approach for injecting application global settings. This approach has the following properties:

* Single top level index.html file (the same for prod and dev) with support for bundles with hash identifiers in the name.
* No build step on start resulting in faster server startup in production.
* Asset build is identical regardless of application options.
* Application settings are consolidated under a single place (window.env) and one server route rather than global replacement and process.env mix/matching and split across webpack configurations. This makes it clearer [maybe] that changes to app code referencing global settings requires updating the route.
* Dockerfile.dev for use with docker-compose which foregoes the build step (this could be consolidated back to a single file single the build step is harmless for development builds).

## To Do

A few outstanding items need attention.

* window.env is a hack/anti-pattern - ideally a source file wraps the access interface to the global settings. Same is true for uses of process.env around the app. I've found it leads to harder to maintain source over a wrapper configuration module.
* bugsnag script tag was removed since the index.html file is no longer environment specific. An easy way to return this would be to have the bundle script inject the script tag as a matter of first action - tho there are tradeoffs with this approach and capturing errors to report. Could use some creative ideas here - could even do the injection similar to the handing of `{{env}}`
* slack app id meta tag was removed - for the same reason as bugsnag. It could use the same approach to return the tag.

---
Please review and let me know if this is heading in a direction that would be acceptable to mere and I will finish off the above ToDo items.


docker-compose.yml
```diff
@@ -23,13 +23,12 @@ services:
...
-      dockerfile: Dockerfile
-      args:
-        pull: 1
+      dockerfile: Dockerfile.dev
     ports:
       - "3000:3000"
     volumes:
-      - .:/opt/outline:cached
+      - .:/opt/outline
+      - /opt/outline/node_modules
```
tommoor: As I understand it moving `node_modules` to a volume means that you're proposing running `yarn build` outside of docker in development?

If that's the case this wouldn't be acceptable as it requires the host machine to have the same version of node installed

defunctzombie: > requires the host machine to have the same version of node installed

It does not - but I will describe in more detail what is happening and see if that answers the question better.

The listing of node modules as a volume without a source (notice no `:`), is a bit of a trick to ensure that the node_modules _of the image_ is used. This is necessary otherwise when you mount the source code into the container for dev it would _not_ have a node modules folder and you wouldn't be able to start the app.

I did not do the same thing for the _dist_ folder since the dev Dockerfile workflow doesn't produce a dist folder. However, should you choose to manually run `yarn build` in the docker environment, the dist folder would end up on the host machine and accessible to the user's dev tools should they want to look at these files in their editor. I can add `/opt/outline/dist` to the volumes list and it would _capture_ the folder within the container in a manner similar to how `/opt/outline/node_modules` is handled. Easy to change to whichever you prefer.

In either case - the user does not run `yarn build` or any other commands on the host machine - everything happens within the docker environment. `make up` was updated to drop the user into a docker shell from where they are able to run `yarn dev` or iterate with other commands.

defunctzombie:
@tommoor still interested in landing a version of this? I have some new changes to the approach (via using docker build args) that works more seamlessly with injecting env vars (for webpack) to the build process during the docker build.

tommoor:
Yea, there is still a lot of interest from folks to have a solution to this (more in the last month actually!) – it's also potentially gotten easier as we no longer use docker for the node app in development so perhaps less exceptions need to be made.

tommoor:
@defunctzombie took across the themes from this PR over here >> https://github.com/outline/outline/issues/1351

PR #861, opened 2019-01-13
Heroku deploy support

bmann:
Hey there! 

I saw a previous attempt to get this working, and you already have a Procfile in the root that Heroku needs.

I have tested this and done a couple of installs. Getting Google or Slack auth login support working is the hardest part, with Slack being easier. I still had a bit of difficulty -- probably walking people through what they need to do would be helpful.

Also, SLACK VERIFICATION is, I think deprecated, and might not be needed anymore -- both in your `env.sample` and in `app.json`.

You can do a test yourself by following this link, which will deploy from the branch that I am PR'ing here: https://heroku.com/deploy?template=https://github.com/bmann/outline/tree/heroku-deploy-support

### DIFF
--- a/app.json
+++ b/app.json
@@ -36,6 +36,7 @@
         "required": true
       },
       "DEPLOYMENT": {
+		"description": "Should be 'self' for self hosted installations, turns off things like pricing pages"
         "value": "self",
         "required": true
       },
@@ -46,7 +47,7 @@
       "SUBDOMAINS_ENABLED": {
         "value": "false",
         "required": true,
-        "description": "Running different subdomains. Not needed in single org mode"
+        "description": "Allows each team to have a different subdomain. Not recommend when self hosting"
       },
       "URL": {
         "description": "https://{your app name}.herokuapp.com",
@@ -97,7 +98,7 @@
         "required": false
       },
       "AWS_S3_UPLOAD_MAX_SIZE": {
-        "description": "",
+        "description": "Maximum file upload size in bytes",
         "value": "26214400",
         "required": false
       },
@@ -130,11 +131,11 @@
         "required": false
       },
       "BUGSNAG_KEY": {
-        "description": "(optional)",
+        "description": "An API key for bugsnag if you wish to collect error reporting (optional)",
         "required": false
       },
       "GITHUB_ACCESS_TOKEN": {
-        "description": "(optional)",
+        "description": "An API token for GitHub, optional for self hosted (optional)",
         "required": false
       }
 
--- a/app.json
+++ b/app.json
@@ -73,7 +73,8 @@
         "required": false
       },
       "SLACK_VERIFICATION_TOKEN": {
-        "description": "Your Slack verification token - PLxk6OlXXXXXVj3YYYY"
+        "description": "Your Slack verification token - PLxk6OlXXXXXVj3YYYY",
+        "required": false
       },
       "SLACK_APP_ID": {
         "description": "A0XXXXXXXXX",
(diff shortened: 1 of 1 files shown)


tommoor:
👋Hey, thanks for putting this together.

`SLACK_VERIFICATION_TOKEN` is required if you want to respond to hooks from Slack, but it's not required for signin.

app.json
```diff
@@ -0,0 +1,142 @@
...
+      "AWS_S3_UPLOAD_BUCKET_NAME": {
+        "description": "yourbucket.example.com",
+        "required": false
+      },
+      "AWS_S3_UPLOAD_BUCKET_URL": {
+        "description": "Live web link to your bucket. For CNAMEs, https://yourbucket.example.com",
+        "required": false
+      },
+      "AWS_S3_UPLOAD_MAX_SIZE": {
+        "description": "",
```
tommoor:
```suggestion
        "description": "Maximum file upload size in bytes",
```

app.json
```diff
@@ -0,0 +1,142 @@
...
+      "SMTP_REPLY_EMAIL": {
+        "description": "wikireply@example.com (optional)",
+        "required": false
+      },
+      "GOOGLE_ANALYTICS_ID": {
+        "description": "UA-xxxx (optional)",
+        "required": false
+      },
+      "BUGSNAG_KEY": {
+        "description": "(optional)",
```
tommoor:
```suggestion
        "description": "An API key for bugsnag if you wish to collect error reporting (optional)",
```

app.json
```diff
@@ -0,0 +1,142 @@
...
+        "required": false
+      },
+      "GITHUB_ACCESS_TOKEN": {
+        "description": "(optional)",
```
tommoor:
```suggestion
        "description": "An API token for GitHub, optional for self hosted (optional)",
```

app.json
```diff
@@ -0,0 +1,142 @@
...
+        "required": true
+      },
+      "ENABLE_UPDATES": {
+        "value": "true",
+        "required": true
+      },
+      "SUBDOMAINS_ENABLED": {
+        "value": "false",
+        "required": true,
+        "description": "Running different subdomains. Not needed in single org mode"
```
tommoor:
```suggestion
        "description": "Allows each team to have a different subdomain. Not recommend when self hosting"
```

app.json
```diff
@@ -0,0 +1,142 @@
...
+      "postdeploy": "yarn sequelize db:migrate"
+    },
+    "env": {
+      "SECRET_KEY": {
+        "description": "A secret key",
+        "generator": "secret",
+        "required": true
+      },
+      "DEPLOYMENT": {
+        "value": "self",
```
tommoor:
```suggestion
		"description": "Should be 'self' for self hosted installations, turns off things like pricing pages"
        "value": "self",
```

bmann:
I haven't used that Github suggested commits before. Super slick! Thanks @tommoor for making that easy! Should be ready to go.

tommoor:
Great, thanks!

satyadeepk:
Awesome! You can also probably add a "Deploy to Heroku" button in the Readme.

bmann:
Yeah. I think especially with your hosting model, sticking it in docs or somewhere else makes more sense than the README.

Very few people are going to be like me and yak shave their way through Slack and/or Google OAuth ;)

Thanks for accepting!

tommoor:
Full transparency – I definitely questioned whether to accept the PR or not 😉. Like you say the long-term business model here is to make hosting and maintaining Outline easy but I believe what you said is correct, there is still enough burden and technical know-how required to maintain your own copy.

PR #896, opened 2019-02-13
Enable read only non admin users

enzolutions:
This is a PoC to try to enable **Read  Only** for non-administrator users.

## Changes

* Disable Edit button to non Admin Users

## Pending
* Throw **access denied** to Non admin users that try to access the edit route for a document. I need some guidance about where this change is required. I am new in this App.


tommoor:
Hi @enzolutions – thanks for PRing your work. As it stands this isn't mergeable, however I'm open to this functionality being added as a team level option. To achieve that it would need:

- An additional setting under "Security" to disable editing for non-admins
- A new database column to track the setting
- Updated API endpoints to allow changing the setting
- Exposing the setting on the team json response so that the frontend can respond correctly
- Disable all editing UI in the frontend based on the setting (partly what you've done here)
- Ensure that all document editing endpoints also obey the setting and return an error

I'm going to close this PR as it stands but feel free to continue the discussion below or in the issues.

PR #905, opened 2019-03-01
Fixes: Unexpected token parsing sessions cookie

tommoor:
### Reproduction Steps

- Sign in to two teams (which will write en-uncoded cookie from the server)
- Sign out of one (which will write an encoded cookie from the client)
- Cookie parsing issues will prevent auth from this point on


app/stores/AuthStore.js
```diff
@@ -138,15 +138,15 @@ export default class AuthStore {
...
 
     // remove authentication token itself
-    Cookie.remove('accessToken', { path: '/' });
+    removeCookie('accessToken', { path: '/' });
 
     // remove session record on apex cookie
     const team = this.team;
     if (team) {
-      const sessions = Cookie.getJSON('sessions') || {};
+      const sessions = JSON.parse(getCookie('sessions') || '{}');
```
tommoor: Both `js-cookie` and `tiny-cookie` automatically run values through `encodeURIComponent` to ensure RFC compliance.

https://www.npmjs.com/package/tiny-cookie#setkey-value-options

server/middlewares/authentication.js
```diff
@@ -115,15 +115,19 @@ export default function auth(options?: { required?: boolean } = {}) {
...
-        const existing = JSON.parse(ctx.cookies.get('sessions') || '{}');
-        const sessions = JSON.stringify({
-          ...existing,
-          [team.id]: {
-            name: encodeURIComponent(team.name),
-            logoUrl: team.logoUrl,
-            url: encodeURIComponent(team.url),
-          },
-        });
+        const existing = JSON.parse(
```
tommoor: Ensure compatibility with the frontend cookie library and compliance with RFC.

PR #941, opened 2019-04-26
chore: Remove marketing material from OSS project

himanshuagarwal:
Allow having a plain team home containing minimalize information and team logo.
Most preferably only for self-hosted single team wiki.
There is no unit test added for the same till now.


tommoor:
Thanks for the PR! I'm going to roll this into https://github.com/outline/outline/issues/945 and make a simple homepage the only option / default for self hosted installations

tommoor:
Thanks for the ping 👍

app.json
```diff
@@ -142,6 +142,14 @@
     "GITHUB_ACCESS_TOKEN": {
       "description": "An API token for GitHub, optional for self hosted (optional)",
       "required": false
+    },
+    "TEAM_NAME": {
```
tommoor: I think this might be confusing because the team name is determined from the login authentication, even on the hosted version. Better to remove as it's only used for ALT text anyway

Issue #759, opened 2018-08-16
File upload in document editor fails with some S3 regions


egetin:
File upload to S3 fails in document editor, when the S3 bucket is in an AWS region that only supports signature version 4. The list of the regions and their signature version support is [here](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region). Avatar upload when logging in first time works fine since it uses `aws-sdk`.

The error message received from AWS is as follows:
```The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256```
In this case I was using `eu-central-1` region.

tommoor:
Thanks for the report @egetin! I'm going to create a new label for issues that only affect self-hosted instances.

defunctzombie:
What would be an appropriate fix for this? Would it be to switch to aws-sdk for the file upload as well? I believe a similar issue affects users who deploy onto GCP with the s3 compatibility layer.

captn3m0:
Hit by this as well. Let me see if I can file a PR for this.

captn3m0:
Based on the AWS S3 Signature Version Doc: https://docs.aws.amazon.com/general/latest/gr/rande.html

| Region Name                    | Region         | Status |
|--------------------------------|----------------|--------|
| Asia Pacific (Seoul)           | ap-northeast-2 | Breaks |
| Asia Pacific (Osaka-Local) *** | ap-northeast-3 | Breaks |
| Asia Pacific (Mumbai)          | ap-south-1     | Breaks |
| Canada (Central)               | ca-central-1   | Breaks |
| China (Beijing)                | cn-north-1     | Breaks |
| China (Ningxia)                | cn-northwest-1 | Breaks |
| EU (Frankfurt)                 | eu-central-1   | Breaks |
| EU (Stockholm)                 | eu-north-1     | Breaks |
| EU (London)                    | eu-west-2      | Breaks |
| EU (Paris)                     | eu-west-3      | Breaks |
| US East (Ohio)                 | us-east-2      | Breaks |
| Asia Pacific (Tokyo)           | ap-northeast-1 | Works  |
| Asia Pacific (Singapore)       | ap-southeast-1 | Works  |
| Asia Pacific (Sydney)          | ap-southeast-2 | Works  |
| EU (Ireland)                   | eu-west-1      | Works  |
| South America (São Paulo)      | sa-east-1      | Works  |
| US East (N. Virginia)          | us-east-1      | Works  |
[...]

PR #949, opened 2019-05-09
Fix: Now supports AWS signature version 4.

himanshuagarwal272:
Fix: https://github.com/outline/outline/issues/759

### DIFF
--- a/server/api/users.js
+++ b/server/api/users.js
@@ -2,7 +2,12 @@
 import uuid from 'uuid';
 import Router from 'koa-router';
 import format from 'date-fns/format';
-import { makePolicy, getSignature, publicS3Endpoint, makeCredential } from '../utils/s3';
+import {
+  makePolicy,
+  getSignature,
+  publicS3Endpoint,
+  makeCredential,
+} from '../utils/s3';
 import { ValidationError } from '../errors';
 import { Event, User, Team } from '../models';
 import auth from '../middlewares/authentication';
--- a/server/api/users.js
+++ b/server/api/users.js
@@ -1,7 +1,8 @@
 // @flow
 import uuid from 'uuid';
 import Router from 'koa-router';
-import { makePolicy, signPolicy, publicS3Endpoint } from '../utils/s3';
+import format from 'date-fns/format';
+import { makePolicy, getSignature, publicS3Endpoint } from '../utils/s3';
 import { ValidationError } from '../errors';
 import { Event, User, Team } from '../models';
 import auth from '../middlewares/authentication';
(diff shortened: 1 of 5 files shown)


server/utils/s3.js
```diff
@@ -20,20 +20,24 @@ export const makePolicy = () => {
...
 };
 
-export const signPolicy = (policy: any) => {
-  invariant(AWS_SECRET_ACCESS_KEY, 'AWS_SECRET_ACCESS_KEY not set');
-  const signature = crypto
-    .createHmac('sha1', AWS_SECRET_ACCESS_KEY)
-    .update(policy)
-    .digest('base64');
+export const getSignature = (policy: any) => {
+  const kSecret = "AWS4" + AWS_SECRET_ACCESS_KEY;
```
tommoor: The inbuild node `crypto` module has support for these hashing methods, so we should use that rather than rely on a third-party npm module that also isn't maintained

server/api/users.js
```diff
@@ -63,7 +64,9 @@ router.post('users.s3Upload', auth(), async ctx => {
 
   const s3Key = uuid.v4();
   const key = `uploads/${ctx.state.user.id}/${s3Key}/${filename}`;
-  const policy = makePolicy();
+  const credential = process.env.AWS_ACCESS_KEY_ID + '/' + format(new Date(), 'YYYYMMDD') + '/' + process.env.AWS_REGION + '/s3/aws4_request';
```
tommoor: Thoughts on moving this into the s3 file too, `makeCredential` or similar?

tommoor:
@himanshuagarwal272 let me know if you're planning on completing this PR or not, I understand if it's good enough for your fork and you don't plan to spend more time on it. Would just be good to get clarity on that 😄

himanshuagarwal272:
> @himanshuagarwal272 let me know if you're planning on completing this PR or not, I understand if it's good enough for your fork and you don't plan to spend more time on it. Would just be good to get clarity on that 😄

I will be working on it.. just not getting enough time from my university examination, It might take a week or two picking this up again.

himanshuagarwal272:
@tommoor Made the suggested changes, Let's work to close this and let me know if any more suggestion

server/utils/s3.js
```diff
@@ -8,9 +8,29 @@ import fetch from 'isomorphic-fetch';
 import bugsnag from 'bugsnag';
 
 const AWS_SECRET_ACCESS_KEY = process.env.AWS_SECRET_ACCESS_KEY;
+const AWS_ACCESS_KEY_ID = process.env.AWS_ACCESS_KEY_ID;
+const AWS_REGION = process.env.AWS_REGION;
```
tommoor: We'll need to add this new one to `.env.sample`, the addition of this environment variable will make upgrades a breaking change for existing installations. Seems like it's required with v4 too?

himanshuagarwal272: Yes we will need AWS_REGION

himanshuagarwal272: Yes we will need AWS_REGION

tommoor:
Looks good, I'll test this out as soon as I'm able

PR #953, opened 2019-05-21
Fix: negative numbers in pagination (limit/offset)

drezo:
Hello 🖖

This method returns an error 😞 (500 HTTP Code): 

POST /api/documents.list?limit=**-20**&offset=**-10**

**Example:**

outline_screenshot

### DIFF
--- a/server/api/middlewares/pagination.js
+++ b/server/api/middlewares/pagination.js
@@ -17,9 +17,9 @@ export default function pagination(options?: Object) {
     let query = ctx.request.query;
     let body = ctx.request.body;
     // $FlowFixMe
-    let limit = parseInt(query.limit || body.limit, 10);
+    let limit = Math.abs(parseInt(query.limit || body.limit, 10));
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks for finding this. I think the better approach here would be to throw an error similar to L26 explaining that limit and offset must be positive numbers, rather than silently reversing them – thoughts?

drezo:
@tommoor 

Yes, as an option. 

But in this case, you need to write an error if the passed parameter is not a number.

_Example:_

POST /api/documents.list?limit=**text20text**&offset=**text20text**

_Response:_

outline_pagination

tommoor:
> But in this case, you need to write an error if the passed parameter is not a number.

I think that would be great if the two were brought in-line. An `isNaN` check would do the trick

tommoor:
Thanks for finding this bug, the issue has been fixed with tests here: 0a8a685c

PR #959, opened 2019-05-31
Update installation guide

mogita:
Hi, just updated a production installation guide. Also updated some of the `.env.sample` fields to be more descriptive.

Please leave comments, thanks!


.env.sample
```diff
@@ -4,11 +4,11 @@
 #
 # Please use `openssl rand -hex 32` to create SECRET_KEY
 
-DATABASE_URL=postgres://user:pass@postgres:5432/outline
-DATABASE_URL_TEST=postgres://user:pass@postgres:5432/outline-test
-SECRET_KEY=F0E5AD933D7F6FD8F4DBB3E038C501C052DC0593C686D21ACB30AE205D2F634B
+DATABASE_URL=postgres://user:pass@localhost/outline
```
tommoor: We should leave the port in the example here

README.md
```diff
@@ -33,6 +35,28 @@ In development you can quickly get an environment running using Docker by follow
    else should work well for development.
 1. Run `make up`. This will download dependencies, build and launch a development version of Outline.
 
+For production deployment by building from scratch, you can follow these steps:
+
+1. Clone this repo and install dependencies with `npm install` or `yarn`
```
tommoor: I think the installation guide should prefer `yarn`, as we use yarn for development and a yarn lockfile is present there's a small possibility that installing with NPM will result in a different set of dependencies being installed

.env.sample
mogita: Fixed

README.md
mogita: Fixed

mogita:
Looks like a type check has failed. What's your idea @tommoor ? Do you need me to fix it for you?

tommoor:
Master was failing for a couple of commits there, sorry about that – I'll fix that up and get this merged, thanks!

PR #963, opened 2019-06-05
Added ability to append to documents. Added test cases for the same

### LINKED ISSUE #950: Add `append` ability to document edit API endpoint
This would be great for gradually building a document over time, think – a Zapier hook or cron job.

I'd imagine it could be easily added as an additional, optional boolean flag.

arpitsingh94:
the documents.update API has been updated to accept a parameter `append`. When present, the text sent in the request will be appended to the existing document's text instead of replacing it.

Test cases to verify this have been added.
Resolves #950

### DIFF
--- a/server/api/documents.js
+++ b/server/api/documents.js
@@ -612,7 +612,7 @@ router.post('documents.update', auth(), async ctx => {
   } = ctx.body;
   ctx.assertPresent(id, 'id is required');
   ctx.assertPresent(title || text, 'title or text is required');
-  if (append) ctx.assertPresent(text, 'text is required while appending');
+  if (append) ctx.assertPresent(text, 'Text is required while appending');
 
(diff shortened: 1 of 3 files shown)


server/api/documents.js
```diff
@@ -615,7 +625,9 @@ router.post('documents.update', auth(), async ctx => {
 
   // Update document
   if (title) document.title = title;
-  if (text) document.text = text;
+  //append to document
+  if (append) document.text += '\n' + text;
```
tommoor: Enforcing a newline here reduces the possibilities for this flag, It seems like it would be fine to give the option to the consumer of the API. Lets change this to use parenthesis to match the rest of the file/codebase.

arpitsingh94:
Hi. I made the changes as requested. Please check.

Issue #1000, opened 2019-07-24
[docs] Document export API


captn3m0:
Would be nice to have an API that triggers an export of all documents.

Was looking at automating exports as an alternative to #997, but since there is no API, we can't script it.

tommoor:
The endpoint you're looking for is called `collections.exportAll` (there's also an equivalent `collections.export` that takes a `collectionId`) right now it will send an email to the authenticated user with a zip of the documents, this is because the export could take some time and has to be done on a background job.

I'll leave this issue open as a documentation request

PR #1001, opened 2019-07-26
feat: Export collection as direct download instead of emailing

tommoor:
@captn3m0 this PR takes a step towards your export requirements, the `collections.export` endpoint allows downloading individual collections as a zip directly with this change – I've also made this behavior the default in the in-app UI.

Next step would be the entire wiki, however I'm going to disable that functionality on the hosted version in favor of keeping email export as It's pretty inadvisable to do so much heavy lifting on a node process that has to continue serving http requests.

closes https://github.com/outline/outline/issues/1000

### DIFF
--- a/app/models/Collection.js
+++ b/app/models/Collection.js
@@ -128,6 +128,10 @@ export default class Collection extends BaseModel {
   };
 
   export = () => {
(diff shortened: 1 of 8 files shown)


captn3m0:
This looks great, thanks :+1:

PR #1003, opened 2019-07-27
fix: document duplicate command

dgadelha:
This is because the front-end sends `collectionId` as `collection`, instead of `collectionId`.

Request:

image

Response:

image


tommoor:
Thanks for catching this, got missed in a refactor 🙏

PR #1004, opened 2019-07-29
chore: add missing EXPOSE rule

shyim:
Otherwise, the running container does not show up in Traefik 😅


tommoor:
Hey, is this needed because you're not using docker-compose to run the image?

shyim:
I don't understand what do you mean :)
I have running [Traefik](https://traefik.io/) on port 80 and this forwards the requests to the Docker container internally. 
To connect from one container to another container in the same network, you have to expose the ports. That's what i did here :)

PR #1036, opened 2019-09-12
fix: Ensure menus are always kept on the screen

MadMango:
completes #981 

still WIP, please see the comments on the files (I was gonna open it as a draft PR but forgot so just tagged it with WIP, soz)

### DIFF
--- a/app/components/DropdownMenu/DropdownMenu.js
+++ b/app/components/DropdownMenu/DropdownMenu.js
@@ -33,12 +33,7 @@ class DropdownMenu extends React.Component<Props> {
   @observable position: 'left' | 'right' | 'center';
   @observable bodyRect: ClientRect;
   @observable labelRect: ClientRect;
-  @observable dropdownRef: { current: null | HTMLElement };
-
-  constructor(props: Props) {
-    super(props);
-    this.dropdownRef = React.createRef();
-  }
+  @observable dropdownRef: { current: null | HTMLElement } = React.createRef();
 
   handleOpen = (
     openPortal: (SyntheticEvent<>) => void,
@@ -91,10 +86,10 @@ class DropdownMenu extends React.Component<Props> {
     const el = this.dropdownRef.current;
 
     const sticksOutPastBottomEdge =
-      el.scrollHeight + this.top > window.innerHeight;
+      el.clientHeight + this.top > window.innerHeight;
     if (sticksOutPastBottomEdge) {
       this.top = undefined;
-      this.bottom = 0;
+      this.bottom = -1 * window.pageYOffset;
     } else {
       this.bottom = undefined;
     }
@@ -179,7 +174,7 @@ const Label = styled(Flex).attrs({
 `;
 
 const Position = styled.div`
-  position: fixed;
+  position: absolute;
   display: flex;
   ${({ left }) => (left !== undefined ? `left: ${left}px` : '')};
   ${({ right }) => (right !== undefined ? `right: ${right}px` : '')};
@@ -187,7 +182,6 @@ const Position = styled.div`
   ${({ bottom }) => (bottom !== undefined ? `bottom: ${bottom}px` : '')};
   max-height: 75%;
   z-index: 1000;
-  box-shadow: ${props => props.theme.menuShadow};
   transform: ${props =>
     props.position === 'center' ? 'translateX(-50%)' : 'initial'};
 `;
@@ -201,6 +195,7 @@ const Menu = styled.div`
   min-width: 180px;
   overflow: hidden;
   overflow-y: auto;
+  box-shadow: ${props => props.theme.menuShadow};
 
   @media print {
     display: none;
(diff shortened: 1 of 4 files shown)


tommoor:
Hey hey, thanks for taking this on – really great stuff. I checked it out and noticed a couple of small issues with the way it works:

- If the menu is attached to a fix item such as the "New doc" button and the page is scrolled then it's placed in the wrong position:
    image

- It seems as though the opening animation is applied to the wrong element, the content of the dropdown moves within the menu bounding.
- We've also lost the subtle border radius on the dropdown, this might be related to the issue above though.

app/components/DropdownMenu/DropdownMenu.js
```diff
@@ -26,9 +26,19 @@ type Props = {
...
+  @observable right: ?number;
+  @observable left: ?number;
+  @observable position: 'left' | 'right' | 'center';
+  @observable bodyRect: ClientRect;
+  @observable labelRect: ClientRect;
+  @observable dropdownRef: { current: null | HTMLElement };
+
+  constructor(props: Props) {
+    super(props);
+    this.dropdownRef = React.createRef();
```
tommoor: There's no need for the constructor here, you can actually just do this on the class itself

```javascript
dropdownRef: React.createRef();
```

app/components/DropdownMenu/DropdownMenu.js
```diff
@@ -62,13 +67,65 @@ class DropdownMenu extends React.Component<Props> {
...
+          this.left = undefined;
+        } else if (this.position === 'left') {
+          this.position = 'right';
+          this.right = undefined;
+        }
+      }
+    }
+
+    this.initPosition();
+    this.forceUpdate();
```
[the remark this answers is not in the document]
tommoor: Updating an observable prop should just magically always re-render so I'm not sure either with this one 🤔

app/components/DropdownMenu/DropdownMenu.js
MadMango: fixed, thanks!

MadMango:
> Hey hey, thanks for taking this on – really great stuff. I checked it out and noticed a couple of small issues with the way it works:
> 
> * If the menu is attached to a fix item such as the "New doc" button and the page is scrolled then it's placed in the wrong position:
>   image
> * It seems as though the opening animation is applied to the wrong element, the content of the dropdown moves within the menu bounding.
> * We've also lost the subtle border radius on the dropdown, this might be related to the issue above though.

I started looking into this and I noticed a couple issues with changing the position to `fixed` and I changed back to `absolute` and I'm now using the scroll position of the window to determine the `bottom` position for the dropdown. However, the menu that you mentioned has the same behavior in master so I think that might be a separate issue? I'm happy to pick it up as well but not sure if it's better to track it as a separate problem or not.

tommoor:
> However, the menu that you mentioned has the same behavior in master so I think that might be a separate issue?

Oh, are you sure? I just tried reproducing on the production website and I was unable to – just opening that menu on the dashboard and then scrolling the page down…

MadMango:
> > However, the menu that you mentioned has the same behavior in master so I think that might be a separate issue?
> 
> Oh, are you sure? I just tried reproducing on the production website and I was unable to – just opening that menu on the dashboard and then scrolling the page down…

Hi Tom, sorry for very inconsistent replies but I've been away for most of the last two weeks, I'll look into it properly tomorrow.

tommoor:
No worries, appreciate your help with this – hoping we can get it squared away for the next release 😄

MadMango:
Hi Tom!

I'm not sure if we're trying the same thing but for me, the behaviour in master and the feature branch is exactly the same for that menu, unless it's something different / I'm using it wrong?

[master](https://drive.google.com/open?id=1zYrb6Z49iLdDVcGl9UrVz-IpOLP_RLbJ)
[feature](https://drive.google.com/open?id=1_OFZWVrPDwxWAv0jwzBSsmutCX9kDt37)

Thanks,
Mat

tommoor:
The issue only shows up when the page is **scrolled** – because the menu takes into account the page scroll but in this case the menu is actually relative to a fixed element. I think the solution would be to find if any parent elements are fixed and then ignore the scroll?

MadMango:
> The issue only shows up when the page is **scrolled** – because the menu takes into account the page scroll but in this case the menu is actually relative to a fixed element. I think the solution would be to find if any parent elements are fixed and then ignore the scroll?

I think for dropdowns that are children of fixed elements we could simply make the dropdown fixed as well, I just tested that and it seems to work fine, can you foresee any problems with that approach?

MadMango:
@tommoor `getComputedStyle` saves us a lot of hassle here, I pushed a change, looks good to me now.

app/components/DropdownMenu/DropdownMenu.js
```diff
@@ -40,18 +46,24 @@ class DropdownMenu extends React.Component<Props> {
...
+        this.bottom = undefined;
+        this.position = this.props.position || 'left';
+
+        if (currentTarget.parentElement) {
+          const triggerParentStyle = getComputedStyle(
+            currentTarget.parentElement
+          );
+
+          if (triggerParentStyle.position === 'static') {
+            this.fixed = true;
```
tommoor:
```suggestion
            this.fixed = true;
            this.top = this.labelRect.bottom;
```

Hey @MadMango – testing this out there still seems to be an issue with the positioning of fixed menus when the body is scrolled – but adding this additional line fixes it for me

Issue #1033, opened 2019-09-05
Show last active date on people list instead of joined date


tommoor:
This will be much more useful

tanvirraj:
@tommoor hey, i want to start  with this

PR #1039, opened 2019-09-15
Show last active date on people list instead of joined date

rylxes:
#1033


tommoor:
Your approach looks exactly right, just gotta update those jest snapshots so that the tests pass 🙏

Issue #1047, opened 2019-09-21
Slack command cannot be used with a Google team


tommoor:
Currently it requires the slack user that initiates the command to have previously signed in to the team.

In hindsight this seems like an unnecessary requirement, it would be nice to enable anyone with access to run the Slack command to get results. This then also accounts for those users that only ever sign in with Google but still use Slack for comms.

tommoor:
Currently the only way to search a wiki is through the `Document.searchForUser` method which requires a user model – this will need to be changed to achieve this.

Precious5050:
Ok thanks

PR #1049, opened 2019-09-22
fix: Unknown Slack users should be able to search team accessible docs

tommoor:
This PR addresses an issue where users that have never signed in to Slack are unable to search team accessible documents.

Previously the search was based entirely off user permissions, now if the passed user is unknown docs that are published in team visible collections will be searchable. This also means that teams using Google to sign in with a Slack integration get a much better experience.

If the user _is_ recognized then their permissions are used to retain searching of their own drafts and private collections.

closes https://github.com/outline/outline/issues/1047

### DIFF
--- a/server/api/hooks.js
+++ b/server/api/hooks.js
@@ -69,7 +69,7 @@ router.post('hooks.interactive', async ctx => {
   if (!team) {
     ctx.body = {
       text:
-        'Sorry, we couldn’t find an integration for your team. Head to yout Outline settings to set one up.',
+        'Sorry, we couldn’t find an integration for your team. Head to your Outline settings to set one up.',
       response_type: 'ephemeral',
       replace_original: false,
     };
@@ -128,7 +128,7 @@ router.post('hooks.slack', async ctx => {
     ctx.body = {
       response_type: 'ephemeral',
       text:
-        'Sorry, we couldn’t find an integration for your team. Head to yout Outline settings to set one up.',
+        'Sorry, we couldn’t find an integration for your team. Head to your Outline settings to set one up.',
     };
     return;
   }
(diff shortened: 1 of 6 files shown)


server/api/hooks.js
```diff
@@ -58,15 +58,18 @@ router.post('hooks.interactive', async ctx => {
...
   ctx.assertPresent(callback_id, 'callback_id is required');
 
-  if (token !== process.env.SLACK_VERIFICATION_TOKEN)
+  if (token !== process.env.SLACK_VERIFICATION_TOKEN) {
     throw new AuthenticationError('Invalid verification token');
+  }
 
-  const user = await User.findOne({
-    where: { service: 'slack', serviceId: data.user.id },
+  const team = await Team.findOne({
```
tommoor: We always trusted the data from Slack here, previously it was the user and now the team.

server/api/hooks.js
```diff
@@ -116,25 +121,33 @@ router.post('hooks.slack', async ctx => {
...
+      teamId: team.id,
+      service: 'slack',
+      serviceId: user_id,
+    },
   });
 
+  const options = {
+    limit: 5,
+  };
+  const results = user
```
tommoor: 🍖 If we recognize the user then use their permissions to return more results

server/api/hooks.test.js
```diff
@@ -156,26 +162,46 @@ describe('#hooks.slack', async () => {
     expect(body.text.includes('How to use')).toEqual(true);
   });
 
-  it('should respond with error if unknown user', async () => {
```
tommoor: 🍖

server/api/users.test.js
```diff
@@ -12,26 +12,32 @@ afterAll(server.close);
...
-    const { user } = await seed();
+    const { user, admin } = await seed();
     const res = await server.post('/api/users.list', {
       body: { token: user.getJwtToken() },
     });
     const body = await res.json();
 
     expect(res.status).toEqual(200);
-    expect(body).toMatchSnapshot();
+    expect(body.data.length).toEqual(2);
```
tommoor: These tests had become flaky

Issue #755, opened 2018-08-13
Microsoft Teams Authentication


ambroselittle:
My org is looking to move off of Slack and onto other platforms, and from Google to Microsoft. 

So, while I think this solution is super slick, I'm curious how much trouble it would be--if we did the modifications in a fork--to remove those dependencies (and possibly swap in a Teams integration). 

We'd like to host internally, so auth is less of a concern for us, FWIW.

Thanks in advance.

jhult:
I would also be interested in authenticating with Microsoft (specifically Teams/Office 365).

snovak7:
I have done part with Azure AD integration, I will make PR soon.

tommoor:
Hello everyone, Microsoft sign-in has now been completed and is available on the cloud hosted version of Outline, here:
https://app.getoutline.com/create

For self-hosted it will be included in the next release `v0.55.0` which is currently estimated to be within the week.

PR #1067, opened 2019-10-16
Microsoft, Office 365 Authentication

snovak7:
This is first version,

There were some assumptions made, only first "Organization" is used.
This solves authentication part of #755


server/auth/azuread.js
```diff
@@ -0,0 +1,211 @@
...
+    const userResource = await client.api('/me').get();
+
+    if (!userResource) {
+      ctx.redirect('/?notice=azuread-no-userinfo');
+      return;
+    }
+
+    userId = userResource.id;
+    userName = userResource.displayName;
+    userEmail = userResource.mail;
```
ahmelsayed:
```suggestion
    userEmail = userResource.mail || userResource.userPrincipalName;
```

`userResource.mail` could be empty if the AAD user doesn't have an email inbox. That will cause an error in `findOrCreate` because it's a required field. 

https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-userprincipalname
https://stackoverflow.com/a/47480542

snovak7: Thank you, I have a mailbox, but I don't use it, haven't checked the docs TBH.

server/auth/azuread.js
```diff
@@ -0,0 +1,211 @@
...
+  return templateAuthzUrl.replace('<state>', state);
+};
+
+var aqcuireTokenAsync = (code, redirectUri, clientId, clientSecret) => {
+  return new Promise((resolve, reject) => {
+    var authenticationContext = new AuthenticationContext(authorityUrl);
+    authenticationContext.acquireTokenWithAuthorizationCode(
+      code,
+      redirectUri,
+      'https://graph.microsoft.com',
```
Xivolkar: Consider making this injectable? constants?

server/auth/azuread.js
```diff
@@ -0,0 +1,211 @@
...
+  authorityUrl
+}//oauth2/authorize?response_type=code&client_id=${clientId}&&redirect_uri=${
+  redirectUri
+}&state=<state>`;
+
+var createAuthorizationUrl = state => {
+  return templateAuthzUrl.replace('<state>', state);
+};
+
+var aqcuireTokenAsync = (code, redirectUri, clientId, clientSecret) => {
```
Xivolkar: typo - acquire

namxam:
What about this PR? Is it still an option to merge this any time soon? I might sign up to fix any open issues to get this merged, as I cannot longer stand to work with Sharepoint 😏  And we have quite a few Office 365 users who would be pain to onboard manually.

snovak7:
I think this is outdated PR, with PR #1183 this should be fixed, and you will have a lot of options for integrating numerous identity providers.

tommoor:
Agreed with @snovak7 – the approach in #1183 is preferred and supersedes this one, many thanks for the efforts here though

Issue #1053, opened 2019-10-01
Guest access


tommoor:
Sometimes it's useful to be able to give access to external contractors or users that aren't part of your main authentication system. We can add a minimal version of this that doesn't require password management using one-time tokens sent via email to login (aka "magic links").

- With guest access enabled the team subdomain signin page will gain an additional option to "signin with email"
- If the user is a pre-registered guest they will receive a link via email with a token to signin

PR #1088, opened 2019-11-26
feat: Guest email authentication 

tommoor:
Adds the ability to invite people that do not have Slack/Google sign-in into your team. This is particularly useful for third party contractors.

Simply invite them via email from the left sidebar, or "People" settings – they will receive an invite link via email to sign-in and are not required to create a password.

This functionality can be disabled with a new setting under "Settings -> Security"

### TODO

- [x] Add "Guest" checkbox on invite form
- [x] Pass guest flag through and choose order of options on login page
- [x] Remove hanging horizontal separator
- [x] Autofocus email input instead of name on invite modal
- [x] Restore copy link invite section
- [x] Allow remove pending invites
- [x] Fix email column restrictions

closes https://github.com/outline/outline/issues/1053


app/scenes/Invite.js
```diff
@@ -67,7 +68,8 @@ class Invite extends React.Component<Props> {
     this.invites.push({ email: '', name: '' });
   };
 
-  handleRemove = (index: number) => {
+  handleRemove = (ev: SyntheticEvent<>, index: number) => {
+    ev.preventDefault();
```
tommoor: This fixes an existing bug where clicking the "X" next to an invite would close the dialog.

Issue #1085, opened 2019-11-17
Add 'Group' permissions


aam1r:
Permissions for collections (re: #989) are now based on a per-user basis. A user is explicitly added to collections and that controls what kind of access they have.

It would be pretty powerful to have Group based memberships. 

Instead of giving one-off permissions to users to individual collections, the administrator would define Groups. For example:

* Engineering
* Finance
* HR
* Operations
* Marketing, Sales, etc.

The collection permissions would then be granted to the group. This makes it much easier to control permissions for larger teams than having to individually grant permissions to users.

tommoor:
Hey @aam1r – this project has been started, progress can be followed here: https://github.com/outline/outline/pull/1102

PR #1102, opened 2019-12-11
feat: adds groups and group permissions

thenanyu:
Groups
- [x] backend for groups
- [x] UI for groups and users
  - [x] group creation and edit workflow
  - [x] group members workflow
  - [x] user counts
  - [x] cleanup

Permissions
- [x] create `collection-groups` model
- [x] modify `User#collectionIds` to take group permissions into account
- [x] create `collections.add_group` endpoint
- [x] create `collections.remove_group` endpoint
- [x] create endpoint to get the groups for a collection
- [x] add web socket events for group membership changes
- [x] add web socket events for collection group changes

closes https://github.com/outline/outline/issues/1085


thenanyu:
@tommoor — getting a failing test that isn't happening locally, look familiar?

server/policies/collection.js
```diff
@@ -14,6 +14,8 @@ allow(User, ['read', 'export'], Collection, (user, collection) => {
   if (
     collection.private &&
     (!collection.memberships || !collection.memberships.length)
+    // XXX: does this need the invariant that is below in publish/update?
+    // it doesn't seem to check the permission type
```
thenanyu: @tommoor looks like this is just checking for the presence of a preloaded membership at all, which I think is ok, but should this have the same invariant as https://github.com/outline/outline/pull/1102/files#diff-240dcfa58e2e777bcc9ac5081c588bfaR30

server/api/documents.js
```diff
@@ -388,6 +392,7 @@ router.post('documents.info', auth({ required: false }), async ctx => {
     document = await Document.findByPk(
       id,
       user ? { userId: user.id } : undefined
+      // XXX: this seems crazy to me, like you can only find the doc if you created it?
```
thenanyu: @tommoor slightly confused by this line, I am under the impression that the option scopes the find to match for userId? This seems like only the author would find the document, but I'm certainly misunderstanding something

server/policies/collection.js
tommoor: I believe there's a reason why, but I don't remember – possibly something to do with allowing read permissions for sharing.

server/api/documents.js
tommoor: It's the second argument, which means it's passed into the `options` and allows us to preload the users memberships, here:
https://github.com/outline/outline/blob/master/server/models/Document.js#L213-L217

server/api/groups.js
```diff
@@ -0,0 +1,250 @@
...
+    },
+  });
+
+  if (!membership) {
+    membership = await group.addUser(user, {
+      through: { createdById: ctx.state.user.id },
+    });
+  }
+
+  await Event.create({
```
tommoor: This event wants to be moved within the `!membership` block above, right?

server/api/groups.js
```diff
@@ -0,0 +1,250 @@
...
+  const { id, name } = ctx.body;
+  ctx.assertPresent(name, 'name is required');
+  ctx.assertUuid(id, 'id is required');
+
+  const user = ctx.state.user;
+  const group = await Group.findByPk(id);
+
+  authorize(user, 'update', group);
+
+  group.name = name;
```
tommoor: Should we check if `group.name !== name` here before performing the update and saving an event?

server/api/groups.js
```diff
@@ -0,0 +1,250 @@
...
+
+router.post('groups.create', auth(), async ctx => {
+  const { name } = ctx.body;
+  ctx.assertPresent(name, 'name is required');
+
+  const user = ctx.state.user;
+
+  authorize(user, 'create', Group);
+  const group = await Group.create({
+    name,
```
tommoor: Thoughts on making name unique within the team?

server/api/__snapshots__/groups.test.js.snap
```diff
@@ -0,0 +1,80 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
thenanyu: not sure what this means, I'm not super familiar with jest snapshots — how should I resolve?

server/api/groups.js
thenanyu: yep, good idea

server/api/groups.js
thenanyu: I am used to depending on the ORM to iron this kind of stuff out, I believe sequelize handles this https://github.com/sequelize/sequelize/commit/8df98ded5c3bbd4f62a37a7fd5ed2c8c6a450b9e

server/api/groups.js
thenanyu: yup yup

server/api/__snapshots__/groups.test.js.snap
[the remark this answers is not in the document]
tommoor: Sorry I mean the `report.20191219.054438.27.0.001.json` file above this in the diff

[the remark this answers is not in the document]
thenanyu: ah yeah got it

app/scenes/Settings/components/GroupListItem.js
```diff
@@ -33,15 +35,20 @@ class GroupListItem extends React.Component<Props> {
...
       <React.Fragment>
         <ListItem
           title={
             <Title onClick={this.handleMembersModalOpen}>{group.name}</Title>
           }
-          subtitle={<React.Fragment>10 members</React.Fragment>}
+          subtitle={
+            <React.Fragment>
+              {membersCount} member{membersCount === 1 ? '' : 's'}
+            </React.Fragment>
```
thenanyu: I didn't see a pluralizer lib, is there one?

server/api/groups.js
```diff
@@ -26,9 +26,18 @@ router.post('groups.list', auth(), pagination(), async ctx => {
     limit: ctx.state.pagination.limit,
   });
 
+  const groupMemberships = await Promise.all(
+    groups.map(async group => {
+      return await group.getGroupMemberships();
+    })
+  );
```
thenanyu: (this is pre-fetched)

server/models/Group.js
```diff
@@ -56,6 +56,15 @@ Group.associate = models => {
...
   });
+  Group.addScope('defaultScope', () => ({
+    include: [
+      {
+        model: models.GroupUser,
+        as: 'groupMemberships',
+        required: false,
+      },
+    ],
+  }));
```
thenanyu: 🍖 — I am a little worried about perf, this seems like it could get quite large

app/scenes/Settings/components/GroupListItem.js
tommoor: Weirdly, there isn't – I've somehow gotten away without it.

app/components/PaginatedList.js
```diff
@@ -67,7 +71,7 @@ class PaginatedList extends React.Component<Props> {
 
     const showLoading =
       this.isFetching && !this.isFetchingMore && !this.isInitiallyLoaded;
-    const showEmpty = !items.length || showLoading;
+    const showEmpty = !items.length && !showLoading;
```
tommoor: I made a similar change on my open branch https://github.com/outline/outline/pull/1156/files#diff-bbe314f2fe1f64f531949f4934bb6f81R90

shared/utils/routeHelpers.js
```diff
@@ -72,3 +72,7 @@ export function signin(service: string = 'slack'): string {
 export function settings(): string {
   return `/settings`;
 }
+
+export function groupSettings(): string {
```
tommoor: How about making the subpage an optional prop for `settings` ?

server/models/Group.js
tommoor: Agreed, seems like it's unsustainable

server/models/Group.js
```diff
@@ -0,0 +1,70 @@
...
+    teamId: {
+      type: DataTypes.UUID,
+      defaultValue: DataTypes.UUIDV4,
+    },
+    name: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+  },
+  {
```
tommoor: You can add custom getters here, perhaps we add one for `membershipCount` that does the count query and can be called from the presenter? I'm not sure there is a more efficient way of doing that, short of caching the count in a column.

thenanyu: investigated the custom getter route, but that would be an `async` getter which seems like creates a bit of complexity in sequelize — I'm gonna try just passing in a number for `memberCount` to the presenter and sending that across

shared/utils/routeHelpers.js
thenanyu: tried that first, but seemed too complex, from a DSL standpoint, this seemed better until we have a couple of more instances

server/models/Group.js
thenanyu: we'd have to prefetch the count at least, gonna try taking that approach outlined here https://stackoverflow.com/questions/37817808/counting-associated-entries-with-sequelize

server/api/groups.js
```diff
@@ -0,0 +1,280 @@
...
+  let group = await Group.findByPk(id);
+  authorize(ctx.state.user, 'update', group);
+
+  const user = await User.findByPk(userId);
+  authorize(ctx.state.user, 'read', user);
+
+  await group.removeUser(user);
+
+  await Event.create({
+    name: 'collections.remove_user',
```
tommoor:
```suggestion
    name: 'groups.remove_user',
```

server/api/collections.js
```diff
@@ -71,6 +81,148 @@ router.post('collections.info', auth(), async ctx => {
...
+  authorize(ctx.state.user, 'update', collection);
+
+  const group = await Group.findByPk(groupId);
+  authorize(ctx.state.user, 'read', group);
+
+  await collection.removeGroup(group);
+
+  await Event.create({
+    name: 'collections.remove_group',
+    collectionId: collection.id,
```
tommoor:
```suggestion
    collectionId: collection.id,
    modelId: group.id,
```

server/api/collections.js
```diff
@@ -71,6 +81,148 @@ router.post('collections.info', auth(), async ctx => {
...
+      createdById: ctx.state.user.id,
+    });
+  } else if (permission) {
+    membership.permission = permission;
+    await membership.save();
+  }
+
+  await Event.create({
+    name: 'collections.add_group',
+    collectionId: collection.id,
```
tommoor:
```suggestion
    collectionId: collection.id,
    modelId: group.id,
```

server/api/groups.js
```diff
@@ -0,0 +1,280 @@
...
+  authorize(user, 'update', group);
+
+  group.name = name;
+
+  if (group.changed()) {
+    await group.save();
+    await Event.create({
+      name: 'groups.update',
+      teamId: user.teamId,
+      actorId: user.id,
```
tommoor:
```suggestion
      actorId: user.id,
      modelId: group.id,
```

server/api/groups.js
```diff
@@ -0,0 +1,280 @@
...
+
+  const { user } = ctx.state;
+  const group = await Group.findByPk(id);
+
+  authorize(user, 'delete', group);
+  await group.destroy();
+
+  await Event.create({
+    name: 'groups.delete',
+    actorId: user.id,
```
tommoor:
```suggestion
    actorId: user.id,
    modelId: group.id,
```

server/models/Group.js
```diff
@@ -0,0 +1,70 @@
...
+    timestamps: true,
```
tommoor:
```suggestion
    timestamps: true,
    paranoid: true,
```

Lets keep deleted groups in the db

server/migrations/20191211044318-create-groups.js
```diff
@@ -0,0 +1,45 @@
...
+        }
+      },
+      createdAt: {
+        type: Sequelize.DATE,
+        allowNull: false
+      },
+      updatedAt: {
+        type: Sequelize.DATE,
+        allowNull: false
+      }
```
tommoor:
```suggestion
      },
      deletedAt: {
        type: Sequelize.DATE,
        allowNull: true
      }
```

app/components/GroupListItem.js
```diff
@@ -0,0 +1,88 @@
...
+              {showFacepile && <Facepile users={users} overflow={overflow} />}
+
+              {renderActions({
+                openMembersModal: this.handleMembersModalOpen,
+              })}
+            </Flex>
+          }
+        />
+        <Modal
+          title="Group Members"
```
tommoor:
```suggestion
          title="Group members"
```

server/models/Group.js
thenanyu: there's a naming uniqueness constraint, how does that play with paranoid?

tommoor: The uniqueness is based on a custom query so it should be a non-issue, deleted groups will be filtered by default

server/models/Group.js
```diff
@@ -0,0 +1,70 @@
...
+    validate: {
+      isUniqueNameInTeam: async function() {
+        const foundItem = await Group.findOne({
+          where: {
+            name: { [Op.iLike]: this.name },
```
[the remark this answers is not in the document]
thenanyu: yeah I think you're right

app/components/GroupListItem.js
```diff
@@ -0,0 +1,88 @@
...
+          isOpen={this.membersModalOpen}
+        >
+          <GroupMembers group={group} onSubmit={this.handleMembersModalClose} />
+        </Modal>
+      </React.Fragment>
+    );
+  }
+}
+
+const Title = styled.span`
```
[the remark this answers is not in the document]
thenanyu: I don't remember honestly, I can take it out and see what happens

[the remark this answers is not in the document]
thenanyu: without the styling there are no hover states. where do you think the hover style would otherwise be coming from?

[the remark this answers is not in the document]
tommoor: Sorry, I added a comment here the other day but I guess github was down.

Agreed, I mistakenly though the hover state was part of `ListItem` – It probably should be, but not for this PR 👍

app/scenes/GroupDelete.js
```diff
@@ -0,0 +1,59 @@
...
+  render() {
+    const { group } = this.props;
+
+    return (
+      <Flex column>
+        <form onSubmit={this.handleSubmit}>
+          <HelpText>
+            Are you sure about that? Deleting the <strong>{group.name}</strong>{' '}
+            will cause its members to lose access to collections and documents
+            that it is assigned to.
```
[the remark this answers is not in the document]
thenanyu: ..."that it is associated with"?

app/stores/CollectionGroupMembershipsStore.js
```diff
@@ -0,0 +1,83 @@
...
+  @action
+  fetchPage = async (params: ?PaginationParams): Promise<*> => {
+    this.isFetching = true;
+
+    try {
+      const res = await client.post(`/collections.group_memberships`, params);
+
+      invariant(res && res.data, 'Data not available');
+
+      runInAction(`/collections.group_memberships`, () => {
```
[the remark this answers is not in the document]
thenanyu: "CollectionGroupMembershipsStore#fetchPage"?

app/stores/GroupMembershipsStore.js
```diff
@@ -0,0 +1,78 @@
...
+  @action
+  fetchPage = async (params: ?PaginationParams): Promise<*> => {
+    this.isFetching = true;
+
+    try {
+      const res = await client.post(`/groups.memberships`, params);
+
+      invariant(res && res.data, 'Data not available');
+
+      runInAction(`/groups.memberships`, () => {
```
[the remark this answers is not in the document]
thenanyu: GroupMembershipsStore#fetchPage ?

server/api/groups.js
```diff
@@ -0,0 +1,280 @@
...
+    offset: ctx.state.pagination.offset,
+    limit: ctx.state.pagination.limit,
+  });
+
+  ctx.body = {
+    pagination: ctx.state.pagination,
+    data: {
+      groups: groups.map(presentGroup),
+      groupMemberships: groups
+        .map(g => g.groupMemberships.slice(0, MAX_AVATAR_DISPLAY))
```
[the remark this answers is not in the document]
thenanyu: we need the full set to get the count

app/scenes/GroupDelete.js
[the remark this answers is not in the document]
tommoor: OKAY

app/stores/CollectionGroupMembershipsStore.js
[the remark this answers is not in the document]
tommoor: It's just for debugging, up to you – that does better match elsewhere 👍

tommoor:
Review in progress…

image

We should change this to "Permissions…"

image

The (+1) facepile addition should be the same size as the avatars, spacing is a little tight with the "...", probably needs 8px between. When you hover over with the mouse it has the hand cursor but clicking does nothing, can we make clicking facepile open the members modal?

image

This is the error when attempting to create the same group name twice, I wonder how we can get rid of that last bit in parenthesis?

image

These avatars should be larger to match those in the "People" list

server/services/websockets.js
```diff
@@ -206,19 +214,232 @@ export default class Websockets {
...
-        // let everyone with access to the collection know a user was removed
-        socketio.to(`collection-${event.collectionId}`).emit(event.name, {
-          event: event.name,
-          userId: event.userId,
-          collectionId: event.collectionId,
+        const membershipUserIds = Collection.membershipUserIds(
+          event.collectionId
+        );
+
+        if (membershipUserIds.includes(event.userId)) {
```
tommoor: image

server/services/websockets.js
```diff
@@ -206,19 +214,232 @@ export default class Websockets {
...
+              event: event.name,
+              collectionId: event.collectionId,
+            });
+          }
+        }
+        return;
+      }
+      case 'groups.add_user': {
+        // do an add user for every collection that the group is a part of
+        const collectionGroupMemberships = CollectionGroup.find({
```
tommoor: Also hitting an error here, I think we want `findAll` ?

image

server/services/websockets.js
```diff
@@ -206,19 +214,232 @@ export default class Websockets {
...
+          // tell any user clients to connect to the websocket channel for the collection
+          return socketio.to(`user-${event.userId}`).emit('join', {
+            event: event.name,
+            collectionId: collectionGroup.collectionId,
+          });
+        }
+        return;
+      }
+      case 'groups.remove_user': {
+        const collectionGroupMemberships = CollectionGroup.find({
```
tommoor: This would be `findAll` again

PR #1114, opened 2019-12-19
feat: Document presence indicator

### LINKED ISSUE #1104: Implement "typing" indicator
- When another user is editing the document we should show an "editing" icon on their avatar
- The tooltip should clearly show "Editing"
- This should also show in "edit" mode

tommoor:
This branch adds an indicator of other users presence inside documents, you can see who else is viewing or editing a document at any given time.

#### Viewing a draft document
image

#### Another user editing
image

### TODO

- [x] Frontend state management
- [x] Frontend UI
- [x] Add storage of editing state in postgres views table
- [x] Update views table on frontend
- [x] Fix delay on entering edit state
- [x] Communicate editing presence 
- [x] Fix delay on leaving edit stay
- [x] Show editing users in edit mode
- [x] Fix editing state not reset on editing client when leaving
- [x] Possible fix room join/leave on changing edit mode
- [x] Test, test, test

Foundational work for #811 
closes #1104


app/components/SocketProvider.js
```diff
@@ -39,164 +46,213 @@ class SocketProvider extends React.Component<Props> {
...
     if (!auth.token) return;
 
     this.socket.on('connect', () => {
+      // immediately  current users token to the websocket backend where it is
+      // verified, if all goes well an 'authenticated' message will be received
+      // in response
       this.socket.emit('authentication', {
         token: auth.token,
       });
+    });
```
tommoor: Note: The majority of changes in this file are whitespace changes as the `on` bindings are no longer nested inside of the `connect` callback handler, this caused the event handlers to be bound multiple times – once for every re-connect.

tommoor:
This was reviewed in person.

PR #1127, opened 2019-12-29
fix: remove transparency from dark theme sidebar background

ezrafree:
Since `black50` sets 50% opacity, it causes the text behind the sidebar to be seen behind the sidebar at smaller/mobile device widths.

This adds a new `veryDarkBlue` color variable to theme.js and uses that for the dark theme sidebar's background color instead.

### DIFF
--- a/shared/styles/theme.js
+++ b/shared/styles/theme.js
@@ -5,6 +5,7 @@ const colors = {
   almostBlack: '#111319',
   lightBlack: '#2F3336',
   almostWhite: '#E6E6E6',
+  veryDarkBlue: '#08090C',
 
   slate: '#9BA6B2',
   slateLight: '#DAE1E9',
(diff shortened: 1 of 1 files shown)


tommoor:
Hey, could you provide a screenshot of the problem this is fixing? Thanks!

ezrafree:
Absolutely, the following screenshot was taken in Google Chrome Version 79.0.3945.88 (Official Build) (64-bit) on Ubuntu 19.10, in the device emulation mode of an iPhone 5/SE:

screencapture-localhost-3000-dashboard-2019-12-29-10_22_49

The same issue occurs for me in Mozilla Firefox 71.0 (64-bit) on Ubuntu 19.10:

Screenshot from 2019-12-29 10-27-01

tommoor:
Perfect, thanks for the fix - looks like the `hex` value is right on and I don't have any issue with the naming 👍

Issue #796, opened 2018-11-06
Store images in private bucket


tommoor:
Right now images are stored publicly on S3 with high entropy URL's to ensure that finding / accessing them is not possible without going through the wiki. However, this can be improved by requiring a signature to access the images…

Implementation would probably look like:

- Change the url that is stored in the document to point to an Outline API url, something like:
`https://www.getoutline.com/api/images.info?id=XXX`
- Have the API forward these requests/response to the backing storage with the correct signature
- On export we have two options, either:
    - download the images, store them in the ZIP and replace the url to a local path
    - embed an access token in the urls so that they work when loaded elsewhere

teddychan:
For self-hosted solution, may be the easiest method is to use S3 bucket policy.
Allow certain IPs to visit those images.

himanshuagarwal272:
Hey, I am thinking of picking this up! Any more suggestions @tommoor?

tommoor:
I still think that the api "proxy" is the way to go for loading authenticated images, even doing this half of the task would be very useful.

The tricky part is what to do on export as it really complicates the existing process somewhat, i'd imagine we'd be able to search the documents for images and include them in the zip as suggested in the PR description.

himanshuagarwal272:
> I still think that the api "proxy" is the way to go for loading authenticated images, even doing this half of the task would be very useful.
> 
> The tricky part is what to do on export as it really complicates the existing process somewhat, i'd imagine we'd be able to search the documents for images and include them in the zip as suggested in the PR description.

Let me start working on the first half and Let's discuss how can we go forward with the other part

Issue #1010, opened 2019-08-07
Cleaning database and S3 bucket on deletion


flaviuse:
Hello !
First thanks for all your work on this open source project !
I have deployed the app on Heroku with Heroku postgres, Heroku redis, Google Oauth and AWS S3 bucket and everything is working well for now.

I have questions about the cleaning of the postgres databse and S3 bucket when deleting a document/collection/image from a document.... 

It seems that deleting an image from a document doesn't delete it from S3 bucket. Do you plan to add S3 bucket deletion feature ? Or am I missing some options to make it done ? It can be a great feature for maintaining a wiki long term.

Deleting (not archiving) a document doesn't delete it in postgres but just soft delete by adding a timestamp to the "deletedAt" field. 
![image](https://user-images.githubusercontent.com/40322270/62637240-5ca46f00-b93b-11e9-862d-23a963e6479b.png)
Is there an option to use real deletion instead of soft deletion to keep the DB clean ? 

Are these features on your road map ?

Once again thank you for this awesome wiki !

tommoor:
> It seems that deleting an image from a document doesn't delete it from S3 bucket. Do you plan to add S3 bucket deletion feature ? Or am I missing some options to make it done ? It can be a great feature for maintaining a wiki long term.

I'd definitely be open to a PR that implements this. I'd imagine there could be a post-document-update job similar to the one that processes backlinks to check if an image was removed and delete from S3. Right now It just hasn't been a priority.


> Deleting (not archiving) a document doesn't delete it in postgres but just soft delete by adding a timestamp to the "deletedAt" field.

Similarly, it's a "want to do" – the idea is to have a "Trash" menu item next to archive that would list the deleted documents for 30 days and have a job that will permanently clear them from the database after that.

flaviuse:
> I'd definitely be open to a PR that implements this. I'd imagine there could be a post-document-update job similar to the one that processes backlinks to check if an image was removed and delete from S3. Right now It just hasn't been a priority.
> 
I'll check when I got time

PR #1137, opened 2020-01-02
feat: private content

### LINKED ISSUE #1131: Serving private content from s3
Hello :wave:, first off this wiki is really quite fantastic! Neat, ergonomic and concise code. Thank you creators and contributors for developing it. I wanted to share some of my thoughts/experiments, on what turned out to be a continuation of this issue #796. I have had to do a fair bit of work with private/protected/restricted assets in the AWS environment specifically, so I thought I'd have a look at what it may take to bring it here.

## Summary
* switch the S3 bucket to be private, and set ACLs when uploading to private
* put an AWS CloudFront (CF) distribution in front of the S3 bucket, with signed cookies as access controls. (ccokies are good as they are passive, and don't need any dynamic rewrite/redirect to provide volatile signed urls for example.)
* simple find/replace the s3 image urls when serving a document with the url of the CF distribution

## Method
This can be broken up into two parts, *outline setup* and *aws setup*. I did this all using the docker-compose development setup for ease of debug.

### outline setup
My demo/mvp changes can be found on a fork [here](https://github.com/HUSSTECH/outline/tree/cf-private-content).

The changes here were not that large, but are done in a bit of a sub-optimal way at the moment. The main things to do in the outline codebase were to
1. setting s3 upload ACLs to `private`
1. be able to set cookies which would be used to authenticate to CF later
1. add a module (`cf.js`) to do all the CF related tasks and other miscellaneous helper functions for now
1. replace the s3 urls on the fly with the CF urls whenever documents text content is called
1. via the addition of 3 new config parameters (`AWS_CLOUDFRONT_URL, AWS_CLOUDFRONT_KEY_PAIR_ID, AWS_CLOUDFRONT_PRIVATE_KEY`), allow this to be toggled on/off

I added an nginx proxy to the compose definition, as per #898, required to make ssl and custom domain SSO work nicely. It is also needed to make the cookies set themselves properly.

### aws setup
This is the worst part! Working with cookies, DNS and CF is always tricky because there is no staging environment of the internet! I'll just leave a diagram below and the links required to replicate the CF setup.

image

- AWS docs [Serving Private Content with Signed URLs and Signed Cookies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/PrivateContent.html) -> follow the links about using signed cookies, and then signing with "custom polices"
- Critical step is to have a trusted signer setup on your account, see AWS docs [Specifying the AWS Accounts That Can Create Signed URLs and Signed Cookies (Trusted Signers)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-trusted-signers.html).
- a nice video step by step from AWS [How to Set up an Amazon CloudFront Distribution for Your Amazon S3 Origin](https://youtu.be/KIltfPRpTi4) (check yes for using signed cookies...the presenter skips this.)
- useful context blog post [Using Cloudfront Signed Cookies](https://www.spacevatican.org/2015/5/1/using-cloudfront-signed-cookies/)
- useful context blog post [Node.js — Serve Private Content using AWS CloudFront](https://medium.com/@apalshah/node-js-serve-private-content-using-aws-cloudfront-992e846259ae)

## TODOs
* Wanted to just stop at a quick MVP working setup, just to prove it could work. I have left TODO comments (some questions!) in my diffs. Mainly I think I'd say my implementation is messy, and likely has stuff in the wrong/inappropriate places.
* Not sure if replacing image urls on the fly is the right path. Maybe they can just be written directly to the committed document text with the CF urls in the first place.
* I don't like how I'm checking for the CF cookies every request. I just piggy backed on the `auth()` middleware and tried to limit the work being done in setting the cookie unless needed. TBH, I have seen cookie checks/sets on every request as a pattern in many other web frameworks...so maybe it has to be that way!

## Other stuff
* I also added the option to set a prefix on the S3 object key. This is not necessary to implement serving private content in this way, but I just found it a nice to have. E.g. you could eventually use the same bucket for different types of assets such as attachments.
* Not on my demo repo at the moment, but I was playing around with eliminating having to explicitly provide AWS credentials in the config when running on an ec2, and instead just letting the instance credentials take care of it. Works fine except only thing stopping it was the `makeCredential` function, as it needs an explicit `AWS_ACCESS_KEY_ID` set. Was thinking as you're already on the server side, it would be possible to use the `AWS.s3.putObject` or `AWS.s3.upload` methods directly, as they don't need a policy to be signed. Or I have often returned a signed POST url which allows the users' browsers to upload directly themselves. Offloading the work from the server entirely.

HUSSTECH:
Implements saving and serving of private images in s3. Also facilitates the archival of private images in exported zip folders.

When the `ENABLE_PRIVATE_CONTENT` environment variable is set to `true` images will be uploaded with a private ACL into s3. When their url is written to the document, it will be written as a call to an api endpoint which resolves to a redirect with a signed s3 url to then access the image.

When exporting collections with private content, the images are downloaded from s3 and saved in a directory alongside the documents. The document text is re-written to replace instances of the image urls with relative paths to the local images.

closes #1131 #796


server/utils/zip.js
```diff
@@ -4,12 +4,26 @@ import JSZip from 'jszip';
...
+const s3KeyRegex = /!\[.*\]\(\/api\/images\.info\?key=(?<key>.*)\)/gi;
+const imgageApiRegex = /(?<=!\[.*\]\()(\/api\/images\.info\?key=)/gi;
 
 async function addToArchive(zip, documents) {
   for (const doc of documents) {
     const document = await Document.findByPk(doc.id);
+    let text = unescape(document.text);
+
+    if (ENABLE_PRIVATE_CONTENT) {
+      const imageKeys = [...text.matchAll(s3KeyRegex)].map(
```
HUSSTECH: `matchAll` is from the es2020 spec, this is a test fail point at the moment - so I could go back to current convention to fix.

.env.sample
```diff
@@ -48,6 +48,10 @@ AWS_REGION=xx-xxxx-x
 AWS_S3_UPLOAD_BUCKET_URL=http://s3:4569
 AWS_S3_UPLOAD_BUCKET_NAME=bucket_name_here
 AWS_S3_UPLOAD_MAX_SIZE=26214400
+# upload images to s3 with private access controls
+# and serve them via signed urls
+# false by default
+ENABLE_PRIVATE_CONTENT=false
```
tommoor: I think private should be the default going forward – so you'd have to explicitly disable it, the only reason I can think to do so would be if you're using an S3 alternative that doesn't support the signing?

With that in mind perhaps this could become `AWS_S3_ACL=private` /  `AWS_S3_ACL=public-read` – i'd like to keep it related in the AWS_S3 namespace, if not set the app would assume private.

server/api/images.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import Router from 'koa-router';
+import auth from '../middlewares/authentication';
+import { getSignedImageUrl } from '../utils/s3';
+
+const router = new Router();
+
+router.post('images.info', auth(), async ctx => {
```
tommoor:
```suggestion
router.get('images.info', auth(), async ctx => {
```

This is one route where only `get` makes sense, I think.

server/utils/zip.js
tommoor: Yea, lets use current convention here.

server/utils/zip.js
```diff
@@ -4,12 +4,26 @@ import JSZip from 'jszip';
 import tmp from 'tmp';
 import unescape from '../../shared/utils/unescape';
 import { Collection, Document } from '../models';
+import { getImageByKey } from './s3';
+
+const ENABLE_PRIVATE_CONTENT = process.env.ENABLE_PRIVATE_CONTENT === 'true';
+const s3KeyRegex = /!\[.*\]\(\/api\/images\.info\?key=(?<key>.*)\)/gi;
+const imgageApiRegex = /(?<=!\[.*\]\()(\/api\/images\.info\?key=)/gi;
```
tommoor:
```suggestion
const imageApiRegex = /(?<=!\[.*\]\()(\/api\/images\.info\?key=)/gi;
```

Although, I wonder if these two regex could be combined :thinking:

server/utils/s3.js
```diff
@@ -112,3 +115,49 @@ export const uploadToS3FromUrl = async (url: string, key: string) => {
...
 };
+
+export const setS3ObjectAcl = () =>
+  ENABLE_PRIVATE_CONTENT ? 'private' : 'public-read';
+
+export const getUrlForImageProxy = (key: string) =>
+  `/api/images.info?key=${key}`;
+
+export const getSignedImageUrl = async (key: string) => {
+  const s3 = new AWS.S3({
```
tommoor: Seems like we could probably move the new `s3` object creation and the existing one https://github.com/outline/outline/pull/1137/files#diff-6445a4e1aa245e5905cb555afdd8ac6bR82-R87 out to the module context?

server/utils/s3.js
```diff
@@ -112,3 +115,49 @@ export const uploadToS3FromUrl = async (url: string, key: string) => {
...
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+  });
+  invariant(AWS_S3_UPLOAD_BUCKET_NAME, 'AWS_S3_UPLOAD_BUCKET_NAME not set');
+
+  const params = {
+    Bucket: process.env.AWS_S3_UPLOAD_BUCKET_NAME,
+    Key: key,
+    Expires: 900,
```
tommoor: Is this 900 seconds? 15 minutes…?

server/api/images.js
tommoor: I'm actually unclear how the auth is succeeding here, I haven't ran this branch locally yet – but it seems like no auth would be available without frontend changes?

tommoor: Thinking about the above suggestion some more it would also help with https://github.com/outline/outline/issues/1010 and form the foundation for file attachments. 

I can go ahead and do this work in a separate branch if you like.

HUSSTECH: I had it as a `get` initially, but it wasn't working. After digging, think it could be because of the `methodOverride` middleware, so I just switched it to `post`. After the initial mystery, I quite liked the uniformity of having everything transformed to a `post`! Do you know a way to override the `methodOverride` here, or is it a case of a whole new `Koa()` instance?

server/utils/s3.js
HUSSTECH: yep, I'll give that a try. I'm bringing baggage from Python where the thread-safety of boto3 is inconclusive!

tommoor: Yea, I think the existing example in the code didn't do any favors either – I'm 90% sure it should be fine defined in the module context.

server/api/images.js
tommoor: Got it, we can leave then 👍

server/utils/s3.js
HUSSTECH: yes the units are seconds. 900 is the default so I could either omit it entirely, or expose it as a config variable? `AWS_S3_VIEW_EXPIRY`..something like that?

server/api/images.js
HUSSTECH: The auth is going through because the image is being requested from the same origin, so all the session cookies etc are going through with the request.

I was going to say actually that having images registered in a table could go towards the file attachment issues I've seen floating about. You could have all uploaded "assets" registered in the new table, and all permission checks, cleanup sweeps, audits etc can happen from that. My questions would be:
- in light of the suggestion to have outline be private by default, should all image urls in the markdown be saved as `/image.info?id=` regardless, and then the proxy can just choose to redirect with the signed url or the public url depending on the setting? (even so, people may as well just let the signed urls happen anyway, unless as you say s3 is not being used)
- doing this does widen the scope of this PR, so how'd you want to proceed? If you think this warrants all going out at once, and not just doing the "simple" permissionsless auth check, then perhaps I can implement/address the rest of the comments here and then it can be merged this into a new branch rather than master to be the actual feature branch for private content.

server/utils/zip.js
HUSSTECH: I'll have a think, I'm sure it is possible to merge. I had 1 regex pulling out just the s3 key part as I need that to get the object from s3, and the second one was specifically to pull out the `/api/images.info?key=` substring for replacement. The complexity with the latter is that I needed to make sure to not accidentally match on someone writing `/api/images.info?key=` in the document text (who would, I know!), hence the lookbehind check to ensure I'm working on the image urls only.

Could be dropped altogether pending our below discussion on an assets table.

server/utils/s3.js
HUSSTECH: Guidance from AWS is to use as short a time as possible. So I think either I leave it as the default (therefore omit the parameter), or shorten it even further. There is no need for this to be long because once the image is loaded, it's in the page. And if the page is reloaded, then a new url is made anyway.

tommoor: Agreed, let's shorten it to 5 minutes.

server/api/images.js
tommoor: > in light of the suggestion to have outline be private by default, should all image urls in the markdown be saved as /image.info?id= regardless

I think I'd prefer that, it should be less edge-cases to deal with.

tommoor: > doing this does widen the scope of this PR, so how'd you want to proceed?

My plan is to start a separate PR with database storage of the image attachments, I'll add you to the review process – once that's approved and on master you can pull into this branch and it should be a relatively simple change to use the new attachment id instead of the path as the key and authenticate access.

tommoor: PR: https://github.com/outline/outline/pull/1144

compengsamit:
Any ETA, when this feature is going to be merged to master?

tommoor:
Unsure, if @HUSSTECH isn't able to take this over the line within a week or two someone on the maintaining team will 👍

HUSSTECH:
Probably got my wires crossed @tommoor, I was waiting for #1144 to go into master, I'm not a required/requested approver on that. I did look at it, and nothing stood out for me to comment on at the moment. I'm ready to wrap up the work for this, just need to know if this PR goes first or the other one :sweat_smile:

tommoor:
Aha! Okay, glad we communicated then – I'll work on landing the other PR to master

compengsamit:
Sounds awesome! 
This will really be helpful in securing the contents :)

server/api/users.js
```diff
@@ -131,7 +135,7 @@ router.post('users.s3Upload', auth(), async ctx => {
       asset: {
         contentType,
         name,
-        url,
+        url: `/api/images.info?id=${id}`,
```
HUSSTECH: this is a question I had. Here I'm returning the url in the form required to hit the image proxy endpoint, so that it can be written in the markdown as the proxy link, but it is being saved in the `Attachments` table as the full url. This is working fine as it is here, but just an open question on whether this is the best way to do it? i.e. what url should be saved vs returned to the front-end and is it different depending on what mode we're in with the `AWS_S3_ACL` value.

server/utils/s3.js
```diff
@@ -78,12 +86,6 @@ export const publicS3Endpoint = (isServerUpload?: boolean) => {
 };
 
 export const uploadToS3FromUrl = async (url: string, key: string) => {
```
HUSSTECH: This method looks like it is only used for Avatar uploads at the moment. The potential issue here is that if I save it as a private acl, the avatars will not load. Couple of options:
1) Save avatars into the attachments table and return proxy urls like the other uploads
2) Save these avatars as special cases as public objects (for these specific objects only)

I'm leaning to option 2. Option 1 needs to rework a fair bit of stuff from adding an `Attachments.create` call, pulling team ID and user ID properties all the way through this and even potentially breaking a couple of the `NOT NULL` constraints in the attachments table. The downside to option 2 is that it won't allow users to fully lock down their *entire* s3 bucket from public access globally. But as long as `AWS_S3_ACL=private` the objects will be secure.

I think option 1 can eventually be done, but doing option 2 does not block the path for taking some more time in thinking how to implement a "proper" solution for everything which goes into the s3 bucket.

server/utils/zip.js
```diff
@@ -3,13 +3,28 @@ import fs from 'fs';
...
+import { Attachment, Collection, Document } from '../models';
+import { getImageByKey } from './s3';
 
 async function addToArchive(zip, documents) {
   for (const doc of documents) {
     const document = await Document.findByPk(doc.id);
+    let text = unescape(document.text);
 
-    zip.file(`${document.title}.md`, unescape(document.text));
+    const attachments = await Attachment.findAll({
```
HUSSTECH: This was quite nice as it let me get rid of the regex and use exact replace methods. Should be better going forward.

server/api/images.js
```diff
@@ -0,0 +1,20 @@
...
+  const { id } = ctx.body;
+  ctx.assertPresent(id, 'id is required');
+  const attachment = await Attachment.findByPk(id);
+  const redirectUrl =
+    attachment.acl === 'private'
+      ? await getSignedImageUrl(attachment.key)
+      : attachment.url;
+  ctx.redirect(redirectUrl);
+});
```
HUSSTECH: three points that I can think of here:
1) not major at the moment, but can this become a more generic `attachments.info` endpoint, as it does not have to just apply to image proxies now?
2) permissions: at the moment no checks happening, it is all passive. i.e. if you are viewing the document then you can get to the images. Where if any should some checks go?
3) shared documents. I think I just totally missed this use-case. Lowest effort outcome, if you are using `AWS_S3_ACL=private` then essentially you can't share publicly. Highest effort outcome, similar to the `documents.info` api, we could do some share checks and remove the `auth()` requirement.

server/api/users.js
```diff
@@ -87,16 +87,20 @@ router.post('users.s3Upload', auth(), async ctx => {
   ctx.assertPresent(size, 'size is required');
 
   const { user } = ctx.state;
+  const id = uuid.v4();
```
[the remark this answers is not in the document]
tommoor: It's probably cleanest to just let the database generate the `id`, this would be in-line with what we do elsewhere.

server/api/users.js
```diff
@@ -88,17 +87,20 @@ router.post('users.s3Upload', auth(), async ctx => {
...
   const key = `uploads/${user.id}/${s3Key}/${name}`;
   const credential = makeCredential();
   const longDate = format(new Date(), 'YYYYMMDDTHHmmss\\Z');
   const policy = makePolicy(credential, longDate);
   const endpoint = publicS3Endpoint();
   const acl = process.env.AWS_S3_ACL || 'private';
-  const url = acl === 'private' ? proxyS3Url(key) : `${endpoint}/${key}`;
+  const url = `${endpoint}/${key}`;
 
   await Attachment.create({
```
tommoor: `const attachment = await Attachment.create({` above to get the attachment record back with the `id` rather than passing it in below.

server/utils/zip.js
tommoor: Nice! This does mean that only image attachments created from this point on would be returned inside the archive, any existing attachments would not – but that's fine by me.

server/utils/zip.js
```diff
@@ -3,24 +3,24 @@ import fs from 'fs';
...
-          const key = match.slice(21);
-          await addImageToArchive(zip, decodeURI(key));
-        }
-        text = text.replace(proxyS3UrlRegex, match => match.slice(21));
+    const attachments = await Attachment.findAll({
+      where: { documentId: document.id },
+      attributes: ['id', 'key'],
+    });
+
+    if (attachments !== null) {
```
tommoor: I'm not sure this would ever be `null` – it would be an empty array if there are no attachments, did you run into this?

server/utils/zip.js
```diff
@@ -3,24 +3,24 @@ import fs from 'fs';
...
+      for (const attachment of attachments) {
+        const key = attachment.key;
+        await addImageToArchive(zip, decodeURI(key));
+        text = text.replace(`/api/images.info?id=${attachment.id}`, key);
```
tommoor: This string appears a couple of times, `/api/images.info?id=${attachment.id}` – how about we move it to a model `getter` so that we can simply do something like `attachment.redirectUrl`

server/api/images.js
tommoor: 1. Agree, how about `attachments.redirect`? We should reserve `info` for a json response with the attachment details.

2. After we query the attachment model on L12, query it's document with the current user, something like:

```
const user = ctx.state.user;
const document = await Document.findByPk(id, { userId: user.id });
authorize(user, 'read', document);
```

I think for now we can say that if you have access to view the document then you have access to read it's attachments without creating a new, separate authentication policy for attachments specifically. 

3. Public shares do complicate things a little. I'm happy to take this one if you like – my current gut is to have `presentDocument` (the document presenter) rewrite the text contents with the signed urls before it sends them down to the client if the document is public.

server/utils/zip.js
HUSSTECH: yes, it is much better with getters....shame they don't do async getters as I found out in the middle of trying to do one to get the signed s3 url!

server/utils/zip.js
HUSSTECH: you're right, would be an empty array, so no need even for an `if` test on the length as the for loop would just not run.

server/api/users.js
HUSSTECH: done

server/utils/zip.js
HUSSTECH: probably also fine because anything before this change would not have an entry in the new attachments table anyway.

server/api/images.js
HUSSTECH: 1) done as per suggestion
2) done as per suggestion
3) I will take your offer of help! Will wrap up anything else you need for this PR and get it ready for merging

0rax:
Any news on when this would hit master ?

I am currently trying out this feature in a fork, though I have some issue with signedUrl generation (upload is fine but retrieval of images are failing).

```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRequest</Code><Message>The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.</Message><RequestId>[READCTED]</RequestId><HostId>[REDACTED]</HostId></Error>
```

Would love to help with some logs if example if needed !

HUSSTECH:
Hey @0rax, that's a great catch! I think this is a signature v2 vs v4 thing. Everything should be switched to v4 by now as AWS plans to retire v2 later this year. From the [blog post](https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/) about the topic, is your s3 bucket in any of these regions below by any chance?

> The Europe (Frankfurt), US East (Ohio), Canada (Central), Europe (London), Asia Pacific (Seoul), Asia Pacific (Mumbai), Europe (Paris), China (Ningxia), Europe (Stockholm), Asia Pacific (Osaka Local), AWS GovCloud (US-East), and Asia Pacific (Hong Kong) Regions were launched after 2013, and support SigV4 but not SigV2. If you have code that accesses S3 buckets in that region, it is already making exclusive use of SigV4.

I have pushed an update which configures the AWS s3 client to use signature v4. You can update your fork, or just apply the chance yourself locally to confirm the fix.

0rax:
Hey @HUSSTECH, I can confirm that my bucket is in the `eu-west-3` (Paris) region ! That's why then.

I just updated and will try to deploy the fix ASAP. Will edit this message with my findings !

Thanks a lot for the catch !

tommoor:
Hmm @0rax that should be expected, the avatar and logo should be public in the db and on S3 still, the feature only applies to document attachments in this first version. Are you suggesting they were also uploaded with private permissions? Perhaps the entire bucket was marked as private?

0rax:
> Hmm @0rax that should be expected, the avatar and logo should be public in the db and on S3 still, the feature only applies to document attachments in this first version. Are you suggesting they were also uploaded with private permissions? Perhaps the entire bucket was marked as private?

To be completely honest, I was expecting to be able to have my whole bucket set to private and have all request proxy-ed through this API layer though it does make sense that the logos and profile pic need to be public in some way.

I will update my bucket accordingly, sorry about that !

BTW, this feature works great for our internal documentation ! Happy to see that much work into this opensource product :)

tommoor:
> I will update my bucket accordingly, sorry about that !

Good to have confirmation. We should split the buckets so you _can_ mark the whole bucket as private for safety.

server/models/User.js
```diff
@@ -128,14 +128,21 @@ const uploadAvatar = async model => {
...
-    );
-    if (newUrl) model.avatarUrl = newUrl;
+    try {
+      const newUrl = await uploadToS3FromUrl(
+        avatarUrl,
+        `avatars/${model.id}/${uuid.v4()}`,
+        'public-read'
+      );
+      if (newUrl) model.avatarUrl = newUrl;
+    } catch (err) {
```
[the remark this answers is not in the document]
tommoor: Okay

server/test/support.js
```diff
@@ -76,12 +76,26 @@ const seed = async () => {
   await document.publish();
   await collection.reload();
 
+  const attachment = await Attachment.create({
```
tommoor: Could we move this into the `factories` file – you'll see the pattern in there. This support method is somewhat deprecated as we're trying to be more specific about which test data is created to make the suite run as fast as possible 😄

server/utils/s3.js
```diff
@@ -112,3 +121,33 @@ export const uploadToS3FromUrl = async (url: string, key: string) => {
     }
   }
 };
+
+export const getSignedImageUrl = async (key: string) => {
```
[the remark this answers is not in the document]
tommoor: I'm wondering if handling `fake-s3` here would also fix the tests that are proving problematic? 🤔 

> An idea is something similar to using an isDocker Boolean (as in the method above this) to skip generating the signed url and just return ${publicS3Endpoint()}/${key}.

This seems like the right thing to do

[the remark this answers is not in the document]
HUSSTECH: added the `isDocker` flag, this now solves serving of locally uploaded images in fake-s3 being shown as it is now serving from `http://localhost:4569/bucket/key`.

However it does still require `http://localhost:4569` to allow the upload in the first place. (Seems like this would have always been the case before this branch anyway.)

server/test/support.js
HUSSTECH: moved to `factories`. Think I followed the pattern correctly, it did make the test a bit more verbose but I assume that is the point.

server/utils/s3.js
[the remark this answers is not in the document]
HUSSTECH: Unfortunately does not fix the redirect tests. I think that is a problem with the test server/library not exposing redirect responses to allow us to assert on. Since the other "first-time" responses of `401` and `403` pass their tests fine.

server/api/attachments.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+  it('should return a redirect for an attachment belonging to a document user has access to', async () => {
+    const user = await buildUser();
+    const attachment = await buildAttachment({
+      teamId: user.teamId,
+      userId: user.id,
+    });
+    const res = await server.post('/api/attachments.redirect', {
+      body: { token: user.getJwtToken(), id: attachment.id },
+    });
```
tommoor:
```suggestion
      redirect: 'manual'
    });
```

server/api/attachments.test.js
```diff
@@ -0,0 +1,84 @@
...
+    const attachment = await buildAttachment({
+      teamId: user.teamId,
+      userId: user.id,
+      documentId: document.id,
+    });
+
+    const res = await server.post('/api/attachments.redirect', {
+      body: { token: user.getJwtToken(), id: attachment.id },
+    });
```
tommoor:
```suggestion
      redirect: 'manual'
    });
```

HUSSTECH:
> Figured it out – these should be the final changes you need to get the test suite passing

Yessss!!!!

server/utils/s3.js
HUSSTECH: current implementation is uploading avatars as `public-read`

server/utils/s3.js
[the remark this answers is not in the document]
HUSSTECH: @tommoor think a quick win is to have `app.js` like so:
```javascript
      connectSrc: compact([
        "'self'",
        process.env.AWS_S3_UPLOAD_BUCKET_URL,
        process.env.NODE_ENV !== 'production' && 
       process.env.AWS_S3_UPLOAD_BUCKET_URL.replace(
          's3:',
         'localhost:'
       ),
        'www.google-analytics.com',
      ])
```
with the `_.compact`,  when not needed the element should be removed.

Then, (fingers crossed) that is it for this PR.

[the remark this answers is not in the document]
tommoor: I'm going to merge and then add that tweak.

klener89:
I just ran into an error on an `self-hosted` s3 instance, where the update of the avatar is not working properly due to fixed public ACL.

Shouldn't the upload of the avatar also be handled with the new api endpoint for attachments instead of the ImageUpload?https://github.com/outline/outline/blob/82433e02a01d7e4603d1a943f3c593bbc0a1be12/app/scenes/Settings/components/ImageUpload.js#L58

Issue #859, opened 2019-01-13
File Attachments


tommoor:
This has been requested several times recently. Initial plan would be to integrate with the two largest storage providers, _Google Drive_ and _Dropbox_, rather than building our own file storage service.

On a document you should be able to:

- Attach a file
- Pick an existing file from GDrive/Dropbox or upload
- See a list of files attached to doc
- See which docs have files attached
- Search for files attached to docs

**As always, please add reactions to this issue to vote for it to be implemented 😄**

kevalkothari:
Hi @tommoor  Count me for this feature :) 
It's very useful feature for attaching Sample Template, Small Code File or Other kind of example demos as part of Wiki Doc.

JoeyJanson:
@tommoor Is there any news about this issue? Need contributors or answers for questions before starting with this feature?

peiying16:
It is good to have versionning.

aam1r:
To scope this down for an initial MVP, I think it would be really cool to have file uploads with simple hyperlinks. 

I wouldn't expect it to do any inline embedding. With this approach, you can support any file extension.

Over time, inline embedding can be added for the more popular filetypes such as PDFs, Word Documents, etc.

victordariovera:
I'm using nextcloud as a google drive alternative, but selfhosted for privacy reasons. How I can integrate Nextcloud share links like Google Drive links?

tommoor:
I'm guessing that the url for every nextcloud installation is unpredictable and different? It would make building an embed a much more involved feat than typical embeds where we can regex match

victordariovera:
Maybe the "plugin" or integration can be configurable with nextcloud's domain. Or, maybe can make a get to a know test url in base domain detecting if it is a nextcloud instance. Every nextcloud installation expose a status endpoint for identification of version.

etehler:
Hi! We are using a self hosted Outline mainly because of the lack of good software hosted inside the EU (GDPR).

The only thing that really limits us is the fact that we cannot uploads PDF:s into Outline, and that we cannot embed any videos except from Vimeo/YouTube. A more flexible video-embed and uploading PDF documents into S3 would make self hosted Outline a nice GDPR compliant alternative.

acrob0t:
[...]

PR #1144, opened 2020-01-05
feat: Store image uploads as attachments in database

tommoor:
This PR provides the foundation for storing images and other file uploads in a private bucket.

Currently some of the data is stored in the `events` table but this is supposed to be an audit log, not a place to query for active data so it makes sense to create a new table for this concept with it's own schema and validation.

related https://github.com/outline/outline/pull/1137
related https://github.com/outline/outline/issues/1010
related https://github.com/outline/outline/issues/859

### DIFF
--- a/server/api/users.js
+++ b/server/api/users.js
@@ -76,7 +76,7 @@ router.post('users.update', auth(), async ctx => {
 });
 
 router.post('users.s3Upload', auth(), async ctx => {
-  let { name, filename, contentType, kind, size } = ctx.body;
+  let { name, filename, documentId, contentType, kind, size } = ctx.body;
 
   // backwards compatability
   name = name || filename;
@@ -100,6 +100,7 @@ router.post('users.s3Upload', auth(), async ctx => {
     size,
     url,
     contentType,
+    documentId,
     teamId: user.teamId,
     userId: user.id,
   });
--- a/server/api/users.js
+++ b/server/api/users.js
@@ -96,7 +96,7 @@ router.post('users.s3Upload', auth(), async ctx => {
   const url = `${endpoint}/${key}`;
 
   await Attachment.create({
-    name,
+    key,
     size,
     url,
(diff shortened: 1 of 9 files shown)


tommoor:
cc @HUSSTECH

server/api/users.js
```diff
@@ -76,29 +76,39 @@ router.post('users.update', auth(), async ctx => {
...
   const credential = makeCredential();
   const longDate = format(new Date(), 'YYYYMMDDTHHmmss\\Z');
   const policy = makePolicy(credential, longDate);
   const endpoint = publicS3Endpoint();
   const url = `${endpoint}/${key}`;
 
+  await Attachment.create({
+    name,
+    size,
+    url,
```
tommoor: I suspect we should also store the `path` – but it's a little redundant. Maybe store path + url, and make `name` a custom getter that extracts the last piece of the path?

server/api/users.js
```diff
@@ -76,29 +76,39 @@ router.post('users.update', auth(), async ctx => {
...
-  const key = `uploads/${ctx.state.user.id}/${s3Key}/${filename}`;
+  const key = `uploads/${user.id}/${s3Key}/${name}`;
   const credential = makeCredential();
   const longDate = format(new Date(), 'YYYYMMDDTHHmmss\\Z');
   const policy = makePolicy(credential, longDate);
   const endpoint = publicS3Endpoint();
   const url = `${endpoint}/${key}`;
 
+  await Attachment.create({
+    name,
```
tommoor: `acl` can be passed here on the https://github.com/outline/outline/pull/1137 branch – for now it defaults to `public-read`

server/api/users.js
tommoor: Changed to store `key` instead of `name`.

tommoor:
@HUSSTECH ready for integration into your branch 👍

PR #1145, opened 2020-01-05
fix: Don't set cookie domain when not using multiple subdomains

### LINKED ISSUE #1142: state cookie not set correctly for heroku domains
I'm trying to use the deploy to heroku button and run an instance of outline on the default herokuapp subdomain.

The URL looks like `https://my-app.herokuapp.com`. I think the outline code makes some assumptions that any subdomain is team-specific and not part of the base application URL.

https://github.com/outline/outline/blob/98dda567c222bcac48f2c2889f58376a73756aad/server/auth/slack.js#L30
Because the subdomain is stripped, the cookie attempts to be set on `.herokuapp.com` which is actually [blacklisted by browser vendors](https://devcenter.heroku.com/articles/cookies-and-herokuapp-com) due to its status as a hosting platform. 

Therefore, the cookie is unavailable to the my-app subdomain, and the result is a state mismatch error.

Instead of using the stripped domain, could we instead use the env var `BASE_URL`? I would imagine a similar issue would occur for the URL `my-app.co.uk`, so it's not just a heroku-specific problem

tommoor:
Allows Outline to work on heroku and other hosting provider subdomains.

closes #1142


thenanyu:
cool cool — I'll deploy this and test it when I get a chance

tommoor:
Did you want to try deploying these last tweaks @thenanyu ?

thenanyu:
@tommoor just redeployed, seems good

PR #1148, opened 2020-01-06
fix: Drafts appear in document insert search

### LINKED ISSUE #1147: Drafts should not show when searching to link from a document
- Create a document
- Add some text and hit CMD+K
- In the document search toolbar drafts should not show as results, they currently do.

tommoor:
closes #1147


app/scenes/Search/Search.js
```diff
@@ -188,6 +188,7 @@ class Search extends React.Component<Props> {
           limit: DEFAULT_PAGINATION_LIMIT,
           dateFilter: this.dateFilter,
           includeArchived: this.includeArchived,
+          includeDrafts: true,
```
tommoor: Drafts should continue to be returned in the regular search

PR #1157, opened 2020-01-12
perf: Reuse redis connections where possible

tommoor:
Does what it says on the tin – reuse redis connections between queues where possible. Should drop the total required connections from 19 -> ~10 per instance.


server/index.js
```diff
@@ -24,7 +21,12 @@ if (process.env.WEBSOCKETS_ENABLED === 'true') {
     cookie: false,
   });
 
-  io.adapter(socketRedisAdapter(process.env.REDIS_URL));
+  io.adapter(
+    socketRedisAdapter({
```
tommoor: Based on https://github.com/socketio/socket.io-redis#sentinel-example

server/utils/queue.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import Redis from 'ioredis';
+import Queue from 'bull';
+import { client, subscriber } from '../redis';
+
+export function createQueue(name: string) {
```
tommoor: Based on https://github.com/OptimalBits/bull/blob/28a2b9aa444d028fc5192c9bbdc9bb5811e77b08/PATTERNS.md#reusing-redis-connections

tommoor:
Prediction was exactly right here, redis connections dropped from 19 -> 10 per server process.

PR #1165, opened 2020-01-22
feat: Private attachments

tommoor:
Continuing the work in https://github.com/outline/outline/pull/1137


tommoor:
@HUSSTECH I'm not able to PR into yours because it's in a different repo, but I have added support for public attachments for avatar uploads here.

Just public shared documents to do.

tommoor:
...aaaand now with support for publicly shared documents. I think this covers the final pieces.

server/presenters/document.js
```diff
@@ -1,24 +1,46 @@
...
 import presentUser from './user';
 
 type Options = {
   isPublic?: boolean,
 };
 
+const attachmentRegex = /!\[.*\]\(\/api\/attachments\.redirect\?id=(?<id>.*)\)/gi;
+
+// replaces attachments.redirect urls with signed/authenticated url equivalents
+async function replaceImageAttachments(text) {
```
tommoor: Not sure if it's worth putting this utility method elsewhere? It's only used here so seems to make sense to co-locate.

HUSSTECH: it is pretty much encapsulated perfectly in here. No other "unit" is publicly shareable except for Documents, so I think it's good here. Also definitely much neater to have it in the `presenter` module.

PR #1167, opened 2020-01-23
WIP: fix: Color the label to be white to contrast

### LINKED ISSUE #1140: The "..." icon on document revisions is the incorrect color when selected
When selected the "..." icon should be white so that it contrasts with the background.

image

mcab:
Closes #1140.

---

hover inconsistencies

Currently, this PR changes the color of the icon to fill with white when the revision is considered active.

It doesn't, however, work well in cases of:
- [ ] clicking directly on the icon (the icon turns white, and then fills back to normal)
- [ ] click-holding onto another revision (the icon turns white before the revision is highlighted)
- [ ] clicking anywhere else in the UI asides from `RevisionList` (the icon fills back to gray)

At this rate, it makes me feel like another approach should be taken, like passing some status (`active`?) down to `Revision`, and then using a specifically styled `MoreIcon` and swap based on that condition. Any thoughts?

### DIFF
--- a/app/components/DocumentHistory/components/Revision.js
+++ b/app/components/DocumentHistory/components/Revision.js
@@ -43,7 +43,7 @@ class RevisionListItem extends React.Component<Props> {
           <StyledRevisionMenu
             document={document}
             revision={revision}
-            label={<MoreIcon color={theme.white} />}
+            label={<MoreIcon />}
           />
         )}
       </StyledNavLink>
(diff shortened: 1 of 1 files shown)


tommoor:
Hey Mark thanks for giving this a go – I agree with you, the history item needs to know if it's selected to be able to style correctly. We could use the `activeStyle` prop on the parent NavLink to add a custom class when it's active which would then override the styles of the "..." menu. Not ideal as it'd be nice to be more declarative within the React component, but it'd work.

tommoor:
Closing as part of cleanup as there was no followup here

PR #1193, opened 2020-02-27
fix: Nested document menu item appears where it shouldn't

### LINKED ISSUE #1191: "New nested document" appears as option on drafts
However it's not possible to create a nested document of a draft currently – it errors. So we should disable the item in this scenario until such functionality exists.

Reported via support.

tommoor:
fix: "Publish" item is disabled when it shouldn't be

closes #1191

### DIFF
--- a/app/menus/DocumentMenu.js
+++ b/app/menus/DocumentMenu.js
@@ -185,7 +185,7 @@ class DocumentMenu extends React.Component<Props> {
             </DropdownMenuItem>
           </React.Fragment>
         )}
-        {can.update && (
+        {can.createChildDocument && (
           <DropdownMenuItem
             onClick={this.handleNewChild}
             title="Create a nested document inside the current document"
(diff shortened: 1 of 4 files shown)


app/menus/DocumentMenu.js
```diff
@@ -185,7 +185,7 @@ class DocumentMenu extends React.Component<Props> {
             </DropdownMenuItem>
           </React.Fragment>
         )}
-        {can.update && (
+        {can.createChildDocument && (
```
tommoor: Introduced a new policy for this 👍

app/scenes/Document/components/Document.js
```diff
@@ -80,6 +80,10 @@ class DocumentScene extends React.Component<Props> {
     this.loadEditor();
   }
 
+  componentDidMount() {
+    this.updateIsDirty();
```
tommoor: Should check policy on mount

Issue #1199, opened 2020-03-03
Documentation for slack slash command setup on self hosted version of Outline


maryolivier:
## Description
Setting up a self hosted version of outline is pretty straightforward. Setting up the accompanying slack app, however, is not. There seems to be some distributed information throughout issues, but no one source.

tommoor:
I wrote up a rough guide here last week, let me know if that helps:
https://www.getoutline.com/share/be25efd1-b3ef-4450-b8e5-c4a4fc11e02a

bryanjos:
Hey @tommoor we got it working! The problem we had was that somehow we ended up with 2 entries in the teams table. One from google and one from slack. I moved the slackId from the slack one to the google one and things seem to be working as expected now

bryanjos:
One more question (that can maybe be cleared up in documentation), can you only have google oauth turned on and still use the slack integration? Are there specific slack environment variables we should set to turn on just the integration without turning on slack oauth?

tommoor:
> Are there specific slack environment variables we should set to turn on just the integration without turning on slack oauth?

You can use Google Auth with the Slack integration but there isn't currently a way to disable the authentication on the homepage but retain the slash command integration – they are both enabled by the existence of the environment variables.

bryanjos:
Thanks for the info. Would that be something you would be interested in a PR to alter? For instance, I think the `SLACK_VERIFICATION_TOKEN` and `SLACK_APP_ID` are what's needed for the integration and `SLACK_KEY` and `SLACK_SECRET` can be used to see if slack authentication should be turned on.

maryolivier:
@tommoor yes that helped us as @bryanjos said! documentation should also include the redirect URLs needed, bot token scopes, and user token scopes

enzolutions:
Hi @maryolivier && @tommoor for this guide

I think I did the integration using my self-hosted outline running in Heroku, and all looks connected as you can see in the following image.
<img width="776" alt="Screen Shot 2020-03-20 at 5 34 48 pm" src="https://user-images.githubusercontent.com/907914/77141629-b116ec00-6ad1-11ea-8308-fe5cd82d3d34.png">

One collection is private, but both post without problems in the designated channels in Slack.

I created a command `/handbook` instead `/outline`, but when I tried to search I got the following "error".

[...]

PR #1207, opened 2020-03-14
feat: Split up check for Slack environment variables

bryanjos:
This allows for the Slack slash command to be used without
needing Slack sign in enabled.

connects to #1199

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,15 +29,23 @@ if (process.env.AWS_ACCESS_KEY_ID) {
 }
 
 if (process.env.SLACK_KEY) {
-  ['SLACK_SECRET', 'SLACK_VERIFICATION_TOKEN', 'SLACK_APP_ID'].forEach(key => {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -29,15 +29,23 @@ if (process.env.AWS_ACCESS_KEY_ID) {
...
+
+if (process.env.SLACK_APP_ID) {
+  if (!process.env.SLACK_VERIFICATION_TOKEN) {
+    console.error(
+      `The SLACK_VERIFICATION_TOKEN env variable must be set when using the Slack slash command`
+    );
+    // $FlowFixMe
+    process.exit(1);
+  }
 }
```
tommoor: To be honest after reassessing the code `SLACK_APP_ID` is not needed for the slash command to work either, I think we can just remove this entire block

bryanjos: Cool! I'll make that update

PR #1208, opened 2020-03-16
fix: Absolute routes won't work when users have subdomains

panphora:
On the Editor page, after I first sign up for an account, the "integrations" link links to https://[mysubdomain].getoutline.com/integrations, which is a 404. 

The link needs to be the full link: https://www.getoutline.com/integrations

### DIFF
--- a/server/onboarding/editor.md
+++ b/server/onboarding/editor.md
@@ -12,6 +12,6 @@ If you’re comfortable writing markdown then all of the usual shortcuts are sup
 
 ## Blocks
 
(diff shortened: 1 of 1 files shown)


tommoor:
Ah thanks, the link should still work but I think it makes sense to make it absolute regardless

panphora:
I tested it. It didn't work for me.

Screen Shot 2020-03-15 at 11 54 56 PM

tommoor:
Ah, yea it's trying to route internally within react router but if you type the address in manually it works. 👍 . Yep this fix definitely makes sense.

Issue #1126, opened 2019-12-25
Timestamps should update while a page stays loaded


tommoor:
The `Time` component should include a shared timer that refreshes any mounted components once per minute so that the timestamps update inline without needing a full page re-render.

TommyJackson85:
I will work on this now.

TommyJackson85:
Where is the Time component? I tried searching it from the text editor and it doesn't seem to be appearing.

TommyJackson85:
I can't find the Time component. Was it removed from another pull request? I tried searching Time.

Any other suggestions on where to start with this issue?

tommoor:
https://github.com/outline/outline/blob/master/shared/components/Time.js

TommyJackson85:
I will try to fix it in a day or 2. If anyone wants to take over and create a PR they can but I would like to atleast try this if its atleast under no pressure to be done.

taylorlapeyre:
I think I can take this

PR #1214, opened 2020-03-19
fix: Update all Time components each second

taylorlapeyre:
Fixes https://github.com/outline/outline/issues/1126

My biggest concern here is the `setInterval` at the global level: once started, it will not stop until the app stops. That seems bad, but I'm not sure if it's actually that bad in practice.

### DIFF
--- a/shared/components/Time.js
+++ b/shared/components/Time.js
@@ -9,12 +9,10 @@ let callbacks = [];
 // This is a shared timer that fires every second, used for
 // updating all Time components across the page all at once.
 setInterval(() => {
-  for (let callback of callbacks) {
-    callback();
-  }
+  callbacks.forEach(cb => cb());
 }, 1000 * 60);
 
-function eachSecond(fn) {
+function eachMinute(fn) {
   callbacks.push(fn);
 
   return () => {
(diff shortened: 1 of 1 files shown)


shared/components/Time.js
```diff
@@ -4,18 +4,60 @@ import Tooltip from 'components/Tooltip';
...
+  state = {
+    autoUpdatingDateTime: this.props.dateTime
+  }
+
+  componentDidMount() {
+    this.removeEachMinuteCallback = eachMinute(() => {
+      this.setState({
+        autoUpdatingDateTime:
+          new Date(
+            Date.parse(this.state.autoUpdatingDateTime) + (1000 * 60)
```
tommoor: Well, we don't want to actually change the timestamp itself – otherwise it will keep moving forward as the component is mounted. We should want to make sure that the component re-renders so that `distanceInWordsToNow` is recalculated. 

I'd suggest changing `autoUpdatingDateTime` -> `key` and then each time the callback is triggered just increment `key` and ensure it's passed as the prop to `<time key={this.state.key}>` ?

taylorlapeyre: Ahh I see. Makes sense. We could also just do a `forceUpdate` every minute as well. That would eliminate the need for internal state altogether. What do you think?

tommoor: Sounds even better

taylorlapeyre: Cool, all set!

PR #1216, opened 2020-03-23
feat: Separate title from body

tommoor:
- `title` field in db finally becomes canonical source
- adds a document `version` for tracking internal document types
- JIT migrate document versions on load from database
- no longer require or store title in body text for version 1 and above
- displays view count and meta data under title

This work lines us up for a tagging interface and also a neater integration with the new [Prosemirror editor](https://github.com/outline/rich-markdown-editor/pull/150) as we won't need a separate schema to handle the required title.


thenanyu:
**Initial Product Feedback**
1. when I create a new doc, the title is filled and the cursor is not on the title line and the title is pre-filled with "untitled". Everyone does this a little differently but I think the title ought to be actually empty with a placeholder text. Where the cursor starts doesn't matter as much in that scenario. The reason is that it almost feels like changing the name lives on a different flow as opposed to being just another part of the editor.
1. when I go into edit mode from an existing document, the cursor isn't inside the editor. I have to focus the edit zone first. It feels like the change to edit mode isn't "done" yet. Practically I still have to place my cursor, but I think there's a psychological aspect of this that would make it feel more secure. It also makes a slightly larger distinction between edit and view mode
1. the default documents end up reading "2 views" when I see them, dunno if thats because I had views from before but they were just hidden or because it's seeded at 1 and my first view bumps that to 2

tommoor:
> when I create a new doc, the title is filled and the cursor is not on the title line and the title is pre-filled with "untitled".

Weird, this definitely shouldn't be the case – the old "Untitled" default was removed.

tommoor:
Looking at your screenshot the placeholder text isn't quite right either, are you sure you're running the latest version?

thenanyu:
> Looking at your screenshot the placeholder text isn't quite right either, are you sure you're running the latest version?

You're right, I wasn't up to date, forget item (1), it's good

app/components/PublishingInfo.js
```diff
@@ -60,7 +65,13 @@ function PublishingInfo({
...
       </span>
     );
-  } else if (publishedAt && (neverUpdated || showPublished)) {
+  } else if (createdAt === updatedAt) {
+    content = (
+      <span>
+        created <Time dateTime={updatedAt} /> ago
+      </span>
+    );
+  } else if (publishedAt && (publishedAt === updatedAt || showPublished)) {
```
thenanyu, line 74: in this else if clause, `publishedAt === updatedAt` can't possibly be true right?

server/models/Document.js
```diff
@@ -427,6 +433,26 @@ Document.addHook('afterCreate', async model => {
...
+
+Document.prototype.migrateVersion = function() {
+  // migrate from document version 0 -> 1 means removing the title from the
+  // document text attribute.
+  if (!this.version) {
+    this.text = this.text.replace(/^#\s(.*)\n/, '');
+    this.version = 1;
+    return this.save({ silent: true, hooks: false });
+  }
+};
```
thenanyu, line 454: The strategy is to add another case in these migration functions for the eventual prosemirror migration?

tommoor: If it's required, yep.

app/components/PublishingInfo.js
tommoor: It's true in the instance that the last action taken was to publish the document

thenanyu: oh I read this wrong, I thought it said `createdAt === updatedAt`

PR #1224, opened 2020-04-05
perf: Optimized Dockerfile

githubsaturn:
Previously "yarn build" was running on every container restart, making the container restart extremely slow. The build process should be part of the Docker build process, not the run process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@ COPY . $APP_PATH
 RUN yarn install --pure-lockfile
 RUN cp -r /opt/outline/node_modules /opt/node_modules
 
-CMD yarn build && yarn start
+RUN yarn build 
+
+CMD yarn start
 
 EXPOSE 3000


tommoor:
Definitely want to  make a change like this however there are a couple of gotchas that were brought up in a similar PR here:
https://github.com/outline/outline/pull/850

githubsaturn:
From the looks of it the concern brought up in #850 is around the dev variant of the Docker, right?

In that case, we shouldn't sacrifice the release version for the dev flow. If my understanding is correct and the concern is around the development flow, we can have a different Dockerfile for release and dev which is a common practice.

tommoor:
Hey @githubsaturn that's fair, perhaps splitting the dev dockerfile out makes it easier to move this forward.

tommoor:
Thanks, this PR has now been superseded and fixed by https://github.com/outline/outline/pull/1353

PR #1242, opened 2020-04-25
fix: Various extra scrollbars when not using mac-style overlaying scrollbars

### LINKED ISSUE #1241: Horizontal scrollbar on long pages.
Unsure whether or not this is an issue just affecting Windows and Linux users or not. It's a little bit of a nit-pick of mine. In documents that extend past the bottom of the screen, there is a horizontal scrollbar when there shouldn't be.

image
_This image is kind of a bad example because there are multiple collapsed scrollbars with tables inside of them._

image
_Here in Firefox_

image
_As well as here in Chrome_

To recreate, on a non-mac make or view a document that extends past the bottom of the screen. It doesn't go away with scaling, or by using a different browser.

Again, it's quite a nit-picky bug, but it does just make scrolling with a touchpad a little janky.

EDIT: It also doesn't go away by switching into Production mode, and is still present on [getoutline.com](https://getoutline.com/)
image

tommoor:
According to [caniuse](https://caniuse.com/#feat=viewport-units) 100vw not considering scrollbars is considered a layout bug in the browsers:

image

closes #1241

### DIFF
--- a/app/components/Layout.js
+++ b/app/components/Layout.js
@@ -142,7 +142,7 @@ const Container = styled(Flex)`
   background: ${props => props.theme.background};
   transition: ${props => props.theme.backgroundTransition};
   position: relative;
-  width: 100vw;
+  width: 100%;
   min-height: 100%;
 `;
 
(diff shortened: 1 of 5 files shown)


app/components/Layout.js
```diff
@@ -142,7 +142,7 @@ const Container = styled(Flex)`
   background: ${props => props.theme.background};
   transition: ${props => props.theme.backgroundTransition};
   position: relative;
-  width: 100vw;
```
tommoor, line 145: 🍖 See description ^

PR #1264, opened 2020-05-08
feat: Implement websocket authentication using api-token

jakoblorz:
When #1263 is merged, private software (soon public and OSS) I wrote will be unable to connect to the websocket using only the `apiToken`. Users will be required to enter their `accessToken` which is highly insecure (under the same argument of "possibly insecure in the future" as in #1263).

I propose that access to the websocket should also be granted by authenticating with the `apiToken`. The changes for that are implemented "adjacent" to the http authentication / token resolve functionality.

https://github.com/outline/outline/blob/48c87a1902f1b97b49d8b6d3bd351bbfc7745413/server/middlewares/authentication.js#L45

While this PR is purely motivated by my own needs, I do think that others may find this useful. Especially as it enables developers to react in near-realtime to changes in outline while web hooks are not available (yet).

### DIFF
--- a/server/middlewares/authentication.js
+++ b/server/middlewares/authentication.js
@@ -8,69 +8,79 @@ import addMonths from 'date-fns/add_months';
 import addMinutes from 'date-fns/add_minutes';
 import { stripSubdomain } from '../../shared/utils/domains';
 
-export default function auth(options?: { required?: boolean } = {}) {
-  return async function authMiddleware(ctx: Context, next: () => Promise<*>) {
-    let token;
+function extractTokenFromCtx(ctx: Context): any {
+  const authorizationHeader = ctx.request.get('authorization');
+  if (authorizationHeader) {
+    const parts = authorizationHeader.split(' ');
+    if (parts.length === 2) {
+      const scheme = parts[0];
+      const credentials = parts[1];
 
-    const authorizationHeader = ctx.request.get('authorization');
-    if (authorizationHeader) {
-      const parts = authorizationHeader.split(' ');
-      if (parts.length === 2) {
-        const scheme = parts[0];
-        const credentials = parts[1];
-
-        if (/^Bearer$/i.test(scheme)) {
-          token = credentials;
-        }
-      } else {
-        throw new AuthenticationError(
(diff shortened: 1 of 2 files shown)


tommoor:
I don't feel great about the repetition between this and the authentication middleware, especially for such sensitive logic. I think it'd make sense to pull it into a shared lib method?

jakoblorz:
@tommoor Will do, I am thinking of a reusable method in /server/middlewares/authentication.js.

server/middlewares/authentication.js
```diff
@@ -8,69 +8,79 @@ import addMonths from 'date-fns/add_months';
...
-        const credentials = parts[1];
-
-        if (/^Bearer$/i.test(scheme)) {
-          token = credentials;
-        }
-      } else {
-        throw new AuthenticationError(
-          `Bad Authorization header format. Format is "Authorization: Bearer <token>"`
-        );
+function extractTokenFromCtx(ctx: Context): any {
```
tommoor, line 11: I'd like to stick to the "getXForY" pattern on these method names

```suggestion
function getTokenForCtx(ctx: Context): ?string {
```

server/middlewares/authentication.js
```diff
@@ -8,69 +8,79 @@ import addMonths from 'date-fns/add_months';
...
+    // $FlowFixMe
+  } else if (ctx.body && ctx.body.token) {
+    return ctx.body.token;
+  } else if (ctx.request.query.token) {
+    return ctx.request.query.token;
+  } else {
+    return ctx.cookies.get('accessToken');
+  }
+
+  return null;
```
tommoor, line 36: Would be fine as…

```suggestion
  return;
```

server/middlewares/authentication.js
```diff
@@ -8,69 +8,79 @@ import addMonths from 'date-fns/add_months';
...
+}
+
+export async function extractUserFromToken(token: any): Promise<User> {
```
tommoor, line 39:
```suggestion
export async function getUserForToken(token: string): Promise<User> {
```

PR #1183, opened 2020-02-22
WIP: feat: Integrated passport.js for authentication

jakoblorz:
Following #1120, here is a WIP to integrate [`passport.js`](http://www.passportjs.org/). 

passport.js supports "sessionless" authentication. By doing that, passport.js acts as a boarding library handling the data-exchange, so the currently login, logout and session processes can be kept as they are. No additional cookies are introduced as well.
If we were to use passport.js like it is shown in most examples, it would act more as a complete serverside state manager, which would require additional changes not only on the serverside but also on the client. E.g. the `accessToken` cookie would be replaced by a `koa-session` managed cookie.

**If you want to support this PR, please have a look at https://github.com/outline/outline/pull/1183#issuecomment-626354357**

### Changes

- [x] Integrate `passport.js` into the Koa stack. Preferably without adding additional cookies or changing current login-flows
- [x] Migrate the Slack OAuth2 flow to the passport module. @MatheusRV suggested [passport-slack](http://www.passportjs.org/packages/passport-slack/)
- [x] Migrate the Google OAuth2 flow to the passport module

Depending on the scope set by @tommoor for this PR:
- [x] Dynamic strategy loading from `./services` or other folder/package.
- ~Database adjustment to replace vendor specific fields (such as `slackId` etc) with generalized fields. A field such as `service` already identifies the vendor.~

### Changelog
- `server/index.js`: Add [`koa-passport`](https://www.npmjs.com/package/koa-passport) to the stack.
- `server/utils/authorize.js`: Create wrapped `passport.authorize` to capture authentication results in an "awaitable" way
- `server/auth/slack.js`: Configure `passport-slack` to be used as an authentication strategy. Migrate `/slack` fully and `/slack.callback` partially to be handled by passport.

### Possible Features / Changes in this or following PRs
- Clean OAuth2 Interfaces / APIs, which would enable users to define (maybe even dynamically-loaded) modules for authentication (like in the `services/` package). https://github.com/outline/outline/pull/939#issuecomment-487390892
[...]

PR #1266, opened 2020-05-11
feat: Gitea as Auth Provider

techknowlogick:
This PR adds the ability to use any Gitea installation as an auth provider. Gitea is a tool that allows for self-hosting of git repos. More info can be found here: https://gitea.io/en-us/

Here is a screenshot of the login experience with Gitea
Screen Shot 2020-05-11 at 4 00 05 PM

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,6 +38,22 @@ if (process.env.SLACK_KEY) {
   }
 }
 
+if (process.env.GITEA_KEY) {
+  if (!process.env.GITEA_SECRET) {
+    console.error(
+      `The GITEA_SECRET env variable must be set when using Gitea Sign In`
+    );
+    // $FlowFixMe
+    process.exit(1);
+  }
+  if (!process.env.GITEA_URL) {
+    console.warn(
+      `The GITEA_URL env variable is not set when using Gitea Sign In and defaulting to https://gitea.com`
+    );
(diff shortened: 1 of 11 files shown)


techknowlogick:
Title and description have been updated to have more info, hopefully the "more information needed" label can be removed now. I've also pushed a new commit that resolves the eslint issues.

techknowlogick:
@tommoor would you be able to provide some guidance on what information I could provide to satisfy the bot?

tommoor:
Hey, thanks for contributing  – code-wise it's definitely all in the right vein and matches the existing implementations!

There is currently a [PR in progress](https://github.com/outline/outline/pull/1183) by several folks that moves the authentication over to passport-js. I think this is a more scalable solution for adding additional auth providers, as it seems like everyone wants something different.

As the maintainer I don't plan to merge this PR as the maintenance burden is too much (this is why we explicitly ask that folks open an issue before submitting code changes), particularly for a service that no-one has requested before. I'd encourage you to maintain your fork, which should be easy to do until the passport-js changes are merged.

Hummer12007:
Thanks for the PR! It works like a charm. @techknowlogick Do you mind me integrating it into my personal MIT-licensed fork at https://github.com/Hummer12007/outline ?

PR #1271, opened 2020-05-17
chore: Improve performance in dev by running Node/Yarn outside of docker

tommoor:
You may have to update your `.env` to point to localhost as in the sample changes, but other than that it should "just work".

One of the points raised in this discussion thread:
https://github.com/outline/outline/discussions/1269

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -6,9 +6,9 @@
 SECRET_KEY=generate_a_new_key
 UTILS_SECRET=generate_a_new_key
 
-DATABASE_URL=postgres://user:pass@postgres:5432/outline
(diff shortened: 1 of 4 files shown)


tommoor:
cc @gempain

gempain:
(whoops, didn't think it'd actually accept my review 😛 , sorry !)

tommoor:
I'm thinking I might change all the ports away from non-standard ones, because anyone that's running postgres/redis locally on their mac already (seems very possible) will have conflicts

gempain:
I would advise against it and leave the default ports. Most people would expect those ports. Most generally, it's better to leave it to end users to customize the ports how they want it IMO.

PR #1280, opened 2020-05-22
fix: Content Security Policy font-src

anteprimorac:
Because `font-src` CSP was not defined, it was falling back to `default-src` which only allowed loading fonts from the same origin. The proposed rule allows loading fonts from any origin and inline `data:`.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -137,6 +137,7 @@ app.use(
       ],
       styleSrc: ["'self'", "'unsafe-inline'", 'github.githubassets.com'],
       imgSrc: ['*', 'data:', 'blob:'],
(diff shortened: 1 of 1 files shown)


tommoor:
What is the reason for this proposed change?

anteprimorac:
@tommoor sorry for taking your time, I got a bunch of CSP issues regarding this. I thought that this application was loading external assets, but it seems like some browser extension was injecting external fonts.

tommoor:
No worries, it sounds like the CSP was doing it's job well then 😆

anteprimorac:
@tommoor yeah 😁, probably the first time

Issue #622, opened 2018-03-24
Reordering Documents in Collection


tommoor:
Requested by Ada Support.

The ability to drag and drop to reorder/move documents in a collection.

Related https://github.com/outline/outline/issues/517

tommoor:
This came recommended and looks great if we restrict the sidebar to just reordering: https://github.com/atlassian/react-beautiful-dnd

To move a document to another parent you'd have to use the existing move functionality, that seems reasonable?

jasonquiz:
Would be great if we could drag it to another Collection. I find myself trying to do this with out thinking about it and then realize it doesn't work.

natefduncan:
My team and I have started using Outline and this is a feature that we'd love to be able to use. I know it's probably low on the priority list, but just wanted to let you know that there's more interest out there.

thomasmalley:
I think #596 might be related somehow ✌🏼️

anteprimorac:
@tommoor any progress on this? If you didn't start building this, I am happy to contribute.

tommoor:
No meaningful progress. The API largely supports the functionality already, there is an `order` property that can be sent. 

Previously I've thought that the sidebar UI component would pretty much need rebuilding from scratch to support drag-and-drop, but that is mainly if we also need to support dragging documents between collections and dragging nested documents to different levels. Whilst that would definitely be nice perhaps the pragmatic path forward is to just add reordering first.

anteprimorac:
I agree we can iterate on this later to add the ability to move the document to a different collection.

Do you have a workflow, like adding special labels and assigning, set in place for managing issues?

tommoor:
I can assign you if you prefer, I'd say 95% of the time folks offer to help but that a PR never materializes – the curse of open source.

tommoor:
This may work, it has no dependencies: 
https://github.com/tajo/react-movable

anteprimorac:
@tommoor I think `react-beautiful-dnd` is a better choice in the long run because it supports moving items between two or more droppable lists.

I have started working on this, actually already have a working prototype for the UI part. I am going to create a WIP PR soon.

MatheusRV:
Are you considering being able to reorder the collections?

tommoor:
Yes, we should allow that for sure, the strategy will have to be different from documents though as the entire document structure for a collection is stored in one place it's easy to do an atomic update – not so much for collections.

[...]

PR #1283, opened 2020-05-22
feat: Reordering documents in collection

anteprimorac:
This PR implements #622.

- [x] sort collection documents
- [x] sort document children
- [x] improve animation by updating UI before server response
- [x] moving document in the hierarchy
  - [ ] ~display placeholder where the document would be dropped~
  - [x] move document anywhere in the hierarchy of the same collection
    - [x] disable moving the document to itself or child droppable
    - [x] expand document hierarchy when dragging over it
    - [x] allow dropping to document that currently doesn't have any child document
  - [x] move the document to another collection
    - [x] expand collection hierarchy when dragging over it
- [ ] improve destination detection and dragging animations


server/presenters/collection.js
```diff
@@ -34,8 +24,7 @@ export default function present(collection: Collection) {
   };
 
   if (collection.type === 'atlas') {
-    // Force alphabetical sorting
-    data.documents = sortDocuments(collection.documentStructure);
+    data.documents = collection.documentStructure;
```
tommoor, line 27: Need to make L23 `documents: []`, and then this something like:

```
if (collection.type === 'atlas' && collection.documentStructure) {
  data.documents = collection.documentStructure;
}
```

The sort that was here previously was converting the null to an empty array.

tommoor:
This is great! Definitely need to figure out how to prevent the jittering in the frontend, but really psyched to see it working – you can even have two tabs open and the existing websocket messages  automatically keep the sidebars in sync 😉

server/presenters/collection.js
anteprimorac: @tommoor thanks for pointing on this. I wasn't aware of that behaviour. It should be resolved now in 660e968

anteprimorac:
@tommoor I've just pushed applying reordering of documents to the UI before sending a request to the server.

The only feature that needs to implement is moving documents between parent documents or collections. But that should be straightforward because the reordering logic on the client and server-side support it already.

tommoor:
> I've just pushed applying reordering of documents to the UI before sending a request to the server.

Nice! Works well. One thing I noticed is that in this situation where you have nested documents when you drag the child documents it will often drag the outer "Documentation" – I think dragging anywhere "inside" should not drag the out doc. Does that make sense?

image

anteprimorac:
@tommoor that make sense and that's how it's supposed to work, but I think it might be a styling issue because the only way that could happen is that keydown event gets propagated to the parent component.

I've pushed an implementation for moving documents in the hierarchy. But it needs more work. The main thing that bothers me is disabling moving document into itself. I will update the PR description to reflect all issues and missing features I found.

anteprimorac:
The implementation of displaying a placeholder where the dragged document would be dropped will be complicated because of `react-beautiful-dnd` limitations. There are some community examples that implement this feature on top of the library but none of them works out of the box.

tommoor:
Hey @anteprimorac – really excited for the potential of this feature, is there anything your waiting on from me as a maintainer here? If you think you won't realistically get time to take this over the finish line let me know and I'll see if we can get other engineers to finish those final TODO's.

Many thanks!

anteprimorac:
Hey @tommoor, thank you for bringing this up. Unfortunately, I didn't have time to finish all TODO's listed in the description, but hopefully, in the next few weeks, I would have time for this.

tommoor:
Please note this recently merged PR (https://github.com/outline/outline/pull/1309) has created a lot of conflicts, but they're easily resolved by following the same steps detailed in the PR in this branch before updating from the default.

anteprimorac:
@tommoor all conflicts should be fixed, and all features I've listed in the description of this PR are implemented now. But currently, the dragging UX is broken because the distinction between droppable components is very small and in some situations, it's hard to drop the item where you want. So, that's currently my main focus.

tommoor:
I'd agree with your assessment, particularly two things stand out:

- Trying to drop into a nested doc is very finicky, there seems to be a very small amount of vertical space where you can drop
- There are a couple of spots you can get in where it will just flicker back and forth.
- Would it be a heavy lift to show a blue line where the doc will be dropped?

anteprimorac:
It's doable, but we will have to position it on our own because the library doesn't have support for it, but hopefully, I've found a few community examples. I think it's worth doing it because it will improve UX a lot and give us better visibility into other UX issues.

tommoor:
Alternatively a "ghosted" version of the document would be good too. Appreciate your willingness to get this to the appropriate level of polish 💪

holtmeister:
Any progress on this one? Am really keen to see it in action. Unfortunately I'm an end user and not a programmer tho :/

tommoor:
Having seen the difficulty of implementing the ghost I'm not so fussed about that, really gotta just fine tune those drop areas though – I'm hoping to take a look this week

icrotz:
@tommoor Hello, do you have any news on this PR ? Can we help on that ?

Thanks !

tommoor:
Yes, @thenanyu is currently rebuilding the sidebar to allow this to be finished – once we tried to get into the details of drag and drop between nested documents it became apparent that the current way it's built was not going to ever work well with `react-beautiful-dnd`.

That work is in progress now.

tommoor:
@anteprimorac wanted to say a heartfelt thanks for the work on this PR, it pushed us forward and helped us to negotiate this work. In the end we had to rebuild the sidebar to not use the nested structure at all in order for dnd to work as users expect which made this PR unworkable.

You can see how it ended up here: https://github.com/outline/outline/pull/1722

anteprimorac:
@tommoor thank you, the nested structure was the main issue with dnd UX. I had started experimenting with `react-dnd` library that is more flexible than `react-beautiful-dnd` but I didn't have enough bandwidth to finish it.

Glad to see that you've managed to add this feature to the project.

Issue #1287, opened 2020-05-27
Cannot read property id of null in groups API endpoint


tommoor:
Most likely this is a deleted user that's still part of a group not being accounted for.

![image](https://user-images.githubusercontent.com/380914/82976237-64e59d80-9f93-11ea-99f4-c87efcf60e0e.png)

PR #1288, opened 2020-05-27
fix: Server error when listing memberships for group with deleted user

tommoor:
Ideally this would be solved at the query level, but adding `required: true` on the GroupUser association causes us to bump into [this old Sequelize bug](https://github.com/sequelize/sequelize/issues/9869) – it generates an invalid query.

closes https://github.com/outline/outline/issues/1287

### DIFF
--- a/server/api/groups.js
+++ b/server/api/groups.js
@@ -40,9 +40,12 @@ router.post('groups.list', auth(), pagination(), async ctx => {
     data: {
       groups: groups.map(presentGroup),
       groupMemberships: groups
-        .map(g => g.groupMemberships.slice(0, MAX_AVATAR_DISPLAY))
+        .map(g =>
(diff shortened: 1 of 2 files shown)


server/api/groups.js
```diff
@@ -42,6 +42,7 @@ router.post('groups.list', auth(), pagination(), async ctx => {
       groupMemberships: groups
         .map(g => g.groupMemberships.slice(0, MAX_AVATAR_DISPLAY))
         .flat()
+        .filter(gM => gM.user)
```
thenanyu, line 45: I think the filter needs to come before the slice — we want to display the first X valid group memberships

tommoor: You're not wrong, this was just easier 😛  – I'll change it.

PR #1297, opened 2020-06-04
fix: Add optional override for internal S3 URL

### LINKED ISSUE #1296: No support for different Internal S3 URL
When you host the S3ish (e.g. minio) next to outline, you need to access it via a local URL (at least in my setup): `http://minio:9000/` - but changing the `AWS_S3_UPLOAD_BUCKET_URL` would break the frontend.

How about adding an optional `AWS_S3_UPLOAD_BUCKET_URL_INTERNAL` that allows to override?

This would also remove the need for the weird hardcoded replacement for the docker setup.

TeNNoX:
Closes #1296

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -44,9 +44,9 @@ SENTRY_DSN=
 AWS_ACCESS_KEY_ID=get_a_key_from_aws
 AWS_SECRET_ACCESS_KEY=get_the_secret_of_above_key
 AWS_REGION=xx-xxxx-x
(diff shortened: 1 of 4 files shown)


TeNNoX:
Tested & working on my dev and production deploy.

Only thing I'm not sure about is why there was the added 's3/' for docker setup. Works for me without it.

tommoor:
We discussed in the issue and decided not to merge, but thanks for your efforts – always appreciated.

PR #1299, opened 2020-06-09
fix: Make MenuItem Links work in Developer Page  (API)

moekhalil:
Menu items pointed to anchors links that were non-existent. 

Adding id's to proper sections (h2 tags) for anchor links in the menu to work.


tommoor:
Great, thanks for the fix!

PR #1302, opened 2020-06-11
fix: Suspended users showing in group and collection member management

### LINKED ISSUE #1301: Suspended users show up when adding to group
**To Reproduce**
1. Suspend a user
2. Go to a group
3. "Add members"
4. Search for the users name, they will appear _and_ not be marked as suspended

tommoor:
closes #1301

### DIFF
--- a/app/scenes/Settings/People.js
+++ b/app/scenes/Settings/People.js
@@ -40,6 +40,10 @@ class People extends React.Component<Props> {
     this.inviteModalOpen = false;
   };
 
+  fetchPage = params => {
+    return this.props.users.fetchPage({ ...params, includeSuspended: true });
+  };
+
   render() {
     const { auth, policies, match } = this.props;
     const { filter } = match.params;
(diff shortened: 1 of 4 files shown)


app/scenes/Settings/People.js
```diff
@@ -40,6 +40,10 @@ class People extends React.Component<Props> {
     this.inviteModalOpen = false;
   };
 
+  fetchPage = params => {
+    return this.props.users.fetchPage({ ...params, includeSuspended: true });
```
tommoor, line 44: 🍖  Because suspended users are no longer included by default we need to explicitly request them on the people management screen.

Note to self: This data loading is due a revamp.

app/stores/UsersStore.js
```diff
@@ -25,6 +25,11 @@ export default class UsersStore extends BaseStore<User> {
     return filter(this.orderedData, user => user.isSuspended);
   }
 
+  @computed
+  get activeOrInvited(): User[] {
```
tommoor, line 29: `active()` filters out invited users too, and we want them to be shown on these screens so admins can organize people before they join.

Issue #1308, opened 2020-06-19
Empty lines cause extraneous backslashes when accessing raw text 


alexieyizhe:
When adding empty lines in the editor, they cause two backslashes `\\` to appear when grabbing the document through the Outline API.

This is the raw text output:
![image](https://user-images.githubusercontent.com/17508679/85185511-38efcc00-b262-11ea-994d-cea9f65523f0.png)

As far as I can tell, this isn't a quirk of Markdown to display new lines, but I may be wrong!

**To Reproduce**
1. Add empty lines in web editor through <kbd>Enter</kbd>
2. Fetch the document with `/api/documents.info?shareId=X`
3. Get raw text on the `response.data.text` property

**Expected behavior**
Empty lines should not contain any extra characters.

**Outline (please complete the following information):**
- Install: self-hosted
- Version: 0.43.2

tommoor:
This is expected behavior right now, a backslash is our way of encoding empty paragraphs – which isn't something that markdown supports. You can "clean up" the raw output using a regex like this:

https://github.com/outline/outline/blob/f3ea02fdd0881fbc86872e198c088806b5739745/shared/utils/unescape.js#L3-L7

alexieyizhe:
Ah that totally makes sense - I do think the additional parameter for a clean output would be helpful like you said. Thanks for clarifying!

tommoor:
We’ll add a new endpoint in the next release and let you know in this issue
once complete

alexieyizhe:
Sounds good! Great work 😄

PR #1343, opened 2020-07-11
feat: Adds documents.export endpoint 

tommoor:
Returns only the cleaned up Markdown, useful for display purposes or import into another system.

closes https://github.com/outline/outline/issues/1308

### DIFF
--- a/server/api/documents.js
+++ b/server/api/documents.js
@@ -366,11 +366,7 @@ router.post("documents.drafts", auth(), pagination(), async ctx => {
   };
 });
 
-router.post("documents.info", auth({ required: false }), async ctx => {
-  const { id, shareId } = ctx.body;
-  ctx.assertPresent(id || shareId, "id or shareId is required");
-
-  const user = ctx.state.user;
+async function loadDocument({ id, shareId, user }) {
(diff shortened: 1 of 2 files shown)


server/api/documents.js
```diff
@@ -366,11 +366,7 @@ router.post("documents.drafts", auth(), pagination(), async ctx => {
   };
 });
 
-router.post("documents.info", auth({ required: false }), async ctx => {
-  const { id, shareId } = ctx.body;
-  ctx.assertPresent(id || shareId, "id or shareId is required");
-
-  const user = ctx.state.user;
+async function loadDocument({ id, shareId, user }) {
```
tommoor, line 369: Because the document load and auth logic is the same for `documents.info` and `documents.export` we pull it out into it's own function to reduce duplication.

server/api/documents.js
```diff
@@ -412,6 +417,18 @@ router.post("documents.info", auth({ required: false }), async ctx => {
   };
 });
 
+router.post("documents.export", auth({ required: false }), async ctx => {
```
tommoor, line 420: 🍖

PR #1346, opened 2020-07-13
feat: Document hover cards

tommoor:
Weekend hack fun with @thenanyu – a first pass at document hover previews. Right now this only works for internal document links, but it should be trivial to update this in the future to be pluggable for integrations to provide info such as Github/JIRA links.

---

image


app/components/HoverPreview.js
```diff
@@ -0,0 +1,182 @@
...
+  const bounds = node.getBoundingClientRect();
+
+  // previews only work for internal doc links for now
+  if (!isInternalUrl(node.href)) {
+    return null;
+  }
+
+  const slug = parseDocumentSlugFromUrl(node.href);
+
+  // $FlowFixMe
```
tommoor, line 34: I hate all these, but we need to upgrade flow and that should happen in a separate PR.

tommoor:
This feels more solid now 👍  cc @thenanyu

MatheusRV:
Awesome!! When you considering merge it?

PR #1347, opened 2020-07-13
feat: progressive web app service worker

chsasank:
Please see discussion here: https://github.com/outline/outline/discussions/1337

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -142,7 +142,7 @@ app.use(
         "www.google-analytics.com",
         "browser.sentry-cdn.com",
         "cdn.jsdelivr.net",
-        "storage.googleapis.com"
+        "storage.googleapis.com",
       ],
       styleSrc: ["'self'", "'unsafe-inline'", "github.githubassets.com"],
       imgSrc: ["*", "data:", "blob:"],
--- a/server/app.js
+++ b/server/app.js
@@ -141,6 +141,8 @@ app.use(
         "gist.github.com",
         "www.google-analytics.com",
         "browser.sentry-cdn.com",
+        "cdn.jsdelivr.net",
+        "storage.googleapis.com"
       ],
       styleSrc: ["'self'", "'unsafe-inline'", "github.githubassets.com"],
       imgSrc: ["*", "data:", "blob:"],
--- a/server/static/dev.html
+++ b/server/static/dev.html
@@ -37,6 +37,12 @@
         window.document.querySelector('#root').style.background = "#111319";
       }
     </script>
+    <script type="module">
+      // PWA: https://www.pwabuilder.com/serviceworker
+      import 'https://cdn.jsdelivr.net/npm/@pwabuilder/pwaupdate';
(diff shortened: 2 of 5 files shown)


server/app.js
```diff
@@ -141,6 +141,8 @@ app.use(
         "gist.github.com",
         "www.google-analytics.com",
         "browser.sentry-cdn.com",
+        "cdn.jsdelivr.net",
+        "storage.googleapis.com",
```
tommoor, line 145: Can we add this without adding two more third-party dependencies? I really don't want to add the entirety of these two CDN's to the CSP allow-list and I think it would be beneficial to keep dependencies bundled with the application.

server/static/dev.html
```diff
@@ -37,6 +37,12 @@
         window.document.querySelector('#root').style.background = "#111319";
       }
     </script>
+    <script type="module">
+      // PWA: https://www.pwabuilder.com/serviceworker
+      import 'https://cdn.jsdelivr.net/npm/@pwabuilder/pwaupdate';
+      const el = document.createElement('pwa-update');
+      document.body.appendChild(el);
+    </script>
```
tommoor, line 45: The UI that this provides doesn't match the rest of the application, I think we'd need to find a way to have it use the regular toast system that is already built in.

I feel like the knock on affects of potentially serving cached assets and js bundles might be quite significant and unpredictable.

server/app.js
chsasank: We certainly can host these in outline in public folder or using yarn. But if only you can guide me, where or how, it'll make my life easy :)

server/static/dev.html
chsasank: Sure, this is not very unrelated to the above right? If you can guide me to a tutorial how to do this, I'll implement it. Thanks.

chsasank:
Hi @tommoor,

I want to revive this PR and make progressive web app well. We can then use this PWA to make both electron (i.e desktop) and android phone apps! Looks like this is going to be future: https://medium.com/pwabuilder/microsoft-and-google-team-up-to-make-pwas-better-in-the-play-store-b59710e487

This is a good direction also because of the fact that outline is based on react and existence of react native!

It'd be great if you can guide me through with how you want to get the PWA integration done! I'll also raise an issue asking for help in PWA builder repo.

tommoor:
In terms of closer integration with the build system, it seems there are two webpack plugins that can do almost 99% of the work:

https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin
https://www.npmjs.com/package/webpack-pwa-manifest

This would keep all of the config inside of webpack and not need additional files and third party script injection

chsasank:
I was considering writing react component to make PWA easy. What do you think?

tommoor:
I’m not clear on how React is related to be honest..

chsasank:
I mean a small react component that we can keep in app.js or something will all the features of pwabuilder.

tommoor:
closing due to inactivity

Issue #1177, opened 2020-02-15
Improve authentication methods & docker setup


gempain:
Hi there,

First, I'd like to say that I tried your wiki with a local install and I loved the experience, really, truly graceful and beautifully designed - kind of has a feeling of Ghost.

I am a bit disappointed that at almost 8k stars, the only way to authenticate is Slack. I would expect email auth first, then cool stuff like Slack. I'd also love to see things like Mattermost, which is a self hosted Slack clone. Storage wise, I'm fine with AS3 because there is a fake AS3 image that we can use as an alternative. Slack, however, can't be swapped.

I've seen issues/PRs on this repo trying to integrate various authentication mechanisms, and I think adding them would really boost adoption of this project. Six months ago, what stopped us from using Outline internally was a lack of authentication mechanisms and ease of docker setup.

In our case, we use Mattermost as a self-hosted Slack alternative. I gave a shot at writing the Mattermost integration, but it seems that the current project structure isn't really friendly for adding new authentication mechanisms, so I gave up.

Docker wise, it's getting there, but I feel like it's still too complex. Users should be able to spin up an outline instance in 5 minutes on any Docker-compatible machine, without having to clone, install dependencies, create a `.env`, etc. I would suggest:
- a Docker image published on Docker Hub (I found [this one](https://hub.docker.com/r/outlinewiki/outline) but there's no docs)
- no need to install dependencies (looking at the Dockerfile of the image mentioned above, I see a `yarn install` but can't be sure of what it really does and how to use the image). For this, just package node_modules with your Dockerfile, or bundle them with webpack.
- no need for a `.env` file, just provide those values as Docker environment variables

You've got awesome features and a great UI/UX, but the devX and sysadminX isn't so smooth, but I'm looking forward to seeing this project grow.

Keep up the great work !

arjunv:
Looks like the pull [#1088](https://github.com/outline/outline/pull/1088) which allows logging in with email accounts was already merged. But it only allows one to add guests who doesn't have google/slack and not to setup the whole thing without those.

This is a really stupid decision for an open source software to take tbh.

It should be as simple as :
- Generate an admin user password in the CLI while starting for the first time
- When the admin logs into the web, have that password changed forcibly
- Allow admin to setup whatever authentication methods he need in the console
- There should be a non smtp based adding of users too. ie the admin generates a user:pass pair which is given to a 3rd party in whatever secure channel he prefers to. On that user's first login, have that password forcibly changed so the admin cannot snoop in.

[...]

PR #1353, opened 2020-07-15
chore: Remove env variables in webpack bundle

### LINKED ISSUE #1351: Remove env variables baked into webpack
A big request from folks self hosting has been to remove the webpack build step inside the docker container for lower memory usage and faster startup.

The main thing that needs to happen to achieve this is to move the dynamic values out of the built javascript bundle and into the rendered html page, inside a script tag.

See this previous (unlanded) PR for reference https://github.com/outline/outline/pull/850

tommoor:
Local testing has gone well, I don't know if any folks self hosting want to give this a try? I'll leave it open for feedback until the weekend. 

closes #1351 
relates #1177
relates #850


defunctzombie:
Sorry I hadn't gotten around to making a PR for the referenced docker image PR - but if you wanted to keep your env vars during the build you can accomplish this using docker build ARGS as seen here:

https://github.com/Polychain/outline/blob/v0.43.3-pc0/Dockerfile

tommoor:
@defunctzombie honestly I prefer the approach of not having them in the bundle at all, it's easier to understand – we've had a few cases of folks tripped up by doing the webpack build in an environment without the variables and it'll be nice to take that out of the equation.

defunctzombie:
Absolutely - you know best what makes sense for the project. I wanted to provide the Dockerfile as an option since I had noted earlier that I would.

On the PR itself - something I've found helpful is to consolidate the config into a single file (config.js, env.js, etc) and use that in various locations rather than "window.env". It allows for a single place to document and simpler management of how the environment configuration is loaded (today from window, maybe tomorrow elsewhere) and avoid the window global sprinkled around the project files. Since you are already touching many of these locations now might be a nice time to make such a change.

tommoor:
Good call on the consolidation 👍

PR #1359, opened 2020-07-16
feat: Add Hotjar Script

MatheusRV:
At this moment make sense use hotjat, in future I promise to create a generic way to add any script inside Outline.

- [ ] Improve hotjar script


tommoor:
Closing as won't merge, we can't add scripts for every possible analytics service.

PR #1360, opened 2020-07-16
fix: Do not load Analytics if env is empty

MatheusRV:
Avoid loading a script unnecessarily


tommoor:
Hi @MatheusRV – this is functionally identical, the script doesn't load either way.

MatheusRV:
The performance may be insignificant, but not letting React carry the class makes sense. Performing a conditional within the class causes React to first create the class in scope, then kill it by returning nothing. But at the end of the day, I don't know how much of an impact this will have on performance gains.

PR #1365, opened 2020-07-18
fix: common misspelling errors

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines

### DIFF
--- a/app/components/ScrollToTop.js
+++ b/app/components/ScrollToTop.js
@@ -13,7 +13,7 @@ class ScrollToTop extends React.Component<Props> {
   componentDidUpdate(prevProps) {
     if (this.props.location.pathname === prevProps.location.pathname) return;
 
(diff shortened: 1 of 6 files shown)


tommoor:
I see you have a _lot_ of this type of PR, I take it you're automating them – I'm happy to merge, but personally i'd be careful as this kind of automated PR behavior has resulted in GitHub accounts getting banned in the past.

Necmttn:
Been experimenting with a script; a.k.a [Grammar Nazi] 😅  original idea was to do a bunch of PR over a weekend. and write a thread on HN on Monday :) 
but getting banned would be a disaster. so I will just make it as `cmd` 

anyway there's good takeaway is I discover a bunch cool new project.

PR #1367, opened 2020-07-19
feat: nicer gradient mask for hover previews

thenanyu:
For previews that do not use all the available space, it's nicer when we don't mask any of the content

---
image

For previews that use all of the space, mask only the last part of the last line for readability and implied continuation

---
image

This is similar to what Wikipedia does

---
image


tommoor:
Thank ya

PR #1376, opened 2020-07-21
refactor: Fix updater's use of the UPDATES_KEY constant

cxcorp:
Was taking a peek at the file and noticed that this literal was probably supposed to be the `UPDATES_KEY` constant instead.


tommoor:
Ah yes, good spot – thanks!

PR #1388, opened 2020-07-24
feat: Sharing improvements

### LINKED ISSUE #1374: If a document is publicly shared, show that somewhere
Right now you can be editing a document and have no idea that there is a public link to it in the wild. Ideally there would be a clear hint that this is the case, I'm thinking a globe icon on the `Share` button is usually the way this is done.

### LINKED ISSUE #1387: Improve the share dialog
- [x] Add public boolean in database
- [x] Add toggle to "publicly viewable" on / off

tommoor:
## Before
image

## After
image

image

closes #1387
closes #1374

### DIFF
--- a/app/components/Editor.js
+++ b/app/components/Editor.js
@@ -95,13 +95,13 @@ const StyledEditor = styled(RichMarkdownEditor)`
 
   p {
     a {
-      color: ${props => props.theme.link};
-      border-bottom: 1px solid ${props => lighten(0.5, props.theme.link)};
+      color: ${props => props.theme.text};
+      border-bottom: 1px solid ${props => lighten(0.5, props.theme.text)};
       text-decoration: none !important;
       font-weight: 500;
 
       &:hover {
-        border-bottom: 1px solid ${props => props.theme.link};
+        border-bottom: 1px solid ${props => props.theme.text};
         text-decoration: none;
       }
     }
--- a/app/components/Tooltip.js
+++ b/app/components/Tooltip.js
@@ -18,6 +18,10 @@ class Tooltip extends React.Component<Props> {
 
     let content = tooltip;
 
+    if (!tooltip) {
+      return this.props.children;
+    }
+
     if (shortcut) {
       content = (
         <React.Fragment>
--- a/app/models/Document.js
+++ b/app/models/Document.js
@@ -41,7 +41,6 @@ export default class Document extends BaseModel {
   deletedAt: ?string;
   url: string;
   urlId: string;
-  shareUrl: ?string;
   revision: number;
 
   get emoji() {
@@ -90,10 +89,7 @@ export default class Document extends BaseModel {
 
   @action
   share = async () => {
-    const res = await client.post("/shares.create", { documentId: this.id });
-    invariant(res && res.data, "Share data should be available");
-    this.shareUrl = res.data.url;
-    return this.shareUrl;
+    return this.store.rootStore.shares.create({ documentId: this.id });
   };
 
   @action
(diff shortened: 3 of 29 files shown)


app/components/Editor.js
```diff
@@ -95,13 +95,13 @@ const StyledEditor = styled(RichMarkdownEditor)`
 
   p {
     a {
-      color: ${props => props.theme.link};
-      border-bottom: 1px solid ${props => lighten(0.5, props.theme.link)};
+      color: ${props => props.theme.text};
```
tommoor, line 98: Unrelated, but noticed that the link colors outside of documents did not have accessible contrast in dark mode.

server/presenters/share.js
```diff
@@ -5,9 +5,11 @@ import { presentUser } from ".";
 export default function present(share: Share) {
   return {
     id: share.id,
+    documentId: share.documentId,
     documentTitle: share.document.title,
     documentUrl: share.document.url,
-    url: `${process.env.URL}/share/${share.id}`,
+    published: share.published,
+    url: `${share.team.url}/share/${share.id}`,
```
tommoor, line 12: fixes: Share links should include team's subdomain to allow authentication to be read if the user is signed in.

server/api/shares.js
```diff
@@ -87,22 +123,25 @@ router.post("shares.create", auth(), async ctx => {
...
-    name: "shares.create",
-    documentId,
-    collectionId: document.collectionId,
-    modelId: share.id,
-    teamId: user.teamId,
-    actorId: user.id,
-    data: { name: document.title },
-    ip: ctx.request.ip,
-  });
+  if (isCreated) {
```
tommoor, line 126: fixes: `shares.create` events were being created whenever the share link dialog is opened.

MatheusRV:
I believe it would be incredible to have a way to share with people who are already using the platform. For example, when clicking on "Share" a multiple selection box will appear and you can write the name of one or more people. I think of two ways to show the documents you were invited to, the first would be notification and the second would be a tab (in sidebar) with all the documents you were invited to.

tommoor:
> For example, when clicking on "Share" a multiple selection box will appear and you can write the name of one or more people. 

Agreed 😄

app/models/Document.js
```diff
@@ -90,10 +89,7 @@ export default class Document extends BaseModel {
 
   @action
   share = async () => {
-    const res = await client.post("/shares.create", { documentId: this.id });
-    invariant(res && res.data, "Share data should be available");
-    this.shareUrl = res.data.url;
-    return this.shareUrl;
+    return this.store.rootStore.shares.create({ documentId: this.id });
```
tommoor, line 92: Store the `share` record in the shares store correctly instead of just the url string on the document

app/components/Tooltip.js
```diff
@@ -18,6 +18,10 @@ class Tooltip extends React.Component<Props> {
 
     let content = tooltip;
 
+    if (!tooltip) {
```
tommoor, line 21: Allow passing an empty string in `tooltip` prop to disable it.

server/migrations/20200723055414-add-published-to-shares.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('shares', 'published', {
+      type: Sequelize.BOOLEAN,
+      allowNull: false,
+      defaultValue: false
+    });
+
+    await queryInterface.sequelize.query(`
```
tommoor, line 11: All existing shares were public

PR #1447, opened 2020-08-12
feat: Add GitHub authentication support

whatl3y:
Followed the Slack authentication flow to build GitHub auth. I didn't add tests because there were no associated tests for the other auth types (slack particularly) that I could see, but all local testing works great. Let me know if you'd like additional info/help/docs/etc. to merge this PR.


MatheusRV:
I prefer to use https://github.com/outline/outline/pull/1183 instead of this.

whatl3y:
That's fair, I actually used [passport-github2](https://github.com/cfsghost/passport-github) to get the various endpoints to support auth in its entirety as outline currently requires, and use passport in all my other projects so could just change this (or create a new fresh PR) to use passport for github auth instead of this homegrown thing here. I just didn't know how far we were from getting outline to support passport and know others have communicated concerns about wanting github integration sooner than later.

Feel free to close this PR if need be and I can work on a passport branch or otherwise in the future if need be, I'll be using this fork for my self hosted instance in the meantime.

tommoor:
Yep, much appreciated PR but I'm closing in favor of #1183 – we don't want to add any more "custom" authentications.

PR #1449, opened 2020-08-12
chore: Upgrade pg and sequelize to support node 14+

### LINKED ISSUE #1373: Add support for Node 14+
Currently we're pinned to < 14 because of postgres dependency. It should all work now once the deps are updated.

This involves updating Sequelize to v6, which introduces a number of failing tests as first blush.

whatl3y:
Will resolve #1373 when we resolve the last two failing tests (info below)

I'm having an issue with 2 tests (`documentMover` suite). I'm not a sequelize expert so having a little trouble navigating that landscape wrt models and `belongsTo` relationships among other things, but any assistance from others would be appreciated.

My issue is hopefully clearly illustrated with the following:

- If I print `collection.documentStructure[0]` at [this line](https://github.com/outline/outline/blob/develop/server/commands/documentMover.test.js#L35), I get the correct expected structure, with `collection. documentStructure[0].children.length === 1` (the only child is `newDocument` we created)
- If I print `collection.documentStructure[0]` at [this line](https://github.com/outline/outline/blob/develop/server/commands/documentMover.js#L29) inside `documentMover`, I'm getting `collection. documentStructure[0].children.length === 0` (i.e. no children underneath the originally created doc). This is unexpected and I can't figure out why `newDocument` does not show up in the collection at this point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     ]
   },
   "engines": {
-    "node": ">= 12 <14"
+    "node": ">= 12"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


tommoor:
Yea, this is pretty much where I got to with the task – those documentMover tests, should get another chance to take a look at this before EOW.

tommoor:
I'm not able to push to your branch, but I got the tests passing over here: 
https://github.com/outline/outline/pull/1462

Many thanks!

PR #1465, opened 2020-08-18
feat: Add ClickUp Embed Service

MatheusRV:
image


app/embeds/ClickUp.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import * as React from "react";
+import Frame from "./components/Frame";
+
+const URL_REGEX = new RegExp(
+  "^https://share.clickup.com/b/h/(.*)/(.*)$"
```
tommoor, line 6: Did you find any documentation on their share/embeds? It would be good to know what the `b/h/` are here and if it's possible that they will be other values.

app/embeds/index.js
```diff
@@ -57,6 +58,13 @@ export default [
     component: Airtable,
     matcher: matcher(Airtable),
   },
+  {
+    title: "ClickUp",
+    keywords: "project",
+    icon: () => <Img src="/images/clickup.png" />,
```
tommoor, line 64: image

Can we get a version of the logo on a transparent background

app/embeds/ClickUp.js
MatheusRV: https://docs.clickup.com/en/articles/3714856-public-sharing

Yes it's possible has others value. I'm chageing it right know.

MatheusRV:
@tommoor could you please review this?

tommoor:
Fixed the lint issue, but your tests arent' passing 🤷

Issue #1463, opened 2020-08-18
Allow unpublishing documents


tommoor:
Split from: https://github.com/outline/outline/issues/1335

Published documents should have the option to "Unpublish" should you accidentally publish a draft or wish to take it back into a private space.

- Show an "Unpublish" option in the document menu
- Unpublish should remove the document from the collection tree structure
- It should set the `publishedAt` to empty
- "Unpublish" should only be available to the original creator and admins
- "Unpublish" will not be available on deleted or archived documents

guilherme-diniz:
I will tackle this issue!

tommoor:
Closed by 0aa338cccca7e83c93783a13a3277ce0a704ff9f

PR #1467, opened 2020-08-18
feat: Allow unpublishing documents

guilherme-diniz:
Implements the action to allow unpublish a document.
This option is only available for a publish document that wasn't archived or deleted.
Should check if the permissions for allowing whom can unpublish are set correctly.

### DIFF
--- a/app/menus/DocumentMenu.js
+++ b/app/menus/DocumentMenu.js
@@ -106,6 +106,11 @@ class DocumentMenu extends React.Component<Props> {
     this.props.ui.showToast("Document restored");
   };
 
+  handleUnpublish = async (ev: SyntheticEvent<>) => {
+    await this.props.document.unpublish();
+    this.props.ui.showToast("Document unpublished");
+  };
+
   handlePin = (ev: SyntheticEvent<>) => {
     this.props.document.pin();
   };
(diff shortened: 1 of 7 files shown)


app/menus/DocumentMenu.js
```diff
@@ -225,6 +230,11 @@ class DocumentMenu extends React.Component<Props> {
               New nested document
             </DropdownMenuItem>
           )}
+          {!document.isArchived && !document.isDeleted && !document.isDraft && (
```
tommoor, line 233:
```suggestion
          {can.unpublish && (
```

`can.unpublish` should be available here as you defined it as a policy.

server/models/Document.js
```diff
@@ -556,6 +555,18 @@ Document.prototype.publish = async function (options) {
   return this;
 };
 
+Document.prototype.unpublish = async function (options) {
+  if (!this.publishedAt) return this.save(options);
```
tommoor, line 559:
```suggestion
  if (!this.publishedAt) return this;
```

Can we just avoid the save here altogether?

server/policies/document.js
```diff
@@ -143,3 +143,19 @@ allow(
...
+    document.collection,
+    "collection is missing, did you forget to include in the query scope?"
+  );
+
+  if (!document.publishedAt || !!document.deletedAt || !!document.archivedAt)
+    return false;
+
+  if (cannot(user, "update", document.collection)) return false;
+
+  // TODO Block unpublish a document that has nested documents
```
tommoor, line 158: Good callout, I missed this on the original spec 👍

tommoor:
closes https://github.com/outline/outline/issues/1463

tommoor:
Sorry I forgot to mention before, can we add tests for the new api endpoint?

The existing pattern should be easy to follow

guilherme-diniz:
Tests added!

tommoor:
Thanks! Will test and (hopefully) merge in next 24h.

PR #1469, opened 2020-08-19
feat: Support importing .docx or .html files as new documents

### LINKED ISSUE #1339: Add support for docx import
Looks like `mammoth` might be a good package to help with this:
https://www.npmjs.com/package/mammoth

Right now import is done by reading the text/markdown file on the client and then creating a new doc with that text content in the same way you'd create any doc on the client, there is no differentiation. To support new file formats we'll want to wrap this in an API endpoint that can gain filetype support over time. eg:

`POST /files.import` with blob of file(s) and recipient `collectionId`, `parentDocumentId`.

The response should be the changed collection and documents. The Ideal UX is that the existing import flows support `.docx` files seamlessly.

whatl3y:
Resolves #1339 

**Usage**
- Import a .docx document using the normal _Import document_ flow
- Also adds support for importing html files as well


tommoor:
Hey Lance, this is a great start – thanks for kicking it off. From a high level we should consider some architectural changes that will simplify things (particularly on the client), and match the approach that was suggested in the issue, albeit very concisely:

- Only a single new API endpoint should be needed, I recommend it be `documents.import`. This endpoint should accept a multipart upload similar to what you have in `files.import`, but instead of returning the markdown it should create the document and return the document/collection the same as `documents.create`. This endpoint will actually end up sharing a lot of logic with documents create.

- No need for the `files.support` endpoint, we can just embed the file support in the frontend code.

whatl3y:
@tommoor makes sense, I made the changes and a new `documents.import` endpoint instead of `files.import`. Let me know what you think.

whatl3y:
I believe I've addressed the outstanding items. Let me know if you see any other adjustments you'd like to see or if I misunderstood anything!

app/menus/CollectionMenu.js
```diff
@@ -114,7 +114,9 @@ class CollectionMenu extends React.Component<Props> {
             ref={(ref) => (this.file = ref)}
             onChange={this.onFilePicked}
             onClick={(ev) => ev.stopPropagation()}
-            accept="text/markdown, text/plain"
+            accept={["text/markdown", "text/plain"]
```
tommoor, line 117: Just noticed in testing there's another spot that needs updating in `DropToImport` component, so best to make this available directly on the store so we can use it twice - `importFiletypesCSV` or something.

app/utils/importFile.js
```diff
@@ -16,35 +16,55 @@ const importFile = async ({
...
+    if (documents.importFiletypes.includes(file.type)) {
+      try {
+        const document = new Document(
+          {
+            parentDocumentId: documentId,
+            collectionId,
+            title: file.name.replace(/\.[^/.]+$/, ""),
+          },
+          documents
+        );
```
tommoor, line 30: This last bit of `Document instantiation is the only thing that's throwing me right now – can we just pass these params all through directly to the `import` store method?

app/utils/importFile.js
```diff
@@ -16,35 +16,55 @@ const importFile = async ({
   documentId,
   collectionId,
 }: Options): Promise<Document> => {
-  return new Promise((resolve, reject) => {
-    const reader = new FileReader();
+  return new Promise(async (resolve, reject) => {
```
tommoor, line 19: We could clean this a little if we move the `Promise` creation down to only wrap the file reader. This first flow for server side import can use async/await as there are no callback involved.

app/menus/CollectionMenu.js
whatl3y: https://github.com/outline/outline/pull/1469/commits/44a861b3af5b5db96d9a0a15c042bda85c2f2b17#diff-5dead950e66336055adf3edbd1311562R117

app/utils/importFile.js
whatl3y: https://github.com/outline/outline/pull/1469/commits/44a861b3af5b5db96d9a0a15c042bda85c2f2b17#diff-538eabbd5033bdf49a45dce7f55b5545R19

app/utils/importFile.js
whatl3y: https://github.com/outline/outline/pull/1469/commits/44a861b3af5b5db96d9a0a15c042bda85c2f2b17#diff-cf91b760df0750438c333d13b1499242R470

app/utils/importFile.js
```diff
@@ -16,26 +16,18 @@ const importFile = async ({
...
-    }
+  // non plain text support
+  if (documents.importFiletypesServer.includes(file.type)) {
+    return await documents.import(
+      { publish: true, file },
+      collectionId,
+      file.name.replace(/\.[^/.]+$/, ""),
+      null,
+      documentId
+    );
```
tommoor, line 27: Why have a `text` argument if we're just going to pass null? Suggest:

```suggestion
    const title = file.name.replace(/\.[^/.]+$/, "");

    return await documents.import(
      {
        title,
        documentId,
        collectionId,
      },
      {
        publish: true,
        file
      }
    );
```

whatl3y: https://github.com/outline/outline/pull/1469/commits/796fb1a563cd7c14cb1bdbb95ea7521912ad4ea2

app/menus/CollectionMenu.js
tommoor: This one was only partly addressed, the `DropToImport` component still not accepting docx types

tommoor:
In testing I realized one more thing, while this does support images the images are embedded in the text data as `data:` urls. We're going to have to post-process the text data before it's saved on the server to convert those into `Attachment`'s so the binary data is stored outside the database.

The `/api/attachments` has examples on how to achieve the upload aspect. Let me know if this isn't something you're comfortable doing, I'm happy to take it on to get this over the finish line.

whatl3y:
@tommoor I'm pretty slammed this week doubt I'll be able to spare any extra time to try and dig into the `data:` URL replacement, I'd be more than happy if you're okay waiting a week or possibly 2 before this gets in, but feel free to take over if you'd like to get this in sooner.

tommoor:
Right on – I appreciate the communication 👍

tommoor:
Superseded by https://github.com/outline/outline/pull/1551

PR #1473, opened 2020-08-20
feat: add filters to drafts

guilherme-diniz:
Include te ability of filter the draft list by collection, author and time.


tommoor:
Functionality wise I think this is fine, I need to work on some of the design though – really the components don't exist to combine filters and sorts in the way you want to do here, but they should.

Please bear with me while that design comes together and we can then apply it to the PR.

tommoor:
closed in favor of https://github.com/outline/outline/pull/1631

PR #1474, opened 2020-08-21
feat: preview draft location in collections tree

guilherme-diniz:
Show draft files in collections tree, previewing where the document will be when published.
Also, add new tab inside a collection, showing all drafts, and highlight nested drafts in a document

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ npm-debug.log
 stats.json
 .DS_Store
 fakes3/*
+.idea
--- a/app/components/Sidebar/components/CollectionLink.js
+++ b/app/components/Sidebar/components/CollectionLink.js
@@ -1,6 +1,7 @@
 // @flow
 import { observable } from "mobx";
 import { observer } from "mobx-react";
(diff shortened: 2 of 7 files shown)


tommoor:
Thanks for the PR – after much deliberation unfortunately we won't be able to merge this change, drafts are very purposefully excluded from the sidebar tree to make it clear which documents are and aren't published. As such, injecting them back into the tree post-load isn't something that makes sense as the product stands today.

I'm closing this out, but open to having a discussion around exactly which workflows this PR was aiming to help with.

guilherme-diniz:
@MatheusRV can you help us on this?

MatheusRV:
After thinking, I have a suggestion. Why we don't just show you the draft we are talking about, exactly where it would be? This makes it easier to identify when I publish it in the correct place.

tommoor:
Yep, that could work for sure if it shows up temporarily. Ideally it would
still get a slightly different treatment, perhaps faded a little.

The sidebar is currently being rebuilt to allow re-ordering so might want
to hold off until that’s done

PR #1475, opened 2020-08-21
feat: Add Header to Document History Sidebar

guilherme-diniz:
Closes #1139

PR #1477, opened 2020-08-22
feat: add document events log menu

guilherme-diniz:
Add events menu to a document, allowing us to see it events.
Screen Shot 2020-08-21 at 22 31 13
Screen Shot 2020-08-21 at 22 31 24

### DIFF
--- a/app/components/DocumentEvents.js
+++ b/app/components/DocumentEvents.js
@@ -1,0 +1,166 @@
+// @flow
+import { action, observable } from "mobx";
+import { inject, observer } from "mobx-react";
+import { CloseIcon } from "outline-icons";
+import * as React from "react";
+import { type Match, Redirect, type RouterHistory } from "react-router-dom";
+import { Waypoint } from "react-waypoint";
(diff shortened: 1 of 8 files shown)


guilherme-diniz:
Added the same fix for header size (#1475 )
Screen Shot 2020-08-25 at 17 42 15

MatheusRV:
Open for feedback @thenanyu @tommoor could review?

MatheusRV:
Friendly ping

tommoor:
Howdy – the idea of being able to see when a document moved is valuable but I don't think it deserves it's own timeline separate to the document history – it should just be included in the same timeline as you commented above.

With this particular implementation it doesn't make sense to render the admin list component in the sidebar here, it means the name of the document is repeated over and over again – the IP address is also kind of rendered half off-screen.

What problem is this solving, which events are you particularly interested in?

tommoor:
closing based on feedback and lack of movement, this idea will be kept around though and I think a combined timeline makes sense for the future

PR #1490, opened 2020-08-25
fix: Unable to view all possible locations when moving document

### LINKED ISSUE #1410: Improve Search Bar when Moving a Doc
When we need to move doc, it has been really hard to find the right "location" once the list of options doesn't show all the collections or docs that have the same title at the same time. 
Like the example below: We have " About Us" in every single collection but when you search is not showing that. 

image

tommoor:
This PR includes a number of improvements to the document move dialog to improve the experience:

image

- There is no longer an artificial cap on results, instead the UI is scrollable. closes #1410 
- The original document name only shows in the path on hover now to reduce visual clutter and add clarity to the end desination.
- Fixed a number of visual and spacing bugs
- Fixed templates should not appear here as an option to move


MatheusRV:
LGFM!

Issue #1501, opened 2020-08-31
Do not force encrypted connection to pgsql (docker)


MDXDave:
I am using outline via docker-compose.
After upgrading to the latest version (0.46.1) i got an error message that says that SSL is not supported on pgsql:

> SequelizeConnectionError: The server does not support SSL connections

My pgsql container does not support SSL because there are no benefits of support it.
I would not force the app to use a encrypted database connection, because its not required when running (local) via docker and the connection to the database is only established locally.

The issue appears due to this commit: c446a91f7db228243664115e9e6638eaee7c0e1e

**Outline (please complete the following information):**
- Install: self-hosted
- Version: 0.46.1

tommoor:
Could you try setting the `PGSSLMODE` env variable to "disable" and see if that helps?
https://ankane.org/postgres-sslmode-explained

MDXDave:
@tommoor Unfortunately it does not work. I've got the same error as before:

> (node:175) UnhandledPromiseRejectionWarning: SequelizeConnectionError: The server does not support SSL connections
outline_1   |     at /opt/outline/node_modules/sequelize/lib/dialects/postgres/connection-manager.js:188:20
outline_1   |     at Connection.connectingErrorHandler (/opt/outline/node_modules/pg/lib/client.js:213:14)
outline_1   |     at Connection.emit (events.js:315:20)
outline_1   |     at Connection.EventEmitter.emit (domain.js:483:12)
outline_1   |     at Socket.<anonymous> (/opt/outline/node_modules/pg/lib/connection.js:76:21)
outline_1   |     at Object.onceWrapper (events.js:422:26)
outline_1   |     at Socket.emit (events.js:315:20)
outline_1   |     at Socket.EventEmitter.emit (domain.js:483:12)
outline_1   |     at addChunk (_stream_readable.js:295:12)
outline_1   |     at readableAddChunk (_stream_readable.js:271:9)

johnmoxley:
Would love to see this in the docker image soon!

PR #1504, opened 2020-08-31
fix: Add PGSSLMODE param to allow unencrypted db access 

MDXDave:
This PR fixes issue #1501 and allows the use of `PGSSLMODE` variable to disable enforced ssl connection to pgsql database

### DIFF
--- a/server/sequelize.js
+++ b/server/sequelize.js
@@ -4,6 +4,8 @@ import Sequelize from "sequelize";
 import EncryptedField from "sequelize-encrypted";
 
 const isProduction = process.env.NODE_ENV === "production";
+const isEncryptedConnection =
+  isProduction && process.env.PGSSLMODE !== "disable";
 
 export const encryptedFields = () =>
   EncryptedField(Sequelize, process.env.SECRET_KEY);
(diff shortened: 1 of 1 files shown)


server/sequelize.js
```diff
@@ -5,6 +5,15 @@ import EncryptedField from "sequelize-encrypted";
...
 const isProduction = process.env.NODE_ENV === "production";
 
+const isPGSqlEnabled =
+  typeof process.env.PGSSLMODE === "undefined" ||
+  process.env.PGSSLMODE === null ||
+  process.env.PGSSLMODE !== "disable";
+
+const isEncryptedConnection = isProduction && isPGSqlEnabled;
+
+console.log(isEncryptedConnection);
```
tommoor, line 15: I think checking for the explicit disable of `process.env.PGSSLMODE !== "disable"` should be enough, no?

tommoor: Pls remove `console.log`

server/sequelize.js
```diff
@@ -5,6 +5,15 @@ import EncryptedField from "sequelize-encrypted";
 
 const isProduction = process.env.NODE_ENV === "production";
 
+const isPGSqlEnabled =
```
tommoor, line 8: I think checking for the explicit disable of `process.env.PGSSLMODE !== "disable"` should be enough, no?

tommoor:
Closed this with a slightly simpler implementation, thanks for filing/inspo.

closed by 6b6d67be

PR #1507, opened 2020-09-01
feat: Visually differentiate unread documents

### LINKED ISSUE #1150: Visually differentiate unread documents in list views
I would like to add two new visual treatments to the document list views to help surface information. On teams were document updates are happening regularly it can be more difficult to find new and relevant docs that are being added.

- A subtle tweak in visual treatment when the current user has not read the document (perhaps bolder?)
- A "New" badge when the above is true and the document was created since users last visit.

To achieve this we would need to query views for the user whenever a document is returned and set a new property on the document.

### Note
I'm mildly concerned about performance of adding a join on the views table – need to sure it doesn't regress in any measurable way. Perhaps this data can be loaded async from the client?

guilherme-diniz:
Closes #1150 

Here an example when we have new Documents and a not seen modified one.
Screen Shot 2020-09-01 at 19 59 11


app/components/Sidebar/components/DocumentLink.js
```diff
@@ -96,7 +97,12 @@ class DocumentLink extends React.Component<Props> {
               state: { title: node.title },
             }}
             expanded={showChildren ? true : undefined}
-            label={node.title || "Untitled"}
+            label={
+              <Title bold={document?.modifiedSinceViewed || document?.new}>
+                {node.title || "Untitled"}
+                {document?.new && <Badge>New</Badge>}
```
MatheusRV, line 103: Could we add a Updated badge?

tommoor: I think this'll be too noisy, I think you'll quickly come to recognize bold as updated as it's a pretty standard treatment eg in email clients.

guilherme-diniz: When is updated i just set title bolder, as requested by @tommoor 
But we. could add like a dot, as in Slack

tommoor:
@guilherme-diniz I don't think that displaying this state in the sidebar is scalable because you'd need to have all of the documents loaded for it to be accurate, and that's never going to be the case. It also introduces extra elements in an already confined space. Ideally we'd add the badge and bold treatment onto the `DocumentPreview` element so it appears everywhere a document is listed in the main content area.

guilherme-diniz:
I understand that.. But thinking as user... they probably will navigate through the collections more than anything.. So i think if we don't put anything on the document tree, the feature won't have its value to them..

But i can change it to meet this requirements.. no problem!!

app/components/Sidebar/components/DocumentLink.js
MatheusRV: Do not to forget to test with large file names... And create this effect when update and created.

MatheusRV:
The implemented way is performative, loading only the first child and recursively loading the adjacent children will work well. @guilherme-diniz could you provide a benchmark after and before for this?

This feature delivers a lot of value, I agree that creating a text badge can be a bit noisy. We can create a small icon in right side using ball for updated and decagon for new.

tommoor:
> The implemented way is performative, loading only the first child and recursively loading the adjacent children

I'm not sure what you're referring to, there is no additional recursive loading code in this PR. If a collection had 100 documents - for example - only some would be correctly displayed just depending on what the document list view has loaded into local cache, this is why the treatment should be limited to the list views where we can guarantee that the data has been loaded in a first version.

---

### Badge treatment for "New"
One key here is that new documents aren't just documents that you've never viewed, otherwise joining an existing knowledge-base all documents would have the "NEW" marker. There needs to be a time component too, say never viewed + created in last 2 weeks to start.

image

### List treatment
Previously viewed and not updated since
image

Previously viewed and updated since
image

Never viewed
image

server/presenters/document.js
```diff
@@ -62,8 +66,14 @@ export default async function present(document: Document, options: ?Options) {
     pinned: undefined,
     collectionId: undefined,
     parentDocumentId: undefined,
+    lastViewedAt: undefined,
   };
 
+  if (!!userId) {
```
tommoor, line 72: When we query a document we already pass the `userId` along to the `findByPk` method in all of the api endpoints, for single documents we can ensure that the view is loaded in scope at that time and avoid an additional query in the presenter.

https://github.com/outline/outline/blob/develop/server/models/Document.js#L226-L228

guilherme-diniz:
Hi @tommoor 
I added the changes you asked. I left the SideBar behaviour just like how I did and let me explain why:
- In my Tests, if i navigate through the collections by clicking in (Click on first collection, then a document inside, then a nested document..) the doc treatment works great, even when i click on the document, the list lose the unseen formatting.
- Based on that, i didn't ran into errors like false positive or false negative (show an unseen doc as a seen doc). And, it didn't impact the performance, since it only loads the view at the level i'm in and I changed to a join instead that query inside the presenter..
I would like you to test and tell me if you ran into any issue.

app/components/DocumentMeta.js
```diff
@@ -103,6 +104,17 @@ function DocumentMeta({
   const collection = collections.get(document.collectionId);
   const updatedByMe = auth.user && auth.user.id === updatedBy.id;
 
+  const timeSinceNow = () => {
+    if (!lastViewedAt)
+      return <Modified highlight={true}>Never Viewed</Modified>;
```
tommoor, line 109:
```suggestion
      return <Modified highlight={true}>Never viewed</Modified>;
```

The color difference on this one is currently too extreme, in light view it's very dark and stands out as much as the document title.

app/components/DocumentPreview/DocumentPreview.js
```diff
@@ -105,6 +105,7 @@ class DocumentPreview extends React.Component<Props> {
           {document.isTemplate && showTemplate && (
             <Badge primary>Template</Badge>
           )}
+          {document.new && <Badge primary>New</Badge>}
```
tommoor, line 108: Hoping to get the different color treatment here, I really need a better system for `Badge`s – adding a prop for `yellow` would probably be fine.

app/models/Document.js
```diff
@@ -24,7 +26,7 @@ export default class Document extends BaseModel {
 
   collaborators: User[];
   collectionId: string;
-  lastViewedAt: ?string;
+  @observable lastViewedAt: ?string;
```
tommoor, line 29:
```suggestion
  lastViewedAt: ?string;
```

`observable` shouldn't be needed here, anything that comes down in the JSON payload is automatically observable

app/models/Document.js
```diff
@@ -73,6 +75,14 @@ export default class Document extends BaseModel {
     return !!this.lastViewedAt && this.lastViewedAt < this.updatedAt;
   }
 
+  @computed
+  get new(): boolean {
```
tommoor, line 79:
```suggestion
  get isNew(): boolean {
```

Suggest `isNew` to match other computed booleans and also not step on the reserved `new` keyword.

app/models/Document.js
```diff
@@ -73,6 +75,14 @@ export default class Document extends BaseModel {
...
+    return (
+      !this.lastViewedAt &&
+      differenceInDays(new Date(), new Date(this.createdAt)) < 14
```
tommoor, line 82: 👍

app/scenes/Document/components/MarkAsViewed.js
```diff
@@ -17,7 +17,7 @@ class MarkAsViewed extends React.Component<Props> {
 
     this.viewTimeout = setTimeout(() => {
       if (document.publishedAt) {
-        document.view();
+        document.view().then((view) => document.updateLastViewed(view));
```
tommoor, line 20:
```suggestion
        const view = await document.view();
        document.updateLastViewed(view);
```

Will need to update function to `async` of course

server/api/documents.js
```diff
@@ -562,7 +571,7 @@ router.post("documents.search", auth(), pagination(), async (ctx) => {
   const documents = results.map((result) => result.document);
   const data = await Promise.all(
     results.map(async (result) => {
-      const document = await presentDocument(result.document);
+      const document = await presentDocument(result.document, user.id);
```
tommoor, line 574:
```suggestion
      const document = await presentDocument(result.document);
```

server/models/Document.js
```diff
@@ -208,11 +208,24 @@ Document.associate = (models) => {
...
+    if (!userId)
+      return {
+        include: [
+          {
+            model: models.View,
+            as: "views",
+            required: false,
+          },
+        ],
+      };
```
tommoor, line 221: Intrigued by this block? I'm concerned that not passing userId could result in accidentally exposing/loading view data for all users in a team.

server/models/View.js
```diff
@@ -52,6 +52,19 @@ View.findByDocument = async (documentId) => {
...
+    order: [["updatedAt", "DESC"]],
+    include: [
+      {
+        model: User,
+        paranoid: false,
+      },
+    ],
+  });
+};
+
```
tommoor, line 67:
```suggestion
```

No longer used, right?

tommoor:
> In my Tests, if i navigate through the collections by clicking in (Click on first collection, then a document inside, then a nested document..)

To reproduce just load a link directly to a document without navigating to it, as if coming from a browser bookmark.

tommoor (changes requested):

app/models/Document.js
guilherme-diniz: I had to add it.. because it wasn't working without it

server/models/Document.js
guilherme-diniz: I add it like this, because i didn't got a way of not loading anything when we don't have a userId.
If you could help me..

app/models/Document.js
guilherme-diniz: Also, i need the observable because i update this property when a user enters the document, without passing by the api..

tommoor: Weird, it should be automatically observable – I must be misunderstanding something

guilherme-diniz: When we dispatch the action to set a document as viewed.. I added a method `document.updateLastViewed(view)` that updates the object stored on memory, setting the lastView property.. Without the @observable, the update wasn't reflecting on the screen.

tommoor:
I want to try something new here in order to get this landed faster.

I believe there is still some visual polish that needs completing but I'm happy to do that, I've created a `release-0.48.0` "release" branch – if you change the base of this PR to that branch then we can merge there first.

guilherme-diniz:
> I want to try something new here in order to get this landed faster.
> 
> I believe there is still some visual polish that needs completing but I'm happy to do that, I've created a `release-0.48.0` "release" branch – if you change the base of this PR to that branch then we can merge there first.

@tommoor  done!

Issue #1095, opened 2019-12-06
UI hangs when updating a page that has been deleted


senyuuri:
**Problem:**
Assume we have two team members A and B, both want to edit a page X. They perform a sequence of operations in the following order:
1. A opens page X in editing mode, start typing
2. B deletes X
3. A clicks on "Done Editing", which sends a POST request to `/api/documents.update`. However, since X has been deleted, the backend returns a 403 error with payload `{"ok":false,"error":"authorization_error","message":"Authorization error"}`. The frontend does not handle the error and stuck in the "loading bar" animation.

![image](https://user-images.githubusercontent.com/2610986/70308012-608ec200-1845-11ea-9915-7ffe1c1b2b08.png)

tommoor:
Thanks for the report!

tommoor:
@senyuuri could you confirm that you're running the latest release? A variation of the bug still exists however it should at least show a banner at the top that tells you the document was deleted

senyuuri:
@tommoor seems we're not running the latest release. Git log shows that our branch is newer than the latest v0.37  (released on 16 Oct).


```
commit 682151554bc69ec49c80e1a792609a3ee1ede52d (grafted, HEAD -> master, origin/master, origin/HEAD)
Author: Tom Moor <tom.moor@gmail.com>
Date:   Mon Nov 18 20:14:55 2019 -0800

fix: Additional SQL cascades required
```

thenanyu:
to clarify, the current version of the error returns an auth error when attempting to edit a deleted document

![image](https://user-images.githubusercontent.com/427579/92316886-cb288600-efae-11ea-8faa-10b42813a7ca.png)

thenanyu:
![image](https://user-images.githubusercontent.com/427579/92331721-8ee83a80-f02d-11ea-97aa-9aaf72b88761.png)

PR #1517, opened 2020-09-06
fix: Improved handling of delete events from collection and document sockets

thenanyu:
fixes https://github.com/outline/outline/issues/1095

### DIFF
--- a/app/components/SocketProvider.js
+++ b/app/components/SocketProvider.js
@@ -174,14 +174,19 @@ class SocketProvider extends React.Component<Props> {
           const collection = collections.get(collectionId) || {};
 
           if (event.event === "collections.delete") {
-            documents.removeCollectionDocuments(collectionId);
-            memberships.removeCollectionMemberships(collectionId);
-
             const collection = collections.get(collectionId);
             if (collection) {
               collection.deletedAt = collectionDescriptor.updatedAt;
             }
 
+            const deletedDocuments = documents.inCollection(collectionId);
+            deletedDocuments.forEach((doc) => {
+              doc.deletedAt = collectionDescriptor.updatedAt;
+              policies.remove(doc.id);
+            });
+
+            documents.removeCollectionDocuments(collectionId);
+            memberships.removeCollectionMemberships(collectionId);
             collections.remove(collectionId);
             policies.remove(collectionId);
             continue;
--- a/app/components/SocketProvider.js
+++ b/app/components/SocketProvider.js
@@ -125,6 +125,8 @@ class SocketProvider extends React.Component<Props> {
             if (document) {
               document.deletedAt = documentDescriptor.updatedAt;
             }
+            policies.remove(documentId);
+
             continue;
           }
 
@@ -173,6 +175,15 @@ class SocketProvider extends React.Component<Props> {
 
           if (event.event === "collections.delete") {
             documents.removeCollectionDocuments(collectionId);
+            memberships.removeCollectionMemberships(collectionId);
+
+            const collection = collections.get(collectionId);
+            if (collection) {
+              collection.deletedAt = collectionDescriptor.updatedAt;
+            }
+
+            collections.remove(collectionId);
+            policies.remove(collectionId);
             continue;
           }
 
@@ -187,9 +198,10 @@ class SocketProvider extends React.Component<Props> {
             await collections.fetch(collectionId, { force: true });
           } catch (err) {
             if (err.statusCode === 404 || err.statusCode === 403) {
-              collections.remove(collectionId);
               documents.removeCollectionDocuments(collectionId);
               memberships.removeCollectionMemberships(collectionId);
+              collections.remove(collectionId);
+              policies.remove(collectionId);
               return;
             }
           }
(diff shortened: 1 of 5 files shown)


app/components/SocketProvider.js
```diff
@@ -173,6 +175,15 @@ class SocketProvider extends React.Component<Props> {
...
             documents.removeCollectionDocuments(collectionId);
+            memberships.removeCollectionMemberships(collectionId);
+
+            const collection = collections.get(collectionId);
+            if (collection) {
+              collection.deletedAt = collectionDescriptor.updatedAt;
+            }
+
+            collections.remove(collectionId);
+            policies.remove(collectionId);
```
thenanyu, line 186: ended up handling this case the same way we do for documents, and letting the component decide what to do with a collection where `deletedAt` is present

app/scenes/Collection.js
```diff
@@ -65,6 +65,15 @@ class CollectionScene extends React.Component<Props> {
...
     const { id } = this.props.match.params;
 
+    if (this.collection) {
+      const { collection } = this;
+      const policy = this.props.policies.get(collection.id);
+
+      if (!policy) {
+        this.loadContent(collection.id);
+      }
+    }
```
thenanyu, line 75: - check for policy presence. this is important for permissions controls too!
- we're omitting the carve-out for `collection.deletedAt` for now, but we can add special handling down the line if that's desired

app/scenes/Collection.js
```diff
@@ -75,18 +84,24 @@ class CollectionScene extends React.Component<Props> {
...
+        this.collection = collection;
+
+        await this.props.documents.fetchPinned({
+          collectionId: id,
+        });
+      }
+    } catch (error) {
+      if (error.name === "AuthorizationError") {
+        this.collection = null;
+      }
```
thenanyu, line 101: catching specific error here, I wanted to do `instanceof AuthorizationError` but I didn't see an example anywhere else, so I went with duck typing instead

tommoor: You can use `instanceof`, it's in use for other error types

app/components/SocketProvider.js
```diff
@@ -125,6 +125,8 @@ class SocketProvider extends React.Component<Props> {
             if (document) {
               document.deletedAt = documentDescriptor.updatedAt;
             }
+            policies.remove(documentId);
```
tommoor, line 128: This is causing the menu with missing items here: 
https://github.com/outline/outline/issues/1095#issuecomment-687852184

I guess this is why the policy wasn't deleted, it should remain but only have the `restore` valid. This is super edge case though as you have to be viewing a document as it gets deleted, going via the trash list would load the correct policy as part of that request

app/components/SocketProvider.js
```diff
@@ -174,14 +174,19 @@ class SocketProvider extends React.Component<Props> {
...
             const collection = collections.get(collectionId);
             if (collection) {
               collection.deletedAt = collectionDescriptor.updatedAt;
             }
 
+            const deletedDocuments = documents.inCollection(collectionId);
+            deletedDocuments.forEach((doc) => {
+              doc.deletedAt = collectionDescriptor.updatedAt;
+              policies.remove(doc.id);
+            });
```
tommoor, line 186: This is what `removeCollectionDocuments` should do on L188, no?

thenanyu: `removeCollectionDocuments` seems like it removes the documents from the store, rather than marks them as deleted. For example, it would do that when someone loses permissions to a collection rather than this more extreme case of the collection being deleted altogether

server/api/documents.js
```diff
@@ -407,11 +407,15 @@ async function loadDocument({ id, shareId, user }) {
...
-    authorize(user, "read", document);
+    document = await Document.findByPk(id, {
+      userId: user ? user.id : undefined,
+      paranoid: false,
+    });
+    if (document.deletedAt) {
+      authorize(user, "restore", document);
+    } else {
+      authorize(user, "read", document);
+    }
```
thenanyu, line 418: 🍖 allow individual docs to be loaded if they are deleted and the user can restore them

server/policies/document.js
```diff
@@ -122,6 +122,7 @@ allow(User, "delete", Document, (user, document) => {
 
 allow(User, "restore", Document, (user, document) => {
   if (!document.deletedAt) return false;
+  if (document.collection.deletedAt) return false;
```
thenanyu, line 125: if the document does not have a collection, it's perma-deleted

app/components/SocketProvider.js
thenanyu: This is now resolved via https://github.com/outline/outline/pull/1517/files#r484107538

server/policies/document.js
tommoor: I'll probably merge that PR tomorrow so might as well just skip this check, what made you put it in rn?

thenanyu: in this current state it would error out if you try to restore a document from a deleted collection

server/api/documents.js
tommoor: cool cool, was thinking of a similar solution

app/components/SocketProvider.js
tommoor: We're still calling `removeCollectionDocuments` below, will that not just remove the docs modified in this loop?

thenanyu: If you're currently looking at one of those documents, removing the collection relationship alone won't quite do the trick. This loop puts those browser windows into the same state as if you had deleted just the single document

tommoor: Understand, works well in testing too

PR #1522, opened 2020-09-09
feat: Add Torii Embed

MatheusRV:
New Torii Catalog


MatheusRV:
@tommoor rdy to be merged

tommoor:
I don't think this is worthy of being in the codebase to be honest, I've never heard of this service and it seems to be enterprise-only. The examples in the test don't lead to valid embed links either.

Issue #1527, opened 2020-09-10
Search existing documents when pasting fully-qualified internal URLs


thenanyu:
![image](https://user-images.githubusercontent.com/427579/92688743-bae81200-f2f2-11ea-823e-9f25844a481f.png)

this search ideally would return a search result

thenanyu:
note: there's already a helper called `isInternalUrl` that helps detect these things

PR #1529, opened 2020-09-10
feat: allow searching for urls of internal documents

thenanyu:
closes https://github.com/outline/outline/issues/1527


thenanyu:
Search vector seems somewhat poor for this — see example
image
image

We may be better off doing something dumber for this class of search

shared/utils/parseDocumentSlug.js
```diff
@@ -1,12 +1,16 @@
...
-  } catch (err) {
-    return;
+  if (url[0] === "/") {
+    parsed = url;
+  } else {
+    try {
+      parsed = new URL(url).pathname;
+    } catch (err) {
+      return;
+    }
```
thenanyu, line 12: allow for parsing of paths as well as fully qualified URLs

app/scenes/Document/components/DataLoader.js
```diff
@@ -70,6 +72,27 @@ class DataLoader extends React.Component<Props> {
...
+          },
+        ];
+      } catch (err) {}
+    } else {
+      try {
+        // don't search at all if term is a non-internal url
+        new URL(term);
+        return;
+      } catch (err) {}
+    }
```
thenanyu, line 93: The try catch look is a little odd, would love some syntax advice

shared/utils/parseDocumentSlug.js
tommoor: Looks good, this util is deserving of a test not sure why it doesn't have one already

app/scenes/Document/components/DataLoader.js
tommoor: I'd consider just dropping this case to be honest – I'm worried that genuine search terms could be successfully parsed as valid URL's and it doesn't do too much harm to just send to the document search – it might even turn up a relevant doc.

app/components/Editor.js
```diff
@@ -34,14 +34,14 @@ class Editor extends React.Component<PropsWithRef> {
...
-  onClickLink = (href: string) => {
+  onClickLink = (href: string, event: MouseEvent) => {
     // on page hash
     if (href[0] === "#") {
       window.location.href = href;
       return;
     }
 
-    if (isInternalUrl(href)) {
+    if (isInternalUrl(href) && !event.metaKey && !event.shiftKey) {
```
tommoor, line 44: Fixes `cmd+click` on an internal link should open a new tab

PR #1532, opened 2020-09-11
feat: Make collection titles searchable

### LINKED ISSUE #1385: Make collection titles searchable
Currently if you search a collection title it isn't included in the list of search results, it would be nice if it did. 

This can be done entirely locally as all collections are loaded on app boot.

renanfilipe:
closes #1385

image

### DIFF
--- a/app/components/CollectionPreview.js
+++ b/app/components/CollectionPreview.js
@@ -4,7 +4,6 @@ import { Link } from "react-router-dom";
 import styled from "styled-components";
 import Collection from "models/Collection";
 import CollectionIcon from "components/CollectionIcon";
-import Flex from "components/Flex";
 import Highlight from "components/Highlight";
 import CollectionMenu from "menus/CollectionMenu";
 
@@ -40,8 +39,6 @@ const CollectionLink = styled(Link)`
   margin: 10px -8px;
   padding: 6px 8px;
   border-radius: 8px;
-  overflow: hidden;
-  position: relative;
 
   ${SecondaryActions} {
     opacity: 0;
--- a/app/components/CollectionPreview.js
+++ b/app/components/CollectionPreview.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from "react";
 import { Link } from "react-router-dom";
-import styled, { withTheme } from "styled-components";
+import styled from "styled-components";
 import Collection from "models/Collection";
 import CollectionIcon from "components/CollectionIcon";
 import Flex from "components/Flex";
@@ -13,24 +13,22 @@ type Props = {
   highlight?: ?string,
 };
 
-class CollectionPreview extends React.Component<Props> {
-  render() {
-    const { collection, highlight } = this.props;
+function CollectionPreview(props: Props) {
+  const { collection, highlight } = props;
 
-    return (
-      <CollectionLink to={collection.url}>
-        <Heading>
-          <IconWrapper>
-            <CollectionIcon collection={collection} />
-          </IconWrapper>
-          <Title text={collection.name} highlight={highlight} />
-          <SecondaryActions>
-            <CollectionMenu collection={collection} position="right" />
-          </SecondaryActions>
-        </Heading>
-      </CollectionLink>
-    );
-  }
+  return (
+    <CollectionLink to={collection.url}>
+      <Heading>
+        <IconWrapper>
+          <CollectionIcon collection={collection} />
+        </IconWrapper>
+        <Title text={collection.name} highlight={highlight} />
+        <SecondaryActions>
+          <CollectionMenu collection={collection} position="right" />
+        </SecondaryActions>
+      </Heading>
+    </CollectionLink>
+  );
 }
 
 const SecondaryActions = styled(Flex)`
@@ -46,9 +44,6 @@ const CollectionLink = styled(Link)`
   margin: 10px -8px;
   padding: 6px 8px;
   border-radius: 8px;
-  max-height: 50vh;
-  min-width: 100%;
-  max-width: calc(100vw - 40px);
   overflow: hidden;
   position: relative;
 
@@ -60,7 +55,6 @@ const CollectionLink = styled(Link)`
   &:active,
   &:focus {
     background: ${(props) => props.theme.listItemHoverBackground};
-    outline: none;
 
     ${SecondaryActions} {
       opacity: 1;
@@ -77,8 +71,6 @@ const Heading = styled.h3`
   overflow: hidden;
   white-space: nowrap;
   color: ${(props) => props.theme.text};
-  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen,
-    Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
 `;
 
 const Title = styled(Highlight)`
@@ -93,4 +85,4 @@ const IconWrapper = styled.span`
   height: 24px;
 `;
 
-export default withTheme(CollectionPreview);
+export default CollectionPreview;
--- a/app/components/CollectionPreview.js
+++ b/app/components/CollectionPreview.js
@@ -1,0 +1,96 @@
+// @flow
+import * as React from "react";
+import { Link } from "react-router-dom";
+import styled, { withTheme } from "styled-components";
+import Collection from "models/Collection";
+import CollectionIcon from "components/CollectionIcon";
+import Flex from "components/Flex";
+import Highlight from "components/Highlight";
+import CollectionMenu from "menus/CollectionMenu";
+
+type Props = {
+  collection: Collection,
+  highlight?: ?string,
+};
+
+class CollectionPreview extends React.Component<Props> {
+  render() {
+    const { collection, highlight } = this.props;
+
+    return (
+      <CollectionLink to={collection.url}>
+        <Heading>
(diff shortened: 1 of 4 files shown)


app/components/CollectionPreview.js
```diff
@@ -0,0 +1,96 @@
...
+import Flex from "components/Flex";
+import Highlight from "components/Highlight";
+import CollectionMenu from "menus/CollectionMenu";
+
+type Props = {
+  collection: Collection,
+  highlight?: ?string,
+};
+
+class CollectionPreview extends React.Component<Props> {
```
tommoor, line 16: Lets make this a functional component, trying to move towards them for new stuff.

app/components/CollectionPreview.js
```diff
@@ -0,0 +1,96 @@
...
+  text-overflow: ellipsis;
+`;
+
+const IconWrapper = styled.span`
+  margin-left: -4px;
+  margin-right: 4px;
+  height: 24px;
+`;
+
+export default withTheme(CollectionPreview);
```
tommoor, line 96: withTheme is not required here

app/components/CollectionPreview.js
```diff
@@ -0,0 +1,96 @@
...
+const Heading = styled.h3`
+  display: flex;
+  align-items: center;
+  height: 24px;
+  margin-top: 0;
+  margin-bottom: 0.25em;
+  overflow: hidden;
+  white-space: nowrap;
+  color: ${(props) => props.theme.text};
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen,
```
tommoor, line 80: No need to override the font-family

app/components/CollectionPreview.js
```diff
@@ -0,0 +1,96 @@
...
+`;
+
+const CollectionLink = styled(Link)`
+  display: flex;
+  margin: 10px -8px;
+  padding: 6px 8px;
+  border-radius: 8px;
+  max-height: 50vh;
+  min-width: 100%;
+  max-width: calc(100vw - 40px);
```
tommoor, line 51: Can we use `Flex` / flexbox to achieve this layout? It's a very simple lockup and should need so many defined widths etc.

app/components/CollectionPreview.js
```diff
@@ -0,0 +1,96 @@
...
+
+  ${SecondaryActions} {
+    opacity: 0;
+  }
+
+  &:hover,
+  &:active,
+  &:focus {
+    background: ${(props) => props.theme.listItemHoverBackground};
+    outline: none;
```
tommoor, line 63: Don't disable the `outline`

app/components/CollectionPreview.js
```diff
@@ -0,0 +1,96 @@
...
+  margin-bottom: 0.25em;
+  overflow: hidden;
+  white-space: nowrap;
+  color: ${(props) => props.theme.text};
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen,
+    Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
+`;
+
+const Title = styled(Highlight)`
+  max-width: 90%;
```
tommoor, line 85: Why 90%?

app/stores/CollectionsStore.js
```diff
@@ -88,6 +91,28 @@ export default class CollectionsStore extends BaseStore<Collection> {
     });
   }
 
+  searchResults(query: string): CollectionSearchResult[] {
+    return this.searchCache.get(query) || [];
+  }
+
+  @action
+  search = async (query: string): Promise<CollectionSearchResult[]> => {
```
tommoor, line 99: No need for this method to be `async` or have the `@action` decorator here

app/stores/CollectionsStore.js
```diff
@@ -88,6 +91,28 @@ export default class CollectionsStore extends BaseStore<Collection> {
...
+        if (!collection.name.toLowerCase().includes(query.toLowerCase())) {
+          return null;
+        }
+
+        return {
+          collection,
+        };
+      })
+    );
+    this.searchCache.set(query, results);
```
tommoor, line 111: Let's just return the results directly from the method? The cache makes sense when searching documents because it's done async against the server – when we're just searching the local store it's unnecessary indirection.

app/scenes/Search/Search.js
```diff
@@ -229,9 +234,16 @@ class Search extends React.Component<Props> {
   };
 
   render() {
-    const { documents, notFound, location } = this.props;
+    const { documents, notFound, location, collections } = this.props;
     const results = documents.searchResults(this.query);
-    const showEmpty = !this.isFetching && this.query && results.length === 0;
+    const collectionResults = !this.collectionId
```
tommoor, line 239: As noted below – just call `const collectionResults = this.props.collections.search(this.query)` directly here 😄

tommoor (changes requested):

app/stores/CollectionsStore.js
renanfilipe: Understood

app/components/CollectionPreview.js
renanfilipe: My mistake, there was no need to set these properties here, so I'll just remove them.

app/components/CollectionPreview.js
renanfilipe: I think it would be nice to have some space between the title and the secondary action, so we can avoid situations like this one here:
image

tommoor: Totally, but this can be achieved with Flexbox only!

Honestly this is our fault for not having a more cohesive component system that you could just use here.

renanfilipe: The icon is now centered and the `max-width: 90%` is gone. Please review the last changes.

app/stores/CollectionsStore.js
```diff
@@ -88,6 +89,22 @@ export default class CollectionsStore extends BaseStore<Collection> {
     });
   }
 
+  search = (query: string): CollectionSearchResult[] => {
```
tommoor, line 92: Just return an array of `Collection[]` from this method, no need for the `SearchResult` object in this case as there is no ranking or context etc.

app/stores/CollectionsStore.js
```diff
@@ -88,6 +89,22 @@ export default class CollectionsStore extends BaseStore<Collection> {
...
+    const results: CollectionSearchResult[] = compact(
+      this.orderedData.map((collection) => {
```
tommoor, line 94: Use `filter` instead of `map` and you can also avoid needing `compact` then…

app/types/index.js
```diff
@@ -54,3 +55,7 @@ export type SearchResult = {
   context: string,
   document: Document,
 };
+
+export type CollectionSearchResult = {
```
tommoor, line 59: This type will no longer be needed

app/scenes/Search/Search.js
```diff
@@ -319,6 +329,18 @@ class Search extends React.Component<Props> {
               mode={ArrowKeyNavigation.mode.VERTICAL}
               defaultActiveChildIndex={0}
             >
+              {collectionResults.map((result) => {
+                const collection = collections.data.get(result.collection.id);
```
tommoor, line 333: No need to get the collection from the store again, we already have it in `collectionResults`

app/components/CollectionPreview.js
```diff
@@ -0,0 +1,81 @@
...
+function CollectionPreview(props: Props) {
+  const { collection, highlight } = props;
+
+  return (
+    <CollectionLink to={collection.url}>
+      <Heading>
+        <IconWrapper>
+          <CollectionIcon collection={collection} />
+        </IconWrapper>
+        <Title text={collection.name} highlight={highlight} />
```
tommoor, line 24: Thoughts on rendering a summary of the collection description to more closely match the document search results?

tommoor (changes requested):
Almost there

app/components/CollectionPreview.js
renanfilipe: It makes more sense to me to also allow the collection description to be searchable instead of just showing it.
cc @MatheusRV

tommoor: I don't mind including it in this PR and just landing things the "right way" the first time to be honest. 

This does mean that we should really use the `searchVector` that's stored for collections in the same way on the server as we do for documents, this simplifies things a little as all search results can come from the server and be treated the same instead of some being local.

But, I'm not sure of the best way to deal with pagination with this approach… any thoughts?

renanfilipe: maybe we can search in the collections first and based on the number of results there we can limit the documents, so we never pass a total of X results for instance.
I just realized that this task is not covering the searches made through slack, is it intentional?

tommoor: I like the idea in theory if you have a good strategy but I any approach that involves coordinating two queries does make pagination extremely difficult… :( I'm struggling to come up with a great way to deal with it beyond the first `offset`.

> I just realized that this task is not covering the searches made through slack, is it intentional?

Yea, we're cool to leave Slack out of this for now.

MatheusRV:
@tommoor could you please finish this for us?

tommoor:
You can now search collections by typing `o` `c`, it will open the command bar in the right place

PR #1551, opened 2020-09-16
feat: Support importing .docx or .html files as new documents

### LINKED ISSUE #1339: Add support for docx import

tommoor:
Based on https://github.com/outline/outline/pull/1469
Closes #1339 

Adds image attachment support and some refactoring


tommoor:
@whatl3y I ended up doing a little refactoring but your PR provided a really solid base, main differences are:

- Moved the majority of the logic to a `documentImporter` command to avoid cluttering the controller layer
- Added parsing and uploading of embedded images
- Moved the old clientside import to the server so all importing logic lives in `documentImporter` now

whatl3y:
@tommoor awesome looks great, thanks for finishing this up and apologies I couldn't wrap it up last/this week. I'm glad you were able to get to it, I believe [this](https://github.com/outline/outline/pull/1551/commits/55b1eef308bf95f56dddb6eea5205632777e29fb#diff-a845a0d8197cedf079618134f5ea8091R754) is what would've taken me a while to wire up since I would either have taken forever to find the existing functions or written my own that wouldn't have been as nicely written and tested as the ones that already exist.

Looking forward to using the new functionality and hopefully contributing again in the future!

PR #1554, opened 2020-09-16
feat: Record search queries

### LINKED ISSUE #1392: Record search queries in the database
So that it is possible to create reports on which terms are most searched within the platform, it would be convenient whenever researched to create a record in the database.

renanfilipe:
closes #1392

We can add tests later when the logic gets approved.

### DIFF
--- a/server/api/documents.test.js
+++ b/server/api/documents.test.js
@@ -8,6 +8,7 @@ import {
   Revision,
   Backlink,
   CollectionUser,
+  SearchQuery,
 } from "../models";
 import {
   buildShare,
@@ -953,6 +954,23 @@ describe("#documents.search", () => {
 
     expect(res.status).toEqual(401);
     expect(body).toMatchSnapshot();
+  });
+
+  it("should save search term, hits and source", async (done) => {
+    const { user } = await seed();
+    await server.post("/api/documents.search", {
+      body: { token: user.getJwtToken(), query: "my term" },
+    });
+
+    setTimeout(async () => {
+      const searchQuery = await SearchQuery.findAll({
+        where: { query: "my term" },
+      });
+      expect(searchQuery.length).toBe(1);
+      expect(searchQuery[0].results).toBe(0);
+      expect(searchQuery[0].source).toBe("app");
+      done();
+    }, 100);
   });
 });
 
--- a/server/api/documents.test.js
+++ b/server/api/documents.test.js
@@ -962,6 +962,8 @@ describe("#documents.search", () => {
       body: { token: user.getJwtToken(), query: "my term" },
     });
 
+    // setTimeout is needed here because SearchQuery is saved asynchronously
+    // in order to not slow down the response time.
     setTimeout(async () => {
       const searchQuery = await SearchQuery.findAll({
         where: { query: "my term" },
--- a/server/models/Document.js
+++ b/server/models/Document.js
@@ -241,14 +241,13 @@ Document.findByPk = async function (id, options = {}) {
   }
 };
 
-type SearchResult = {
-  data: {
+type SearchResponse = {
+  results: {
     ranking: number,
     context: string,
     document: Document,
   }[],
   totalCount: number,
-  currentPage: number,
 };
 
 type SearchOptions = {
@@ -271,7 +270,7 @@ Document.searchForTeam = async (
   team,
   query,
   options: SearchOptions = {}
-): Promise<SearchResult> => {
+): Promise<SearchResponse> => {
   const limit = options.limit || 15;
   const offset = options.offset || 0;
   const wildcardQuery = `${escape(query)}:*`;
@@ -279,7 +278,7 @@ Document.searchForTeam = async (
 
   // If the team has access no public collections then shortcircuit the rest of this
   if (!collectionIds.length) {
-    return { data: [], totalCount: 0, currentPage: 0 };
+    return { results: [], totalCount: 0 };
   }
 
   // Build the SQL query to get documentIds, ranking, and search term context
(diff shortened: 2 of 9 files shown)


server/models/SearchQuery.js
```diff
@@ -0,0 +1,65 @@
...
+    as: "user",
+    foreignKey: "userId",
+  });
+  SearchQuery.belongsTo(models.Team, {
+    as: "team",
+    foreignKey: "teamId",
+  });
+};
+
+SearchQuery.saveQuery = async function (searchQuerie) {
```
renanfilipe, line 43: Should this method be a `beforeCreate` hook?

server/models/SearchQuery.js
```diff
@@ -0,0 +1,65 @@
...
+    as: "user",
+    foreignKey: "userId",
+  });
+  SearchQuery.belongsTo(models.Team, {
+    as: "team",
+    foreignKey: "teamId",
+  });
+};
+
+SearchQuery.saveQuery = async function (searchQuery) {
```
renanfilipe, line 43: Should this method be a beforeCreate hook?

tommoor: I don't think this method needs to exist, `SearchQuery.create` can be called directly from the controllers.

The key thing that was misunderstood is that the `results` should record the number of search results that were returned to the user, not the number of times the user has performed the search

renanfilipe: Got it, I'll fix it.

tommoor:
Everything here looks good, however with recording only `results.length` it means we'll be recording only a page of results so if the limit is set to `5` as it is with the Slack integration the most it will ever be is 5, when really what we want to know is what is the **total number** of results for the query at that time.

To achieve this we'd need to update the `searchForUser` and `searchForTeam` methods to return a total count as well as the current page of results – It's a bit tricky but not too bad, there's an inbuilt helper in sequelize that does similar:

https://github.com/sequelize/sequelize/blob/a2d75e6edd477b6167636d2a7e446346d666b9d9/src/model.js#L2193

Unfortunately we can't use this here because the search queries are so custom, but it could use a similar approach where there is a count version and the normal query that are run in parallel with `Promise.all`

renanfilipe:
> Everything here looks good, however with recording only `results.length` it means we'll be recording only a page of results so if the limit is set to `5` as it is with the Slack integration the most it will ever be is 5, when really what we want to know is what is the **total number** of results for the query at that time.
> 
> To achieve this we'd need to update the `searchForUser` and `searchForTeam` methods to return a total count as well as the current page of results – It's a bit tricky but not too bad, there's an inbuilt helper in sequelize that does similar:
> 
> https://github.com/sequelize/sequelize/blob/a2d75e6edd477b6167636d2a7e446346d666b9d9/src/model.js#L2193
> 
> Unfortunately we can't use this here because the search queries are so custom, but it could use a similar approach where there is a count version and the normal query that are run in parallel with `Promise.all`

Hello, I've updated the methods to return the `totalCount` and `currentPage` as requested. Now we are for sure saving the total number of results. :)

server/models/Document.js
```diff
@@ -242,9 +242,13 @@ Document.findByPk = async function (id, options = {}) {
 };
 
 type SearchResult = {
```
tommoor, line 244: This type is no longer named well, `SearchResponse` perhaps?

server/models/Document.js
```diff
@@ -422,13 +471,17 @@ Document.searchForUser = async (
...
+  return {
+    data: map(results, (result) => ({
+      ranking: result.searchRanking,
+      context: removeMarkdown(unescape(result.searchContext), {
+        stripHTML: false,
+      }),
+      document: find(documents, { id: result.id }),
+    })),
+    totalCount: count,
+    currentPage: Math.floor(offset / limit),
```
tommoor, line 483: I think we can just lose `currentPage` if we're not using it for anything – it's just unused code to maintain 😉

server/models/Document.test.js
```diff
@@ -201,7 +201,8 @@ describe("#searchForTeam", () => {
       title: "test",
     });
 
-    const results = await Document.searchForTeam(team, "test");
+    const response = await Document.searchForTeam(team, "test");
+    const { data: results } = response;
```
tommoor, line 205: Seems like we're mapping to `results` everywhere, maybe just make `data` -> `results` in the response to begin with?

tommoor (commented):
Just some minor nits, ready for tests for sure – nice work

server/models/Document.js
renanfilipe: yep, I'll change it.

server/models/Document.js
renanfilipe: sure, np

server/models/Document.test.js
renanfilipe: I don't mind changing it, I used data because it sounded more standard than results for an API response.

server/api/documents.test.js
```diff
@@ -954,6 +955,23 @@ describe("#documents.search", () => {
...
     expect(body).toMatchSnapshot();
   });
+
+  it("should save search term, hits and source", async (done) => {
+    const { user } = await seed();
+    await server.post("/api/documents.search", {
+      body: { token: user.getJwtToken(), query: "my term" },
+    });
+
+    setTimeout(async () => {
```
tommoor, line 965: These setTimeout's shouldn't be needed, take a look at the other tests for patterns. Other than that I think we're good to go here!

renanfilipe: I'm sorry, but can you please point an example? the setTimeout is here because we save the search asynchronously, which means that the request ends before the record gets saved into the table.

tommoor: Ah I see, yea this is non-obvious but makes sense now, perhaps just a comment will do then to explain why the timeout is needed.

renanfilipe: got it

tommoor (approved):

PR #1573, opened 2020-09-25
feat: Improved search results when finding links in document editor

tommoor:
Moves searching for a document via the link editor to it's own endpoint. There are a couple of reasons for this:

- We're now recording search queries and these shouldn't be mixed in with link searches, it will improve the signal to noise ratio.
- Returning results that match based on content of the document doesn't make much sense for this context
- Faster!

### DIFF
--- a/app/scenes/Document/components/DataLoader.js
+++ b/app/scenes/Document/components/DataLoader.js
@@ -1,6 +1,7 @@
 // @flow
 import distanceInWordsToNow from "date-fns/distance_in_words_to_now";
 import invariant from "invariant";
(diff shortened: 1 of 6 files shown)


app/scenes/Document/components/DataLoader.js
```diff
@@ -73,6 +74,8 @@ class DataLoader extends React.Component<Props> {
   }
 
   onSearchLink = async (term: string) => {
+    term = term.trim();
```
tommoor, line 77: Remove with next prerelease of rich-markdown-editor

PR #1579, opened 2020-09-28
fix: spotify widget wrong size

obynio:
Hello, while using outline I noticed the spotify iframe elements were using a wrong size. The spotify widgets are all using the same size either for tracks, playlist, album and podcast. This results in a lot of wasted space on the wiki page.

This fix allows to create widgets with specific size depending if it's a podcast episode, a song track, a playlist or an album following [the spotify recommandations](https://developer.spotify.com/documentation/widgets/).

Feel free to comment 👍  

## Before
before

## After
after

### DIFF
--- a/app/embeds/Spotify.js
+++ b/app/embeds/Spotify.js
@@ -25,11 +25,21 @@ export default class Spotify extends React.Component<Props> {
   render() {
     const normalizedPath = this.pathname.replace(/^\/embed/, "/");
 
+    var height;
+
+    if (normalizedPath.includes("episode") || normalizedPath.includes("show")) {
+      height = 232;
+    } else if (normalizedPath.includes("track")) {
+      height = 80;
+    } else {
+      height = 380;
+    }
+
     return (
       <Frame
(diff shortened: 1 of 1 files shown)


tommoor:
Hey! Just got to reviewing this… definitely like the idea but when I test locally I'm seeing podcasts have a large gap at the bottom – I believe it's the difference between an entire podcast and an individual show?

https://open.spotify.com/show/1x8V6y5CPDh2aYPThICNXR?si=qOgQS0vqQdeU-fRD_jqVqA

image

obynio:
Oh damn, why this gap ? 🤔 
I'll take a look asap to try to fix this 🙏

obynio:
Ah yes you're totally right, I'm checking for the string `episode` but I was not expecting the whole podcast thing with the string `show`. I'm fixing this.

obynio:
@tommoor should now be fixed, give it a try 👍

tommoor:
Great, thanks for the contribution!

PR #1594, opened 2020-10-17
add backlinks to slack unfurls

thenanyu:
- [ ] will need to update instructions in readme and verify smallest possible set of oauth permissions to configure
- [ ] update the UI to ask the user to re-connect their slack app if their tokens are not up to date

---

- [ ] change the disconnect button to warning and the connect with slack button to primary maybe
- [ ] make user user id optional on backlinks
- [ ] make reverse document Id optional
- [ ] add type, json blob, timestamp, permalink, to backend
- [ ] add backlinks api endpoint
- [ ] backlink presenter schema
- [ ] backlink UI renderers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-
-
 <p align="center">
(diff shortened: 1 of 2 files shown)


tommoor:
Still liking this idea, pretty unique

PR #1596, opened 2020-10-19
fix: Dark mode inline code styling fixed #1593

### LINKED ISSUE #1593: dark mode inline monospace styling is very low contrast
image

This might be an editor issue?

RotonEvan:
background changed to pitch black

### DIFF
--- a/app/scenes/Settings/Slack.js
+++ b/app/scenes/Settings/Slack.js
@@ -136,7 +136,7 @@ const ListItem = styled.li`
 const Code = styled.code`
   padding: 4px 6px;
   margin: 0 2px;
-  background: #eaebea;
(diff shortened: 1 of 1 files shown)


app/scenes/Settings/Slack.js
```diff
@@ -136,7 +136,7 @@ const ListItem = styled.li`
 const Code = styled.code`
   padding: 4px 6px;
   margin: 0 2px;
-  background: #eaebea;
+  background: #2c2b2c;
```
thenanyu, line 139: I think this might break light mode? I believe this should take into consideration theme

tommoor: `props.theme.codeBackground` should do it

PR #1598, opened 2020-10-20
feat: Debounce notification emails by 5 minutes

### LINKED ISSUE #1002: Throttle email notifications
We've reached the limit of what the current dumb system is capable of, it's time to put email notifications on a delayed job to ensure that we don't over-notify.

tommoor:
This work took priority as setting up these debounced systems is an enabler for realtime collaboration.

closes #1002

### DIFF
--- a/server/api/notificationSettings.js
+++ b/server/api/notificationSettings.js
@@ -2,7 +2,7 @@
 import Router from "koa-router";
 
 import auth from "../middlewares/authentication";
-import { NotificationSetting } from "../models";
+import { Team, NotificationSetting } from "../models";
 import policy from "../policies";
 import { presentNotificationSetting } from "../presenters";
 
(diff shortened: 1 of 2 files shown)


server/api/notificationSettings.js
```diff
@@ -62,16 +62,23 @@ router.post("notificationSettings.unsubscribe", async (ctx) => {
...
+        model: Team,
+        required: true,
+        as: "team",
+      },
+    ],
+  });
 
+  if (setting && setting.unsubscribeToken === token) {
     await setting.destroy();
+    ctx.redirect(`${setting.team.url}/settings/notifications?success`);
```
tommoor, line 77: fix: Only tangentially related fix – I noticed we were redirecting to the main domain instead of the correct team subdomain here.

server/services/notifications.js
```diff
@@ -9,29 +10,21 @@ import {
...
+notificationsQueue.process(async (job) => {
+  const event = job.data;
 
+  try {
     const document = await Document.findByPk(event.documentId);
     if (!document) return;
 
+    // If the document has been updated since we initially queued a notification
+    // abort sending a notification – this functions as a debounce.
+    if (document.updatedAt > new Date(event.createdAt)) return;
```
tommoor, line 26: 🍖  Compare the last updated at for the document to the event, if it was updated after the event was fired then we can discard this job.

tommoor: @thenanyu I think this is a pretty simple approach to a debounce – see anything amiss?

PR #1622, opened 2020-11-05
fix: gist not support username contains numbers

n3n:
Problem REGEX: https://regex101.com/r/Tm3EFc/1/

Fixed REGEX: https://regex101.com/r/3es1Mf/2


tommoor (approved):
Wow, good find – it's been like that for _years_ 😢

PR #1624, opened 2020-11-06
feat: i18n

glaand:
i've created a small MVP of outline with i18n. It uses i18next package along with a parser, which reads all .js files and generates a JSON with all keys and values.

With the command `yarn build:i18n` you can build the JSON Translation file automatically.

To try it out, please set the env variable `DEFAULT_LANGUAGE` to either `en_US` or `de_DE`.

Since outline can manage more than one team, i've added a variable named as `language`of type *string* to the model: *Team*.

The migration file sets the default language value depending on the env variable ``DEFAULT_LANGUAGE`.

At the moment, only the `app` is ready to be translated, the server still needs some bone structuring.

If the team admin wants to change the language, he can on the `DetailsPage` change it with the InputSelect Element.

Can someone please try this out?

grafik


tommoor:
Hey, thanks for taking this on – I have two main bits of feedback:

- Language should be a user setting, not a team one. I already know of several mixed language teams that use Outline heavily. On those lines does it need to be a database column or can we just rely on reading the language that the browser tells us to use?

- I'd prefer to keep dynamic/shared content out of the translations, here that mainly means the onboarding documents and default collection description. Other UI copy makes total sense to have in there.

glaand:
Hey @tommoor 

Thanks for the feedback. 

 - In this way it makes sense that the langauge can be changed at user-level. My concern was that if multiple persons in the same team use different languages, they won't understand functionality between each other but that is an organisational problem out of scope from outline to be solved. 
- Yes, using the browser to recognize the language and override it on database-level should be the correct solution.
- Alright, i will remove the dynamic content out of dynamic translation ( Error of mine 🤦 )

glaand:
I guess for reducing complexity for now, we can let English as default language and the user could change it afterwards on profile.

The browser detection feature we could add in the future in a new `feat`. What do you say @tommoor 

Try to check te new commit :) , now you can change the language on the `ProfilePage`

tommoor:
> The browser detection feature we could add in the future in a new feat.

That sounds like the sensible way to go, probably best to keep language changing optional for a few months until more translations are built out anyway.

app/scenes/Settings/Profile.js
```diff
@@ -44,8 +48,12 @@ class Profile extends React.Component<Props> {
     await this.props.auth.updateUser({
       name: this.name,
       avatarUrl: this.avatarUrl,
+      language: this.language,
     });
-    this.props.ui.showToast("Profile saved");
+    this.props.ui.showToast(t("Profile saved"));
+    setTimeout(function () {
+      window.location.reload();
+    }, 1000);
```
tommoor, line 56: If we use the hook from `react-i18next` then that should avoid the need to reload the entire app here, we can just call `changeLanguage` on save.

i18next-parser.config.js
```diff
@@ -0,0 +1,83 @@
...
+    // see below for more details
+    lexers: {
+      hbs: ['HandlebarsLexer'],
+      handlebars: ['HandlebarsLexer'],
+  
+      htm: ['HTMLLexer'],
+      html: ['HTMLLexer'],
+  
+      mjs: ['JavascriptLexer'],
+      js: ['JavascriptLexer'], // if you're writing jsx inside .js files, change this to JsxLexer
```
tommoor, line 33: Looks like this should be "JsxLexer"

server/models/Team.js
```diff
@@ -142,6 +142,7 @@ Team.prototype.provisionFirstCollection = async function (userId) {
     name: "Welcome",
     description:
       "This collection is a quick guide to what Outline is all about. Feel free to delete this collection once your team is up to speed with the basics!",
+    type: "atlas",
```
tommoor, line 145: Can remove this, guess you might have been on an old fork at one point?

shared/translations/de_DE.json
```diff
@@ -0,0 +1,19 @@
+{
+    "Home": "Startseite",
+    "Recently updated": "Zuletzt geändert",
+    "Recently viewed": "Zuletzt gesehen",
+    "Created by me": "Von mir erstellt",
+    "Profile saved": "Profil gespeichert",
+    "Profile picture updated": "Profilfoto wurde aktualisiert.",
+    "Unable to upload new avatar": "Neuer Avatar kann nicht hochgeladen werden.",
```
tommoor, line 8: While we're here this should change to profile picture to match the other language

```suggestion
    "Unable to upload new profile picture": "Neuer profilfoto kann nicht hochgeladen werden",
```

shared/translations/de_DE.json
```diff
@@ -0,0 +1,19 @@
+{
+    "Home": "Startseite",
+    "Recently updated": "Zuletzt geändert",
+    "Recently viewed": "Zuletzt gesehen",
+    "Created by me": "Von mir erstellt",
+    "Profile saved": "Profil gespeichert",
+    "Profile picture updated": "Profilfoto wurde aktualisiert.",
```
tommoor, line 7:
```suggestion
    "Profile picture updated": "Profilfoto wurde aktualisiert",
```

shared/translations/de_DE.json
```diff
@@ -0,0 +1,19 @@
...
+    "Unable to upload new avatar": "Neuer Avatar kann nicht hochgeladen werden.",
+    "Profile": "Profil",
+    "Photo": "Foto",
+    "Upload": "Hochladen",
+    "Full name": "Vorname",
+    "Language": "Sprache",
+    "Saving…": "Am speichern…",
+    "Save": "Speichern",
+    "Delete Account": "Konto löschen",
+    "You may delete your account at any time, note that this is unrecoverable": "Sie können Ihr Konto jederzeit löschen. Beachten Sie, dass dies nicht wiederherstellbar ist",
```
tommoor, line 17: German question – is it normal to have capital letters in the middle of sentences like this, or a mistake?

shared/translations/default.json
```diff
@@ -0,0 +1,19 @@
+{
+  "Home": "Home",
+  "Recently updated": "Recently updated",
+  "Recently viewed": "Recently viewed",
+  "Created by me": "Created by me",
+  "Profile saved": "Profile saved",
+  "Profile picture updated": "Profile picture updated",
+  "Unable to upload new avatar": "Unable to upload new avatar",
```
tommoor, line 8:
```suggestion
  "Unable to upload new profile picture": "Unable to upload new profile picture",
```

shared/translations/i18n.js
```diff
@@ -0,0 +1,29 @@
...
+    en_US: {
+      translation: en_US,
+    },
+    de_DE: {
+      translation: de_DE,
+    },
+  },
+});
+
+const t = i18n.getFixedT();
```
tommoor, line 27: As mentioned above, lets use the hook from the react-i18n instead of exporting a fixed t.

tommoor (commented):
This is getting me excited for the multi-lingual future 🙏

app/scenes/Settings/Profile.js
glaand: done - changed to HOC *withTranslation*

i18next-parser.config.js
glaand: changed

server/models/Team.js
glaand: ups yes :D

shared/translations/de_DE.json
glaand: Left, the english version, is the string key which is in the source. Cannot change the text as you wish since its a key. gonna let it as it is

shared/translations/de_DE.json
glaand: changed

shared/translations/de_DE.json
glaand: Haha - i feel you. Yes, its a very german thing to do :)

The funny part is, that most foreigners have trouble writing german because of misusing of capitalisation :D

shared/translations/default.json
glaand: Again - CAn't be changed

shared/translations/i18n.js
glaand: changed

glaand:
@tommoor For some reason the test build is not working, do you have any ideas why?

tommoor:
Run `yarn run test:server -u` locally to update the snapshots should do it.

shared/translations/default.json
tommoor: I realize it shouldn't be changed here, but in the source file?

package.json
```diff
@@ -5,11 +5,12 @@
   "main": "index.js",
   "scripts": {
     "clean": "rimraf build",
+    "build:i18n": "i18next 'app/**/*.js' 'server/**/*.js' ",
     "build:server": "babel -d ./build/server ./server && babel -d ./build/shared ./shared && cp package.json ./build && ln -sf \"$(pwd)/webpack.config.dev.js\" ./build",
     "build:webpack": "webpack --config webpack.config.prod.js",
     "build": "yarn clean && yarn build:webpack && yarn build:server",
     "start": "node ./build/server/index.js",
-    "dev": "nodemon --exec \"yarn build:server && node build/server/index.js\" -e js --ignore build/ --ignore app/",
+    "dev": "nodemon --exec \"yarn build:server && cp -R shared/translations/*.json build/shared/translations && node build/server/index.js\" -e js --ignore build/ --ignore app/",
```
tommoor, line 13: This copy step should probably happen as a sub-command of `build:server`, as it'll be needed in production too

tommoor:
Looking good, the question becomes how much translation should be done before merging this. I think at least the core menu's and perhaps the editor copy would be a reasonable minumum

shared/translations/default.json
glaand: Yes sure, i will do it

package.json
glaand: done

glaand:
> 
> 
> Looking good, the question becomes how much translation should be done before merging this. I think at least the core menu's and perhaps the editor copy would be a reasonable minumum

I will try to label the maximum as possible and translate everything to german and portuguese :)

I will also write tests for i18n.

tommoor:
Note for the editor, it has support for passing a `dictionary` prop – which will be used to translate all the UI internal to the editor itself:
https://github.com/outline/rich-markdown-editor#dictionary

glaand:
There are thousand of labels, i really would appreciate help from something wrapping labels with the `t` function from i18n.
Now we have translations for German (Germany) and Portuguese (Portugal)

glaand:
@tommoor After the merge we should think on setting up a process for translators  to be able to translate the documents without technical knowledge?

tommoor:
> There are thousand of labels, i really would appreciate help from something wrapping labels with the t function from i18n.

I've spotted some patterns already that can reduce the number of labels to translate, I'll comment in the code. I do think it's wise to get this merged asap as it'll cause conflicts with almost all other work, in fact there is already some conflicts with a refactor I did to the menus in another PR.

In order to achieve this, I might suggest that we aim to merge in the next few days – but with the UI for switching languages commented out or marked as experimental/beta. Once the `t` method is on the default branch it'll be easier.

> setting up a process for translators to be able to translate the documents without technical knowledge?

Yep, absolutely. I guess first version they can take the en_US and translate the contents into a new file.

app/components/Avatar/AvatarWithPresence.js
```diff
@@ -37,20 +39,28 @@ class AvatarWithPresence extends React.Component<Props> {
...
             <Centered>
-              <strong>{user.name}</strong> {isCurrentUser && "(You)"}
-              <br />
-              {isPresent
-                ? isEditing
-                  ? "currently editing"
-                  : "currently viewing"
-                : `viewed ${distanceInWordsToNow(new Date(lastViewedAt))} ago`}
+              {t(
+                "<strong>{{ userName }} {{ you }} <br /> {{ action }}</strong>",
```
tommoor, line 51: We don't want to include JSX in this strings, right?

app/components/Breadcrumb.js
```diff
@@ -21,6 +22,8 @@ import Flex from "components/Flex";
 import BreadcrumbMenu from "./BreadcrumbMenu";
 import { collectionUrl } from "utils/routeHelpers";
 
+const t = (k) => i18n.t(k);
```
tommoor, line 25: Why not use the hook in here?

app/components/DropdownMenu/DropdownMenu.js
```diff
@@ -51,8 +53,9 @@ class DropdownMenu extends React.Component<Props> {
   ) => {
     return (ev: SyntheticMouseEvent<HTMLElement>) => {
       ev.preventDefault();
+      const { t } = this.props;
       const currentTarget = ev.currentTarget;
-      invariant(document.body, "why you not here");
+      invariant(document.body, t("why you not here"));
```
tommoor, line 58: This is a development warning, no need to translate

app/components/InputSearch.js
```diff
@@ -51,13 +53,13 @@ class InputSearch extends React.Component<Props> {
...
   render() {
-    const { theme, placeholder = "Search…" } = this.props;
+    const { theme, t } = this.props;
 
     return (
       <InputMaxWidth
         ref={(ref) => (this.input = ref)}
         type="search"
-        placeholder={placeholder}
+        placeholder={t("Search...")}
```
tommoor, line 62: Need to retain the ability to pass `placeholder` as a prop here. It's an ellipses (…) at the end of the string too, not three periods.

shared/translations/i18n.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+import i18n from "i18next";
+import { initReactI18next } from "react-i18next";
+
+import de_DE from "./de_DE.json";
+import en_US from "./default.json";
+import pt_PT from "./pt_PT.json";
+
+const outlineTranslation = {
```
tommoor, line 9: can we just export the `init` method here?

app/scenes/Collection.js
```diff
@@ -188,26 +190,31 @@ class CollectionScene extends React.Component<Props> {
             {collection.isEmpty ? (
               <Centered column>
                 <HelpText>
-                  <strong>{collection.name}</strong> doesn’t contain any
-                  documents yet.
-                  <br />
-                  Get started by creating a new one!
+                  {t(
```
tommoor, line 193: Another spot where JSX snuck into the string

app/components/Avatar/AvatarWithPresence.js
glaand: This double curly brackets are not JSX, those are only placeholders for the t function to render the variables dynamically.

grafik

app/components/Breadcrumb.js
glaand: I think the hook can only be used on functional components - Maybe i am wrong?

Can you help me with this?

app/components/DropdownMenu/DropdownMenu.js
glaand: removed

app/components/InputSearch.js
glaand: ups, my bad ... fixed

shared/translations/i18n.js
glaand: yes, done

app/components/Avatar/AvatarWithPresence.js
glaand: The HTML code is ok to be included

app/scenes/Collection.js
glaand: Again, same as above.

glaand: gonna fix it

app/components/Avatar/AvatarWithPresence.js
glaand: gonna fix it

app/components/Breadcrumb.js
tommoor: These components _are_ functional! 😄

glaand: Looool sry haha :D should have tried it first out haha

glaand: Fixed

app/components/Avatar/AvatarWithPresence.js
```diff
@@ -37,20 +39,28 @@ class AvatarWithPresence extends React.Component<Props> {
...
             <Centered>
-              <strong>{user.name}</strong> {isCurrentUser && "(You)"}
-              <br />
-              {isPresent
-                ? isEditing
-                  ? "currently editing"
-                  : "currently viewing"
-                : `viewed ${distanceInWordsToNow(new Date(lastViewedAt))} ago`}
+              <Trans
+                defaults="<strong>{{ userName }} {{ you }} <br /> {{ action }}</strong>"
```
tommoor, line 51: While I realise you _can_ put html in these strings, I don't think it's best practice and certainly stands out in the translation files as being odd.

Any reason we can't just translate the four strings separately here?

"you"
"currently editing"
"currently viewing"
"viewed {{ timeAgo }} ago"

app/components/Sidebar/components/DocumentLink.js
```diff
@@ -96,7 +99,7 @@ class DocumentLink extends React.Component<Props> {
         this.isActiveDocument())
     );
     const document = documents.get(node.id);
-    const title = node.title || "Untitled";
+    const title = node.title || t("Untitled");
```
tommoor, line 102: Nice benefit that we didn't store "Untitled" in the database here!

app/menus/TemplatesMenu.js
```diff
@@ -42,7 +44,10 @@ class TemplatesMenu extends React.Component<Props> {
             <div>
               <strong>{template.titleWithDefault}</strong>
               <br />
-              <Author>By {template.createdBy.name}</Author>
+              <Author>
+                {" "}
```
tommoor, line 48: Extra space was added here

app/scenes/Collection.js
```diff
@@ -188,26 +190,26 @@ class CollectionScene extends React.Component<Props> {
...
               <Centered column>
                 <HelpText>
-                  <strong>{collection.name}</strong> doesn’t contain any
-                  documents yet.
-                  <br />
-                  Get started by creating a new one!
+                  <Trans
+                    components={{ collectionName: collection.name }}
+                    defaults="<strong>{{ collectionName }}</strong> doesn’t contain any documents yet. <br /> Get started by creating a new one!"
+                  />
```
tommoor, line 196: If we wrap `Trans` around the components here the output in the translations will be nicer, see:

https://react.i18next.com/latest/trans-component

server/api/users.js
```diff
@@ -62,10 +62,11 @@ router.post("users.info", auth(), async (ctx) => {
 
 router.post("users.update", auth(), async (ctx) => {
   const { user } = ctx.state;
-  const { name, avatarUrl } = ctx.body;
+  const { name, avatarUrl, language } = ctx.body;
 
   if (name) user.name = name;
   if (avatarUrl) user.avatarUrl = avatarUrl;
+  if (language) user.language = language;
```
tommoor, line 69: Note: We'll want to validate in the model that the `language` key is one of the available languages, not a blocker for initial landing.

app/components/Avatar/AvatarWithPresence.js
tommoor: Note, the `<strong>` tag should only be around the username

glaand: The reason of this is because in some languages the order of the words changes.

In this case i guess the order doesn't change so much but i am not sure for asian languages like japanese and chinese .

If a translator wants to change the respective order, he only needs to do it in the translation files and not in the code.

Example:
EN: "You have edited the document successfully"
DE: "Du hast die Datei erfolgreich bearbeitet" -> word translated to EN would be "You have the document successfully edited".

glaand: For now i fixed the **strong**

app/components/Sidebar/components/DocumentLink.js
glaand: Indeed :)

app/menus/TemplatesMenu.js
glaand: fixed

app/scenes/Collection.js
glaand: fixed

server/api/users.js
glaand: Good point, i will connect it with the env variables

tommoor:
Do you mind If I merge this into a branch on the `outline` repo? I can contribute some changes before it reaches the default branch easier then. You can PR any further changes or improvements into the new branch

glaand:
No, just do it :)

Issue #880, opened 2019-01-22
Internationalization & Localization


glaand:
Title says everything!

tommoor:
Well, it says something 😆

What language are you after? Would you be willing to help translate?

glaand:
Hey Tom

Yes! I would help with European Portuguese and High German

Best regards

MSG-maniac:
Chinese，I can translate

Vinetos:
I can translate in French

tommoor:
Awesome folks, having these offers definitely bumps translation in priority 😄

vitorleal:
Brazilian Portugues and Spanish

MatheusRV:
@tommoor What about this implementation? Need some help?

tommoor:
Hey @MatheusRV , would be happy if you’d like to work on a PR for the foundation of this if you have experience. Currently no work has been prioritized on this front

MatheusRV:
@tommoor could you rename this as Internationalization & Localization (l10n) is so important as I18N.

https://www.w3.org/International/questions/qa-i18n/

dmumedical:
We have translation team and can help with german and russian, maybe also with arabic language.

glaand:
I will get my hands dirty with this feature. Anyway, i was the one who opened the issue 😆

Sunday you are going to hear/read more from me !

glaand:
Being developed at https://github.com/outline/outline/pull/1624

tommoor:
Hey folks, we're really close to closing this out now – if you'd like to help contribute some of the initial translations there are instructions here:

https://github.com/outline/outline/blob/9b573a2b8bc88c5374056fd707978be532283bc7/CONTRIBUTING.md#how-to-translate-outline-via-crowdin

mustafa-online:
I wanna start with Arabic Language, but it's now exist yet! how to add it ??

baditaflorin:
@tommoor I would want to help with Romanian translation, could you please add it to https://crowdin.com/ ?

tommoor:
@baditaflorin added now

PR #1653, opened 2020-11-14
feat: I18n

tommoor:
Integration branch for https://github.com/outline/outline/pull/1624
closes https://github.com/outline/outline/issues/880

image

**Additional translation and i18n work should be PR'd into this branch**


tommoor:
@glaand the frustrating thing right now is that flow still doesn't understand decorators, which means we have to use the hoc either in the non-decorator syntax or manually type `t` in every component that uses it, the former is probably the least of two evils

tommoor:
@glaand I also have a question about the tests – it seems like they are enforcing that all strings in the default language exist in all languages? This seems wrong unless the default language gets copied across to all languages when missing, we don't want to be in a situation where you can't add new text unless it's translated into all languages.

glaand:
> 
> 
> @glaand I also have a question about the tests – it seems like they are enforcing that all strings in the default language exist in all languages? This seems wrong unless the default language gets copied across to all languages when missing, we don't want to be in a situation where you can't add new text unless it's translated into all languages.

You are right. The reason i did this for now was to ensure that the languages were working fine without checking every string manually. 

Do you want to remove this tests or should i do it? @tommoor

shared/translations/i18n.js
```diff
@@ -20,18 +32,10 @@ const initI18n = () => {
...
-      },
-    },
+    resources,
   });
 };
 
-export { initI18n, i18n, en_US, de_DE, pt_PT };
+const languages: string[] = Object.keys(resources);
+
+export { initI18n, languages, i18n, en_US, de_DE, pt_PT };
```
glaand, line 41: Very good one here :P @tommoor

server/models/User.js
```diff
@@ -39,6 +40,9 @@ const User = sequelize.define(
     language: {
       type: DataTypes.STRING,
       defaultValue: process.env.DEFAULT_LANGUAGE,
+      validate: {
+        isIn: [languages],
```
glaand, line 44: So i guess this does the auto check of user.language is in languages @tommoor ?

tommoor: Yep, validation at the ORM level a bit more resilient. It throws a specific error type which returns the right error to the client.

tommoor:
**Do you want to remove this tests or should i do it? @tommoor**

Tests removed 👍

tommoor:
I think the main thing I'd like to figure out before merging this first pass is what the contribution flow looks like – what is a good way to work through missing translations? 🤔

glaand:
> 
> 
> I think the main thing I'd like to figure out before merging this first pass is what the contribution flow looks like – what is a good way to work through missing translations? 🤔

Right, better look it now. I've made a prototype of the future translation process :P , please take a look.

What do you think?

Please follow from process start til process end

Translation Process - Outline

tommoor:
Thanks for the diagram, that's very useful. The "push" step can certainly happen from the `ci` servers, perhaps in the future the "pull" step could be an automated PR too.

It's a shame that localize doesn'ts seem to offer a free plan for open source projects for their TMS seeing as it seems to be well integrated with react-i18n

glaand:
> 
> 
> Thanks for the diagram, that's very useful. The "push" step can certainly happen from the `ci` servers, perhaps in the future the "pull" step could be an automated PR too.
> 
> It's a shame that [locize](https://locize.com/) doesn't seem to offer a free plan for open source projects for their TMS seeing as it seems to be well integrated with react-i18next. Still, it's cheap so that might be the one to start with.

Yeah true. A bad business move but anyways... It looks like ABB uses locize, ABB its a well known company here in Switzerland (In the billions category) so i guess it's a good referal to take them.

Looks like it costs 5 dollars per month. How do you handle expenses at Outline @tommoor 

I guess we could choose locize for this yes

tommoor:
$5 is base price and charges per-word above that, but as long as it integrates well it's fine. I will setup an account that the Outline SaaS business will sponsor.

glaand:
> 
> 
> $5 is base price and charges per-word above that, but as long as it integrates well it's fine. I will setup an account that the Outline SaaS business will sponsor.

Alright cool thank you very much :)

tommoor:
I split the translations out from the js bundle to reduce initial load time, they're now served via HTTP. I'm happy to merge this now but we do still need a strategy for identifying which translations are needed

glaand:
@tommoor Thank you for the changes. Yes indeed we need!

Im gonna take a look at it today, then i will post a proposal here.

MatheusRV:
Is this ready for new languages? I can put in PT BR, if yes.

tommoor:
It is, however this branch will be merged to `develop` very soon. Translations should now be provided through the portal, see this file for instructions:

https://github.com/outline/outline/blob/8aa215dce350f209481444ee699e4e6da25fa214/CONTRIBUTING.md

Thanks!

MatheusRV:
@tommoor I can't add new languages.

image

tommoor:
Okay added pt_BR just now 😌

PR #1663, opened 2020-11-18
fix: Prevent API request for views data for deleted documents

### LINKED ISSUE #1659: Prevent API request for views data for deleted documents
**To Reproduce**
1. Go to a deleted document
2. Note that a request to `views.list` fails in the network inspector

**Expected behavior**
Don't make the request

iamsaumya:
This pull request Fixes #1659 . 

Thanks!

### DIFF
--- a/app/components/Collaborators.js
+++ b/app/components/Collaborators.js
@@ -20,7 +20,7 @@ type Props = {
 @observer
 class Collaborators extends React.Component<Props> {
   componentDidMount() {
-    if (!this.props.document.deletedAt) {
+    if (!this.props.document.isDeleted) {
       this.props.views.fetchPage({ documentId: this.props.document.id });
     }
   }


iamsaumya:
I think I screwed up with the commits in my forked repo. Please let me know if there is any workaround. 

Thanks!

app/components/Collaborators.js
```diff
@@ -20,7 +20,9 @@ type Props = {
 @observer
 class Collaborators extends React.Component<Props> {
   componentDidMount() {
-    this.props.views.fetchPage({ documentId: this.props.document.id });
+    if (!this.props.document.deletedAt) {
```
tommoor, line 23:
```suggestion
    if (!this.props.document.isDeleted) {
```

Lots of pre-computed attributes available on the model :) 

https://github.com/outline/outline/blob/develop/app/models/Document.js#L95

tommoor:
Thanks 👍

PR #1666, opened 2020-11-19
chore: Change response of shares.info response for unshared document

### LINKED ISSUE #1658: Change response of shares.info response for unshared document
It seems that this request being a 404 for unshared documents is confusing and has been reported as a bug on more than one occasion. 

Instead we should make it a `200` or `204` with an empty response.

iamsaumya:
This pull request Fixes #1658 

I have updated server/api/share.js to send 204 status for unshared documents.

### DIFF
--- a/app/stores/SharesStore.js
+++ b/app/stores/SharesStore.js
@@ -1,6 +1,6 @@
 // @flow
 import invariant from "invariant";
-import { sortBy, filter, find } from "lodash";
+import { sortBy, filter, find, isUndefined } from "lodash";
 import { action, computed } from "mobx";
 import Share from "models/Share";
 import BaseStore from "./BaseStore";
@@ -47,6 +47,8 @@ export default class SharesStore extends BaseStore<Share> {
 
     try {
       const res = await client.post(`/${this.modelName}s.info`, { documentId });
+      if (isUndefined(res)) return;
+
       invariant(res && res.data, "Data should be available");
 
       this.addPolicies(res.policies);
--- a/app/utils/ApiClient.js
+++ b/app/utils/ApiClient.js
@@ -110,6 +110,8 @@ class ApiClient {
 
       download(blob, trim(fileName, '"'));
       return;
+    } else if (success && response.status === 204) {
+      return;
     } else if (success) {
       return response.json();
     }
--- a/server/api/shares.js
+++ b/server/api/shares.js
@@ -30,7 +30,7 @@ router.post("shares.info", auth(), async (ctx) => {
         },
   });
   if (!share || !share.document) {
-    throw new NotFoundError();
+    return (ctx.response.status = 204);
   }
 
   authorize(user, "read", share);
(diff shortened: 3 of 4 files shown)


iamsaumya:
I see a lot of tests failing because they expect 404.

Following are the tests that are probably failing:
1. should not find share for different user
2. should not find revoked share
3. should not find share for deleted document

let me know how to go about this.

Thanks!

tommoor:
Update the tests to expect a 204, we also need to check that the frontend is capable of handling the 204 response – it might be expecting json and throw an error.

iamsaumya:
Updating the shares.info tests to expect 204 passes all the server tests. Though all the tests are passed, I am getting this weird error in a few tests and idk what to do with this. 

Screenshot from 2020-11-20 13-59-35

iamsaumya:
In frontend, I can observe some breaking. 

This is a snippet from ApiClient which expects the success response to always have some sort of data.

```js

    if (options.download && success) {
      const blob = await response.blob();
      const fileName = (
        response.headers.get("content-disposition") || ""
      ).split("filename=")[1];

      download(blob, trim(fileName, '"'));
      return;
    } else if (success) {
      return response.json();
    }
```
Responding with 204 without making any change will result in this error. 

Screenshot from 2020-11-20 13-56-21

So, I updated the logic a bit to this, trying to mock an empty JSON.

```js
    if (options.download && success) {
      const blob = await response.blob();
      const fileName = (
        response.headers.get("content-disposition") || ""
      ).split("filename=")[1];

      download(blob, trim(fileName, '"'));
      return;
    } else if (success && response.status === 204) {
      return {};
    } else if (success) {
      return response.json();
    }
```

However, then I am getting an error from invariant.

Screenshot from 2020-11-20 13-56-40

I am not sure where should I write the logic. One possible solution could be to check in the SharesStore if the res has data in it. I will update ApiClient and SharesStore respectively to the following. 

```js
    if (options.download && success) {
      const blob = await response.blob();
      const fileName = (
        response.headers.get("content-disposition") || ""
      ).split("filename=")[1];

      download(blob, trim(fileName, '"'));
      return;
    } else if (success && response.status === 204) {
      return { data: "no content" };
    } else if (success) {
      return response.json();
    }
```
```js
 try {
      const res = await client.post(`/${this.modelName}s.info`, { documentId });
      invariant(res && res.data, "Data should be available");

      if (res.data === "no content") {
        return;
      }

      return this.add(res.data);
    } finally {
      this.isFetching = false;
    }
```

iamsaumya:
If we will follow the above approach, I think it will intuitive to change the status to `200` with some message from the server itself.

tommoor:
I think ApiClient can be updated to check for the `204` response code and not attempt to parse the body as JSON

iamsaumya:
Do you mean `return undefined`? I am afraid the SharesStore expects the res to be defined and have `data` key when the response from the server is successful.

tommoor:
Right, that expectation would have to be changed too

iamsaumya:
Hey @tommoor, please review this pull request.

Thanks!

tommoor (approved):
Thanks, this is a good improvement 👍

Issue #1186, opened 2020-02-24
Add total user count to people management screen


tommoor:
It's useful to know. This request came through support.

mo-jinbuu:
Hey @tommoor is this still available? I'd like to implement this if possible.

thenanyu:
this looks to still be needed @mo-jinbuu — design is a bit of a question, I would say, start by adding counts to the `active` `admins` and `invited` tabs on the people management screen

mo-jinbuu:
Alright I'll get on it, I'll let you know if I need anything

PR #1672, opened 2020-11-24
feat: add total users to people management screen

timc1:
Closes https://github.com/outline/outline/issues/1186.

This PR adds the total users to the people management screen:

https://user-images.githubusercontent.com/12195101/106837862-3a42fd00-6669-11eb-9475-8af84734e49d.mp4

#### Testing
- Add an invite
- Add multiple invites
- Remove invites


server/api/users.js
```diff
@@ -47,9 +47,14 @@ router.post("users.list", auth(), pagination(), async (ctx) => {
...
   });
 
+  const total = await User.count({ where });
+
   ctx.body = {
-    pagination: ctx.state.pagination,
-    data: users.map((listUser) =>
+    pagination: {
+      ...ctx.state.pagination,
+      total,
```
timc1, line 55: 🍖

app/stores/UsersStore.js
```diff
@@ -77,6 +77,7 @@ export default class UsersStore extends BaseStore<User> {
     runInAction(`invite`, () => {
       res.data.users.forEach(this.add);
     });
+    this.total += res.data.sent.length;
```
timc1, line 80: I think this is the right way of handling the total count since it's an observable?

app/scenes/Settings/People.js
```diff
@@ -87,6 +89,10 @@ class People extends React.Component<Props> {
           Invite people…
         </Button>
 
+        <TotalCount $loading={totalUsers === 0}>
```
timc1, line 92: Looks like loading is managed within `Pagination`. This seems like a viable way to check since the totalUsers will always be at least 1

app/stores/UsersStore.js
tommoor: Yep, this is fine – would be slightly improved by being inside the `runInAction`

app/scenes/Settings/People.js
tommoor: Is using the dollar sign a pattern I might have missed with styled-components?

tommoor:
I'm guessing you skipped the design as mocked by Nan because there's no clear way to separate the counts for the different tabs?

app/scenes/Settings/People.js
timc1: dollar sign represent transient properties that only styled-components use, and will not transfer to the dom https://styled-components.com/docs/api#transient-props

timc1:
> I'm guessing you skipped the design as mocked by Nan because there's no clear way to separate the count requests for the different tabs?

Yeah, unless we query for all user records

MatheusRV:
Could you please provide a counter to each group on title left on the tab bar.

image

timc1:
> Could you please provide a counter to each group on title right on the tab bar.
> 
> 

That's what we were referencing in the comment [above](https://github.com/outline/outline/pull/1672#issuecomment-733851242) – since this screen is paginated we won't have the absolute total for each filter until all users are loaded. Though I'm happy to take suggestions for efficiently handling that!

app/scenes/Settings/People.js
tommoor: nice, TIL

tommoor:
Okay, so pondered this for a couple of days and I think we have two similar options to add the counts to the tabs:

- Add a param to `users.list`, `includeCounts` – which will result in the counts for active/all/invited etc being included in the data response (at the cost of additional queries)
- Add a new endpoint `users.counts`  just to get this payload separately to the list itself, this would allow the majority of data to load faster so I'd probably err towards this

Storing the counts can be done in an object on the UsersStore pretty much the same as you have it. Thoughts?

server/models/User.js
```diff
@@ -43,6 +43,9 @@ const User = sequelize.define(
       isSuspended() {
         return !!this.suspendedAt;
       },
+      isInvited() {
+        return !this.lastActiveAt;
+      },
```
timc1, line 48: @tommoor I was able to get `isInvited` on the server by adding this getter – I _think_ this is okay pls lmk otherwise

server/api/users.js
```diff
@@ -47,14 +48,41 @@ router.post("users.list", auth(), pagination(), async (ctx) => {
...
+    },
+  });
+
+  const count = {
+    active: filter(users, (user) => !user.isSuspended && user.lastActiveAt)
+      .length,
+    admins: filter(users, (user) => user.isAdmin).length,
+    all: users.length,
+    invited: filter(users, (user) => user.isInvited).length,
+    suspended: filter(users, (user) => user.isSuspended).length,
```
timc1, line 78: Same logic as on the `UserStore`

app/stores/UsersStore.js
```diff
@@ -76,10 +87,45 @@ export default class UsersStore extends BaseStore<User> {
     invariant(res && res.data, "Data should be available");
     runInAction(`invite`, () => {
       res.data.users.forEach(this.add);
+      this.total += res.data.sent.length;
+      this.count.invited += res.data.sent.length;
+      this.count.all += res.data.sent.length;
```
timc1, line 92: First time working with mobx – lmk if there's a better way of managing this observable

server/api/users.js
```diff
@@ -47,14 +48,41 @@ router.post("users.list", auth(), pagination(), async (ctx) => {
...
+    },
     data: users.map((listUser) =>
       presentUser(listUser, { includeDetails: user.isAdmin })
     ),
   };
 });
 
+router.post("users.count", auth(), async (ctx) => {
+  const user = ctx.state.user;
+  const users = await User.findAll({
```
tommoor, line 66: Chatted with Nan about this, it was a red flag because you're querying every single user on the team and loading them into memory here.

We think that this is actually a perfect spot to break out of the ORM and just do a raw query to count up the various types of user inside PG. Here's an example:

https://dba.stackexchange.com/a/23074

server/models/User.js
tommoor: lgtm

MatheusRV:
@timc1 friendly ping. Any news?

app/scenes/Settings/People.js
```diff
@@ -87,27 +93,31 @@ class People extends React.Component<Props> {
           Invite people…
         </Button>
 
-        <Tabs>
+        <TotalCount $loading={!count.all}>
+          {count.all} total {count.all > 1 ? "users" : "user"}
+        </TotalCount>
```
tommoor, line 98: Lets remove this for now, it isn't super useful called out and is included on the "Everyone" tab below

app/stores/UsersStore.js
```diff
@@ -52,16 +60,19 @@ export default class UsersStore extends BaseStore<User> {
...
 
   @action
   demote = (user: User) => {
+    this.count.admins -= 1;
     return this.actionOnUser("demote", user);
   };
 
   @action
   suspend = (user: User) => {
+    this.count.suspended += 1;
```
tommoor, line 75: `activate` method below is the opposite of suspend so need the `-= 1` there

app/stores/UsersStore.js
```diff
@@ -76,10 +87,44 @@ export default class UsersStore extends BaseStore<User> {
...
+    invariant(res && res.data, "Data should be available");
+
+    this.addPolicies(res.policies);
+    this.count = res.data.count;
+    return res.data;
+  };
+
+  @action
+  async delete(user: User) {
+    super.delete(user, { confirmation: true });
```
tommoor, line 112: Shouldn't we pass the second parameter through as well rather than hardcoding?

server/models/User.js
tommoor: Doesn't look like this is needed anymore but I'm fine with leaving it

server/models/User.js
```diff
@@ -267,4 +270,23 @@ User.afterCreate(async (user, options) => {
...
+User.prototype.getCount = async function () {
+  const countSql = `
+    SELECT 
+      COUNT(CASE WHEN "suspendedAt" IS NOT NULL THEN 1 END) as "suspendedCount",
+      COUNT(CASE WHEN "isAdmin" = true THEN 1 END) as "adminCount",
+      COUNT(CASE WHEN "lastActiveAt" IS NULL THEN 1 END) as "invitedCount",
+      COUNT(CASE WHEN "suspendedAt" IS NULL AND "lastActiveAt" IS NOT NULL THEN 1 END) as "activeCount",
+      COUNT(*) as count
+    FROM users
+    WHERE "deletedAt" IS NULL
```
tommoor, line 282: Your missing a critical part of the `WHERE` query, need to pass in and include the `teamId` otherwise we're counting the entire userbase.

server/models/User.js
```diff
@@ -267,4 +270,23 @@ User.afterCreate(async (user, options) => {
   ]);
 });
 
+User.prototype.getCount = async function () {
```
tommoor, line 273:
```suggestion
User.prototype.getCounts = async function (teamId: string) {
```

server/models/User.js
```diff
@@ -267,4 +270,23 @@ User.afterCreate(async (user, options) => {
...
+      COUNT(*) as count
+    FROM users
+    WHERE "deletedAt" IS NULL
+  `;
+
+  const results = await sequelize.query(countSql, {
+    type: sequelize.QueryTypes.SELECT,
+  });
+
+  return results[0];
```
tommoor, line 289: Could we move the `parseInt` logic you have in the controller in here too, I think it makes sense for this method to return integers rather than strings

server/api/users.test.js
```diff
@@ -353,3 +353,16 @@ describe("#users.activate", () => {
...
+describe("#users.count", () => {
+  it("should return the count", async () => {
+    const user = await buildUser();
+    const res = await server.post("/api/users.count", {
+      body: { token: user.getJwtToken(), id: user.id },
+    });
+    const body = await res.json();
+
+    expect(res.status).toEqual(200);
+    expect(body).toMatchSnapshot();
```
tommoor, line 366: Trying to move away from snapshots, they don't make the expectations of the test as explicit

```suggestion
    expect(body.data.count.all). toEqual(1);
    expect(body.data.count.admins). toEqual(0);
    expect(body.data.count.invited). toEqual(0);
    expect(body.data.count.suspended). toEqual(0);
    expect(body.data.count.active). toEqual(1);
```

tommoor (changes requested):
Thanks again – We nearly there but can't approve with the missing where clause 😉

timc1:
@tommoor added `teamId` & pluralized `count` – context switching between getCounts and `this.count` justified that. Lmk if you're cool with it 👍

tommoor:
@timc1 I made some critical fixes on the above branch, thanks for all the upfront work

PR #1686, opened 2020-11-30
Deploy using PM2 cluster mode & updated README

rjsgn:
I just updated package.json to use PM2 cluster mode with number of instances set to number of available CPUs. Let me know if it looks good.


tommoor:
Hey @rjsgn – thanks for this, I'll check this in a production environment.

I think with this change we should change the readme to remove the mention of installing and using pm2 and just make it an explicit dependency, for one thing the docker environment won't run without that change

rjsgn:
Yeah. Also I realized that we should update Dockerfile to add PM2 `RUN yarn add global pm2`. I can do that if you want.

tommoor:
We can just add it to the regular dependencies in package.json? No need to globally add separately.

rjsgn:
No because PM2 has to be installed globally. I just realized that we should change some parts of the Dockerfile too. Since PM2 is daemon based PM, docker container stops as soon as yarn start is run.

It would be better to use nginx as a reverse proxy. IMO, It'd be even a good idea for production use.

rjsgn:
I added PM2 to package.json. It seems to work on my machine.

Sylvenas:
Why is this `PR` closed？ I think it's a good idea to use `PM2` as cluster management

PR #1689, opened 2020-11-30
chore: Flatten left nav in preparation to refactor drag to reorder

thenanyu:
a bit of a messy commit history but I think it's the right approach.
- main concept is to flatten nested components into a non-nested html structure to play nice with `react-beautiful-dnd`
- also adds more clear visual representation of drag to import


app/components/Sidebar/Sidebar.js
```diff
@@ -26,7 +26,7 @@ function Sidebar({ location, children }: Props) {
     if (location !== previousLocation) {
       ui.hideMobileSidebar();
     }
-  }, [ui, location]);
+  }, [ui, location, previousLocation]);
```
thenanyu, line 29: I had a linter warning here, it's generally unrelated to the rest of the changes

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -34,61 +31,63 @@ class CollectionLink extends React.Component<Props> {
...
+              key={node.id}
+              node={node}
+              collection={collection}
+              activeDocument={activeDocument}
+              prefetchDocument={prefetchDocument}
+              canUpdate={canUpdate}
+              depth={1.5}
+            />
+          ))}
+      </>
```
thenanyu, line 90: 🍖 nested/recursive calls to Collection and Document Link do not have nested HTML elements

app/components/Sidebar/components/DocumentLink.js
```diff
@@ -25,139 +23,150 @@ type Props = {|
...
-const DocumentChildren = styled(Flex)``;
+const Disclosure = styled(CollapsedIcon)`
+  position: absolute;
+  left: -24px;
+
+  ${({ expanded }) => !expanded && "transform: rotate(-90deg);"};
+`;
 
-export default DocumentLink;
+const ObservedDocumentLink = observer(DocumentLink);
```
thenanyu, line 171: 🍗 the recursive call needs to be wrapped with observer; this seems like the least mysterious way to do it

app/components/Sidebar/components/SidebarLink.js
```diff
@@ -29,46 +26,25 @@ function SidebarLink({
...
-      setExpanded(!expanded);
-    },
-    [expanded]
-  );
-
-  const handleExpand = React.useCallback(() => {
-    setExpanded(true);
-  }, []);
-
-  const showDisclosure = !!children && !hideDisclosure;
```
thenanyu, line 71: 🍖 all of this logic is moved into `DocumentLink`, which was the only place that was using it at all

tommoor (approved):
love it, this should provide the foundation we need to finally build a real draggable sidebar.

PR #1694, opened 2020-12-01
feat: Retrieve AWS credentials from IAM role in ECS & EC2

syphernl:
This PR adds support for retrieving AWS Credentials from more sources than just the environment.

It will search for credentials is:
- The environment 
- ECS IAM Role
- EC2 IAM Role

The advantage of this is that no hardcoded environment variables (either directly or via the Parameter Store) have to be provided to the ECS Task and that the credentials in this case are rotated automatically thus increasing security.

### DIFF
--- a/server/utils/s3.js
+++ b/server/utils/s3.js
@@ -6,48 +6,19 @@ import addHours from "date-fns/add_hours";
 import format from "date-fns/format";
 import fetch from "isomorphic-fetch";
 
-const AWS_REGION = process.env.AWS_REGION;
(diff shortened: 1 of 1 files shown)


syphernl:
I'm running into issues with the fact that these tokens expire, and the AWS SDK doesn't seem to refresh this by its own. 
I haven't yet found a solution for that.

tommoor:
My guess would be this is because the AWS instance is instantiated in the module context once right now.

If it was moved to create a new instance on each upload it would probably pull fresh credentials?

tommoor:
Did you figure out the issue here?

PR #1696, opened 2020-12-01
fix: One source of transaction deadlock when invites > pg pool

### LINKED ISSUE #1695: Intermitent transaction deadlock
**To Reproduce**

In production we've seen an occasional transaction deadlock, logged as `SequelizeConnectionAcquireTimeoutError`. This causes the node process to lock up and timeout for other queries. The process often needs to be restarted to fully recover.

**Expected behavior**

No deadlocks would be nice.

- Install: 0.50.0

tommoor:
closes #1695

### DIFF
--- a/server/commands/userInviter.js
+++ b/server/commands/userInviter.js
@@ -47,48 +47,33 @@ export default async function userInviter({
   let users = [];
 
   // send and record remaining invites
-  await Promise.all(
-    filteredInvites.map(async (invite) => {
-      const transaction = await sequelize.transaction();
-      try {
-        const newUser = await User.create(
-          {
-            teamId: user.teamId,
-            name: invite.name,
-            email: invite.email,
-            service: null,
-          },
-          { transaction }
-        );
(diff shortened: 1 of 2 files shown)


server/commands/userInviter.js
```diff
@@ -47,48 +47,33 @@ export default async function userInviter({
   let users = [];
 
   // send and record remaining invites
-  await Promise.all(
-    filteredInvites.map(async (invite) => {
-      const transaction = await sequelize.transaction();
```
tommoor, line 52: In hindsight the transaction was unnecessary here

server/commands/userSuspender.js
```diff
@@ -16,29 +17,27 @@ export default async function userSuspender({
     throw new ValidationError("Unable to suspend the current user");
   }
 
-  let transaction;
-  try {
-    transaction = await sequelize.transaction();
```
tommoor, line 21: Transaction was created but not committed or used here, big oversight

Issue #1697, opened 2020-12-02
Upgrade react dropzone


tommoor:
- We're 7 major versions behind
- We can move to the hooks API
- We can stop it from showing the active state when dragging unacceptable items over the sidebar

PR #1699, opened 2020-12-02
chore: React-Dropzone version bump

thenanyu:
fixes the internal hover state issue!
closes https://github.com/outline/outline/issues/1697


tommoor:
AND gets rid of some global styles?! Be still my beating heart

tommoor (approved):

PR #1715, opened 2020-12-11
fix: Attachments should not always be deleted with their original document

### LINKED ISSUE #1709: Images copied from a page are deleted along with the original page (image reference counting?)
When content is copied from one page to another that contains images and then then original page is deleted, the images are deleted from S3.

**To Reproduce**
1. Create `Page 1`, insert an image.
2. Select all of the content of `Page 1` and copy to the clipboard
3. Create `Page 2`, paste that content
4. Delete `Page 1`
5. View `Page 2` and see a broken image.

**Expected behavior**
Image content would be tracked based on how many pages reference a given image, or each page that contains an image would get its own copy of the image. Image copy/paste would detect internally-hosted images instead of copying the HTML directly to the new page.

**Screenshots**

image

- Install: tested on both getoutline.com and a self hosted instance
- Version: 0.50.0

 - OS: macOS
 - Browser Safari
 - Version 14.0.1

tommoor:
- Remove the cascading delete from documents -> attachments as attachments can appear in multiple docs
- Fix the utils code not correctly querying with paranoid (trash never emptied)
- Check for attachments in other docs before deleting
- Ensure attachment is not deleted from S3 prematurely when moved to trash
- Ensure attachment _is_ removed from S3 with utils

closes #1709

### DIFF
--- a/server/api/attachments.js
+++ b/server/api/attachments.js
@@ -12,6 +12,7 @@ import {
   publicS3Endpoint,
   makeCredential,
   getSignedImageUrl,
+  deleteFromS3,
 } from "../utils/s3";
 
 const { authorize } = policy;
@@ -104,6 +105,7 @@ router.post("attachments.delete", auth(), async (ctx) => {
   authorize(user, "update", document);
 
   await attachment.destroy();
+  await deleteFromS3(attachment.key);
 
   await Event.create({
     name: "attachments.delete",
--- a/server/api/attachments.js
+++ b/server/api/attachments.js
@@ -12,7 +12,6 @@ import {
   publicS3Endpoint,
   makeCredential,
   getSignedImageUrl,
-  deleteFromS3,
 } from "../utils/s3";
 
 const { authorize } = policy;
@@ -105,7 +104,6 @@ router.post("attachments.delete", auth(), async (ctx) => {
   authorize(user, "update", document);
 
   await attachment.destroy();
-  await deleteFromS3(attachment.key);
 
   await Event.create({
     name: "attachments.delete",
--- a/server/api/utils.js
+++ b/server/api/utils.js
@@ -4,13 +4,14 @@ import debug from "debug";
 import Router from "koa-router";
 import { AuthenticationError } from "../errors";
 import { Document, Attachment } from "../models";
-import { Op } from "../sequelize";
+import { Op, sequelize } from "../sequelize";
+import parseAttachmentIds from "../utils/parseAttachmentIds";
 
 const router = new Router();
 const log = debug("utils");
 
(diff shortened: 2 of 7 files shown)


server/api/attachments.js
```diff
@@ -104,6 +105,7 @@ router.post("attachments.delete", auth(), async (ctx) => {
   authorize(user, "update", document);
 
   await attachment.destroy();
+  await deleteFromS3(attachment.key);
```
tommoor, line 108: Previously this used the `afterDestroy` hook in the model

server/api/utils.js
```diff
@@ -4,44 +4,82 @@ import debug from "debug";
...
+        } else {
+          log(`Unknown attachment ${attachmentId} ignored`);
+        }
+      }
+    }
+  }
+
   await Document.scope("withUnpublished").destroy({
-    where,
+    where: {
```
tommoor, line 76: Re-using the `where` clause from above actually introduced a race condition here before where any documents that had become gc-able in the time between the start of this process and now would also be deleted without the associated attachment cleanup.

server/migrations/20201211080408-attachment-no-cascade.js
```diff
@@ -0,0 +1,37 @@
+const tableName = 'attachments';
```
tommoor, line 1: Remove cascade on delete association as attachment<->document relationships are not 1-1 in reality, this was a mistake.

server/models/Attachment.js
```diff
@@ -51,10 +50,6 @@ const Attachment = sequelize.define(
   }
 );
 
-Attachment.beforeDestroy(async (model) => {
```
tommoor, line 54: This was being triggered when the document was placed in the trash, resulting in the file being immediately deleted from s3. You can reproduce this in production today by createing a doc with an image, putting it in the trash and reloading moments later – the image is gone and there's no way to recover.

server/utils/parseAttachmentIds.js
```diff
@@ -0,0 +1,8 @@
+// @flow
```
tommoor, line 1: Just moved from `server/presenters/document` so it can be reused

server/api/utils.js
```diff
@@ -4,44 +4,82 @@ import debug from "debug";
...
-      documentId: documentIds,
+      deletedAt: {
+        [Op.lt]: subDays(new Date(), 30),
+      },
     },
+    paranoid: false,
+    limit,
   });
 
+  const query = `
```
tommoor, line 34: 🍖  Find if any other documents reference this attachment, thankfully the searchVector works just fine for this, combined with limiting by `teamId` it's an efficient query.

PR #1717, opened 2020-12-12
feat: Moving documents via drag and drop in sidebar

thenanyu:
WIP...

A new attempt at implementing drag and drop reordering on the sidebar. Uses `react-dnd`, a lower-level [library](https://github.com/react-dnd/react-dnd) for handling drag and drop functionality.

Inspired by https://github.com/outline/outline/pull/1283, will likely be doing at least some copy-pasta, and will credit original author

I would like to produce a UI that has a higher sense of precision, similar to how Notion's reordering UI looks and feels. The plan is to approach this in a few stages.

1. drag and drop for re-nesting documents under other documents or collections
2. manual sorting and drag and drop re-ordering of list items

This feature can ship at both stages in either beta or GA.

- [ ] placeholder after every document
- [ ] placeholder before first document
- [x] up one level (move out of nesting)
- [x] down one level (move into nesting)

### Permissions

- [x] Drag from
- [x] Drag to


thenanyu:
image

Quick functional spec of intended re-ordering heuristics

app/components/Sidebar/components/DocumentLink.js
```diff
@@ -101,48 +108,87 @@ function DocumentLink({
...
+      pathToNode && !pathToNode.includes(monitor.getItem().id),
+    collect: (monitor) => ({
+      isOver: !!monitor.isOver(),
+      item: monitor.getItem(),
+      canDrop: monitor.canDrop(),
+    }),
+  });
+
+  // /DnD
+  // cant drag onto itself or inside its own shit
```
tommoor, line 135: lol

app/components/Sidebar/components/DocumentLink.js
```diff
@@ -101,48 +108,87 @@ function DocumentLink({
...
-                <DocumentMenu
-                  position="right"
-                  document={document}
-                  onOpen={() => setMenuOpen(true)}
-                  onClose={() => setMenuOpen(false)}
+    <div
+      key={node.id}
+      ref={drag}
+      style={{
+        opacity: isDragging ? 0.8 : 1, // TODO put this into a styled component
```
tommoor, line 142: We should hide the document menu in the dragging state too

app/components/Sidebar/components/DocumentLink.js
thenanyu: That was a todo which is actually done in this commit already!

tommoor:
feels solid so far, totally agree with the diagram – based on your experiments think it would be easy enough to have collections open up if you hover over them while dragging?

app/components/Sidebar/components/DocumentLink.js
tommoor: - [ ] placeholder after every document
- [ ] placeholder before first document
- [ ] up one level (move out of nesting)
- [ ] down one level (move into nesting)

app/components/Sidebar/components/DocumentLink.js
```diff
@@ -101,50 +108,93 @@ function DocumentLink({
...
+                      onOpen={() => setMenuOpen(true)}
+                      onClose={() => setMenuOpen(false)}
+                    />
+                  </Fade>
+                ) : undefined
+              }
+            />
+          </DropToImport>
+        </div>{" "}
+      </div>
```
thenanyu, line 195: closed the draggable element right here, so that's the object that gets captured as the default drag layer, seems straightforward without having to deal with the custom drag layer stuff. I'm fine either way; maybe start with this and do the custom thing later if we need it?

tommoor:
Tested the additional guard ✅

PR #1731, opened 2020-12-20
chore: Enable automatic generation of email server in non production env

tommoor:
Setting up a test email environment is a pain, this PR means it'll happen automatically if no email server details are set in development.

Unfortunately `ethereal.email` has chosen today to have SSL issues, so leaving this in draft until the service is resolved.


tommoor:
ethereal.email is back, baby

PR #1747, opened 2020-12-24
feat: Batch Import

### LINKED ISSUE #914: Export and import data from Outline (different instances)
Hello,

I would like to suggest an easy to use tool that allows users to export data from an Outline instance and import in another instance of Outline.

E.g:
- Export all content/data from self-hosted Outline
- Import all content/data from the self-hosted Outline into Outline paid service (Outline hosted solution).

### LINKED ISSUE #1846: Update attachment record when upload fails
Currently `attachments.create` is called to get a pre-signed url which is then used to upload the attachment from the client side to S3-compatible storage. If the upload fails for some reason then the attachment remains in the DB. It should be marked as failed.

tommoor:
### TODO

- [x] Upload attachments
- [x] Remap attachment paths
- [x] Parent document relationships
- [x] API endpoint
- [x] Merge work with UI branch
- [x] Improve UI feedback
- [x] Handle collection export
- [x] Matching server side validation of collections > 1, documents > 1
- [x] Unlink tmp file

closes #1846
closes #914


Maronato:
I know that it’s a desirable feature, but wouldn’t it allow for people to avoid paying for the hosted version? They could just keep exporting/importing every month

tommoor:
I guess, not something we'll worry about for now - you had me at "it's a desirable feature" 😉

PR #1751, opened 2020-12-25
feat: Compress avatar images before upload

Maronato:
Since avatar images are never displayed in full size, this PR compresses and reduces them to 256x256.


tommoor:
Nice, pretty down for this improvement – I suggest that we double the size to 512px to future proof though, that will enable 256x256 @ 2x.

Maronato:
> Nice, pretty down for this improvement – I suggest that we double the size to 512px to future proof though, that will enable 256x256 @ 2x.

Done :)

PR #1755, opened 2020-12-30
feat: Sidebar collection sort

tommoor:
### What

- [x] Add `sort.direction` and `sort.field` to `collections.update` API endpoint
- [x] Allow collection update to be passed partial data
- [x] Added "Sort" menu item to change sort on collection
- [x] Default new collections to custom sort
- [x] Add sort option in collection modal too

### Why

It became clear that we cannot just stop automatically ordering collections by title over night to allow custom ordering as existing collections will become "unorganized" (they are JIT sorted in the presenter layer). As such, we should have a new option to choose the sort mode for the collection and default to unsorted for new collections.

### Screenshot

image


tommoor:
cc @thenanyu as discussed yesterday

tommoor:
Once this is merged to the parent we'll need to make sure that the drag and drop functionality is disabled when "sort by title" is enabled.

thenanyu:
> Once this is merged to the parent we'll need to make sure that the drag and drop functionality is disabled when "sort by title" is enabled.

I added a commit for this — note that only the drop to reorder cursors need to be disabled. Drop to re-parent is still valid behavior regardless of sort option

PR #1756, opened 2020-12-30
fix: miro - use the incoming domain to ensure access to logged in boards works

cliftonc:
I've started trying to use Outline (it's great and has huge potential - thanks!), and we use Miro a lot.  I notice that the embed forces the embed url to `realtimeboard.com` - which breaks access to logged in boards, as we login via miro.com.  This change uses the incoming url domain to generate the iframe src url domain.


tommoor (approved):
lgtm and makes sense, thanks for tackling this!

cliftonc:
Great! Glad to help, will send PRs your way as I find things.

PR #1758, opened 2020-12-31
fix: 'New' badge should never show to document creator

tommoor:
It's currently possible to see the yellow "New" badge as a document creator depending on how the doc is created. These should be reliably only displayed when the doc is created by another user.

As a nicety this works well with https://github.com/outline/outline/pull/1747 to prevent all newly imported documents as showing "New" to the importer.


app/components/DocumentList.js
```diff
@@ -2,7 +2,7 @@
 import ArrowKeyNavigation from "boundless-arrow-key-navigation";
 import * as React from "react";
 import Document from "models/Document";
-import DocumentPreview from "components/DocumentPreview";
+import DocumentListItem from "components/DocumentListItem";
```
tommoor, line 5: Wanted to do this rename for a while

app/components/DocumentListItem.js
```diff
@@ -0,0 +1,242 @@
...
+
+const SEARCH_RESULT_REGEX = /<b\b[^>]*>(.*?)<\/b>/gi;
+
+function replaceResultMarks(tag: string) {
+  // don't use SEARCH_RESULT_REGEX here as it causes
+  // an infinite loop to trigger a regex inside it's own callback
+  return tag.replace(/<b\b[^>]*>(.*?)<\/b>/gi, "$1");
+}
+
+function DocumentListItem(props: Props) {
```
tommoor, line 39: Converted to functional component so it's easy to access the `currentUser`

PR #1764, opened 2021-01-03
feat: Add 'archive' option in delete confirmation modal

tommoor:
When deleting a document we suggest archiving it instead, this change updates the delete confirmation modal to our latest best practices (functional, translations) and adds the option to archive inline:

---

image

- chore: Move to functional component
- chore: Add translation tags to delete confirmation
- chore: Move neutral button to shadow instead of border so that it's size visually aligns with primary/danger buttons


MatheusRV:
image
image

tommoor:
These style of confirmations make sense when the data is unrecoverable, that's not the case here as the deleted document can be restored from "Trash" for a month.

PR #1768, opened 2021-01-05
fix: Keyboard accessible context menus

### LINKED ISSUE #1766: Refactor popover menus to use Reakit
We've been gradually improving the keyboard accessibility of Outline but one place stands out as being a major problem is the popover menus, where a lot of functionality is still kept.

I'd like to start moving to the [Reakit](https://reakit.io/docs/menu/) framework for controls where possible to ensure we're following best practices and aria standards here. This should be relatively straight forward as the `DropdownMenu` code is already centralized, and will likely just mean removing a lot of custom logic for menu positioning and state management.

tommoor:
- Makes menus fully accessible and keyboard driven
- Currently adds 2.8% to initial bundle size due to the inclusion of Reakit and its dependency, popperjs.
- Converts all menus to functional components
- Remove old custom menu system
- Various layout and flow improvements around the menus

closes #1766


tommoor:
This still isn't mergable as it's not as performant due to the menus rendering all of their contents before they are clicked, which is different to the old system. I think in order to get this feeling good the key menus that are on every page (DocumentListItem, DocumentLink, CollectionLink) will need to be rendered only when visible which will neccessitate moving hover/focus into React state.

app/components/Button.js
```diff
@@ -42,7 +46,7 @@ const RealButton = styled.button`
   }
 
   ${(props) =>
-    props.neutral &&
+    props.$neutral &&
```
thenanyu, line 49: what does the leading `$` mean generally?

tommoor: Stops it being added to the dom by styled-components: https://styled-components.com/docs/api#transient-props

PR #1776, opened 2021-01-07
fix: Deleting a collection should not deleted archived documents within

### LINKED ISSUE #1775: Deleting a collection deletes archived documents
**To Reproduce**
1. Archive a document
2. Delete collection
3. Archived documents are deleted

**Expected behavior**

It's probably more expected that archived documents hang around in the archive even if their original collection is deleted, as they don't appear in the sidebar it's otherwise easy to delete documents accidentally in this manner.

tommoor:
closes #1775

### DIFF
--- a/app/components/Breadcrumb.js
+++ b/app/components/Breadcrumb.js
@@ -89,8 +89,6 @@ const Breadcrumb = ({ document, onlyText }: Props) => {
 
   let collection = collections.get(document.collectionId);
   if (!collection) {
-    if (!document.deletedAt) return <div />;
-
     collection = {
       id: document.collectionId,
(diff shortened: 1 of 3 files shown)


tommoor:
@thenanyu make sense to you in terms of desired behavior?

thenanyu:
This seems like a good patch and worth shipping, but there's a consideration to change the default behavior to archive documents instead of deleting

Then a checkbox can be added in the delete dialog which allows true deletion of included documents 

This way we can say something like: "also delete 3 active and 5 archived documents"

tommoor:
Yea, I like that – default to a soft delete in the future and make this the hard delete option.

Issue #461, opened 2017-12-01
Indicator for starred state of document


anchepiece:
While viewing a document, if you `star` it using the `Edit` menu, there is no indication that its status was changed. 

A better experience for starring documents could be to have a star outline icon, that when clicked changes to a filled in version. Similar to how the chrome bookmark tool in the address bar behaves.

I like how clean the document appears as it is though, maybe the indicator can sit in the DropDownMenu item itself (to the left of Star/Unstar label).

tommoor:
Totally agree - starring when viewing a document is definitely a secondary experience to starring from the document list right now. The UX is a little rough. 

We have a design for a sticky header which might provide somewhere for the star to live more easily next to the document title:

![image](https://user-images.githubusercontent.com/380914/33518975-942c4b04-d752-11e7-9ffd-a587f17f9ace.png)
https://share.goabstract.com/28fa0a5d-cbad-488b-9b09-6be9dff20e35

anchepiece:
That mock looks fantastic and it looks like you have a clear direction. I'll hold off any UI comments until the project reaches that point.

ThomasBem:
The new header looks nice! I really like the profile icons which I guess is listing who has contributed to said document? 

When it comes down to the star feature. I guess a fairly common way of doing this is to have a star icon next to the documents on the left side. Something like a transparent star by default, then when you select it it adds to Starred and changes color.

We have done something like this in our latest app when we allow our users to save their favoritt searches/filters.

![screen shot 2018-01-31 at 09 04 39](https://user-images.githubusercontent.com/3599597/35611633-f19d9918-0665-11e8-888f-c683d2c85f06.png)
![screen shot 2018-01-31 at 09 04 50](https://user-images.githubusercontent.com/3599597/35611634-f1b8fd16-0665-11e8-81fd-9b434e560aff.png)

PR #1785, opened 2021-01-08
fix: Add indicator of starred status when viewing a document

tommoor:
image

related #461


MatheusRV:
Is possible to add a list who star this document?

tommoor:
It's not possible to query other users stars, they're considered private – like bookmarks

Issue #1728, opened 2020-12-19
Update relative timestamps to account for i18n


tommoor:
We use the `date-fns` library, it includes support for passing in the current users language and will then format relative timestamp appropriately: https://date-fns.org/

Timestamps are centralized in the `Time` component

leviscar:
can i help to fix this issue?

rubenmoya:
Hello! I was thinking about picking up this one, this is how I think I'd approach it, I'd love some feedback before starting:

- Either inject `AuthStore` to the `Time` component or convert it to a functional component and use the `useCurrentUser` hook.
- Get the language from the user and split it (for example, get `en` from `en_US` and `es` from `es_ES`
- Dynamically require ``date-fns/locale/${locale}`` and pass it to the `distanceInWordsToNow`

That _should_ work, but I have some questions since I've been playing with Outline for a few minutes:
- How could we handle the `shorten` prop? it's really coupled to english right now, and I can't think of a good approach for it
- Injecting the whole auth to the Time component feels weird, is there a better way I could do it? Also depending on the current user feels weird...

tommoor:
This sounds exactly like what I had in mind – I'd convert to a functional component, in fact here's a commit/branch where I already did some of that work several weeks back and could be a jumping off or reference point: 
https://github.com/outline/outline/compare/functional-time?expand=1

> How could we handle the shorten prop? it's really coupled to english right now, and I can't think of a good approach for it

In the near-term it's probably fine to leave it as English only, it was really a work around for the default english translations being quite verbose. 

> Injecting the whole auth to the Time component feels weird, is there a better way I could do it? Also depending on the current user feels weird...

Not that I can think of, we could store the users language in another mobx object but that's probably overkill – the `auth` store rarely changes after all. One thing I'd note is that we need to make sure that the component still works correctly when there is no current user as we have public-facing pages that use the `Time` component.

PR #1793, opened 2021-01-10
fix: take into account user lang in Time component

rubenmoya:
Closes https://github.com/outline/outline/issues/1728

This PR takes into account the user selected language to format the time in the `Time` component. It uses the changes done in the [functional-time](https://github.com/outline/outline/compare/functional-time?expand=1) branch.

Also, I've added Tom as a Co-Author, since he was the one who transformed it into a functional component.

I moved the logic to a method outside the component, I wasn't sure if you prefer it on top of the comment or at the bottom, or even if you just prefer to have that logic inside the component, so any feedback is welcome.

I haven't tested it in a public page, since I'm not sure how to do it, how do I find a public page to test it? I've tried by publishing and sharing a document but I didn't see any dates.

image


app/components/Time.js
```diff
@@ -28,44 +29,53 @@ type Props = {
...
+}
+
+function getLocaleForUser(user) {
+  if (!user) {
+    return undefined;
   }
+
+  const userLanguage = user.language.split("_")[0];
+  return require(`date-fns/locale/${userLanguage}`);
 }
```
tommoor, line 79: > I moved the logic to a method outside the component, I wasn't sure if you prefer it on top of the comment or at the bottom, or even if you just prefer to have that logic inside the component, so any feedback is welcome.

How about we pull this out to `hooks/useUserLocale` ?

app/components/Time.js
```diff
@@ -28,44 +29,53 @@ type Props = {
...
+    </Tooltip>
+  );
+}
+
+function getLocaleForUser(user) {
+  if (!user) {
+    return undefined;
   }
+
+  const userLanguage = user.language.split("_")[0];
```
tommoor, line 77: This made me realise that the flowtype for `user.language` is actually wrong, it can be `null` on older accounts.

I think we can `en_US` as a default in the presenter here to fix that:

https://github.com/outline/outline/blob/develop/server/presenters/user.js#L27

tommoor (changes requested):
Thanks for taking this on!

app/components/Time.js
rubenmoya: Should we use `process.env.DEFAULT_LANGUAGE` just in case?

tommoor: We can use it in the sever side, yea that's a good idea.

rubenmoya:
Tests are failing because of the snapshots related to the users api and user presenter tests, if we hardcode `en_US` to the presenter we'd make sure the snapshots are the same, using the `DEFAULT_LANGUAGE` env var it could end up being `undefined` (which is breaking the tests in CI) I ended up keeping `en_US` after the check of `DEFAULT_LANGUAGE` just in case. What do you think?

tommoor:
Sounds good 👌

tommoor:
Functionally this is good, but looking at the bundle size audit the require isn't working how we might have expected – all of the translations are getting included in an existing chunk because the require isn't async, causing a 3.5% increase to bundle size.

https://app.relative-ci.com/projects/TMqufq6bi8qzsOjEHSWY/jobs/545-8eb133a6-a48b-486e-84d7-8756f2013c21/modules


In order to have the translations dynamically load we probably need to make these changes unless you have another idea:

- Create a new component, call it `LocaleTime`
- Import only the languages we support into `LocaleTime` using a regular import, we might be able to use the values from here: https://github.com/outline/outline/blob/develop/shared/i18n/index.js#L43 or it might be copy/paste.
- In `Time` we use React Suspense to load `LocaleTime` as it's own chunk, and have the fallback be English during loading

Does this make sense?

rubenmoya:
Yes, it makes a lot of sense, what I'm not sure about is what to move exactly to the `LocaleTime` component, at first I thought about the `time` element, but now I realised that we should also translate the tooltip, since it will show the months in english otherwise.

I've never used Suspense before, so let me get this right, should we keep what we had before as the fallback, and create a new component which also has the tooltip and the time?

Example:

```ts
<React.Suspense fallback={
  <Tooltip ... >
    <time  ... />
  </Tooltip>
}>
    <LocaleTime />
</React.Suspense>
```

tommoor:
> should we keep what we had before as the fallback, and create a new component which also has the tooltip and the time?

I'd be happy if the fallback was a dumb html `<time>` tag in English with no time-based updating, that should prevent any sort of UI flashing in most usecases and keep the `<Time>` component really simple, it's really just a suspense wrapper around `LocaleTime` then.

rubenmoya:
Hmm, I'm not sure what I've done wrong, but it still shows `+3.74%` :sweat_smile:

rubenmoya:
Yeah that looks better, 0.49%, I'm not sure if there is a better way to load them, I don't really like loading them manually, it's pretty easy to add a new language to `shared/i18n` and forget to add it in this component, it won't break, but it won't be translated...

Another thing I don't like either is how it flahes from the fallback to the LocaleTime, but I'm not sure what else we can do...

Let me know what you think!

app/components/Time.js
```diff
@@ -28,44 +12,27 @@ type Props = {
   shorten?: boolean,
 };
 
-class Time extends React.Component<Props> {
-  removeEachMinuteCallback: () => void;
+function Time(props: Props) {
+  let content = distanceInWordsToNow(props.dateTime, {
+    addSufix: props.addSuffix,
```
tommoor, line 17:
```suggestion
    addSuffix: props.addSuffix,
```

app/components/LocaleTime.js
```diff
@@ -0,0 +1,81 @@
+// @flow
+import distanceInWordsToNow from "date-fns/distance_in_words_to_now";
+import format from "date-fns/format";
+import * as React from "react";
+import Tooltip from "components/Tooltip";
+import useUserLocale from "hooks/useUserLocale";
+
+const locales = {
```
tommoor, line 8: Lets add a note in `shared/i18n` to point to here, something like:

> Note: Updating the available languages? Make sure to also update the locales array in `app/components/LocaleTime.js` to enable translation for timestamps.

tommoor (changes requested):
> Another thing I don't like either is how it flahes from the fallback to the LocaleTime, but I'm not sure what else we can do...

I think this will be better in production where the chunks are cached by the browser.

app/components/Time.js
rubenmoya: Oops, my bad!

tommoor (approved):
Great! Thank you.

rubenmoya:
It took a while but we got there! :rocket: 

Let's hope it doesn't come back to bite us, locales are never easy :sweat_smile:

tommoor:
Absolutely, thanks for your help

Issue #1111, opened 2019-12-18
Migrate creatorId to createdById


tommoor:
This old field is the odd one out.

PR #1794, opened 2021-01-10
chore: rename collection creatorId to createdById

rubenmoya:
Closes https://github.com/outline/outline/issues/1111

This PR renames the `Collection` `creatorid` to `createdById`.

It seems to be a pretty straightforward change, but I'm new to the project so I'm not sure. Tests are failing in local because of the user language, I'm not sure what's going on, could be something related to the env vars so let's see what CircleCI says.


tommoor (approved):
This looks great, I don't see any issues – I guess it's handy that this is the only place `creatorId` was used 😄

PR #1807, opened 2021-01-14
fix: Unintended scroll reset when switching between view / edit

tommoor:
- fix: Don't remount document when switching between view and edit
- fix: Don't remount `MarkAsRead` when switching from view to edit
- fix: Button vertical alignment when using `as={Link}`
- Update latest editor for related fixes switching between edit/read-only. This brings in new functionality like the image toolbar and float controls too.

### DIFF
--- a/app/components/Button.js
+++ b/app/components/Button.js
@@ -102,7 +102,7 @@ export const Inner = styled.span`
   line-height: ${(props) => (props.hasIcon ? 24 : 32)}px;
   justify-content: center;
   align-items: center;
-  min-height: 30px;
+  min-height: 32px;
 
   ${(props) => props.hasIcon && props.hasText && "padding-left: 4px;"};
   ${(props) => props.hasIcon && !props.hasText && "padding: 0 4px;"};
--- a/app/scenes/Document/components/DataLoader.js
+++ b/app/scenes/Document/components/DataLoader.js
@@ -240,13 +240,11 @@ class DataLoader extends React.Component<Props> {
     }
 
     const abilities = policies.abilities(document.id);
-    const key = this.isEditing ? "editing" : "read-only";
 
     return (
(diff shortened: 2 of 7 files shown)


app/components/Button.js
```diff
@@ -102,7 +102,7 @@ export const Inner = styled.span`
   line-height: ${(props) => (props.hasIcon ? 24 : 32)}px;
   justify-content: center;
   align-items: center;
-  min-height: 30px;
+  min-height: 32px;
```
tommoor, line 105: This was misaligned when displaying an `a` tag a a button

app/scenes/Document/components/DataLoader.js
```diff
@@ -240,13 +240,11 @@ class DataLoader extends React.Component<Props> {
     }
 
     const abilities = policies.abilities(document.id);
-    const key = this.isEditing ? "editing" : "read-only";
```
tommoor, line 243: This is the meat of the fix.

app/scenes/Document/components/Document.js
```diff
@@ -440,7 +440,7 @@ class DocumentScene extends React.Component<Props> {
                     ui={this.props.ui}
                   />
                 </Flex>
-                {readOnly && !isShare && !revision && (
```
tommoor, line 443: Don't mark viewed when switching between edit/read

app/scenes/Document/components/Header.js
```diff
@@ -118,8 +105,6 @@ class Header extends React.Component<Props> {
   };
 
   render() {
-    if (this.redirectTo) return <Redirect to={this.redirectTo} push />;
```
tommoor, line 121: Without the remount this state caused problems – why was it built like this instead of a regular link? Must have been having a weird day.

PR #1814, opened 2021-01-16
feat: Google Drawings Integration

milesstoetzner:
Hello, 

I implemented a Google Drawings Integration.
The integration embeds the drawing using the url of the file as iframe src.
For example, `https://docs.google.com/drawings/d/1zDLtJ4HSCnjGCGSoCgqGe3F8p6o7R8Vjk8MDR6dKf-U/edit`.

The file must be published if the outline user does not have access to the file using his google account.

Screenshot from 2021-01-16 13-10-03

Google Drawings could also be integrated using an image tag.
The corresponding HTML code can be retrieved using `File > Publish in Web` inside the Goolge Drawings Editor.
For example:
```
<img src="https://docs.google.com/drawings/d/e/2PACX-1vRtzIzEWN6svSrIYZq-kq2XZEN6WaOFXHbPKRLXNOFRlxLIdJg0Vo6RfretGqs9SzD-fUazLeS594Kw/pub?w=960&amp;h=720">
```

The img src link is not the same link as the file link which is used as iframe src.
Since I want to bind the edit link to the "Open" button of the frame I chose the iframe approach.
The regex does not allow these img src links.

Note: I recreated the Goolge Drawings Logo by myself because I did not find a fitting image online.

EDIT: mention that img src is not equals frame src


mxzinke:
Here is a fitting icon, not made up by my self 😄 https://en.wikipedia.org/wiki/Google_Drawings#/media/File:Google_Drawings_2015_Logo.svg

app/embeds/GoogleDrawings.test.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+import GoogleDrawings from "./GoogleDrawings";
+
+describe("GoogleDrawings", () => {
+  const match = GoogleDrawings.ENABLED[0];
+  test("to be enabled on share link", () => {
+    expect(
+      "https://docs.google.com/drawings/d/1zDLtJ4HSCnjGCGSoCgqGe3F8p6o7R8Vjk8MDR6dKf-U/edit".match(
```
mxzinke, line 8: I think the `/edit` link is not the shared link. From my experience, it is with ending `?usp=sharing` only.

app/embeds/GoogleDrawings.test.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+import GoogleDrawings from "./GoogleDrawings";
+
+describe("GoogleDrawings", () => {
+  const match = GoogleDrawings.ENABLED[0];
+  test("to be enabled on share link", () => {
```
mxzinke, line 6: What about links like: ```https://docs.google.com/drawings/d/XXX/view?usp=sharing```.

milesstoetzner: I am not sure. Google provides a link to copy when publishing the file and the link ends with `/edit...`.

app/embeds/GoogleDrawings.test.js
milesstoetzner: I dont thin this is correct.

The following links is provided by my gdrive app when creating a link. Even if the link provides read only access.

```
https://docs.google.com/drawings/d/1zDLtJ4HSCnjGCGSoCgqGe3F8p6o7R8Vjk8MDR6dKf-U/edit?usp=drivesdk
```

milesstoetzner:
Yes and no. I think they use the icon I recreated when a small and square icon is required. 

See e.g. 
Screenshot_20210116-192316_Drive

app/embeds/GoogleDrawings.test.js
mxzinke: Then please put it also into your tests.

app/embeds/GoogleDrawings.test.js
mxzinke: From my latest researches in this topic, I have got that Google does add always the "usp" parameter when creating a shared URL. The parameter can have multiple values.

So, checking in the regex on the usp parameter would be good.

mxzinke:
I don't know if google has changed the icon lately, but if you look closer, you will notice a difference between the one which I posted and the one which you have.

app/embeds/GoogleDrawings.test.js
milesstoetzner: it is already part of the tests.

see https://github.com/outline/outline/pull/1814/files#diff-dad70d662be2d7eda7c97042f853eff786b32a266e7a93f8581bc032d8ee575cR13

i explicit want to allow links without the parameter to be able to integrate drawings that are not published.

app/embeds/GoogleDrawings.test.js
milesstoetzner: > i explicit want to allow links without the parameter to be able to integrate drawings that are not published.

as I mentioned above

tommoor:
It looks like they use both icons on the website so I think this works 👍

mxzinke:
@tommoor Just a reminder, the PR needs to be merged, if okay.

app/embeds/GoogleDrawings.js
```diff
@@ -0,0 +1,38 @@
...
+export default class GoogleDrawings extends React.Component<Props> {
+  static ENABLED = [URL_REGEX];
+
+  render() {
+    return (
+      <Frame
+        {...this.props}
+        src={this.props.attrs.href.replace("/edit", "/preview")}
+        icon={
+          <img
```
tommoor, line 25: Please pull in `develop` and use the `Image` component here now, it added support for loading images from a CDN since this fork was created. see:

https://github.com/outline/outline/blob/develop/app/embeds/GoogleDocs.js#L24

milesstoetzner: done :)

tommoor (approved):

Issue #817, opened 2018-12-15
Embed Services


tommoor:
Continuing from https://github.com/outline/outline/pull/680 – Please add requests for additional services as comments on this issue.

- [x] Prezi
- [x] Abstract
- [ ] Speakerdeck
- [ ] Leanboard
- [ ] Twitter
- [ ] Google Maps
- [ ] Looker
- [ ] Overflow.io

nnblack1980:
Abstract please :)

tommoor:
I can promise that Outline will be the first service to have Abstract embeds 😉

nnblack1980:
Excellent news - thanks @tommoor

jbrewer:
[ ] Google Slides 
[ ] Notion
[ ] Asana

coleylogan:
Github to see a brief summary of any issues I reference in docs.

locomot1f:
- [ ] Mindmeister

locomot1f:
so... I love the option to embed google slides.
One option that would be awesome would be to embed the "presentation" format of the slide, as opposed to the "edit".
so, if I try to embed this link - https://docs.google.com/presentation/d/1a_18aJ8rshQWK8q5bwT5wG2z97o9Lb315-uTEfXN-eo/**edit**?usp=sharing - it's fine
but if I try this link - https://docs.google.com/presentation/d/1a_18aJ8rshQWK8q5bwT5wG2z97o9Lb315-uTEfXN-eo/**present**?usp=sharing - i get a blank box where the presentation should be.

tommoor:
That would be nice, unfortunately it looks like Google sets some security headers that prevent the presentation version from being embedded as an iframe 😢 

![image](https://user-images.githubusercontent.com/380914/51625847-1c007d80-1f36-11e9-856a-65607996ed26.png)

atomobianco:
- [ ] LaTex (through https://katex.org/ or else)

aam1r:
* [ ] Whimsical or [mermaid](https://mermaidjs.github.io/#/)

MatheusRV:
- [X] PowerBI (done in https://github.com/outline/outline/pull/1325)
- [ ] Linkedin SlideShare
- [ ] SlideShare
- [ ] DailyMotion
- [ ] Flickr
- [ ] Giphy
- [ ] Kickstarter Campaign
- [ ] Kickstater Catalogue
- [ ] Vimeo
- [ ] Twitter Tweet
- [ ] Twitter HashTag Feed
[...]

PR #1815, opened 2021-01-16
feat: generic frame

milesstoetzner:
Hello, 

I implemented an Integration that basically accepts any URL as iframe src.
See "Any URL" in https://github.com/outline/outline/issues/817#issuecomment-653703111.

Screenshot from 2021-01-16 14-08-10


milesstoetzner:
okay there might be a problem. 

other integrations are replaced by this generic frame integration after rerendering a page.
I assume that this happens since the generic frame basically accepts any url.

Not familiar with the integration selection algorithm but I guess some kind of longest prefix matching would be working.

tommoor:
Yes, I think you have found the problem with this – not only other integrations will be replaced but every url in all existing documents which would cause many to break.

In order to achieve this we'd need a way to mark that urls should be considered an "embed" beyond the current method of checking if the url and alt text are identical. We could probably use a similar method to what was [just used for image alignment](https://github.com/outline/rich-markdown-editor/pull/345), storing the metadata in the title part of the markdown string. This work would need to be done in the `rich-markdown-editor` repo.

MatheusRV:
@milesstoetzner nice work, you could try to start with others simplest embedded. But still helping make Outline the best knowledge base.

tommoor:
closing this out as it'll need more work in the editor before being mergable, feel free to reopen if it gets there

matbrgz:
@tommoor  Is possible to re-open this? Since #2302 is done.

michalkos:
Would love to see this re-opened.

PR #1841, opened 2021-01-23
chore: Sentry performance monitoring

tommoor:
Adds Sentry [performance monitoring](https://docs.sentry.io/platforms/javascript/guides/react/#monitor-performance) when a sentry dsn is provided.


shared/i18n/index.js
```diff
@@ -24,7 +24,7 @@ export const initI18n = () => {
       },
       lng,
       fallbackLng: lng,
-      debug: process.env.NODE_ENV !== "production",
+      debug: process.env.NODE_ENV === "development",
```
tommoor, line 27: Stops i18n logging in CI

PR #1865, opened 2021-02-05
feat: Inline collection editing

### LINKED ISSUE #1802: Move collection description editing inline
Currently editing the description of a collection happens in the "Edit" modal with name/icon/color etc. Because the description is a rich text area this doesn't work too well and unnecessarily disconnects the contents from where it appears – it should be more like a document. I suggest the following:

- Add an "Edit description…" option to the collection menu (ensure policies enable)
- Have it navigate to collection/<uuid>/edit
- Show the collection as normal except with the description area editable, focused, and with a "Save" button (design needed?)
- Remove the description from the "Edit collection" modal

tommoor:
closes #1802

image

image

image


matbrgz:
AWESOME. CONGRATZ TOM.. Fascinating design!! How about start improvements on the design system? I could help with Guidelines and Principe, Accessibility, UX Research etc.

tommoor:
A more defined design system would be great, but it is a lot of work – I think a good starting point would be to implement [Storybook](https://www.learnstorybook.com/intro-to-storybook/react/en/get-started/) and start documenting the existing components

PR #1867, opened 2021-02-05
fix: Pass credentials with API requests when required to work with Cloudflare Access

### LINKED ISSUE #1850: CORS errors when loading manifest.json behind Cloudflare Access
**To Reproduce**
1. Deploy Outline on your own self-hosted deployment
2. Use Cloudflare for DNS and point a DNS record to your Outline deployment
3. Enable Cloudflare Access

**Expected behavior**

`manifest.json` will fail to load:

image

- Install: self-hosted v0.52.0

**Similar issues:**

https://community.cloudflare.com/t/cloudflare-access-manifest-json-issues/228716

### LINKED ISSUE #1866: Outline not working when self-hosted behind Cloudflare Access
We have a number of web apps self-hosted in AWS and Google Cloud behind Cloudflare. We also use [Cloudflare Access](https://blog.cloudflare.com/introducing-cloudflare-access/) to control access to these web apps.
Recently we decided to deploy Outline the same way and found out it's not working - the main screen just renders blank.

How Cloudflare Access works:
* Domain names of protected resources resolve to CF edge proxies
* When requesting a protected URL, an edge proxy verifies an authentication cookie in the request
* If the cookie is missing or invalid/expired, the client is redirected to a CF service domain  that offers a choice of auth/SSO options
* After selecting the auth method and successful authentication, the client is redirected to a CF service URL on the original domain, where the auth cookie is set, and then back to the original URL
* Edge proxy verifies the auth cookie and passes the request to a backend where the web app is actually deployed (e.g. AWS load balancer)

What happens when Outline is deployed behind CF Access:
* The index HTML and main app bundles are loaded normally by the browser
* The app tries to fetch() `/auth.config` using an [ApiClient](https://github.com/outline/outline/blob/v0.52.0/app/utils/ApiClient.js)
* Since `ApiClient` always calls fetch() with [credentials: "omit"](https://github.com/outline/outline/blob/v0.52.0/app/utils/ApiClient.js#L94), the auth cookie set by CF Access is not passed in the request
* The request is then redirected to the CF Access auth landing page as described above
* The unexpected response breaks the app

How to fix the issue:
* Make `ApiClient` either always use `credentials: "same-origin"` or use `credentials: "omit"` for cross-origin requests and `credentials: "same-origin"` for same-origin requests. (Quick grepping of the source code gives an impression that `ApiClient` is only used for same-origin requests at the moment.) Please see [fetch() docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters) for reference.

This also affects the web manifest provided [in the index HTML](https://github.com/outline/outline/blob/v0.52.0/server/static/index.html#L13) with `<link rel="manifest" href="/manifest.json" />`. Behind CF Access it fails to load because without an explicit `crossorigin` attribute the auth cookie is not passed in the request - please refer to the docs on deployment of [web manifests behind authentication](https://developer.mozilla.org/en-US/docs/Web/Manifest#deploying_a_manifest_with_the_link_tag).

Should be fixed by replacing the tag with `<link rel="manifest" href="/manifest.json" crossorigin="use-credentials" />`.
This issue has already been reported in #1850 

I'm going to file a pull request that fixes these issues and has been verified as working in our local environment.

**Outline:**
- Install: self hosted
- Version: v0.52.0

**Desktop:**
 - OS: macOS Big Sur 
 - Browser: Chrome 88.0, Firefox 85.0

sangwa:
Makes Outline work correctly behind Cloudflare Access proxies which require passing credentials (cookies) with API requests.

Fixes #1866
Fixes #1850

### DIFF
--- a/app/utils/ApiClient.js
+++ b/app/utils/ApiClient.js
@@ -1,9 +1,9 @@
 // @flow
 import invariant from "invariant";
 import { map, trim } from "lodash";
+import { getCookie } from "tiny-cookie";
 import stores from "stores";
 import download from "./download";
-import { getCookie } from "tiny-cookie";
 import {
   AuthorizationError,
   BadRequestError,
--- a/app/utils/ApiClient.js
+++ b/app/utils/ApiClient.js
@@ -3,6 +3,7 @@ import invariant from "invariant";
 import { map, trim } from "lodash";
 import stores from "stores";
 import download from "./download";
+import { getCookie } from "tiny-cookie";
 import {
   AuthorizationError,
   BadRequestError,
@@ -17,6 +18,11 @@ import {
 type Options = {
   baseUrl?: string,
 };
+
+// authorization cookie set by a Cloudflare Access proxy
+const CF_AUTHORIZATION = getCookie("CF_Authorization");
+// if the cookie is set, we must pass it with all ApiClient requests
+const CREDENTIALS = CF_AUTHORIZATION ? "same-origin" : "omit";
 
 class ApiClient {
   baseUrl: string;
@@ -91,7 +97,7 @@ class ApiClient {
         body,
         headers,
         redirect: "follow",
-        credentials: "omit",
+        credentials: CREDENTIALS,
         cache: "no-cache",
       });
     } catch (err) {
--- a/server/static/index.html
+++ b/server/static/index.html
@@ -12,7 +12,7 @@
       href="/favicon-32.png"
       sizes="32x32"
     />
-    <link rel="manifest" href="/manifest.json" />
+    <link rel="manifest" href="/manifest.json" crossorigin="use-credentials" />
     <link
       rel="search"
       type="application/opensearchdescription+xml"


tommoor (commented):
Hey, this looks sensible – do you know are there any other clues that the app is being Cloudflare Access? I'm tempted to actually look for the cookie and only turn on credential sending if it exists because `omit` is used here as a performance measure to stop cookies being sent with every API request.

sangwa:
> I'm tempted to actually look for the existence of the _specific_ cookie and only turn on credential sending if it exists

Indeed, the name of the cookie [is documented](https://developers.cloudflare.com/access/setting-up-access/json-web-token) as `CF_Authorization`.
I've pushed a more generic patch that considers it and looks working in my local env - please review.

app/stores/AuthStore.js
```diff
@@ -36,10 +36,17 @@ export default class AuthStore {
...
+  externalAuthCookie: ?string;
 
   constructor(rootStore: RootStore) {
     this.rootStore = rootStore;
 
+    // attempt to fetch an authorization cookie potentially set by an external
+    // authorizing proxy (if configured), to use with ApiClient
+    if (env && env.EXTERNAL_AUTH_COOKIE) {
+      this.externalAuthCookie = getCookie(env.EXTERNAL_AUTH_COOKIE);
+    }
```
tommoor, line 48: I don't think this needs to be in the auth store to be honest, we can evaluate it once in the context of the ApiClient module on load, I'd also be totally fine with `CF_Authorization` being the default, it's never come up before so the need for it being configurable seems slim.

How does that sound?

sangwa: The motivation to use the auth store was that it's already imported and used in ApiClient and this could also be useful for possible future integration of external authorization with Outline authorization.
Though I agree that's a kind of over-engineering.

sangwa:
Pushed a much simpler fix that specifically targets the issue with Cloudflare Access.
@tommoor what do you think?

tommoor:
I think this is good, will test and get merged this weekend ✌️

app/stores/AuthStore.js
tommoor: Always a balancing act, particularly in long running projects like this I'm wary of introducing config that I don't know if it's used but have to support forever 😅

Issue #1856, opened 2021-01-30
Update hotkeys to use a hooks-based solution


tommoor:
One of the items keeping us in a component-based world is usage of `react-hotkey`, we should remove it and update to something with a modern API like https://johannesklauss.github.io/react-hotkeys-hook/

There are only 5 components using this right now, so it should be a manageable PR to convert them to functional.

iamsaumya:
Hey, Should I take over this or wait for my other PR to get resolved first?

tommoor:
I'm unlikely to tackle this in the coming week 👍

iamsaumya:
Great!, this is on my plate now :smile: .

PR #1722, opened 2020-12-16
feat: reordering documents in collection

thenanyu:
WIP

- so far, just a sketch of some gap-cursor interactions
- tuned the interaction to feel more natural by having a larger hover footprint than the visual gap cursor

Inspired by #1283
Closes #622

---

Quick functional spec of intended re-ordering heuristics

image

PR #1870, opened 2021-02-07
feat: Allow sorting collections in sidebar

### LINKED ISSUE #1759: Allow sorting collections in sidebar
### Why

Now that document drag and drop is complete, as an admin I should be able to choose the order of collections in the sidebar in the same manner.

### How

Probably a new `collections.move` endpoint to match `documents.move` and a new corresponding policy as well that only allows admins in a first version.

To avoid having to pass the entire list of collections back to the server we can use the concept of fractional indexing. I'll link the library here to achieve this soon.

iamsaumya:
Closes #1759 

Todo
- [x]  Take care of the top cursor.
- [x]  Introduce an index attribute in the collection table.
- [x]  Create a middleware to sanitize the index` attribute of all the collections. 
- [x]  Update `collections.create` endpoint to create an index as well. 
- [x]  Create a `collections.move` endpoint 
- [x]  Add a policy to only admin to move the collection.
- [x]  Add test for `collection.move`
- [x]  Integrate with websockets.


iamsaumya:
A possible algorithm to use fractional indexing, by default all the collections will have their fractional index `null`.

1. Fetch all the collections.
2. Iterate through the collections. If two or more collections have the same index, leaving a collection to reinitialize all other to `null`.
For example: if 3 collections have the same index, then reset the index of two of them to `null`.
3. Sort the collections such that all the collections with `null` index are at the bottom of the list. For example: `["23.32","32.11",null,null]`
4. Pass the collection list to a function which will create generate a new index for the collection with `null` index value.

iamsaumya:
Tom, please let me know if the todos and algorithm look decent enough. 

If we are introducing a new attribute in the collection table then what are things I should take care of in the codebase?

tommoor:
Looking good – note that because the "cursor" is shown below, moving to the top of the list is a special case - for documents the additional "top" cursor is here: https://github.com/outline/outline/pull/1722/files#diff-c6ce8fd8c9fd8e2796d41937514439cadff8e2d64b3306caf77a4e4fb07d9901R106

> Tom, please let me know if the todos and algorithm look decent enough.

If the existing collection indexes are null then it does introduce a lot of annoying edge-cases. For example it won't really be possible to calculate a valid index on the client without at least calculating indexes for every item above it. This makes me think that we should create indexes "just in time" when `collections.list` is called if they don't exist, so the new client code or `collections.move` never has to deal with collections lacking an index.

> If we are introducing a new attribute in the collection table then what are things I should take care of in the codebase?

- I suggest we call the column `index`, it should be text
- See docs on [migrations](https://github.com/outline/outline#migrations)
- Add the field to the [model](https://github.com/outline/outline/blob/main/server/models/Collection.js)
- Add the field to the [presenter](https://github.com/outline/outline/blob/main/server/presenters/collection.js)

iamsaumya:
While converting the `Collections.js` to a functional component to takecare of the top cursor, I am not sure how to implement the @keydown('n') decorator. Is there a way to implement it in a function component?

iamsaumya:
Also, I am unable to trigger the top cursor with a document. When I try to hover in-between the first document and collection name, the drop to import functionality gets triggered. 

https://user-images.githubusercontent.com/40148448/107362701-1cf5af00-6aff-11eb-93db-0310da8a083f.mp4

tommoor:
>  I am not sure how to implement the @keydown('n') decorator. Is there a way to implement it in a function component?

I have an issue for this, but didn't expect it to be a blocker here – https://github.com/outline/outline/issues/1856 – I think the pragmatic solution is to just move that keyboard shortcut to `Layout.js` for now, it doesn't really have to live in `Collections.js`, it was just handy because the new document function already existed.


> When I try to hover in-between the first document and collection name, the drop to import functionality gets triggered.

I'm not able to reproduce this in production, any chance it's been broken on your branch?

iamsaumya:
> any chance it's been broken on your branch?

Yeah, It was. I fixed it.

iamsaumya:
Hey Tom, There are some warning in the console after pulling from the main branch. You might want to check on it.

iamsaumya:
> This makes me think that we should create indexes "just in time" when collections. the list is called if they don't exist, so the new client code or collections. the move never has to deal with collections lacking an index.

Keeping your remarks in mind, I have put some thought together and here is the solution that I propose. 

All the collections irrespective of being private or public have something in common i.e `teamId`. Therefore, we can keep all the collections of a team sorted and it will take care of all the edges cases involved in a collection that going private from public or vice versa.

We can introduce an `index` column of `text` datatype with the default value `null` ( lemme know if it has to be `""`). 

In `collections.list`,

1. We will first fetch all the collections of the user's team.
1. Validate that there are no index collision or a collection with the default value.
1. If there is, then we will fix it. 
1. Usual `collections.list`.

How will the `collections.move`  work?

`collection.move` will take 2 parameters. 
1. The collection that is to be moved.
2. The collection that will be present above the new position of the dragged collection. 

`collection.move( draggedCollection , aboveCollection )`

In the backend, we will fetch all the collections of user's team and iterate through it till we find the collection just next to the `aboveCollection` in the fetched list. If  `aboveCollection`  is the last collection then the second index will be null. 

Now we have two indexes and we calculate the new fractional index and assign it to `draggedCollection`. 

If the `draggedCollection` is moved to the top in the sidebar then the `aboveCollection` will be null and now we know that `draggedCollection` has to be at the top.

tommoor:
`collections.list` approach and migration sounds correct.

`collections.move` can be simpler, the beauty of using the fractionalIndex approach is that you don't need to send any more data than the index of the collection being moved (no need to send the "above" collection) ([reference](https://www.figma.com/blog/realtime-editing-of-ordered-sequences/)).

`collections.move` can accept a `collectionId` and `index` in the request, on the server we must check if the collection index is existing in the team context (in the case that two people move at once, or there is a private collection that the user doesn't have access to with that index), in this case we then get the above collection and generate an index between the two. These can probably be subsequent queries as this case will be very rare.

iamsaumya:
I am writing the logic to sanitize the `index` field of the collections for `collections.list` in a middleware so that we plug it wherever we want in future.

iamsaumya:
whoopsie! A circleci check is failing. What exactly is the failed check about and What should I do to fix it?

server/api/middlewares/collectionIndexing.js
```diff
@@ -0,0 +1,67 @@
+// @flow
+import fractionalIndex from "fractional-index";
+import { type Context } from "koa";
+import { Collection } from "../../models";
+export default function collectionIndexing() {
+  return async function collectionIndexingMiddleware(
+    ctx: Context,
```
tommoor, line 7: Needs to be `ContextWithState` imported from the internal types file to fix your CI error

server/api/middlewares/collectionIndexing.js
```diff
@@ -0,0 +1,67 @@
...
+    next: () => Promise<*>
+  ) {
+    const teamId = ctx.state.user.teamId;
+    const collections = await Collection.findAll({
```
tommoor, line 11: The problem with this approach is you're pulling all the collections from the db whether they're needed or not on every request that uses the middleware. Consider we already load collections inside of `collections.list` why not perform this as a one-off when those collections are loaded just check to see if first has a `null` index property?

iamsaumya:
Oh god please! all this time I thought the `pending` tag meant that I am waiting for your response. Had posted this review a week before and was waiting for you 🤦

server/api/middlewares/collectionIndexing.js
tommoor: I think we can use the users collections from the existing query as a "hint" to whether the migration needs to happen if the first index is `null`, and then in only that case we can query _all_ collections and fix them.

iamsaumya:
Hey Tom, Once this feat is in place, Are we going to discard listing the collections in descending order by `updatedAt` or both the options are going to coexist?

tommoor:
The functionality will remain in the API to sort collections by various means but the UI will only allow the manual ordering

iamsaumya:
Alright, I will update `colletions.list` to return collections in asc order by index.

iamsaumya:
Almost done! :man_dancing:  

Have added a few more to-dos to the list, please let me know they suffice. Thanks!!

iamsaumya:
On hold because of https://github.com/goabstract/fractional-index/issues/1

iamsaumya:
I think we will have to sort the collections in the app using normal `sort()`. In the package, If a is `"2"` and b is `"11"` then it will throw an error ( `a>=b` error). And because we are sorting the collections using natural sort, this error is very likely to happen in the app.

tommoor:
Yea, that sounds right – natural sort is for human-like sorting, in this case we want rigid logical machine sorting :)

iamsaumya:
Which comment did the outline delete? :thinking:

iamsaumya:
I modified `collections.create` endpoint to take `index` in the body. However, I did not make it a required field as it will help the API users to create a collection without bothering about `index`. 

Please guide me on how to update the API docs to show the `index` in request and responses.

iamsaumya:
Yay :beers:, Worked on everything I think there was to work on. If I have left something, please let know :smiley:

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -136,6 +194,11 @@ function CollectionLink({
   );
 }
 
+const Draggable = styled("div")`
+  opacity: ${(props) => (props.$isDragging || props.$isMoving ? 0.5 : 1)};
+  pointer-events: ${(props) => (props.$isMoving ? "none" : "all")};
```
thenanyu, line 199: this is a small nit — would `"auto"` be a better default value than `"all"` here

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -74,49 +78,103 @@ function CollectionLink({
...
+    }),
+    canDrag: (monitor) => {
+      return can.move;
+    },
+    begin: (monitor) => {
+      ui.activeCollectionId = "";
+    },
+    end: (monitor) => {
+      ui.activeCollectionId = monitor.activeCollectionId;
+    },
```
thenanyu, line 119: What I think you're attempting here is to temporarily hide the collection contents when a drag is active.
- modifying central state like this is a bit of a roundabout way to accomplish this goal
- instead, move the calculation for `expanded` https://github.com/outline/outline/pull/1870/files#diff-c6ce8fd8c9fd8e2796d41937514439cadff8e2d64b3306caf77a4e4fb07d9901R47 below here and simply change its definition to also include `!isCollectionDragging`

server/api/collections.js
```diff
@@ -608,4 +632,31 @@ router.post("collections.delete", auth(), async (ctx) => {
...
+  ctx.assertPresent(index, "index is required");
+  ctx.assertUuid(id, "id must be a uuid");
+
+  const user = ctx.state.user;
+  const collection = await Collection.findByPk(id);
+  authorize(user, "move", collection);
+
+  collection.index = index;
+
+  await collection.save();
```
tommoor, line 646:
```suggestion
  await collection.update({ index });
```

app/scenes/CollectionNew.js
```diff
@@ -38,13 +38,15 @@ class CollectionNew extends React.Component<Props> {
...
     this.isSaving = true;
+    const { collections } = this.props;
     const collection = new Collection(
       {
         name: this.name,
         sharing: this.sharing,
         icon: this.icon,
         color: this.color,
         private: this.private,
+        index: collections.indexForNew,
```
tommoor, line 49: What do you think about moving this to the server? After all the client might not be aware of all the collections… which would result in the new collection potentially being inserted part way down the list.

I think _allowing_ `index` in the create is okay from an API pov though

app/stores/CollectionsStore.js
```diff
@@ -33,10 +34,24 @@ export default class CollectionsStore extends BaseStore<Collection> {
...
+        return 0;
+      }
+      return a.index < b.index ? -1 : 1;
+    });
+  }
+
+  @computed
+  get indexForNew(): string {
+    const collections = this.orderedData;
+    return fractionalIndex(collections[collections.length - 1].index, null);
```
tommoor, line 54: See comment above about moving this to the server, we'd have to query the last collection in the team

app/components/Sidebar/components/Collections.js
```diff
@@ -1,98 +1,91 @@
...
-
-  componentDidMount() {
-    const { collections } = this.props;
-
+function Collections({ onCreateCollection }: Props) {
+  const { ui, policies, documents, collections } = useStores();
+  const isPreloaded: boolean = !!collections.orderedData.length;
+  const { t } = useTranslation();
+  const orderedCollections = collections.orderedData;
+  React.useEffect(() => {
```
tommoor, line 25:
```suggestion

  React.useEffect(() => {
```

Nit: A little extra space between methods and `const` would be nice :)

server/utils/collectionIndexing.js
```diff
@@ -0,0 +1,58 @@
+// @flow
+import fractionalIndex from "fractional-index";
+import { Collection } from "../models";
+export default async function collectionIndexing(teamId: string) {
```
tommoor, line 4:
```suggestion

export default async function collectionIndexing(teamId: string) {
```

server/api/collections.js
```diff
@@ -571,6 +574,27 @@ router.post("collections.list", auth(), pagination(), async (ctx) => {
...
+  const indexSet = new Set();
+
+  const repeatedIndex = collections.some((collection) => {
+    if (indexSet.has(collection.index)) {
+      return true;
+    }
+    indexSet.add(collection.index);
+    return false;
+  });
+
```
tommoor, line 590: I'm not a huge fan of this much logic running in the controller method, is the repeated index check necessary?

I was previously thinking the "migration" would be a one-time thing, perhaps the repeat check should happen as part of `collections.move` before writing the new index?

tommoor (changes requested):
Thanks so much for the work here 💪

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -74,49 +78,103 @@ function CollectionLink({
...
+    collect: (monitor) => ({
+      isCollectionDragging: monitor.isDragging(),
+    }),
+    canDrag: (monitor) => {
+      return can.move;
+    },
+    begin: (monitor) => {
+      ui.activeCollectionId = "";
+    },
```
tommoor, line 116:
```suggestion
    begin: (monitor) => {
      setTimeout(() => {
        ui.clearActiveCollection();
      }, 0);
    },
```

In my testing this needs to be moved to next frame otherwise collections below the active collection are not draggable, this seems like it _might_ be a browser quirk

server/api/collections.js
iamsaumya: I agree, let's move this logic elsewhere. 

I think the scenario which will lead to the repeated index is when a private collection is made public. 

How about we run this logic to specifically when the visibility of collections is changed? 

I think It will be unpredictable for the user when they move a collection `X` and they see other collections move as well because of index collision.

app/components/Sidebar/components/CollectionLink.js
iamsaumya: Done, Thanks Nan!

app/components/Sidebar/components/CollectionLink.js
iamsaumya: Indeed, it was a bug. I should have checked it. Thanks :smile:

app/scenes/CollectionNew.js
iamsaumya: Let's move this to the server. If the index is `undefined` in the request payload we will insert the collection at the bottom of the collections (this is will be the default behaviour for the app). 

If the index is specified we will also need to check if the index characters are between the ASCII characters that we allow.

I was wondering, what If the index in the req is already taken? Trello API doesn't specify its behaviour for this edge case.

tommoor: > I was wondering, what If the index in the req is already taken?

I think this is where moving the index de-duplication out of the `collections.list` method comes in. If we have a de-dupe check as part of create and move then we can just adjust the item downward by one if that ever happens.

server/api/collections.js
```diff
@@ -51,6 +54,29 @@ router.post("collections.create", auth(), async (ctx) => {
...
+  if (index) {
+    const allowedASCII = new RegExp(/^[\x21-\x7E]+$/);
+    if (!allowedASCII.test(index)) {
+      throw new ValidationError(
+        "Index characters must be between x21 to x7E ASCII"
+      );
+    }
+  } else {
+    collections.sort((a, b) => {
```
iamsaumya, line 70: I had to sort the collections manually because `order by` wasn't returning the collections in the order presented in the app.

tommoor: Do you mean the postgres order wasn't working as expected? I don't think its feasible to just query all the collections and sort them in js, I can look into the postgres ordering if that's the problem.

iamsaumya: Sorry to put you through this work, but I think you should take a look as well. This was my query.

```js
  const collections = await Collection.findAll({
    where: { teamId: user.teamId, deletedAt: null },
    attributes: ["id", "index", "updatedAt"],
    order: [["index", "DESC"]],
    limit: 1,
  });
```

iamsaumya:
Moved collections don't reflect in other users sidebar, unlike other operations which do. How does this work? We should implement it for reordering collections as well.

server/api/collections.js
tommoor: Try the following:

`order: ['index COLLATE "C"', "ASC"]],`

Using collate makes postgres sort by the byte-order ([ref](https://dba.stackexchange.com/questions/94887/what-is-the-impact-of-lc-ctype-on-a-postgresql-database)), which is what we want in this case. I'd switch it round so that new collections go to the top – they could be lost easily at the bottom, this will also match the behavior of new docs.

iamsaumya: This was very informative, Thanks Tom! 

Yeah sure, the current implementation was inspired by Trello's way of creating cards. Now, Let's make the new collections go to the top.

iamsaumya:
I grasped a bit of how `socket.io` is being used in the codebase. I Will work on it too. 

The todos for a feat req keeps increasing as I connect more dots in the codebase :muscle:. I think this is a good sign :sweat_smile:

iamsaumya:
@tommoor please guide on what will be the best approach to let the client know that there was an index collision when `collections.create` or `collections.move` was fired.

For `collections.move` we can send the collectionId and their new index to client in the response. However, sending these details on the `collections.create` request as well feels odd. What do you think?

iamsaumya:
Pls ignore my last message :sweat_smile: 

I think this is it. I want to roll out this feature in the March release.

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -74,49 +78,105 @@ function CollectionLink({
     }),
   });
 
+  //Drop to reorder Collection
```
tommoor, line 81:
```suggestion
  // Drop to reorder Collection
```

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -74,49 +78,105 @@ function CollectionLink({
...
+      id: collection.id,
+    },
+    collect: (monitor) => ({
+      isCollectionDragging: monitor.isDragging(),
+    }),
+    canDrag: (monitor) => {
+      return can.move;
+    },
+    begin: (monitor) => {
+      setTimeout(() => {
```
tommoor, line 115: I almost forgot why we did this already, needs a comment.

```suggestion
      // setTimeout is required to move this to next render frame otherwise collections
      // below the active collection are not draggable, we believe this is a browser quirk
      setTimeout(() => {
```

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -74,49 +78,105 @@ function CollectionLink({
...
-              <>
-                {can.update && (
-                  <CollectionSortMenuWithMargin
+        <Draggable
+          key={collection.id}
+          ref={dragToReorderCollection}
+          $isDragging={isCollectionDragging}
+          $isMoving={isCollectionDragging}
+        >
+          <DropToImport key={collection.id} collectionId={collection.id}>
```
tommoor, line 133: `key` no longer used here

```suggestion
          <DropToImport collectionId={collection.id}>
```

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -74,49 +78,105 @@ function CollectionLink({
...
+            <SidebarLinkWithPadding
+              key={collection.id}
```
tommoor, line 135: `key` not used here either

```suggestion
```

app/components/Sidebar/components/CollectionLink.js
tommoor: Actually, not sure this is even working any more as dragging is not collapsing the active collection at all.

app/components/Sidebar/components/Collections.js
```diff
@@ -1,98 +1,92 @@
...
+      );
+    },
+    canDrop: (item, monitor) => {
+      return item.id !== orderedCollections[0].id;
+    },
+    collect: (monitor) => ({
+      isCollectionDropping: monitor.isOver(),
+    }),
+  });
+  const content = (
```
tommoor, line 47:
```suggestion

  const content = (
```

app/stores/CollectionsStore.js
```diff
@@ -95,6 +103,18 @@ export default class CollectionsStore extends BaseStore<Collection> {
...
+  @action
+  move = async (collectionId: string, index: string) => {
+    const res = await client.post("/collections.move", {
+      id: collectionId,
+      index,
+    });
+    invariant(res && res.success, "Collection could not be moved");
+    const collection = this.get(collectionId);
+    collection.updateIndex(index);
+    this.addPolicies(res.policies);
```
tommoor, line 115:
```suggestion
```

server/api/collections.js
```diff
@@ -49,6 +55,27 @@ router.post("collections.create", auth(), async (ctx) => {
   const user = ctx.state.user;
   authorize(user, "create", Collection);
 
+  const collections = await Collection.findAll({
+    where: { teamId: user.teamId, deletedAt: null },
+    attributes: ["id", "index", "updatedAt"],
+    order: [
+      //using LC_COLLATE:"C" because we need byte order to drive the sorting
+      sequelize.literal('"collection"."index" collate "C"'),
```
tommoor, line 63: 👍

server/api/collections.js
```diff
@@ -49,6 +55,27 @@ router.post("collections.create", auth(), async (ctx) => {
   const user = ctx.state.user;
   authorize(user, "create", Collection);
 
+  const collections = await Collection.findAll({
+    where: { teamId: user.teamId, deletedAt: null },
+    attributes: ["id", "index", "updatedAt"],
+    order: [
+      //using LC_COLLATE:"C" because we need byte order to drive the sorting
```
tommoor, line 62:
```suggestion
      // using LC_COLLATE:"C" because we need byte order to drive the sorting
```

server/api/collections.js
```diff
@@ -49,6 +55,27 @@ router.post("collections.create", auth(), async (ctx) => {
   const user = ctx.state.user;
   authorize(user, "create", Collection);
 
+  const collections = await Collection.findAll({
+    where: { teamId: user.teamId, deletedAt: null },
+    attributes: ["id", "index", "updatedAt"],
+    order: [
```
tommoor, line 61: We only need the first collection, right?

```suggestion
    limit: 1,
    order: [
```

server/api/collections.js
```diff
@@ -59,14 +86,17 @@ router.post("collections.create", auth(), async (ctx) => {
     private: isPrivate,
     sharing,
     sort,
+    index,
   });
 
+  const collectionsIdWithIndex = await removeIndexCollisions(user.teamId);
```
tommoor, line 92: The use of this method is pretty much the only thing I'm left questioning in this PR, it feels very brute force. We already know the new `index` we're saving here, why don't we just check if it exists in the team already with a specific query rather than querying all collections in the team..

server/api/collections.test.js
```diff
@@ -130,6 +130,38 @@ describe("#collections.import", () => {
...
+    const { collection } = await seed();
+
+    const res = await server.post("/api/collections.move", {
+      body: { token: user.getJwtToken(), id: collection.id, index: "P" },
+    });
+
+    expect(res.status).toEqual(403);
+  });
+
+  it("should return success", async () => {
```
tommoor, line 153: We should add a test for the expectation of what happens when the index is already taken here.

server/api/collections.js
iamsaumya: Actually, this is how it should have been called

```js
  const collectionsIdWithIndex = await removeIndexCollisions(
    user.teamId,
    index,
    collections
  );
```
But again, all this logic was for my previous implementation. Earlier, I thought to remove the index collision of not only the `new index` but other indexes as well. Now that I think of it I unable to create a specific scenario where this is necessary. 

I will push your suggestion in an hour if I cannot contrive an example.

server/api/collections.test.js
iamsaumya: Done. Does it look good?

tommoor:
Made some naming tweaks and flow fixes on this branch, you able to cherry-pick the commits across?

https://github.com/outline/outline/compare/iamsaumya-reorder-collection-dev?expand=1

I can't PR into your branch unfortunately.

tommoor (approved):
Much appreciated @iamsaumya – great work 👏

server/utils/removeIndexCollisions.js
```diff
@@ -0,0 +1,64 @@
...
+import { Collection } from "../models";
+import { sequelize } from "../sequelize";
+
+/**
+ *
+ * @param teamId The team id whose collections has to be fetched
+ * @returns An array or undefined. If array, each array element is an array of collided collection id and its new index [id,newIndex]
+ */
+export default async function removeIndexCollisions(teamId: string) {
+  let collections = await Collection.findAll({
```
tommoor, line 12: This is still a bit brute-force for my liking, we really shouldn't need to query _all_ collections like this as we're always constraining collisions to a collection that's being added or updated.

Given that this is an infrequent operation I don't think it should hold up merging the PR though

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -18,10 +18,12 @@ import CollectionSortMenu from "menus/CollectionSortMenu";
 
 type Props = {|
   collection: Collection,
-  ui: UiStore,
```
tommoor, line 21: I think you brought across some but not all of my changes here, we should either pass `ui` as a prop or take it from `useStores`

iamsaumya: Done. Fetching from `useStores`

server/utils/removeIndexCollisions.js
iamsaumya: I am pushing an update, pls bear with me.

iamsaumya:
Removed the brute force approach, I wrote such code thinking of too much edges cases which in reality I think will never happen. 

Thanks for the reviews and your patience 😄

server/api/collections.test.js
```diff
@@ -198,6 +198,78 @@ describe("#collections.move", () => {
     expect(updateCreatedCollection.data.index).toEqual("h");
     expect(updateCreatedCollection.data.index > "Q").toBeTruthy();
   });
+
+  it("if index collision with an extra collection, should updated index of other collection", async () => {
```
tommoor, line 202: Is it possible to tweak the remove collision method so that the collection that gets the updated `index` is the one being acted upon, rather than the one it happens to be "next" to? 

I think that'll make a lot more sense. The `index` you provide to create/update is a clue as to where the collection should land visually rather than a need for it to be that specific character.

iamsaumya: I made it work this way because when I send an index through API as a developer I expect the collection's index to be that specific value. I don't care about the index of other collections. So I changed the index of other collections giving precedence to the one being acted upon.

iamsaumya: Tom, What do you think? I can always rewrite the function to behave the way you want.

tommoor: Personally I think the specific value is not important, the thing that's important is the position in the list but I'm very conscious of the amount of iteration we've already asked you to put into the PR.

One thing I wonder that might tip your thinking is that if we went the route of modifying the incoming `index` does that simplify the event data format too also as you wouldn't need to store an array of tuples? As only one collection index would ever be changed…

iamsaumya: Indeed, this will definitely simplify things. I will jump right into it.

iamsaumya: Done 🎉

tommoor: so CLEAN

iamsaumya: Ikr! Thanks for the nudge. I don't mind iterating a hundred times if the code quality improves.

tommoor (approved):
We did it 👏

PR #1875, opened 2021-02-08
feat: Added ability to disable sharing at collection

### LINKED ISSUE #840: Allow disabling public share links on individual collections
Mentioned on the original collection permissions thread, [here](https://github.com/outline/outline/issues/668#issuecomment-451714644). It should be possible to disable public sharing for private collections

Please vote if you would find this feature useful.

tommoor:
- Adds a new option to collections to disable sharing on a per-collection basis:
  image

- When disabled any previous public share links no longer work
- The new option in collection edit is disabled when sharing is already disabled for the team

closes #840


app/scenes/CollectionEdit.js
```diff
@@ -17,13 +18,15 @@ import Switch from "components/Switch";
 type Props = {
   collection: Collection,
   ui: UiStore,
+  auth: AuthStore,
   onSubmit: () => void,
   t: TFunction,
 };
 
 @observer
 class CollectionEdit extends React.Component<Props> {
```
tommoor, line 27: Need to refactor these components soon, getting quite dated.

PR #1885, opened 2021-02-15
Update app.json

bglidwell:
Currently, running the deploy to Heroku button results in the following

image

It's possible this is due to the postgres version heroku is now using by default?

### DIFF
--- a/app.json
+++ b/app.json
@@ -23,7 +23,10 @@
       "plan": "heroku-redis"
     },
     {
-      "plan": "heroku-postgresql"
+      "plan": "heroku-postgresql",
+      "options": {
+        "version": "10"
+      }
     }
   ],
   "scripts": {


tommoor:
Oh, this change fixes it for you?

bglidwell:
> Oh, this change fixes it for you? Or are you just guessing it might?

Guessing it might. Has worked for another wiki having the same issue in heroku

tommoor:
Just had a look and I have Outline running with v11, 12.5, and v13 in various environments so I don't think this is the source of the problem.

My guess is you should set `heroku config:set PGSSLMODE=require`

bglidwell:
Can that env variable be added to the app.json? Right now the deploy button just fails for me, I don't even get to a point where I can add env vars because it never gets that far. Thanks.

tommoor:
Try now, looks like `NODE_ENV` was not set to production so it would default to non-SSL

https://github.com/outline/outline/commit/887e341e4822189093c2182a88241c082f240789

bglidwell:
image
image

tommoor:
Where is it that you're deploying from? That version of the application is very out of date – it looks like it's pulling from the old `master` branch rather than the current `main` branch.

bglidwell:
https://elements.heroku.com/buttons/outline/outline

Just searched outline wiki on heroku elements.

tommoor:
Gotcha, I have contacted their support to see how it can be changed. In the meantime you can edit the url from that page to point at the `main` branch.

tommoor:
I got a reply from Heroku support, looks like you found a bug with their platform 😆 :

> Buttons are listed in our Elements Marketplace as long as they register at least one deploy to Heroku in the last 45 days. The button description page is automatically generated and updated from the linked Github repository at least once a day, but as you noticed the deploy URL always points to the master branch on the repository.
> 
> We have added a task to our backlog in order to fix that behavior and read the default branch from Github, so the correct deploy URL is used. I'll keep you posted when the change is ready.

PR #1890, opened 2021-02-15
feat: Embed Cawemo

PedroSeda:
Embed service for Cawemo.

Cawemo is a free Web-based BPMN platform.

I might have skipped a few step as to how to proceed with the feature request, but we are running a working version with cawemo in production, and we figured more people could use this.


matbrgz:
LGFM

app/embeds/Cawemo.js
```diff
@@ -0,0 +1,31 @@
...
+const URL_REGEX = new RegExp("https?://cawemo.com/(?:share|embed)/(.*)$");
+
+type Props = {|
+  attrs: {|
+    href: string,
+    matches: string[],
+  |},
+|};
+
+export default class Abstract extends React.Component<Props> {
```
tommoor, line 14:
```suggestion
export default class Cawemo extends React.Component<Props> {
```

tommoor:
Thanks!

I've never heard of this tool – it makes me think that we soon need to have a way to classify those integrations that are big enough to appear in the insert menu and ones that are supported for embedded but not necessarily advertised in that menu.

PedroSeda:
> Thanks!
> 
> I've never heard of this tool – it makes me think that we soon need to have a way to classify those integrations that are big enough to appear in the insert menu and ones that are supported for embedded but not necessarily advertised in that menu.

One of the things that drew us to use outline was the abundance of embeds, and the recent refactor made it absurdly easy to add new ones, kudos to you guys for that.
Maybe some logic to allow for a custom list of embeds to be displayed might allow the community to keep contributing with integrations and still keep the list tidy.
I'm not entirely familiar with the full outline architecture and might need some guidance, but I might be able to help with that.

PR #1918, opened 2021-02-22
fix: Drag and drop images in editor, conflict with sidebar react-dnd

tommoor:
see: https://github.com/react-dnd/react-dnd/pull/3052
see: https://github.com/react-dnd/react-dnd/issues/3099


app/components/Sidebar/components/Collections.js
```diff
@@ -35,7 +35,7 @@ class Collections extends React.Component<Props> {
   componentDidMount() {
     const { collections } = this.props;
 
-    if (!collections.isFetching && !collections.isLoaded) {
```
thenanyu, line 38: assuming this check is done internally on the store?

tommoor: `isFetching` is also true when a single collection is fetching, so there was the poss of a race condition which started happening with some of the tweaks on this branch

PR #1919, opened 2021-02-22
fix: Update hotkeys to use a hooks-based solution

### LINKED ISSUE #1856: Update hotkeys to use a hooks-based solution

iamsaumya:
Fixes #1856

### DIFF
--- a/app/components/DocumentListItem.js
+++ b/app/components/DocumentListItem.js
@@ -39,7 +39,7 @@ function replaceResultMarks(tag: string) {
   return tag.replace(/<b\b[^>]*>(.*?)<\/b>/gi, "$1");
 }
 
-function DocumentListItem(props: Props) {
+const DocumentListItem = React.forwardRef((props: Props, ref) => {
   const { t } = useTranslation();
   const currentUser = useCurrentUser();
   const [menuOpen, setMenuOpen] = React.useState(false);
@@ -69,6 +69,7 @@ function DocumentListItem(props: Props) {
         pathname: document.url,
         state: { title: document.titleWithDefault },
       }}
+      ref={ref}
     >
       <Content>
         <Heading>
@@ -136,7 +137,7 @@ function DocumentListItem(props: Props) {
       </Actions>
     </DocumentLink>
   );
-}
+});
 
 const Content = styled.div`
   flex-grow: 1;
--- a/app/components/InputSearch.js
+++ b/app/components/InputSearch.js
@@ -1,89 +1,85 @@
 // @flow
-import { observable } from "mobx";
 import { observer } from "mobx-react";
 import { SearchIcon } from "outline-icons";
 import * as React from "react";
-import { withTranslation, type TFunction } from "react-i18next";
-import keydown from "react-keydown";
-import { withRouter, type RouterHistory } from "react-router-dom";
-import styled, { withTheme } from "styled-components";
+import { useHotkeys } from "react-hotkeys-hook";
+import { useTranslation } from "react-i18next";
+import styled, { useTheme } from "styled-components";
 import Input from "./Input";
-import { type Theme } from "types";
+import useStores from "hooks/useStores";
 import { meta } from "utils/keyboard";
 import { searchUrl } from "utils/routeHelpers";
 
 type Props = {
-  history: RouterHistory,
-  theme: Theme,
   source: string,
   placeholder?: string,
   label?: string,
   labelHidden?: boolean,
   collectionId?: string,
-  t: TFunction,
 };
 
-@observer
-class InputSearch extends React.Component<Props> {
-  input: ?Input;
-  @observable focused: boolean = false;
+function InputSearch(props: Props) {
+  let input: ?Input;
+  const { history } = useStores();
+  const { t } = useTranslation();
+  const theme = useTheme();
 
-  @keydown(`${meta}+f`)
-  focus(ev: SyntheticEvent<>) {
+  const {
+    source,
+    placeholder = `${t("Search")}…`,
+    label,
+    labelHidden,
+    collectionId,
+  } = props;
+  const [focused, setFocused] = React.useState(false);
+
+  useHotkeys(`${meta}+f`, (ev: SyntheticEvent<>) => {
     ev.preventDefault();
-
-    if (this.input) {
-      this.input.focus();
+    if (input) {
+      input.focus();
     }
-  }
+  });
 
-  handleSearchInput = (ev: SyntheticInputEvent<>) => {
+  const handleSearchInput = (ev: SyntheticInputEvent<>) => {
     ev.preventDefault();
-    this.props.history.push(
+    history.push(
       searchUrl(ev.target.value, {
-        collectionId: this.props.collectionId,
-        ref: this.props.source,
+        collectionId: collectionId,
+        ref: source,
       })
     );
   };
 
-  handleFocus = () => {
-    this.focused = true;
-  };
+  const handleFocus = React.useCallback(() => {
+    setFocused(true);
+  }, []);
 
-  handleBlur = () => {
-    this.focused = false;
-  };
+  const handleBlur = React.useCallback(() => {
+    setFocused(false);
+  }, []);
 
-  render() {
-    const { t } = this.props;
-    const { theme, placeholder = `${t("Search")}…` } = this.props;
-
-    return (
-      <InputMaxWidth
-        ref={(ref) => (this.input = ref)}
-        type="search"
-        placeholder={placeholder}
-        onInput={this.handleSearchInput}
-        icon={
-          <SearchIcon
-            color={this.focused ? theme.inputBorderFocused : theme.inputBorder}
-          />
-        }
-        label={this.props.label}
-        labelHidden={this.props.labelHidden}
-        onFocus={this.handleFocus}
-        onBlur={this.handleBlur}
-        margin={0}
-      />
-    );
-  }
+  return (
+    <InputMaxWidth
+      ref={(ref) => (input = ref)}
+      type="search"
+      placeholder={placeholder}
+      onInput={handleSearchInput}
+      icon={
+        <SearchIcon
+          color={focused ? theme.inputBorderFocused : theme.inputBorder}
+        />
+      }
+      label={label}
+      labelHidden={labelHidden}
+      onFocus={handleFocus}
+      onBlur={handleBlur}
+      margin={0}
+    />
+  );
 }
 
 const InputMaxWidth = styled(Input)`
   max-width: 30vw;
 `;
 
-export default withTranslation()<InputSearch>(
-  withTheme(withRouter(InputSearch))
-);
+export default observer(InputSearch);
--- a/app/components/InputSearch.js
+++ b/app/components/InputSearch.js
@@ -4,9 +4,9 @@ import { SearchIcon } from "outline-icons";
 import * as React from "react";
 import { useHotkeys } from "react-hotkeys-hook";
 import { useTranslation } from "react-i18next";
+import { useHistory } from "react-router-dom";
 import styled, { useTheme } from "styled-components";
 import Input from "./Input";
-import useStores from "hooks/useStores";
 import { meta } from "utils/keyboard";
 import { searchUrl } from "utils/routeHelpers";
 
@@ -19,8 +19,8 @@ type Props = {
 };
 
 function InputSearch(props: Props) {
-  let input: ?Input;
-  const { history } = useStores();
+  const input = React.useRef<Input>(null);
+  const history = useHistory();
   const { t } = useTranslation();
   const theme = useTheme();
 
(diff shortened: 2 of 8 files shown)


iamsaumya:
The first commit settles a warning in the console, it used to come because we were using ref attribute to a functional component (took me 3 hours to understand `ref` and fix it!). 

Before I go edit the other two files(which are huge!) I have a few doubts :sweat_smile:  

If I use the `useLocation` hook(inside `Search.js`) the return type will not be `LocationWithState`, which is I assume our custom location type.  

1. Why do we have this custom type?
2. Where is the `custom location` being injected?

These might sound dumb questions, but I couldn't comprehend the way things are. So :man_shrugging:

tommoor:
> Why do we have this custom type?
> Where is the custom location being injected?

The type provided in `react-router-dom` is actually incorrect so the custom type is a hack for that. If it's still needed in those components we can cast the type, something like `const location: LocationWithState = useLocation();`

app/components/InputSearch.js
```diff
@@ -1,89 +1,85 @@
...
   collectionId?: string,
-  t: TFunction,
 };
 
-@observer
-class InputSearch extends React.Component<Props> {
-  input: ?Input;
-  @observable focused: boolean = false;
+function InputSearch(props: Props) {
+  let input: ?Input;
```
tommoor, line 22: Lets use `React.useRef()` for this now

app/components/InputSearch.js
```diff
@@ -1,89 +1,85 @@
...
+  useHotkeys(`${meta}+f`, (ev: SyntheticEvent<>) => {
+    ev.preventDefault();
+    if (input) {
+      input.focus();
     }
-  }
+  });
 
-  handleSearchInput = (ev: SyntheticInputEvent<>) => {
+  const handleSearchInput = (ev: SyntheticInputEvent<>) => {
```
tommoor, line 43: All callbacks in functional component need to use `React.useCallback` 👍

app/components/Layout.js
```diff
@@ -32,126 +29,114 @@ import {
...
-  @keydown(["t", "/", `${meta}+k`])
-  goToSearch(ev: SyntheticEvent<>) {
+  const handleCloseKeyboardShortcuts = React.useCallback(() => {
+    setKeyboardShortcutsOpen(false);
+  }, []);
+
+  useHotkeys("t", "/", `${meta}+k`, (ev: SyntheticEvent<>) => {
     ev.preventDefault();
     ev.stopPropagation();
     this.redirectTo = searchUrl();
```
tommoor, line 68: Sorry, this `redirectTo` approach is a bit of an old pattern that we need to remove, can use a standard `history.push(route)` like you've done in `InputSearch` already for these 😄

app/components/Sidebar/components/Collections.js
```diff
@@ -1,98 +1,80 @@
...
-@observer
-class Collections extends React.Component<Props> {
-  isPreloaded: boolean = !!this.props.collections.orderedData.length;
-
-  componentDidMount() {
-    const { collections } = this.props;
+const Collections = ({ onCreateCollection }: Props) => {
+  const { collections, ui, policies, documents } = useStores();
+  const { t } = useTranslation();
+  const isPreloaded: boolean = !!collections.orderedData.length;
```
tommoor, line 23: I believe we should set this in state so it's not re-evaluated on every render

```suggestion
  const [isPreloaded] = React.useState(!!collections.orderedData.length);
```

tommoor (commented):
Some comments but this looks like a great start

iamsaumya:
Having hard time rewriting `search.js`. Stuck with an infinite loop. 😅

iamsaumya:
What should I do with functions decorated with `get' and `action`? I converted them to normal callback functions using `useCallback`. Is it the correct approach?

iamsaumya:
> Want to push up changes so I can take a look?

Giving it another shot with a fresh mind and reading Dan Abramov's `useEffect` blog again. I will push changes before calling it a day. 

> useCallback is only needed if they are being used as a prop into another React component, if not a regular function or even just const assignment within render is probably fine.

Understood.

iamsaumya:
Wow, all checks have passed.

iamsaumya:
Wow, all the checks have passed. 

I am not very confident with the code. Please scrutinize it to the fullest. The `useEffect` for `componentDidMount` was making an never-ending search requests. I just removed it and everything went smooth.

app/scenes/Search/Search.js
```diff
@@ -33,349 +28,352 @@ import DateFilter from "./components/DateFilter";
...
-      this.isLoading = true;
-      this.lastQuery = this.query;
-      this.lastParams = params;
+      setIsLoading(true);
+      lastQuery.current = query;
+      lastParams.current = params;
 
       try {
-        const results = await this.props.documents.search(this.query, params);
+        setResults(await documents.search(query, params));
```
tommoor, line 118: results does not need to be set in state, this might have been the source of your loop 😄 

It's only used to know if we received less than the pagination limit, aka there are more results to load.

app/scenes/Search/Search.js
```diff
@@ -33,349 +28,352 @@ import DateFilter from "./components/DateFilter";
...
+      locationSearchRef.current = location.search;
+    }
+  }, [handleQueryChange, location.search, locationSearchRef]);
+
+  React.useEffect(() => {
+    if (matchParamsTermRef.current !== match.params.term) {
+      handleTermChange();
+      matchParamsTermRef.current = match.params.term;
+    }
+  }, [handleTermChange, match.params.term]);
```
tommoor, line 188: These refs shouldn't be needed, with hooks you can just make eg `location.search` one of the deps and the hook will only be called when it changes – so there's no need to store or compare against the previous value like with `componentDidUpdate`

app/scenes/Search/Search.js
```diff
@@ -33,349 +28,352 @@ import DateFilter from "./components/DateFilter";
...
-    // Don't paginate if there aren't more results or we’re in the middle of fetching
-    if (!this.allowLoadMore || this.isLoading) return;
+  }, [params]);
 
-    // Fetch more results
-    await this.fetchResults();
-  };
+  const includeArchived = React.useCallback(() => {
+    return params.get("includeArchived") === "true";
+  }, [params]);
```
tommoor, line 93: These could all be simple `const` now, eg:

```
const includeArchived = params.get("includeArchived") === "true";
```

Same applies for the other params

app/scenes/Search/Search.js
iamsaumya: We are using `results` for `showEmpty` variable and to show all the documents.  We have to use `useState` or `useRef`(if we don't want to re-render when results are updates).

app/scenes/Search/Search.js
```diff
@@ -33,349 +28,352 @@ import DateFilter from "./components/DateFilter";
...
+    [location.search, history]
+  );
+
+  const setFirstDocumentRef = (ref: any) => {
+    firstDocument.current = ref;
   };
 
-  render() {
-    const { documents, notFound, location, t } = this.props;
-    const results = documents.searchResults(this.query);
```
tommoor, line 259: @iamsaumya ^^^

app/scenes/Search/Search.js
tommoor: The results comes from the documentstore, not the return of the API request directly (for better or worse). Just pinged you on the spot.

app/scenes/Search/Search.js
iamsaumya: Apologies, I should have checked this.

iamsaumya:
I think it's impossible to press `Ctrl + Ctrl + h` for table of contents, isn't it? I am unable to do it. Also, `h` kind of counterintuitive for a table of contents. Maybe we can change this to `Ctrl + c`?

tommoor:
`ctrl + c` is already copy, I believe I based on the shortcut on Google Docs.

iamsaumya:
Just `c`?

app/scenes/Search/Search.js
```diff
@@ -33,349 +28,352 @@ import DateFilter from "./components/DateFilter";
...
+    [location.pathname, location.search, history]
+  );
+
+  const handleNewDoc = React.useCallback(() => {
+    if (collectionId()) {
+      history.push(newDocumentUrl(collectionId));
+    }
+  }, [collectionId, history]);
+
+  // const isFiltered = React.useCallback(() => {
```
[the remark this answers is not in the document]
tommoor: If it's no longer used!

app/scenes/Document/components/Document.js
```diff
@@ -54,157 +50,181 @@ Are you sure you want to discard them?
...
-  @observable editor = React.createRef();
-  @observable isUploading: boolean = false;
-  @observable isSaving: boolean = false;
-  @observable isPublishing: boolean = false;
-  @observable isDirty: boolean = false;
-  @observable isEmpty: boolean = true;
-  @observable moveModalOpen: boolean = false;
-  @observable lastRevision: number = this.props.document.revision;
-  @observable title: string = this.props.document.title;
-  getEditorText: () => string = () => this.props.document.text;
```
iamsaumya, line 82: `getEditorText` was used in a cryptic way. I was confused about whether it is a function or a variable. So, I rewrote it as a state.

app/scenes/Document/components/Document.js
```diff
@@ -54,157 +50,181 @@ Are you sure you want to discard them?
...
 
-@observer
-class DocumentScene extends React.Component<Props> {
-  @observable editor = React.createRef();
-  @observable isUploading: boolean = false;
-  @observable isSaving: boolean = false;
-  @observable isPublishing: boolean = false;
-  @observable isDirty: boolean = false;
-  @observable isEmpty: boolean = true;
-  @observable moveModalOpen: boolean = false;
```
iamsaumya, line 79: Removed from the code because it was unused.

app/scenes/Document/components/Document.js
```diff
@@ -54,157 +50,181 @@ Are you sure you want to discard them?
...
       ui.showTableOfContents();
     }
-  }
+  };
+
+  useHotkeys("m", goToMove);
+  useHotkeys("e", goToEdit);
 
-  handleCloseMoveModal = () => (this.moveModalOpen = false);
-  handleOpenMoveModal = () => (this.moveModalOpen = true);
```
iamsaumya, line 191: No longer used functions.

iamsaumya:
This PR can be considered done. Do let me know if you want something to be changed.

app/components/DocumentListItem.js
```diff
@@ -69,6 +69,7 @@ function DocumentListItem(props: Props) {
         pathname: document.url,
         state: { title: document.titleWithDefault },
       }}
+      ref={ref}
```
tommoor, line 72: 👍

app/components/Sidebar/components/Collections.js
```diff
@@ -1,98 +1,80 @@
...
-    const { collections } = this.props;
+const Collections = ({ onCreateCollection }: Props) => {
+  const { collections, ui, policies, documents } = useStores();
+  const { t } = useTranslation();
+  const [isPreloaded] = React.useState(!!collections.orderedData.length);
+  const history = useHistory();
 
+  React.useEffect(() => {
     if (!collections.isFetching && !collections.isLoaded) {
       collections.fetchPage({ limit: 100 });
```
tommoor, line 28: At some point we'll have to account for more than 100 collections 😆

app/components/Sidebar/components/Collections.js
```diff
@@ -1,98 +1,80 @@
...
     if (!activeCollectionId) return;
 
-    const can = this.props.policies.abilities(activeCollectionId);
+    const can = policies.abilities(activeCollectionId);
     if (!can.update) return;
 
-    this.props.history.push(newDocumentUrl(activeCollectionId));
-  }
+    history.push(newDocumentUrl(activeCollectionId));
+  });
```
tommoor, line 40: I think we'll need to make `ui.activeCollectionId` a memo dependency here otherwise `useHotkeys` will memoize the first state and always create a new document in that single collection.

```suggestion
  }, [ui.activeCollectionId]);
```

app/scenes/Document/components/Document.js
tommoor: `getEditorText` is a function that returns the editor text as markdown, the reason it's a function is because converting the entire document state to markdown can be a relatively expensive operation so we only want to do it when neccessary, not every time the doc changes.

It might want to be `setGetEditorText`, lol.

app/scenes/Document/components/Document.js
```diff
@@ -221,242 +241,209 @@ class DocumentScene extends React.Component<Props> {
...
-      this.isPublishing = false;
+      setIsSaving(false);
+      setIsPublishing(false);
     }
   };
 
-  autosave = debounce(() => {
-    this.onSave({ done: false, autosave: true });
+  const autosave = debounce(async () => {
+    await onSave({ done: false, autosave: true });
```
tommoor, line 271: Why the `async/await` addition here?

iamsaumya: because `onSave` is `async`?

app/components/Sidebar/components/Collections.js
iamsaumya: Ikr!

iamsaumya:
Please don't merge, the document editor breaks while editing.

iamsaumya:
I acknowledge the effort Tom has put in helping me through this. There is a lot of bug in the PR, I believe I have to understand react and codebase a little better. 

I am closing this PR for now with the assurance that I will tackle this issue soon. Thanks!

PR #1929, opened 2021-02-28
chore: Migrate authentication to new tables

### LINKED ISSUE #1317: Migrate team authentication methods to new table
In order to support the ongoing work for additional pluggable authentication methods [here](https://github.com/outline/outline/pull/1183) – we need to move authentication details away from being stored on the `teams` table.

Some things to consider as this project starts up:

- A new table along the lines of `authentication_services`
  - `id`
  - `service` e.g. 'slack'
  - `service_id` e.g. 'T234234'
- Migrate existing data on teams to new table
- Migrate all queries to use new table
- Migrate users to add one-to-many relationship to new table

This will also unlock the ability for a single user to have multiple authentication methods.

tommoor:
Currently user authentication is stored on the user record. While this is efficient and simple it is not scalable beyond the two auth services we currently support – it's not wise to keep adding columns for each additional sign-in method. There are also a few other needs that are reaching the top of the roadmap:

- Store access and refresh tokens so that we can validate whether the user still has access automatically, in the case that they are revoked from the SSO.
- Add the ability for a single user to have multiple authentication methods into a team. This allows for a user to sign-in with Google, but potentially "link" a Slack account so they can still search private collections.
- Allow team admins to link multiple sign-in methods

This work provides a foundation for a more pluggable authentication system such as the one outlined in https://github.com/outline/outline/issues/1317.

closes #1317


server/api/utils.js
```diff
@@ -3,20 +3,108 @@ import subDays from "date-fns/sub_days";
...
+  const { token } = ctx.body;
 
   if (process.env.UTILS_SECRET !== token) {
     throw new AuthenticationError("Invalid secret token");
   }
 
+  return next();
+}
+
+router.post("utils.migrateAuthentication", authMiddleware, async (ctx) => {
```
tommoor, line 30: Note: This may move to a script rather than endpoint before merge.

server/auth/google.js
```diff
@@ -147,6 +115,7 @@ router.get("google.callback", auth({ required: false }), async (ctx) => {
       if (exists) {
         ctx.redirect(`${team.url}?notice=email-auth-required`);
       } else {
+        console.error(err);
```
thenanyu, line 118: send to sentry?

server/commands/teamCreator.js
```diff
@@ -6,6 +6,12 @@ import { generateAvatarUrl } from "../utils/avatars";
 
 const log = debug("server");
 
+type TeamCreatorResult = {|
+  team: Team,
+  authenticationProvider: AuthenticationProvider,
+  isNew: boolean,
```
thenanyu, line 12: `isNewTeam` is fine here, no need to be quite so general. That ways you can use `isNewTeam` and `isNew User` and the commands can be symmetrical and completely semantically coherent

server/commands/userCreator.js
```diff
@@ -124,7 +138,11 @@ export default async function userCreator({
       }
     );
     await transaction.commit();
-    return { user, isFirstSignin: true };
+    return {
+      user,
+      authentication: user.authentications[0],
+      isFirstSignin: true,
```
thenanyu, line 144: this wants to be `isNewUser`

tommoor:
This will be merged in ~12 hours.

Issue #862, opened 2019-01-14
Enforce single team when self-hosting


bmann:
I recently installed Outline on Heroku. I set `DEPLOYMENT=self` and `SUBDOMAINS_ENABLED=false`.

I initially setup Slack authentication, and then setup Google Authentication.

When I logged in with Google Authentication, Outline created a new team for me, linked to my login domain (`mypersonalgoogle.com`). Each user that logs in from a different Google domain (I have three in my `GOOGLE_ALLOWED_DOMAINS`) creates a new team.

My expectation was that with subdomains disabled, I was essentially running Outline in "single team" mode.

A workaround would be explaining how I can invite people to the same Outline team, even if they have a different Google domain. It makes sense to me for a Slack workspace that different Slack == different Outline team, but for Google, I mix and match which accounts I login to, and definitely want to share with users without having them to give them an account on my Google domain.

A feature request would be single team mode, where either no additional teams are created or are only created by an admin.

Thanks for any thoughts on this!

tommoor:
Hi @bmann 

Whilst Outline is an open source product (I wrote [about why this is the case](https://blog.tommoor.com/2019/01/01/you-should-open-source-your-product/) here)  there are really not many special accommodations made in the codebase for self hosting right now.

As the solo maintainer all of my efforts go into the core product experience and the hosted service at http://getoutline.com . I'm not against others taking on these changes however they are unlikely to be priority for myself, hope that makes sense!

tommoor:
> I still see a use case where you have people from multiple Google domains wanting to login to one Outline team (hosted version). Which doesn’t seem possible right now. Can you confirm?

Yes, this seems reasonable. There could be an env flag to force everyone onto the same team.

andermurias:
> Great, thanks for the response! I wanted to share my experience and wonder out loud what made sense.
> 
> I still see a use case where you have people from multiple Google domains wanting to login to one Outline team (hosted version). Which doesn’t seem possible right now. Can you confirm?
> 
> (and feel free to close this issue)

Hi!

I needed something similar, if you don't mind to change a few lines, you can achive if pretty easily.

My case was a bit different but a part of it could work for you, I wanted to allow @gmail users to login in the outline, and also some domains using gsuite to be allowed to login the tool and aloow all of them mto work with the same documents.

My work arround was something like this, 

I've created a variable in `.env`
```
TEAM_ID=teamname
```

[...]

PR #1934, opened 2021-03-04
feat: Move to passport for authentication

### LINKED ISSUE #1120: Changing towards passport.js for authentication
Lots of different OAuth providers are being requested, e.g. in #755, #962, or some other issues.

I'm in the process of working for #962 and was wondering about moving towards passport.js, which would make for a much easier provider addition procedure.

tommoor:
Based on #1929
Supersedes https://github.com/outline/outline/pull/1183
Closes #1120


ulentini:
Hi, this is a great feature, we really needed it. 
With very little effort and using a few ENV variable, self-hosted "single team" could be implemented (ie: all configured signin method would be always available and all users would join the same single team, like discussed on #862).
I'm already testing it locally and it works pretty fine, I think I'm gonna test it on a few of our outline installations.
Would you be interested in a PR about this?

tommoor:
> "single team" could be implemented (ie: all configured sign-in method would be always available

Yes, ideally this would be the default behavior to be honest, but such a change will have to wait until this is all landed.

> on a few of our outline installations.

You have multiple installations? 🤔

ulentini:
> > "single team" could be implemented (ie: all configured sign-in method would be always available
> 
> Yes, ideally this would be the default behavior to be honest, but such a change will have to wait until this is all landed.
> 
> > on a few of our outline installations.
> 
> You have multiple installations? 🤔
> You have multiple installations? 🤔
Yes 😄

server/commands/accountProvisioner.js
```diff
@@ -0,0 +1,105 @@
...
+      subdomain: teamParams.subdomain,
+      avatarUrl: teamParams.avatarUrl,
+      authenticationProvider,
+    });
+  } catch (err) {
+    throw new AuthenticationError(err.message);
+  }
+  invariant(team, "Team must exist");
+
+  const authP = team.authenticationProviders[0];
```
thenanyu, line 57: feels unsafe — this is an explicit guarantee but it violates demeter — better to use `_.find(match the provider id)`

server/commands/accountProvisioner.js
```diff
@@ -0,0 +1,105 @@
...
+          email: userParams.email,
+          teamId: team.id,
+        },
+      });
+
+      if (exists) {
+        throw new EmailAuthenticationRequiredError(
+          "Email authentication required",
+          team.url
+        );
```
thenanyu, line 97: this is slightly suspect but at worst it's doing pretty much nothing, maybe note it in a comment or something

jakoblorz:
Great work on this!

jdaviescoates:
This is awesome! @tommoor will you be releasing a new version that includes this in the near future? Many thanks for your work on this!

tommoor:
New releases are cut once/month, generally around the ~15th depending on what's in flight.

VityaSchel:
image

guys.... it's been more than a year... can someone tell me is this feature in beta or like how to enable it? me and some other people are still subscribed and waiting for this feature, what is the current status? was it implemented? is there something we have to change in docker.env?

VityaSchel:
image

wtf

PR #1935, opened 2021-03-04
PLAT-177: Initial separation of FE from BE

DanAtSpark:
* Renamed folders
* Removed server
* Configured gatsby build
* Cleared unused environment


tommoor:
Hey Dan,not sure what this one is about, but based on the changes and the org you're working for i will take this opportunity to remind you that the BSL license does not allow reselling of the Outline codebase in any form.

PR #1948, opened 2021-03-10
fix: Remove double separator in collection menu

### LINKED ISSUE #1947: bug: Remove double separator in collection menu
A separator is present if the user cannot delete a collection in collection menu. 

**To Reproduce**
1. Go to '...' of a collection that you cannot delete.

**Expected behavior**
Don't show the separator if the user cannot delete a collection in collection menu

**Screenshots**
Screenshot from 2021-03-11 02-41-07

iamsaumya:
fixes #1947

### DIFF
--- a/app/menus/CollectionMenu.js
+++ b/app/menus/CollectionMenu.js
@@ -160,7 +160,6 @@ function CollectionMenu({
             },
             {
               type: "separator",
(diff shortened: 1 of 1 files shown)


tommoor:
Looks like the problem is the double separator itself, no need for the `visible` check as useless separators are filtered by the context menu component.

Issue #1881, opened 2021-02-12
Local Authentication


iojanis:
I don't like the fact that you need a Google or Slack account. I'm sure you'll scare off a lot of people with that. Have a lot of interest in this app, but would never think of hosting something myself, but still have the users go through another service.

matbrgz:
#1183

routerino:
Honestly I think local authentication is actually a bad thing. It's hard to get right and has large reaching consequences if done wrong.

Once the oauth (or potentially SAML) authentication is set to be more generic, you can also self-host a dedicated authentication provider like Keycloak or Authelia, who are going to be able to devote more resources to doing it right.

arjunv:
#1177 

> It should be as simple as :
> 
> * Generate an admin user:password in the CLI while starting for the first time
> * When the admin logs into the web, have that password changed forcibly
> * Allow admin to setup whatever authentication methods he needs from the web console
> * Additionally, there should be a non smtp based adding of users too. ie the admin generates a user:pass pair which is given to a 3rd party in whatever secure channel he prefers to. On that user's first login, have that password forcibly changed so the admin cannot snoop in.

Voila, a fully self-hosted instance with no external dependencies like web-oauth or smtp.

slurdge:
Looking at passport.js and the current providers, it seems a local datastore would be needed to store/validate the local users. Unfortunately, passport doesn't seem to provide this (only verification).
So the effort would be:

- Create a schema for user credentials in the DB (with sequelize)
- A command line for create & delete an user with optional admin rights.

From this point on, the community could test this functionality. The UI for password reset/change can then be added at a later step.

>  Allow admin to setup whatever authentication methods he needs from the web console

This is already done with env variables if I'm not mistaken.

tommoor:
I don't see the CLI as being needed, the first ever sign-in on self-hosted automatically creates the team and becomes admin already. I see something like this as a bare minimum set of functionality, you can't have password auth without the ability to reset passwords...

## Backend
- Allow for password auth to be turned on or off via ENV in the same way as all other authentication integrations
- Add a new `password` column to `user_authentications`,
- Create a new `password` strategy inline with the existing three for password auth, it will need to handle password being passed to `accountProvisioner` and `userCreator` commands
- For user creator we'll need a reasonable default for a minimum password length
- Add `/auth/password.reset` endpoint for requesting a password reset with email
- Add rate limiting for password reset endpoint
- Add email for password reset with one-time use JWT
- Tests for all of the above

[...]

PR #1953, opened 2021-03-13
feat: Microsoft authentication

### LINKED ISSUE #755: Microsoft Teams Authentication

tommoor:
[>> Setup Guide <<](https://wiki.generaloutline.com/share/dfa77e56-d4d2-4b51-8ff8-84ea6608faa4)

If anyone wants to follow the guide and try this branch out on their own instance it would be much appreciated, I only have one Microsoft team to test with and it would be good to get some broader coverage before merging

closes #755


server/auth/azure.js
```diff
@@ -0,0 +1,88 @@
...
+
+const router = new Router();
+const providerName = "azure";
+const AZURE_CLIENT_ID = process.env.AZURE_CLIENT_ID;
+const AZURE_CLIENT_SECRET = process.env.AZURE_CLIENT_SECRET;
+
+const scopes = [];
+
+export const config = {
+  name: "Office 365",
```
tommoor, line 21: "Microsoft", "Azure", "Teams", "Office 365" – I'm tempted to just do "Microsoft", it seems to cover all the possible options.

server/auth/azure.js
```diff
@@ -0,0 +1,88 @@
...
+    },
+    async function (req, accessToken, refreshToken, params, _, done) {
+      try {
+        // see docs for what the fields in profile represent here:
+        // https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens
+        const profile = jwt.decode(params.id_token);
+
+        const domain = profile.email.split("@")[1];
+        const subdomain = domain.split(".")[0];
+        const teamName = capitalize(subdomain);
```
tommoor, line 44: Ideally we'd grab the organization name with an API request, haven't quite figured that out yet. The accessToken returned from oauth isn't working with Microsoft' graph API

ulentini: @tommoor the callback signature seems to be wrong.
I tried to implement the same strategy, but as you can see on the strategy page (http://www.passportjs.org/packages/passport-azure-oauth2/), there's no `req` object in the callback, the first parameter is the accessToken. So I guess you're trying to query the graph API with the refresh token

tommoor: The startegy callbacks are standardized – `passReqToCallback: true` in the strategy options makes the first parameter always the request.

ulentini: Yes, you're right. The problem actually is that that package use the old Azure AD Graph API, instead of the new Microsoft Graph API

tommoor: Could you point to that? I didn't think it was using either graph api.

I'll probably fork the whole thing, it's also point to an old login url.

ulentini: Sure, here:
https://github.com/AndrewKeig/passport-azure-oauth2/blob/master/lib/index.js#L9
The base URL for oauth requests is `login.windows.net` instead of `graph.microsoft.com`.

server/auth/azure.js
```diff
@@ -0,0 +1,88 @@
...
+      store: new StateStore(),
+      scope: scopes,
+    },
+    async function (req, accessToken, refreshToken, params, _, done) {
+      try {
+        // see docs for what the fields in profile represent here:
+        // https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens
+        const profile = jwt.decode(params.id_token);
+
+        const domain = profile.email.split("@")[1];
```
ulentini, line 42: Keep in mind that `profile.email` is not always present. In my company profile I can only see `profile.upn`, which contains my email address. In my personal account (which is a mix of password and github authentication) I have a mess.
I guess a request to `https://graph.microsoft.com/v1.0/me` would be better, since all data seem to be normalized, there

tommoor:
Waiting on Microsoft Partner approval before moving this forward

bglidwell:
Will be able to purchase subscription once this is merged.

bglidwell (approved):
Successfully authenticated for me

bglidwell:
Any idea when this might be merged in?

ulentini:
Hi, I have 2 issues with this branch (but maybe it's just me)
1. If I try to setup the azure oauth with one of my company azure apps, I cannot login since I get the "Invalid audience" error. This could depend on the different endpoints used to perform the initial authentication (from the passport strategy) VS the following requests to MS Graph api
2. Even after I fixed the first problem, only the first login after the server starts has been successful. After that it fails with a "state mismatch" error. I was able to fix this changing [this line](https://github.com/auth0/passport-azure-ad-oauth2/blob/master/lib/passport-azure-ad-oauth2/strategy.js#L116) in the passport azure package. It's like there's some sort of closure on the `options` object and once the first user generate the state cookie, it gets used for all following users. So, basically, changing that return statement to `return { ...options }` solved the issue for me.

tommoor:
RE 2, weird but it could be the case – other packages do produce a new object, eg:

https://github.com/jaredhanson/passport-google-oauth2/blob/2399ef47f00b47ac587f056f2f100a2c4db81928/lib/strategy.js#L134

I'll fork the dependency it's clearly not maintained and a poor thing to rely on for such a key aspect of the application

tommoor:
@ulentini moved to a fork, and included your fix for the state and the updated authentication domain. I checked that the state is not retained between sign-ins, so looking good now.

tommoor:
@bglidwell this is now deployed to the cloud hosted version of Outline

fooness:
That’s a good step forward!

When might we get authentication methods which are not externalized to Google, Microsoft, or Slack?

Local accounts with Postgres? LDAP? SAML?

https://github.com/outline/outline/pull/1183

https://github.com/outline/outline/issues/1881

tommoor:
LDAP and SAML are included in the enterprise edition and will not be included in the open source codebase – you (or anyone in the future that reads this) can contact hello@getoutline.com to arrange an enterprise license key.

I have detailed what is required for local accounts in the ticket you linked, if the community wants to build that feature now that the authentication providers are pluggable they are welcome to do so!

Issue #1945, opened 2021-03-10
Add authentication provider management


tommoor:
A new screen under team settings "Authentication" lists the providers connected to the team, it should:

- Allow connecting providers from those available.
- Allow disabling providers that were previously connected (`enabled` column already exists and defaulted to true for this). Disabled providers should prevent further logins and show an error message on the login screen that explains to the user.

almereyda:
- #2494

PR #1954, opened 2021-03-14
feat: Enforce single team when self-hosted

### LINKED ISSUE #862: Enforce single team when self-hosting

tommoor:
Behavior is as follows:

- On first install signup is open (apart from `ALLOWED_DOMAINS` restriction)
- After first team is registered no other teams can be created
- If attempting to signin with another google hd that's within `ALLOWED_DOMAINS` it will be assigned to the existing team, otherwise an error will be returned
- The addition of https://github.com/outline/outline/issues/1945 will enable adding multiple authentication methods in the future.

closes #862

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -71,7 +71,7 @@ DEBUG=cache,presenters,events,emails,mailer,utils,multiplayer,server,services
 
 # Comma separated list of domains to be allowed to signin to the wiki. If not
 # set, all domains are allowed by default when using Google OAuth to signin
-GOOGLE_ALLOWED_DOMAINS=
+ALLOWED_DOMAINS=
 
 # For a complete Slack integration with search and posting to channels the 
 # following configs are also needed, some more details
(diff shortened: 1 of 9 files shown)


benmort:
NICE ONE!

Issue #1158, opened 2020-01-12
Serialize "class level" policies onto the team policy object


thenanyu:
Some policies, especially admin ones, are effectively duplicated across the primary domain and the team domain. For example — whether to allow a user to invite new members to the team.

We need to do this in order to expose the policy on the frontend, because every policy must be tied to an entity object. Policies for domains (like "Users") rather than specific records don't get serialized in the normal process.

It would be more DRY to keep the policy definitions in their domain and just attach them to the Team entity.

tommoor:
Totally agreed, if we can expose these on the team in an automated way that would be great. We could also consider having the team policy simply call the underling users policy, for example. so at least the logic isn't duplicated.

iamsaumya:
From what I understand we don't want to use `user.isAdmin` on the frontend rather something very specific like `can.canInvite` or `can.createDocument`.  However, I guess I am missing some building blocks here. 

What is a primary domain and a team domain? 

What is the difference between these two policies? when to use the latter policy? 
https://github.com/outline/outline/blob/ad8c08497c6d477dad878fefdf4d73ca375556ca/server/policies/team.js#L20-L23
https://github.com/outline/outline/blob/ad8c08497c6d477dad878fefdf4d73ca375556ca/server/policies/user.js#L15-L17

An example of what is wrong and what needs to be done will help me I guess.

tommoor:
> What is the difference between these two policies? when to use the latter policy?

The latter policy is just wrong and not currently used (thankfully). Let me take a look at this tomorrow, I should get time to whip up a PR that we can use for the read-only work.

PR #1955, opened 2021-03-14
feat: Read-only users

### LINKED ISSUE #1873: Read-only users
We'd like to add the ability to have read-only users, currently the option exists on a per-collection basis however there is a need to also optionally have this ability at the user-level.

### Requirements

- A new database column `isViewer`
- A new control on `settings -> people -> user menu` to change user to a viewer or back to a regular member.
- Must have a confirmation dialog for this operation
- Viewers do not have the ability to modify anything except their own profile

iamsaumya:
Closes #1873


iamsaumya:
I removed a validation in this commit [`c13bc01` (#1955)] (https://github.com/outline/outline/pull/1955/commits/c13bc014db8d72f82546e3e9f1e4914c5b5afc1a) which stops someone from demoting themselves. I find this more intuitive and this is already in place when suspending a user. 

We can remove the [query](https://github.com/iamsaumya/outline/blob/3d71e22e0f3de66169deec5b9813250e6fcd89e4/server/models/Team.js#L198) to look if at least an admin is present in the DB if we have this validation. What do you think?

iamsaumya:
Hey Tom, Could you review this PR as well, please? I will make all the necessary changes quickly so that we can ship the feature as well.

iamsaumya:
@tommoor here 🖐🏻 this PR needs your review as well 😅

app/components/DocumentListItem.js
```diff
@@ -111,21 +114,24 @@ function DocumentListItem(props: Props) {
...
-            >
-              {t("New doc")}
-            </Button>
-            &nbsp;
-          </>
-        )}
+        {document.isTemplate &&
+          !document.isArchived &&
+          !document.isDeleted &&
+          !user.isViewer && (
```
tommoor, line 120: We should use `can.update` from the related collection policy here, the `document.collectionId`

app/components/Sidebar/Main.js
```diff
@@ -123,23 +123,25 @@ function MainSidebar() {
...
-                }
-                active={
-                  documents.active
-                    ? !documents.active.publishedAt &&
-                      !documents.active.isDeleted &&
-                      !documents.active.isTemplate
-                    : undefined
-                }
-              />
+              {!user.isViewer && (
```
tommoor, line 126: In order to use a policy here we might have to tackle this one first: https://github.com/outline/outline/issues/1158

So that we can have `can.createDocument` or something similar on the team policy 🤔

app/scenes/Settings/People.js
```diff
@@ -113,6 +115,11 @@ class People extends React.Component<Props> {
               {t("Suspended")} <Bubble count={counts.suspended} />
             </Tab>
           )}
+          {can.update && (
+            <Tab to="/settings/people/viewers" exact>
+              {t("Viewers")} <Bubble count={counts.viewers} />
+            </Tab>
+          )}
```
tommoor, line 122:
```suggestion
          <Tab to="/settings/people/viewers" exact>
            {t("Viewers")} <Bubble count={counts.viewers} />
          </Tab>
```

tommoor (commented):
I started looking through and my feedback was going to be very repetitive. The core here is totally right, new column, new controls etc 👍 

The key thing is that there are lots of spots where we're checking `user.isViewer` directly in the client that we need to change to using a policy like `can.createDocument` instead. I added a comment about #1158, and it's feeling like it might be a blocker to get the policies needed here.

server/models/Team.js
```diff
@@ -195,10 +195,10 @@ Team.prototype.provisionFirstCollection = async function (userId) {
 };
 
 Team.prototype.addAdmin = async function (user: User) {
-  return user.update({ isAdmin: true });
+  return user.update({ isAdmin: true, isViewer: false });
 };
 
-Team.prototype.removeAdmin = async function (user: User) {
+Team.prototype.demoteUser = async function (user: User, to: string) {
```
tommoor, line 201: Argh, I hate these methods – why are they on team? Makes no sense – but not your fault 😉

server/policies/share.js
```diff
@@ -5,14 +5,13 @@ import policy from "./policy";
...
 const { allow } = policy;
 
-allow(
-  User,
-  ["read", "update"],
-  Share,
-  (user, share) => user.teamId === share.teamId
-);
+allow(User, ["read", "update"], Share, (user, share) => {
+  if (user.isViewer) return false;
```
tommoor, line 9: I think viewers can read share information, that seems expected or they won't be able to see if a doc is shared publicly and access that link

tommoor (commented):
Most excellent work, some minor feedback but nothing huge at all.

I do think I'll want to iterate on the people management interface some after this but it makes sense to get the functionality landed and then refine that UI.

app/components/Sidebar/components/Collections.js
```diff
@@ -14,19 +14,15 @@ import DropCursor from "./DropCursor";
...
   onCreateCollection: () => void,
 };
 
 function Collections({ onCreateCollection }: Props) {
-  const { ui, policies, documents, collections } = useStores();
+  const { ui, policies, documents, collections, auth } = useStores();
   const isPreloaded: boolean = !!collections.orderedData.length;
   const { t } = useTranslation();
   const orderedCollections = collections.orderedData;
+  const user = auth.user;
```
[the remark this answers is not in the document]
iamsaumya: I did find about it after making the commits, was looking for your feedback. Consider it done.

app/menus/UserMenu.js
```diff
@@ -98,9 +118,16 @@ function UserMenu({ user }: Props) {
...
                 userName: user.name,
               }),
-              onClick: handleDemote,
+              onClick: handleMember,
               visible: can.demote,
             },
+            {
+              title: t("Make {{ userName }} a viewer...", {
+                userName: user.name,
+              }),
```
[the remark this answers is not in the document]
iamsaumya: Let's fix both?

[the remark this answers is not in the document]
tommoor: Let's fix 'er up

server/api/users.js
```diff
@@ -131,14 +131,22 @@ router.post("users.promote", auth(), async (ctx) => {
...
   const teamId = ctx.state.user.teamId;
+  let { to } = ctx.body;
+
   const actor = ctx.state.user;
   ctx.assertPresent(userId, "id is required");
 
+  const possibleTos = ["member", "viewer"];
+  if (possibleTos.findIndex((possibleTo) => possibleTo === to) === -1) {
+    to = "member";
+  }
```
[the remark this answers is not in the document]
iamsaumya: Yes! this is good for now.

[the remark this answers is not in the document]
iamsaumya: @tommoor The fallback is `viewer` here. Previously `users.demote` would demote to `member`, Do you suggest changing the ternary to make `member` a fallback? 

We have a test failing because of this, I am confused between changing the testcase or the code 😅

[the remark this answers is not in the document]
tommoor: I'm not against making the default "member" if that helps reduce test churn

iamsaumya:
The active user count doesn't change when a user is suspended. Is this a bug?

tommoor:
On the UI or the result from the API? Either would be a bug though

iamsaumya:
On the UI, we are not updating the active count.

iamsaumya:
Your advice to reload the counts from the server is on point. 

I was wondering if it's okay to add the counts in the response of the operations. It will save us an API call after each of these operations.

iamsaumya:
I did this and the app feels a bit slow after removing the optimistic update.

server/models/Team.js
iamsaumya: Do you feel better now?

server/policies/share.js
iamsaumya: Done

server/models/User.js
```diff
@@ -292,10 +298,45 @@ User.getCounts = async function (teamId: string) {
...
     active: parseInt(counts.activeCount),
     admins: parseInt(counts.adminCount),
+    viewers: parseInt(counts.viewerCount),
     all: parseInt(counts.count),
     invited: parseInt(counts.invitedCount),
     suspended: parseInt(counts.suspendedCount),
   };
 };
 
+User.prototype.demote = async function (teamId: string, to: string) {
```
tommoor, line 308: I think we need tests for this method now, currently it's not going to work as expected as it assumes you're trying to demote an `admin` to a lower role, but now you could be demoting a member to a viewer – it would stop you doing that if there was only one admin on the team 😉

server/models/User.js
```diff
@@ -292,10 +298,45 @@ User.getCounts = async function (teamId: string) {
...
+      return this.update({ isAdmin: false, isViewer: false });
+    } else if (to === "viewer") {
+      return this.update({ isAdmin: false, isViewer: true });
+    }
+  } else {
+    throw new ValidationError("At least one admin is required");
+  }
+};
+
+User.prototype.makeAdmin = async function () {
```
tommoor, line 331: Should we convert this to `promote` ? The asymmetry is awkward as I'm not sure how we would make a `viewer` -> `member` with this API?

server/models/User.js
iamsaumya: Actually, it working! We are validating if there exists at least one admin in the team while not considering the rank of the current user.

server/models/User.js
iamsaumya: The `users.promote` endpoint is only to used to make the user an `admin`. `users.demote` is used to make a user `viewer` -> `member` or `member` to `viewer`. 

I know this is awkward as well. What do you think about having a single endpoint? That will change the rank of the current user based on the `to` parameter in the `req` body. 

`to = ['admin','member','viewer']`

server/models/User.js
iamsaumya: We can remove `at least one admin is required` validation if in the policies we can change to not let a user act on himself. 

This is confusing because in the app I cannot take any action on ourselves but I can through the API.

server/models/User.js
tommoor: It's tempting… but it's probably going to make life easier to keep the endpoints aligned with the policies, otherwise you need to calculate if the `to` parameter is promoting or demoting. Plus it enables a roll out without a breaking API change.

iamsaumya: I was conscious of the breaking API change and that's why I didn't put forward the idea before.

iamsaumya: In conclusion, I'm renaming the function to `promote`.

tommoor:
I am amazed but pleased that this has no conflicts with the merge of https://github.com/outline/outline/pull/1991

app/stores/UsersStore.js
```diff
@@ -61,13 +67,22 @@ export default class UsersStore extends BaseStore<User> {
...
   @action
-  demote = (user: User) => {
-    this.counts.admins -= 1;
-    return this.actionOnUser("demote", user);
+  demote = (user: User, to: string) => {
+    if (user.isAdmin) {
+      this.counts.admins -= 1;
+    }
+
+    if (to === "member" && user.isViewer) {
```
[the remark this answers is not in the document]
iamsaumya: Reloading the count or recalculating the count after change feels a slow visually wrt current approach. What do you suggest here?

server/models/User.js
iamsaumya: Resolving the conversation because the function works as expected.

iamsaumya:
> I am amazed but pleased that this has no conflicts with the merge of #1991

because the codebase is really well architectured.

app/stores/UsersStore.js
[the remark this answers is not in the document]
iamsaumya: @tommoor I believe this is the only thing holding us back here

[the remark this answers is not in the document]
tommoor: Okay, I think we can deal with it on the client with a little refactoring. Especially if promote and demote here are updated to work like the server side now does. In two clear steps of logic:

- You just need to check what type of user they currently are and -1
- Check what user they're going to be and +1
- We could also cache `this.counts` before and reset it back if actionOnUser throws

app/stores/UsersStore.js
```diff
@@ -66,35 +68,46 @@ export default class UsersStore extends BaseStore<User> {
 
   @action
   promote = (user: User) => {
-    this.counts.admins += 1;
-    if (user.isViewer) this.counts.viewers -= 1;
-    return this.actionOnUser("promote", user);
+    try {
+      this.updateCounts(Rank.ADMIN, getUserRank(user));
+      this.actionOnUser("promote", user);
```
tommoor, line 73: Applies to all of these updates – need to `await` `this.actionOnUser` otherwise this `catch` will never be applicable.

app/utils/getUserRank.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import User from "models/User";
+import Rank from "../../shared/utils/rankEnum";
+
+export default function getUserRank(user: User) {
```
tommoor, line 5: This could be a `computed` accessor on the user model?

shared/utils/rankEnum.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+export default Object.freeze({
```
tommoor, line 2: I'm not a fan of the `enum`, I think it's unnecessary when we're using types

app/stores/UsersStore.js
iamsaumya: oh my gosh! how did I miss this!

app/utils/getUserRank.js
iamsaumya: Logical 👍🏻

shared/utils/rankEnum.js
iamsaumya: will use plain object 👍🏻

shared/utils/rank.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+export type Rank = "Admin" | "Viewer" | "Member";
```
tommoor, line 2: Can move this inside the `types.js` file now

app/components/Sidebar/Main.js
```diff
@@ -123,23 +123,25 @@ function MainSidebar() {
...
-                }
-                active={
-                  documents.active
-                    ? !documents.active.publishedAt &&
-                      !documents.active.isDeleted &&
-                      !documents.active.isTemplate
-                    : undefined
-                }
-              />
+              {can.createDocument && (
```
tommoor, line 126: Let's also hide the "Templates" item if you can't create a document, after testing it out locally it felt odd to see that, a bit like being "behind the scenes".

tommoor (commented):
This is great, just these two small final tweaks and we'll get it merged in time for next release.

shared/utils/rank.js
iamsaumya: It is also being used on the front end. Should I create `shared/types.js`? otherwise we will have to keep it in `app/types/index.js` and `server/types.js`

tommoor: >  Should I create `shared/types.js? `

Yes let's do that, I'm amazed it hasn't been needed yet.

tommoor (approved):

PR #1959, opened 2021-03-15
feat: use flexbox to center the floating title

thenanyu:
Very wide viewport; the title is visually centered in the content area
image

When the viewport is made smaller, we use overflow ellipses to condense the title
image

Mobile breakpoint now displays the title
image

Saving indicator does not affect visual center...
image

...and it causes sensible overflow behavior
image


tommoor:
Its so beautiful

tommoor (approved):

PR #1960, opened 2021-03-15
fix: correct GoogleSheets component class name

msynk:
The GoogleSheets component has an incorrect name which is "GoogleSlides". Its probably result of a copy/paste mistake.


tommoor (approved):
Thanks!

PR #1961, opened 2021-03-16
fix: touch up UI for compact breadcrumbs

thenanyu:
image
image

1. adds ellipses to last segment of list item metadata
2. makes the slashes more visible, especially in dark mode and fixes vertical alignment


tommoor:
Ah, thanks – the slash color was bugging me in dark mode

tommoor (approved):

PR #1966, opened 2021-03-18
feat: Add "Import document" menu item to document menu

### LINKED ISSUE #1940: Add "Import document" menu item to document menu
Currently the only way to import inside another document is to drag a file onto it in the sidebar, this has confused some folks who thought it wasn't possible. We should add an "Import document…" option in the document menu when the user has permissions to create a child document.

iamsaumya:
closes #1940

### DIFF
--- a/app/menus/DocumentMenu.js
+++ b/app/menus/DocumentMenu.js
@@ -4,6 +4,7 @@ import * as React from "react";
 import { useTranslation } from "react-i18next";
 import { useHistory } from "react-router-dom";
 import { useMenuState, MenuButton } from "reakit/Menu";
+import { VisuallyHidden } from "reakit/VisuallyHidden";
 import styled from "styled-components";
 import Document from "models/Document";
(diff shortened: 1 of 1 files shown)


tommoor (approved):
You're taking all the issues before I can get to them 😆

app/menus/DocumentMenu.js
```diff
@@ -243,6 +309,11 @@ function DocumentMenu({
               }),
               visible: !!can.createChildDocument,
             },
+            {
+              title: t("Import document"),
```
tommoor, line 313: I feel like I want to clarify this menu item but can't quite figure out the wording, perhaps

"Import nested document"

cc @thenanyu

iamsaumya:
> You're taking all the issues before I can get to them 😆

Tackling the small issues so you can focus on the bigger ones 😄

app/menus/DocumentMenu.js
tommoor: Leaving it how it is for now

PR #1970, opened 2021-03-21
chore: Serialize domain policies on team

### LINKED ISSUE #1158: Serialize "class level" policies onto the team policy object

tommoor:
Exposing model creation policies on the team allows us to gate functionality on the frontend in a scalable way. This is particularly important as we introduce read-only roles.

closes #1158

### DIFF
--- a/server/policies/team.js
+++ b/server/policies/team.js
@@ -1,5 +1,4 @@
 // @flow
-import { AdminRequiredError } from "../errors";
 import { Team, User } from "../models";
 import policy from "./policy";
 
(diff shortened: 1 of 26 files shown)


tommoor:
cc @iamsaumya

server/policies/team.js
```diff
@@ -12,24 +11,7 @@ allow(User, "share", Team, (user, team) => {
...
-  if (user.isAdmin) return true;
-  throw new AdminRequiredError();
-});
-
-allow(User, ["update", "export"], Team, (user, team) => {
+allow(User, ["update", "export", "manage"], Team, (user, team) => {
   if (!team || user.teamId !== team.id) return false;
-  if (user.isAdmin) return true;
-  throw new AdminRequiredError();
+  return user.isAdmin;
```
iamsaumya, line 16: Is there any specific reason we don't throw the `AdminRequiredError` here?

tommoor: Yea but it's hard to explain, "manage" is a special case in the authorization library.

If we throw here it the other team policies are not considered 😔

iamsaumya: I think It's indeed hard to explain 😅

PR #1971, opened 2021-03-21
feat: github oauth strategy

razor950:
These two 
GITHUB_KEY = GITHUB_CLIENT_ID
GITHUB_SECRET = GITHUB_CLIENT_SECRET 

This doesn't let you choose a organization but it does create a team based on the github oauth application you create. It is also following the newer passport integration as close as possible with the working github v3 oauth strategy package I could find.


razor950:
I am not sure what domain and subdomain should be for github strategy, the data you get from github doesn't relay any of this information as far as I can tell.

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+
+  // start the oauth process and redirect user to Github
+  // signin callback from Github
+  router.get(
+    "github.callback",
+    auth({ required: false }),
+    passportMiddleware(providerName)
+  );
+
+  router.get("github.commands", auth({ required: false }), async (ctx) => {
```
tommoor, line 84: This method and `github.post` don't seem to make sense, they're only applicable to the Slack integration

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+      passReqToCallback: true,
+      store: new StateStore(),
+      scope: scopes,
+    },
+    async function (req, accessToken, refreshToken, profile, done) {
+      try {
+        const result = await accountProvisioner({
+          ip: req.ip,
+          team: {
+            name: profile.displayName,
```
tommoor, line 40: `subdomain` is required, it should be set to a lowercase string that would make sense as a subdomain – I'd use the team username.

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+            avatarUrl: profile._json.avatar_url,
+          },
+          user: {
+            name: profile.username,
+            email: profile.emails[0].value,
```
tommoor, line 45: Is the first email always the default email? Or does the object include a key to identify which email in the array is default…

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+import passportMiddleware from "../middlewares/passport";
+import { Authentication, Collection, Integration, Team } from "../models";
+import { StateStore } from "../utils/passport";
+
+const router = new Router();
+const providerName = "github";
+const GITHUB_CLIENT_ID = process.env.GITHUB_KEY;
+const GITHUB_CLIENT_SECRET = process.env.GITHUB_SECRET;
+
+const scopes = ["repo", "read:org", "user"];
```
tommoor, line 18: `repo` scope shouldn't be needed?

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+import env from "../env";
+import * as Github from "../github";
+import auth from "../middlewares/authentication";
+import passportMiddleware from "../middlewares/passport";
+import { Authentication, Collection, Integration, Team } from "../models";
+import { StateStore } from "../utils/passport";
+
+const router = new Router();
+const providerName = "github";
+const GITHUB_CLIENT_ID = process.env.GITHUB_KEY;
```
tommoor, line 15: Let's have the `env` variables match (`GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET`), it looks like you copied the Slack integration here but it's not a great example because those are so old and we've had to maintain backwards compatibility.

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+const GITHUB_CLIENT_SECRET = process.env.GITHUB_SECRET;
+
+const scopes = ["repo", "read:org", "user"];
+
+export const config = {
+  name: "Github",
```
tommoor, line 21:
```suggestion
  name: "GitHub",
```

server/auth/github.js
```diff
@@ -0,0 +1,199 @@
...
+        });
+        return done(null, result.user, result);
+      } catch (err) {
+        return done(err, null);
+      }
+    }
+  );
+
+  // For some reason the author made the strategy name capatilised, I don't know
+  // why but we need everything lowercase so we just monkey-patch it here.
```
tommoor, line 67:
```suggestion
```

This comment doesn't make sense for GitHub integration

server/github.js
```diff
@@ -0,0 +1,55 @@
+// @flow
```
tommoor, line 1: This file should not be needed at all

tommoor (changes requested):
> This doesn't let you choose a organization

We'll need to figure out how to enable that before this is mergable, might be a config on GitHub's side?

server/auth/github.js
razor950: I only have one email on the account, I can try to add more emails and see what I find.

server/auth/github.js
razor950: I will look into this as well.

PR #2005, opened 2021-04-05
feat: Drop to import onto collection scene

### LINKED ISSUE #1569: Allow dropping files onto collection scene
Currently you can only drop files into the sidebar, this is a little finicky – dropping onto a collection itself is easier and gives the opportunity to display a state that shows which file types are supported.

tommoor:
- Pull import file logic into reusable hook (we may be able to use this in the menus too?)
- Add a Dropzone around the entire collection view
- Add some additional error handling
- Convert related components to functional to use hooks

closes #1569


app/components/Mask.js
```diff
@@ -23,7 +23,7 @@ class Mask extends React.Component<Props> {
   }
 
   render() {
-    return <Redacted width={this.width} />;
+    return <Redacted width={this.width} height={this.props.height} />;
```
tommoor, line 26: Unrelated bug I spotted, the `height` prop wasn't plumbed through

app/components/Scene.js
```diff
@@ -38,7 +40,11 @@ function Scene({
         actions={actions}
         breadcrumb={breadcrumb}
       />
-      <CenteredContent withStickyHeader>{children}</CenteredContent>
+      {centered !== false ? (
```
tommoor, line 43: Allows opting out of the centering when we need content that spans width

app/components/Sidebar/components/DropToImport.js
```diff
@@ -0,0 +1,83 @@
...
+
+type Props = {|
+  children: React.Node,
+  collectionId: string,
+  documentId?: string,
+  disabled: boolean,
+  staticContext: Object,
+|};
+
+function DropToImport({ disabled, children, collectionId, documentId }: Props) {
```
tommoor, line 19: 🍖  Moved core of logic to reusable hook  `useImportDocument`, turned to function and moved to `Sidebar` as it's only used there and has sidebar-specific styles

app/hooks/useImportDocument.js
```diff
@@ -0,0 +1,69 @@
+// @flow
```
tommoor, line 1: logic from `DropToImport`

app/scenes/Collection.js
```diff
@@ -34,315 +29,371 @@ import Subheading from "components/Subheading";
...
+                />
               )}
-            </Heading>
-            <CollectionDescription collection={collection} />
+            />
+          </Action>
+        </>
+      }
+    >
+      <Dropzone
```
tommoor, line 169: 🍖  Wrap entire collection scene in a dropzone

PR #2051, opened 2021-04-20
feat: New keyboard shortcuts guide

tommoor:
### Before
Full screen modal, very inefficient use of space.

image

### After
Partial screen dialog, includes search(!), efficient use of space.

image


app/scenes/KeyboardShortcuts.js
```diff
@@ -3,185 +3,416 @@ import * as React from "react";
...
-          <Key>{metaDisplay}</Key> + <Key>z</Key>
-        </Keys>
-        <Label>{t("Undo")}</Label>
-        <Keys>
-          <Key>{metaDisplay}</Key> + <Key>Shift</Key> + <Key>z</Key>
-        </Keys>
-        <Label>{t("Redo")}</Label>
-      </List>
+  const handleChange = React.useCallback((event) => {
+    setSearchTerm(event.target.value.toLowerCase());
```
tommoor, line 350: Neat thing here is that the filtering works on the translated strings

PR #2075, opened 2021-04-26
feat: Nested document sharing

### LINKED ISSUE #1141: Allow sharing of entire document tree
**Allow for sharing of a document plus all child documents in a share link**
- Add an option in the share dialog to support this
- Add breadcrumbs to the shared view when viewing a nested document
- Display "nested documents" footer when sharing a document tree

tommoor:
- [x] Migration to store nested sharing preference
- [x] Update UI to allow toggling sharing nested docs
- [x] Update API to allow `shareId` to load nested docs
- [x] Frontend routing
- [x] Tests for `documents.info`
- [x] Return nested documents tree
- [x] Re-writing of links to point to nested share urls
- [x] Loading of nested documents list – not sure how yet, we need to expose just the sub-document tree as usually this information is loaded from the collection when authenticated
- [x] Hide/disable nested sharing toggle when not publicly shared
- [x] Add breadcrumbs to header

image

image

closes #1141


app/scenes/Document/components/PublicBreadcrumb.js
```diff
@@ -0,0 +1,84 @@
...
+            ...item,
+            url: `/share/${shareId}${item.url}`,
+          };
+        }),
+    [sharedTree, shareId, documentId]
+  );
+  const isNestedDocument = path.length > 2;
+  const firstPath = path[0];
+  const lastPath = path.length ? path[path.length - 1] : undefined;
+  const menuPath = isNestedDocument ? path.slice(1, -1) : [];
```
thenanyu, line 56: this area could use some commenting. 2 seems like a bit of a magic number that corresponds to an expected structure

app/scenes/Document/components/PublicBreadcrumb.js
```diff
@@ -0,0 +1,84 @@
...
+        .slice(0, -1)
+        .map((item) => {
+          return {
+            ...item,
+            url: `/share/${shareId}${item.url}`,
+          };
+        }),
+    [sharedTree, shareId, documentId]
+  );
+  const isNestedDocument = path.length > 2;
```
thenanyu, line 53: I think this naming is confusing me; it's a nested document as long as firstPath and lastPath are not the same right? This is like... isNestedMoreThanOnce?

I think it might be too hard to explain in a variable name, might be more self-documenting to simply have a case for `path.length === 2` and `path.length > 2` in the return statement.

app/scenes/Document/components/PublicBreadcrumb.js
thenanyu, line 56: this brings up a possibility for the case statement to be more literal `menuPath.length && ...`

app/scenes/Document/components/PublicBreadcrumb.js
tommoor: I went ahead and completely refactored the Breadcrumb component 💪

tommoor:
@thenanyu any opinion on if we need to expose the information that a parent document is shared in child documents before deploying this?

thenanyu:
@tommoor I think a little notice in the "publish to internet" section (maybe inline with the other text) that says "This document is also published via a [parent doc]" with a link to the parent would be nice and close the loop for a while

tommoor:
Well, that "little notice" took a lot more work than expected

PR #2097, opened 2021-04-29
fix: Speed up stripping of markdown

marksteve:
We were encountering huge CPU spikes that would cause our outline server to stall for an hour when our wiki users searched for certain keywords. I identified the cause of the high CPU usage after some digging. It was when the search endpoint calls `removeMarkdown()` to render search results context. 

My fix was to replace the package used with remark and the [strip-markdown](http://remarkjs.github.io/strip-markdown/) plugin. Do note that the plugin doesn't have the option to disable stripping of HTML and has some [quirks](https://github.com/remarkjs/strip-markdown/issues/19). Not really sure what HTML is being whitelisted here.


tommoor:
This is a great find, any idea if it's certain keywords and if so which?

marksteve:
Ooh. I thought it was just because of big documents. Updating your fork would be a better fix!

tommoor:
I'll pull in the fix from the other repo that hasn't been merged 🙄  – you're right I think less churn would be good here and being able to retain the `stripHTML` option. Regardless just finding this is huge.

marksteve:
And I guess the HTML that needs to be whitelisted are for emphasizing the matching terms?

tommoor:
That's right – pg returns html tags for that, lol

marksteve:
Got it! Closing this then. Thank you!

tommoor:
By the way, I suppose this was also the issue you were seeing with timeout's when searching from Slack on the cloud version.

marksteve:
Yep! Most probably

PR #2098, opened 2021-04-30
feat: Signup query params tracking

tommoor:
In order to know where new teams are coming from we want to add minimal tracking so that we can quantify the value of different traffic sources.

This doesn't have any use that I can think of for self-hosted installations as it only applies to team creation, which will only happen once for these folks.

### DIFF
--- a/server/utils/authentication.js
+++ b/server/utils/authentication.js
@@ -1,6 +1,9 @@
 // @flow
+import querystring from "querystring";
+import * as Sentry from "@sentry/node";
 import addMonths from "date-fns/add_months";
 import { type Context } from "koa";
+import { pick } from "lodash";
 import { User, Event, Team } from "../models";
 import { getCookieDomain } from "../utils/domains";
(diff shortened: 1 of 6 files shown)


server/utils/authentication.js
```diff
@@ -10,17 +13,36 @@ export function getAllowedDomains(): string[] {
...
+  if (isNewTeam) {
+    // see: scenes/Login/index.js for where this cookie is written when
+    // viewing the /login or /create pages. It is a URI encoded JSON string.
+    const cookie = ctx.cookies.get("signupQueryParams");
+
+    if (cookie) {
+      try {
+        const signupQueryParams = pick(
+          JSON.parse(querystring.unescape(cookie)),
+          ["ref", "utm_content", "utm_medium", "utm_source", "utm_campaign"]
```
tommoor, line 37: Some filtering here to at least have control over which keys end up in this object

PR #2105, opened 2021-05-02
fix: Add application/octet-stream as a valid mimetype for docx uploads

### LINKED ISSUE #2104: docx import fails on Chrome and Edge
#1551 added support for docx files, but it only seems to work on Firefox. On Chrome and Edge (latest versions on Win10), it fails with `File type application/octet-stream not supported`.

**To Reproduce**
1. From a collections page, click `Import Document`
2. From the file picker dialog, change file type to `All files (*.*)`
3. Select a `.docx` file

**Expected behavior**

Outline shows .docx files in the file dialog and allows importing of docx files.

**Actual behavior**

docx upload fails with `File type application/octet-stream not supported`

**More details**

The bug is on [this line of documentImporter.js](https://github.com/outline/outline/blob/49ffcda8e051abf9429a74782f3ce6af417253ec/server/commands/documentImporter.js#L49). It assumes that all browsers upload `.docx` files with mimetype `application/vnd.openxmlformats-officedocument.wordprocessingml.document`. Firefox does use that mimetype, but both Chrome and Edge upload `.docx` files as `application/octet-stream`.

**Screenshots**

image

image

- Install: getoutline.com
- Version: n/a

 - OS: Win10 x64
 - Browser Chrome
 - Version 90

tommoor:
closes #2104


app/stores/DocumentsStore.js
```diff
@@ -11,6 +11,8 @@ import Document from "models/Document";
 import type { FetchOptions, PaginationParams, SearchResult } from "types";
 import { client } from "utils/ApiClient";
 
+const MAXIUMUM_IMPORT_SIZE = 1024 * 1000 * 5;
```
iamsaumya, line 14: A typo in Maximum?

tommoor: 🤦

iamsaumya:
If we refer [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), especially this point 

> `application/octet-stream` is the default value for all other cases. An unknown file type should use this type. Browsers pay particular care when manipulating these files, attempting to safeguard the user to prevent dangerous behaviours.

and later in the table `application/octet-stream` is mapped to `.bin` type. 

Idk how, but I feel this could be a potential security risk.

iamsaumya:
Maybe for the files that have `application/octet-stream` we should check if they have `.docx` in name.

tommoor:
> Maybe for the files that have application/octet-stream we should check if they have .docx in name.

We could certainly do that, I don't see any harm in it as an extra filter on the server 👍 . All of these parameters can be manipulated on the client with ease so I don't think it's introducing any security issues to handle this type either way.

iamsaumya:
I'm confused about whether to update this error message to `File type X with extension Y is not supported" or not! What are your thoughts? 

https://github.com/outline/outline/blob/db6a9e03c5c394f71083196434e747bfbda5f073/server/commands/documentImporter.js#L164

tommoor:
> whether to update this error message

I think the messaging is okay in the context, they'll know what file they tried to import. Side note: Next time please create your own branch and propose the changes via PR rather than working on the same branch, thanks!

iamsaumya:
> Next time please create your own branch

Sorry, I will keep this in mind in the future.

PR #2119, opened 2021-05-05
fix: Added onClick event listener to show Appearance Menu options in mobile

jvmsangkal:
Changes:
=====
- Added onClick event listener to show Appearance Menu options in mobile

### DIFF
--- a/app/menus/AccountMenu.js
+++ b/app/menus/AccountMenu.js
@@ -33,7 +33,7 @@ const AppearanceMenu = React.forwardRef((props, ref) => {
 
   return (
     <>
-      <MenuButton ref={ref} {...menu} {...props}>
+      <MenuButton ref={ref} {...menu} {...props} onClick={menu.show}>
         {(props) => (
           <MenuAnchor {...props}>
             <ChangeTheme justify="space-between">


app/components/ContextMenu/index.js
```diff
@@ -99,8 +99,7 @@ const Background = styled.div`
   border-radius: 6px;
   padding: 6px 0;
   min-width: 180px;
-  overflow: hidden;
-  overflow-y: auto;
```
tommoor, line 103: I'm not sure when this regressed yet, but this won't be the fix – the appearance menu should be a portal so the overflow of the parent has no effect. We still need these rules to ensure that the content of the menu scrolls correctly when it won't fit.

tommoor: I bisected and amazingly it was this change: https://github.com/outline/outline/commit/8009e8f6917ae03a8a00a7d8ec0a02ca42bbc739

Fixed on main as this is a pretty serious issue:
https://github.com/outline/outline/commit/974d6b2cbed8476d771f887211bea93ba6a6394e

I think we could keep this PR with your improvement for the click behavior though…

jvmsangkal: Thank you! I've updated the PR title and description and removed the commit relating to this fix.

tommoor:
Thanks 😄

PR #2126, opened 2021-05-09
fix: Remove application/octet-stream as valid frontend mimetype

### LINKED ISSUE #2123: Remove application/octet-stream as valid frontend mimetype
Follow up to https://github.com/outline/outline/pull/2105

After further testing on Windows it seems that using this mimetype results in .bin, .com, .exe files showing as available options. Instead lets just whitelist .doc, .docx extensions explicitly. This is a frontend only change, all of the updated handling on the backend can remain the same.

iamsaumya:
Closes #2123

### DIFF
--- a/app/stores/DocumentsStore.js
+++ b/app/stores/DocumentsStore.js
@@ -24,12 +24,13 @@ export default class DocumentsStore extends BaseStore<Document> {
 
   importFileTypes: string[] = [
     ".md",
+    ".doc",
+    ".docx",
     "text/markdown",
     "text/plain",
     "text/html",
     "application/msword",
     "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-    "application/octet-stream",
   ];
 
   constructor(rootStore: RootStore) {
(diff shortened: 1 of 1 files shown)


tommoor:
I don't think this will work alone because of this check:

https://github.com/outline/outline/blob/69802cc9850a07fd986a9dcfda1b50549f81debd/app/stores/DocumentsStore.js#L535

Probably needs to be updated to check if the file name ends with as an alternative.

iamsaumya:
If I am not mistaken, we have to update this for the scenario where mimetype is `application/octet-stream` and the file is `.docx`. Isn't it? Because for most of the cases the two possible mimetype of word doc is present in the array.

iamsaumya:
I think `.includes()` alone can handle the empty string scenario here, what are your thoughts? 
https://github.com/outline/outline/blob/69802cc9850a07fd986a9dcfda1b50549f81debd/app/stores/DocumentsStore.js#L535

tommoor:
I don't think the extension check should be docx specific, it'll be useful to be able to add other additional files by extension-only here in the future.

tommoor (approved):
great, thanks!

PR #2136, opened 2021-05-13
fix: Moving documents between collections does not update attachment permissions

### LINKED ISSUE #2107: Images permissions are shared when document is moved
User with lower access rights than in the original document cannot view image on a duplicated document.

**To Reproduce**
1. Create a document with images in it, inside a restricted access collection
2. Duplicate this document to another collection with wider / larger access or public
3. Open the duplicated document **as a user** that does not have access to the first and restricted collection where the  document were originally created.
4. You does not have access to the images (403)

**Expected behavior**
When a document is duplicated, image right should have their own access data. I suspect they share the same permissions data.

**Workaround** 
Re-upload images in the duplicated document.

**Screenshots**

You get a 403 for each image in the duplicated document :(

- Install: self hosted
- Version: [Docker Image 0.55.0](https://hub.docker.com/layers/outlinewiki/outline/version-0.55.0/images/sha256-1af6794e2a290e0c42b5272f2cd9cd08ca4b6f6d4e48e87070226750e8e2dddf?context=explore)
- We use Minio as image storage service

 - Browser : Firefox (87.0), Chrome (90.0.4430.93)

tommoor:
closes #2107

### DIFF
--- a/server/test/factories.js
+++ b/server/test/factories.js
@@ -242,11 +242,6 @@ export async function buildAttachment(overrides: Object = {}) {
     overrides.userId = user.id;
   }
 
-  if (!overrides.collectionId) {
-    const collection = await buildCollection(overrides);
-    overrides.collectionId = collection.id;
-  }
-
   if (!overrides.documentId) {
     const document = await buildDocument(overrides);
     overrides.documentId = document.id;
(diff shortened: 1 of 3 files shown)


server/test/factories.js
```diff
@@ -242,11 +242,6 @@ export async function buildAttachment(overrides: Object = {}) {
     overrides.userId = user.id;
   }
 
-  if (!overrides.collectionId) {
```
tommoor, line 245: Unused – attachments don't have collection association

PR #2139, opened 2021-05-14
feat: Rebuilt member admin

### LINKED ISSUE #2077: Switching between tabs on "People" screen in admin does not load additional data
**To Reproduce**
1. Go to the "People" screen in admin
2. Click on tabs
3. Network requests are not performed to get more users with that specific filter
4. On large teams this results in incorrect results

tommoor:
Moving admin screens to use filterable, sortable, tables where appropriate. The screen that needs this the most right now is member management. For teams with hundreds of users it's very difficult to find or understand users currently without lots of scrolling.

#### Before (main)
image

#### After (this branch)
image

#### Search query
image

#### Ordered by recently active
image

closes #2077


app/components/FilterOptions.js
```diff
@@ -50,30 +51,49 @@ const FilterOptions = ({
...
-                menu.hide();
-              }}
-              active={option.key === activeKey}
-              {...option}
-              {...menu}
-            />
-          ))}
-        </List>
+        {options.map((option) => (
+          <MenuItem
```
tommoor, line 55: I was able to change search filters to use the standard context menu component with these tweaks, and then reuse the search filter dropdown on the members page, all-in-all nice improvements

app/components/Mask.js
```diff
@@ -8,6 +8,8 @@ import Flex from "components/Flex";
 type Props = {|
   header?: boolean,
   height?: number,
+  minWidth?: number,
```
tommoor, line 11: Allows tweaking the size of the placeholder mask for specific usecases where content length is known

app/components/Table.js
```diff
@@ -0,0 +1,235 @@
+// @flow
```
tommoor, line 1: 🍖  Reusable `Table` component for admin screens

app/stores/BaseStore.js
```diff
@@ -175,7 +175,7 @@ export default class BaseStore<T: BaseModel> {
         res.data.forEach(this.add);
         this.isLoaded = true;
       });
-      return res.data;
+      return res;
```
tommoor, line 178: Nowhere else uses this

PR #2162, opened 2021-06-01
fix: Use friendly urls for collections

### LINKED ISSUE #1995: Use friendly urls for collections
We've been storing `urlId` in the database for collections since day 1, in the same way we do for documents – but we've never used it! It should be relatively simple to update the `findByPk` method to accept a urlId (basically copy and paste from the Document one) and then have the frontend use them.

This way we change urls from:

`https://app.getoutline.com/collections/8e6edb23-e4b7-474c-97b4-5e7fec49df56` ->
`https://app.getoutline.com/collections/product-97b4ih`

iamsaumya:
Closes #1995

Todos
- [x] Expose collection urlId to frontend
- [x] Update URL getterMethods to return new URL type
- [x] Update findByPk(similar to doc)
- [x] Add getByUrl in collection store
- [x] Add useEffect to redirect to new URL type
- [x] Overried fetch in collection store
- [x] fix bug, to return null when the urlId and slugified name do match to a collection.
- [x] Migrate from /collections/ to /collection/


app/stores/BaseStore.js
```diff
@@ -140,6 +140,11 @@ export default class BaseStore<T: BaseModel> {
     }
 
     let item = this.data.get(id);
+
+    if (this.modelName === "Collection") {
+      item = this.rootStore.collections.getByUrl(id);
+    }
+
```
iamsaumya, line 147: @tommoor, Is this okay, or Should I consider creating fetch in the collection store?

server/models/Collection.js
```diff
@@ -223,6 +226,17 @@ Collection.addHook("afterCreate", (model: Collection, options) => {
 
 // Class methods
 
+Collection.findByPk = async function (id, options = {}) {
+  if (isUUID(id)) {
+    return this.findOne({ where: { id }, ...options });
```
iamsaumya, line 231: We are using `findByPk` with two `scopes`. So I used `this` here. All the server test cases are passing, so I think `this` is using the scope returned by the `scopes`.

app/stores/BaseStore.js
tommoor: Override fetch in collection store pls :) This feature is likely to mean a bit of copy/paste from documents but that's okay.

iamsaumya:
This looks good to me. Please give it a review whenever you're unoccupied to help me corner cases.

app/stores/CollectionsStore.js
```diff
@@ -135,6 +159,12 @@ export default class CollectionsStore extends BaseStore<Collection> {
     if (path) return path.title;
   }
 
+  getByUrl(url: string): ?Collection {
+    return find(Array.from(this.data.values()), (col: Collection) =>
```
tommoor, line 163:
```suggestion
    return find(this.orderedData, (col: Collection) =>
```

`orderedData` is already a pre-computed array we can use here

app/scenes/Collection.js
```diff
@@ -38,18 +48,31 @@ import { newDocumentUrl, collectionUrl } from "utils/routeHelpers";
...
+  const collection = collections.getByUrl(id);
+  const can = policies.abilities(collection?.id || "");
   const canUser = policies.abilities(team.id);
-  const { handleFiles, isImporting } = useImportDocument(collectionId);
+  const { handleFiles, isImporting } = useImportDocument(collection?.id || "");
+
+  React.useEffect(() => {
+    async function load() {
+      if (isUUID(id)) {
+        const collection = await collections.fetch(id);
```
tommoor, line 69: The `collection` model is already guaranteed to be fetched for this screen, I wonder if we could we just add `collection` as a dependency of useEffect and wait for it to exist?

app/scenes/Collection.js
```diff
@@ -38,18 +48,31 @@ import { newDocumentUrl, collectionUrl } from "utils/routeHelpers";
...
   const team = useCurrentTeam();
   const [isFetching, setFetching] = React.useState();
   const [error, setError] = React.useState();
   const [permissionsModalOpen, setPermissionsModalOpen] = React.useState(false);
 
-  const collectionId = params.id || "";
-  const collection = collections.get(collectionId);
-  const can = policies.abilities(collectionId || "");
+  const id = params.id || "";
+  const collection = collections.getByUrl(id);
```
tommoor, line 61:
```suggestion
  const collection = collections.getByUrl(id) || collections.get(id);
```

Would it be useful for backwards compat and avoid the extra fetch call when loading an old-style url?

app/scenes/DocumentNew.js
```diff
@@ -1,58 +1,52 @@
...
-    } catch (err) {
-      this.props.ui.showToast("Couldn’t create the document, try again?", {
-        type: "error",
-      });
-      this.props.history.goBack();
+  useEffect(() => {
+    async function createDocument() {
+      const params = queryString.parse(location.search);
+      try {
+        const collection = await collections.fetch(match.params.id || "");
```
tommoor, line 23: We should be able to make the fetch here a fallback if collection can't be found in the store, 99.9% of the time the collection will be loaded locally already

server/models/Collection.js
tommoor: This makes sense, a couple of explicit tests for the cases of passing a uuid / urlId here  would be good.

server/models/Collection.js
```diff
@@ -1,12 +1,13 @@
...
 import randomstring from "randomstring";
-import slug from "slug";
+import isUUID from "validator/lib/isUUID";
 import { Op, DataTypes, sequelize } from "../sequelize";
+import slugify from "../utils/slugify";
 import CollectionUser from "./CollectionUser";
 import Document from "./Document";
 
-slug.defaults.mode = "rfc3986";
+const URL_REGEX = /^[0-9a-zA-Z-_~]*-([a-zA-Z0-9]{10,15})$/;
```
tommoor, line 10: I wonder if we should centralize this now it's used in two places, perhaps export from shared/routeHelpers?

app/utils/routeHelpers.js
```diff
@@ -14,10 +14,19 @@ export function newCollectionUrl(): string {
...
-export function collectionUrl(collectionId: string, section: ?string): string {
-  const path = `/collections/${collectionId}`;
-  if (section) return `${path}/${section}`;
-  return path;
+export function collectionUrl(url: string, section: ?string): string {
+  if (section) return `${url}/${section}`;
+  return url;
+}
+
+export function updateCollectionUrl(oldUrl: string, newUrl: string): string {
```
tommoor, line 22: Wondering if this would be more maintainable as a string replacement, looking for `collections/${collection.id}` -> `collections.url` ?

server/models/Collection.js
iamsaumya: Agree, Will do that.

app/scenes/DocumentNew.js
iamsaumya: We do `collection.get` and `collections.getByUrl` calls inside the fetch.

app/utils/routeHelpers.js
iamsaumya: I am sorry I didn't get it. Isn't this what we are doing?

tommoor: I think that's what this is achieving but in a slightly more fragile way by splitting and recombining the parts. Could it not be?

```javascript
  return oldUrl.replace(`collections/${collection.id}`, collections.url);
```

tommoor (commented):
Working great while testing locally, haven't found an edgecase yet

tommoor:
Now that I see it in the url I'm thinking it would make sense to take this opportunity to singularize the new routes to "collection" the same as "doc"

iamsaumya:
I would be happy to make another PR with those changes. It will help us keep this PR focused on collections.

app/utils/routeHelpers.js
iamsaumya: Do you suggest modify the `updateDocumentUrl`? This function is an exact copy of it, and just the name is different 😅

app/scenes/DocumentNew.js
```diff
@@ -1,58 +1,52 @@
...
+          collectionId: collection.id,
+          parentDocumentId: params.parentDocumentId,
+          templateId: params.templateId,
+          template: params.template,
+          title: "",
+          text: "",
+        });
+        history.replace(editDocumentUrl(document));
+      } catch (err) {
+        ui.showToast("Couldn’t create the document, try again?", {
```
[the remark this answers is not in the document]
iamsaumya: Thanks for pointing out 👍🏻

app/utils/routeHelpers.js
tommoor: Hah, maybe! My old code can always be improved... but maybe I'm missing something here, it's late

iamsaumya: I will update it if it doesn't break anything.

app/routes/authenticated.js
```diff
@@ -46,9 +46,9 @@ export default function AuthenticatedRoutes() {
           <Route exact path="/drafts" component={Drafts} />
           <Route exact path="/archive" component={Archive} />
           <Route exact path="/trash" component={Trash} />
-          <Route exact path="/collections/:id/new" component={DocumentNew} />
-          <Route exact path="/collections/:id/:tab" component={Collection} />
-          <Route exact path="/collections/:id" component={Collection} />
```
tommoor, line 51: We'll need to retain a redirect from `/collections/*` -> `/collection/*` here

app/scenes/DocumentNew.js
```diff
@@ -1,58 +1,60 @@
...
-  async componentDidMount() {
-    const params = queryString.parse(this.props.location.search);
+  useEffect(() => {
+    async function createDocument() {
+      const params = queryString.parse(location.search);
+      try {
+        const collection =
+          collections.getByUrl(id) ||
+          collections.get(id) ||
+          (await collections.fetch(id));
```
tommoor, line 29:
```suggestion
        const collection = await collections.fetch(id);
```

Checking for id is part of `fetch`, right? No need for these extra checks

server/models/Collection.test.js
```diff
@@ -416,3 +418,63 @@ describe("#membershipUserIds", () => {
     expect(membershipUserIds.length).toBe(6);
   });
 });
+
+describe("#findByPk", () => {
+  test("should return collection with collection Id", async () => {
+    const { collection } = await seed();
```
tommoor, line 424:
```suggestion
    const collection = await buildCollection();
```

`seed` is quite slow as it builds a bunch of different models, I'm trying to avoid using it in new tests in favor of individual factories like above

server/models/Collection.test.js
```diff
@@ -416,3 +418,63 @@ describe("#membershipUserIds", () => {
...
+  });
+
+  test("should return null when urlId is present but corresponding collection name is incorrect", async () => {
+    const { collection } = await seed();
+    const id = `${slugify("not present collection")}-${collection.urlId}`;
+
+    const response = await Collection.findByPk(id);
+
+    expect(response).toBe(null);
+  });
```
tommoor, line 479: Hmm, this is unexpected behavior – as long as the `urlId` is correct the slug portion can be anything. This enables the title to be changed and existing links that might exist to continue working

server/models/Document.js
```diff
@@ -217,12 +217,16 @@ Document.findByPk = async function (id, options = {}) {
...
       where: {
-        urlId: id.match(URL_REGEX)[1],
+        urlId: id.match(URL_REGEX)[2],
       },
       ...options,
     });
+    if (document && slugify(document.title) !== id.match(URL_REGEX)[1]) {
+      return null;
+    }
+    return document;
```
tommoor, line 229: See comment on test, this addition is undesired

app/scenes/Collection.js
```diff
@@ -38,18 +49,28 @@ import { newDocumentUrl, collectionUrl } from "utils/routeHelpers";
...
+  const id = params.id || "";
+  const collection: ?Collection =
+    collections.getByUrl(id) || collections.get(id);
+  const can = policies.abilities(collection?.id || "");
   const canUser = policies.abilities(team.id);
-  const { handleFiles, isImporting } = useImportDocument(collectionId);
+  const { handleFiles, isImporting } = useImportDocument(collection?.id || "");
+
+  React.useEffect(() => {
+    if (isUUID(id) && collection) {
```
tommoor, line 69: Rather than checking `isUUID` here we should check if the current url !== canonicalUrl and redirect in that instance, this allows for old slugs to be updated if, for example, there is an external link somewhere to a collection that gets renamed.

(This is the same pattern used on documents)

shared/utils/routeHelpers.js
```diff
@@ -61,3 +61,5 @@ export function settings(): string {
 export function groupSettings(): string {
   return `/settings/groups`;
 }
+
+export const URL_REGEX = /^([0-9a-zA-Z-_~]*)-([a-zA-Z0-9]{10,15})$/;
```
tommoor, line 65:
```suggestion
export const SLUG_URL_REGEX = /^([0-9a-zA-Z-_~]*)-([a-zA-Z0-9]{10,15})$/;
```

Might be clearer in the future now it's no longer within the context of documents specifically.

tommoor (changes requested):

server/models/Collection.test.js
iamsaumya: Oh! For some reason, I thought seed should be fast 😅

server/models/Collection.test.js
iamsaumya: Noted. 👍🏻

app/utils/routeHelpers.js
```diff
@@ -25,7 +25,10 @@ export function updateCollectionUrl(
   collection: Collection
 ): string {
   // Update url to match the current one
-  return oldUrl.replace(`/collection/${collection.id}`, collection.url);
```
iamsaumya: This approach was not updating the URL if it is `wrong-collection-r9QsabjdGv` to `correct-collection-r9QsabjdGv`.

tommoor: aha!

tommoor (approved):
I think this is ready to go

PR #2164, opened 2021-06-02
fix: Add Portugese, Brazil to language options

### LINKED ISSUE #2163: Add Portugese, Brazilian to language options
Enough translations have been completed to make this a selectable language

iamsaumya:
Closes #2163


iamsaumya:
Silly me 😞 

I will have to update the package to support the `pt_BR`.

tommoor:
This is great, the work really spiralled with the upgrade of date-dns huh? 😅  

I'm struggling to understand why the upgrade caused the size to jump so much, according to RelativeCI the entire application got 12% larger by upgrading the date library, wtf!

iamsaumya:
Strange! I was hoping for the size to decrease because the package claims to be the smallest date library out there.

iamsaumya:
I'm pushing up a code. Let's see if it solves the issue.

iamsaumya:
It's still 6% 😕

iamsaumya:
I'm not sure how to decrease the bundle size further 😕

tommoor:
Updated with all of @vio 's improvements to make sure that RelativeCI is displaying the diff correctly, it seems that 6% larger might just be as small as it goes :( Kind of frustrating that the lib increased so much in size with v2

iamsaumya:
It's clearly a huge increase in bundle size! Earlier I was wrongly importing the functions; hence the tree shaking was not happening. I guess we have to deal with a 6% increase :(

iamsaumya:
Btw going through modules, I see some modules that are not used at all and still included in the bundle. 

Idk if these modules are transitively used in the functions that we use. 

Screenshot from 2021-06-04 11-23-08

tommoor:
It could be the use of `format`, it looks like there are more specific format functions eg `formatRFC3339` that can be used for creating the credential expiration timestamps. Using these more specific ones will probably cut a few percent.

https://date-fns.org/v2.22.1/docs/formatRFC3339

iamsaumya:
For revisions, it's not possible to format like `June 3rd, 2021 7:15 PM` without using `format`.

vio:
@iamsaumya you are right about `getWeekYear` module, it is not removed by the tree-shaking optimization. 

```json
ModuleConcatenation bailout: Module is not an ECMAScript module
```
[webpack stats module entry](https://gist.github.com/vio/fbd52519943caccbe75c55e75bdd4258)

I suspect an issue with webpack loading commonjs modules and not being able to tree shake. I used a small example app to import all the fns that you are using, downgrade webpack and match the babel configuration: https://app.relative-ci.com/projects/QYHKPo5Xwt85dKsTmV83/jobs/1152-ce92703b-36af-4c86-987c-12c6038d709d/modules?bm=%7B%22search%22%3A%22node_modules%2Fdate-fns%2F%22%2C%22filters%22%3A%7B%22changed%22%3Afalse%7D%7D

- the stats size is ~ 4-5x smaller (extra locales excluded)
- modules are included from `esm` folder

iamsaumya:
Thanks @vio for the help here!

iamsaumya:
Loading from `esm` folder in the `server` files is giving an error while building, probably because it doesn't support it.

Loading fns from `esm` is working fine on the `app` side but the jest is giving an error while testing because it doesn't support it.

vio:
@iamsaumya got the same error locally on build

However, i saw that webpack is explicitly using only `browser` and `main` package.json fields - https://github.com/outline/outline/blob/main/webpack.config.js#L48. It's not a standard, but most of the packages([including `date-fns`](https://github.com/date-fns/date-fns/blob/master/package.json#L21)) are publishing esm versions using `module` field.

I pushed a change for the config in a different branch - https://github.com/outline/outline/pull/2173:
1. many packages had their total size decreased 
2. couple packages increased with more than 150%

iamsaumya:
~-6% is a decrease in the bundle size is phenomenal. Thanks for the help here, @vio. 

I got to admit. I don't know what these fields do and how exactly they impact importing packages. Now I know I need to learn frontend tooling and performance. 😄

tommoor:
Urgh, back to 12% 😠

iamsaumya:
Is it because the import format is changed? Feels cryptic.

vio:
@iamsaumya I don't think the bundle size increase is directly related to the code changes, but the date-fns upgrade might be affecting the way webpack is chunking the code.

Couple things in the latest report were not making a lot of sense for me, i tried to spend a bit more time debugging the output.

1. [`authenticated-routes` view](https://bit.ly/3ijdRBi) was showing a minor total delta compared with the chunk size difference. 

  This was a bug with the filters that were applied after the comparison, so the the modules available on the baseline side were not part of the same chunk. This was fixed and now the filter is respecting the module chunkId for both baseline and current job.

2. [`vendors~locale-time.js` view](https://bit.ly/3fTI1tn) is showing a large positive change while the resulting chunk is reporting an increase in size (negative). 

  This appears to be caused by the chunkId changing between builds - more info on the PR #2181

3. [`date-fns` size increased with `+117.45%`](https://bit.ly/34OsV23
) 
  
  [Excluding locales](https://bit.ly/3ptjdeL
), the size is close with [the standalone test](https://bit.ly/3vUK8T7
) and much better compared with size before the package.json field fix - https://bit.ly/2RocHt8. Maybe worth looking if the format fn change requires more dependencies from the library or if the locales data can be loaded asynchronously based on the user.

iamsaumya:
The `date-fns` upgrade is indeed something we cannot put out of the picture. 

As you mentioned, we should do something about `locales`. Going through their documentation, I had found their solution for dynamic import with webpack [here](https://date-fns.org/v2.22.1/docs/webpack).

A couple of days ago, I did implement their approach (https://github.com/outline/outline/pull/2164/commits/7f0fc684cf37b924f17de4bdb5dad3a41ae3a63d), but there was no improvement. 

With the improvements you have made in the last few days, I will give their dynamic imports another shot.

tommoor:
By all means give it a shot. If the individual languages are async loaded then we can probably remove LocaleTime being an async chunk as this was just a cheap way of achieving similar by moving all the language translations into that one place.

vio:
@tommoor thanks for merging the latest changes! Based on the report, it looks like the difference is caused by [duplicate app code in `authenticated-routes` chunk](https://bit.ly/34VIjcE).

Screenshot 2021-06-05 at 23 45 28

`vendors~authenticated-routes~locale-time` contains [`date-fns` updates](https://app.relative-ci.com/projects/TMqufq6bi8qzsOjEHSWY/jobs/1096-2e971fff-a155-4e46-90b8-2ca4ad3012a5/modules?bm=%7B%22filters%22%3A%7B%22c.0%22%3Afalse%2C%22c.authenticated%22%3Afalse%2C%22c.authenticated-routes%22%3Afalse%2C%22c.keyed-document%22%3Afalse%2C%22c.locale-time%22%3Afalse%2C%22c.login%22%3Afalse%2C%22c.logout%22%3Afalse%2C%22c.main%22%3Afalse%2C%22c.runtime%22%3Afalse%2C%22c.settings%22%3Afalse%2C%22c.table%22%3Afalse%2C%22c.vendors%22%3Afalse%2C%22c.vendors~authenticated-routes%22%3Afalse%2C%22c.vendors~authenticated-routes~keyed-document%22%3Afalse%2C%22c.vendors~authenticated-routes~keyed-document~login%22%3Afalse%2C%22c.vendors~authenticated-routes~locale-time%22%3Atrue%2C%22c.vendors~locale-time%22%3Afalse%2C%22c.vendors~rich-markdown-editor%22%3Afalse%2C%22c.vendors~settings%22%3Afalse%2C%22c.vendors~table%22%3Afalse%2C%22c.vendors~twitter-picker%22%3Afalse%2C%22mft.CSS%22%3Afalse%2C%22mft.JS%22%3Atrue%2C%22mft.OTHER%22%3Afalse%2C%22changed%22%3Atrue%7D%2C%22search%22%3A%22%22%7D) and `vendors~locale-time` contains [`date-fns` locales](https://app.relative-ci.com/projects/TMqufq6bi8qzsOjEHSWY/jobs/1096-2e971fff-a155-4e46-90b8-2ca4ad3012a5/modules?bm=%7B%22filters%22%3A%7B%22c.0%22%3Afalse%2C%22c.authenticated%22%3Afalse%2C%22c.authenticated-routes%22%3Afalse%2C%22c.keyed-document%22%3Afalse%2C%22c.locale-time%22%3Afalse%2C%22c.login%22%3Afalse%2C%22c.logout%22%3Afalse%2C%22c.main%22%3Afalse%2C%22c.runtime%22%3Afalse%2C%22c.settings%22%3Afalse%2C%22c.table%22%3Afalse%2C%22c.vendors%22%3Afalse%2C%22c.vendors~authenticated-routes%22%3Afalse%2C%22c.vendors~authenticated-routes~keyed-document%22%3Afalse%2C%22c.vendors~authenticated-routes~keyed-document~login%22%3Afalse%2C%22c.vendors~authenticated-routes~locale-time%22%3Afalse%2C%22c.vendors~locale-time%22%3Atrue%2C%22c.vendors~rich-markdown-editor%22%3Afalse%2C%22c.vendors~settings%22%3Afalse%2C%22c.vendors~table%22%3Afalse%2C%22c.vendors~twitter-picker%22%3Afalse%2C%22mft.CSS%22%3Afalse%2C%22mft.JS%22%3Atrue%2C%22mft.OTHER%22%3Afalse%2C%22changed%22%3Atrue%7D%2C%22search%22%3A%22%22%7D).

Not clear what change triggered the code duplication, definitely worth investigating. Also, it is possible to configure webpack to extract the common modules between `authenticated-routes` and `keyed-document` and avoid the extra bundle-size.

If you need any help with that, will be more than happy to have a look ;)

tommoor:
I would appreciate your help but you've already done so much. If you have any tips happy to investigate here too

vio:
@tommoor glad i could help & really useful to test relative-ci usage on a different setup ;) I will keep an eye on this issue and if still need it I will have a look at setting a [default / common splitChunk group](https://v4.webpack.js.org/plugins/split-chunks-plugin/#optimizationsplitchunks). Extracting the common application code into different chunks should remove the bundle-size increase.

In case it helps, i'm usually using a local workflow to test the changes (it's faster than waiting for the CI build) - https://gist.github.com/vio/af8c5e75930e11dc0d718629e3ff2807

tommoor:
I've tried various configs of splitChunks on this branch without much luck: https://app.relative-ci.com/projects/TMqufq6bi8qzsOjEHSWY/tree/fix%2Fissue-2163

I did notice that KeyedDocument was async in one spot and not in another, so this knocked a couple of percent off.

tommoor:
Okay, happy with this! And because of @vio's help we're now much smaller than a week ago…

image

tommoor (approved):

PR #2170, opened 2021-06-03
build: Add async chunk names

vio:
Prevent webpack from renaming chunks between builds:

1. improve chunk modules tracking between builds
2. deterministic cache invalidation - chunks will be invalidated only when the content will change


tommoor:
Nice, thank you!

tommoor (approved):

PR #2173, opened 2021-06-04
build: Webpack config - allow package.json module field usage

vio:
- revert resolve.mainFields to [default](https://webpack.js.org/configuration/resolve/#resolvemainfields)
- revert boundless-* package aliases to commonjs folder to avoid transpiling

rel: https://github.com/outline/outline/pull/507/commits/fe00e1f76d080107bf7f0c4ec40e24c754efc521

@tommoor I pushed the config change to see the impact. There are many changes across the modules, please consider merging only if you don't see any side-effects.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -45,9 +45,12 @@ module.exports = {
       path.resolve(__dirname, 'app'),
       'node_modules'
     ],
-    mainFields: ["browser",  "main"],
     alias: {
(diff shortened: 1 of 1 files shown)


tommoor:
Testing this locally I haven't found any side effects yet, It makes sense that the default webpack config for this field should be relatively "safe"

PR #2176, opened 2021-06-04
fix: Failure loading collections on frontend results in loading loop

### LINKED ISSUE #2161: Failure loading collections on frontend results in loading loop
**To Reproduce**
1. Go to /home
2. If collections.info call fails
3. Request will be made repeatedly in a tight loop until success

iamsaumya:
Closes #2161

### DIFF
--- a/app/components/Sidebar/components/Collections.js
+++ b/app/components/Sidebar/components/Collections.js
@@ -38,10 +38,10 @@ function Collections({ onCreateCollection }: Props) {
           setFetching(true);
           await collections.fetchPage({ limit: 100 });
         } catch (error) {
-          ui.showToast(error.message + t(", try again?"), {
+          ui.showToast(t("Collections could not be loaded, please reload the app"), {
             type: "error",
           });
-          setFetchError(true);
+          setFetchError(error);
         } finally {
           setFetching(false);
         }
--- a/app/components/Sidebar/components/Collections.js
+++ b/app/components/Sidebar/components/Collections.js
@@ -20,7 +20,7 @@ type Props = {
 
 function Collections({ onCreateCollection }: Props) {
   const [isFetching, setFetching] = React.useState(false);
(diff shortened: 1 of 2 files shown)


app/components/Sidebar/components/Collections.js
```diff
@@ -20,6 +20,7 @@ type Props = {
 
 function Collections({ onCreateCollection }: Props) {
   const [isFetching, setFetching] = React.useState(false);
+  const [fetchError, setFetchError] = React.useState(false);
```
tommoor, line 23:
```suggestion
  const [fetchError, setFetchError] = React.useState();
```

`undefined` is better then a boolean in this case

app/components/Sidebar/components/Collections.js
```diff
@@ -32,17 +33,22 @@ function Collections({ onCreateCollection }: Props) {
...
-      if (!collections.isLoaded && !isFetching) {
+      if (!collections.isLoaded && !isFetching && !fetchError) {
         try {
           setFetching(true);
           await collections.fetchPage({ limit: 100 });
+        } catch (error) {
+          ui.showToast(error.message + t(", try again?"), {
+            type: "error",
+          });
+          setFetchError(true);
```
tommoor, line 44:
```suggestion
          ui.showToast(t("Collections could not be loaded, please reload the app"), {
            type: "error",
          });
          setFetchError(error);
```

The error message isn't guaranteed to make sense to the end user, probably clearer to just hard code a message here. "Try again?" doesn't make as much sense for an action that was automatic to begin with.

tommoor (changes requested):
Some tweaks, but this is the right approach 👍

app/components/Sidebar/components/Collections.js
iamsaumya: Yes Yes, taking note that it was automatic, tru will not definitely make sense.

PR #2187, opened 2021-06-06
fix: change `metaDisplay` key to Alt for "Table of contents"

xphoniex:
`metaDisplay` is **Ctrl** on non-Mac systems so this key becomes "Ctrl+ctrl+h" which doesn't work. Changing `metaDisplay` to Alt seems to be a safe choice that should work on Mac too.

I also tried creating `KeyboardShortcuts.test.js` to detect duplicate keys in `KeyboardShortcuts` component's config but somehow can't install `enzyme` with yarn.

### DIFF
--- a/app/scenes/Document/components/Document.js
+++ b/app/scenes/Document/components/Document.js
@@ -174,7 +174,7 @@ class DocumentScene extends React.Component<Props> {
     this.onSave({ publish: true, done: true });
   }
 
-  @keydown(`${meta}+ctrl+h`)
+  @keydown(`${meta}+alt+h`)
   onToggleTableOfContents(ev) {
     if (!this.props.readOnly) return;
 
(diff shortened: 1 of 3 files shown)


app/scenes/Document/components/Document.js
```diff
@@ -174,7 +174,7 @@ class DocumentScene extends React.Component<Props> {
     this.onSave({ publish: true, done: true });
   }
 
-  @keydown(`${meta}+ctrl+h`)
+  @keydown(`${meta}+alt+h`)
```
tommoor, line 177:
```suggestion
  @keydown("ctrl+alt+h")
```

This binding is different to help text because it maps to "cmd+alt+h" on Mac.

app/scenes/Document/components/Header.js
```diff
@@ -83,7 +83,7 @@ function DocumentHeader({
   const toc = (
     <Tooltip
       tooltip={ui.tocVisible ? t("Hide contents") : t("Show contents")}
-      shortcut={`ctrl+${metaDisplay}+h`}
+      shortcut={`Ctrl+alt+h`}
```
tommoor, line 86:
```suggestion
      shortcut="ctrl+alt+h"
```

tommoor (changes requested):
Good find

Issue #2015, opened 2021-04-07
InputSelect uses a white background for options in dark mode.


iamsaumya:
InputSelect uses a white background for options in dark mode.

**To Reproduce**
Steps to reproduce the behaviour( in dark mode):
1. Open the modal for creating a collection
2. Click on the default access dropdown
3. Options have a white background

**Expected behaviour**
Use custom select input to show the options. 

Settings -> Members page
![Screenshot from 2021-06-07 00-40-13](https://user-images.githubusercontent.com/40148448/120937067-fea86480-c728-11eb-8605-5061c155c82f.png)

**Screenshots**
![Screenshot from 2021-04-07 17-40-48](https://user-images.githubusercontent.com/40148448/113864434-85dc6900-97c8-11eb-9adc-3f9efbb0d596.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Brave

tommoor:
Interesting, we don't attempt to style this aspect of the select – on macOS it looks great in the system style

![image](https://user-images.githubusercontent.com/380914/114253292-8575d680-995e-11eb-9164-7ba32353ae99.png)

iamsaumya:
Should I go ahead and make a PR? This issue is stranded for quite some time now.

tommoor:
I wonder if the only way to properly solve this is to use a custom select input 😬

iamsaumya:
Makes sense 👍🏻. I'll remove the "good first issue" label.

PR #2188, opened 2021-06-06
style: add option background for InputSelect

xphoniex:
This should fix https://github.com/outline/outline/issues/2015.
Please check with a Mac and leave a comment if it works.

### DIFF
--- a/app/components/InputSelect.js
+++ b/app/components/InputSelect.js
@@ -16,6 +16,10 @@ const Select = styled.select`
   color: ${(props) => props.theme.text};
   height: 30px;
 
+  option {
+    background: ${(props) => props.theme.buttonNeutralBackground};
+  }
+
   &:disabled,
   &::placeholder {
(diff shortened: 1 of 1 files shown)


iamsaumya:
Hey @xphoniex, thanks for making an effort to put up a PR. 

The maintainer of the project had expressed his interest in solving this issue using custom input select ([here](https://github.com/outline/outline/issues/2015#issuecomment-851019380)). 

It's my fault for not updating the **Expected Behaviour** of the issue, and I will update it right away.

tommoor:
If something this simple _does_ fix the background on Windows and not affect macOS then i'd be down with it as a patch until something custom is built

iamsaumya:
On Ubuntu 

Screenshot from 2021-06-07 12-42-50

tommoor:
image

MacOS continues to ignore it, so honestly this seems like a great stop-gap.

tommoor (approved):

PR #2192, opened 2021-06-07
fix: Add ability to permanently delete documents in trash

### LINKED ISSUE #2057: Add ability to permanently delete documents in trash
- Currently when you first delete a document it is added to the trash section
- Deleted documents in trash are permanently deleted via a cron job after 30 days
- There is currently no way to delete prior to that

We would like to have a "Delete…" option in the "..." overflow menu for documents in the trash that immediately and permanently deletes the document before the 30 days after a confirm modal.

iamsaumya:
Closes #2057


iamsaumya:
@tommoor, with the progress here, when I delete and then permanently delete (via API) a document that has nested documents, its nested documents also happen to get completely destroyed.

I am unable to figure out how it is happening. Please help me out here!

iamsaumya:
I see, the `parentDocumentId` has a `on delete cascade` constraint. Because in the `Document` model, the `parentDocumentId` didn't have any associations declared similar to `templateId`, the idea of it being a foreign key was discard from my mind. 

Any reason why we don't mention its association in the model file? 

Do you want to keep this cascade behavior when the user permanently deletes a parent doc?

tommoor:
> Any reason why we don't mention its association in the model file?

I think this is just because it's super old and at the beginning we perhaps weren't using all the Sequelize best practices.

> Do you want to keep this cascade behavior when the user permanently deletes a parent doc?

I believe so, it's necessary to not leave miss-parented documents behind and the behavior hasn't been a problem that I can think of.

iamsaumya:
It felt odd seeing docs disappearing when I did not delete them. Probably because when they are in the trash, we lose the state of their nestedness and assume that docs are free of hierarchy. 


Maybe we should let the user know that deleting this doc will permanently delete the following docs, and the confirmation modal will show the list of docs. 

Or we can make the parentdocumentId of its child `null`. 

What do you think?

tommoor:
> we lose the state of their nestedness and assume that docs are free of hierarchy

This is a good point, it does make the behavior hard to grasp as an outside observer. Clearing the `parentDocumentId` of any immediate children and leaving them soft deleted seems safe, I can agree with this approach thanks for pushing back 👍

app/stores/DocumentsStore.js
```diff
@@ -616,8 +616,8 @@ export default class DocumentsStore extends BaseStore<Document> {
   }
 
   @action
-  async delete(document: Document) {
-    await super.delete(document);
+  async delete(document: Document, options?: Object) {
```
tommoor, line 619:
```suggestion
  async delete(document: Document, options?: {| permanent: boolean |}) {
```

server/services/websockets.js
```diff
@@ -79,6 +79,11 @@ export default class Websockets {
             ],
           });
       }
+      case "documents.permanent_delete": {
+        return socketio.to(`team-${event.teamId}`).emit(event.name, {
```
tommoor, line 83:
```suggestion
        return socketio.to(`collection-${document.collectionId}`).emit(event.name, {
```

I believe this should be sent down the collection channel otherwise the event is propagated to users that don't have permission to know about the document. (It's only an id down a websocket, but same principles apply)

server/commands/documentPermanentDeleter.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import debug from "debug";
+import { Document, Attachment } from "../models";
+import { sequelize } from "../sequelize";
+import parseAttachmentIds from "../utils/parseAttachmentIds";
+
+export async function documentPermanentDeleter(
```
tommoor, line 7: Could we pull some of the tests across from `utils` to test the command? Specifically those around attachment handling can be fairly easily moved.

server/commands/documentPermanentDeleter.js
```diff
@@ -0,0 +1,59 @@
...
+  documents: Document[],
+  log?: debug.Debugger
```
tommoor, line 9: No need to pass the log through, can use `const log = debug("commands")`

server/commands/documentPermanentDeleter.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  const countDeletedDocument = await Document.scope("withUnpublished").destroy({
+    where: {
+      id: documents.map((document) => document.id),
+    },
+    force: true,
+  });
+
+  return countDeletedDocument;
```
tommoor, line 58:
```suggestion
  return Document.scope("withUnpublished").destroy({
    where: {
      id: documents.map((document) => document.id),
    },
    force: true,
  });
```

app/scenes/DocumentPermanentDelete.js
```diff
@@ -0,0 +1,67 @@
...
+      try {
+        setIsDeleting(true);
+        await documents.delete(document, { permanent: true });
+        showToast(t("Document permanently deleted"), { type: "success" });
+        onSubmit();
+
+        if (documents.deleted.length) {
+          history.push("/trash");
+        } else {
+          history.push(collectionUrl(collection?.url || "/"));
```
tommoor, line 38: What's the thinking here? My immediate thought was "nice detail", but on further thought I think it would be better to stay in the trash and let the user see that it's empty?

app/scenes/DocumentPermanentDelete.js
```diff
@@ -0,0 +1,67 @@
...
+import Button from "components/Button";
+import Flex from "components/Flex";
+import HelpText from "components/HelpText";
+import useStores from "hooks/useStores";
+import { collectionUrl } from "utils/routeHelpers";
+
+type Props = {
+  document: Document,
+  onSubmit: () => void,
+};
```
tommoor, line 16:
```suggestion
type Props = {|
  document: Document,
  onSubmit: () => void,
|};
```

Sealing the object stops random props from being passed in without a warning

tommoor (changes requested):
Glad you could find and reuse the logic in the utils endpoint, makes a big difference to the maintainability of this feature!

server/commands/documentPermanentDeleter.js
iamsaumya: Sure

app/scenes/DocumentPermanentDelete.js
iamsaumya: learned something new!

app/scenes/DocumentPermanentDelete.js
iamsaumya: I thought of saving a click for the user 😝 . But, now that I think more, I agree it will be strange for the user to be navigated to a random collection.

iamsaumya:
Tests are passing on local 🤔

server/commands/documentPermanentDeleter.test.js
```diff
@@ -0,0 +1,119 @@
...
+      deletedAt: new Date(),
+    });
+
+    const countDeletedDoc = await documentPermanentDeleter([document]);
+
+    expect(countDeletedDoc).toEqual(1);
+    expect(await Document.unscoped().count({ paranoid: false })).toEqual(0);
+  });
+
+  it("should destroy documents even if not deleted", async () => {
```
iamsaumya, line 33: Is this okay?

tommoor:
> Tests are passing on local 🤔

I think it's the same sporadic failure, will try and address this today.

server/commands/documentPermanentDeleter.test.js
tommoor: This isn't the intent, seems like a sideeffect of querying the document outside of the command. We could have a check inside the for loop in the command and just throw if a document is passed that isn't already deleted?

server/commands/documentPermanentDeleter.test.js
```diff
@@ -0,0 +1,119 @@
...
+  const mS3 = { deleteObject: jest.fn().mockReturnThis(), promise: jest.fn() };
+  return {
+    S3: jest.fn(() => mS3),
+    Endpoint: jest.fn(),
+  };
+});
+
+beforeEach(() => {
+  flushdb();
+});
```
tommoor, line 18:
```suggestion
beforeEach(() => flushdb());
```

This is the cause of test failures, need to return the promise

iamsaumya: Thanks!!

server/commands/documentPermanentDeleter.js
```diff
@@ -7,6 +7,14 @@ import parseAttachmentIds from "../utils/parseAttachmentIds";
 const log = debug("commands");
 
 export async function documentPermanentDeleter(documents: Document[]) {
+  const index = documents.findIndex((doc) => !doc.deletedAt);
+
+  if (index !== -1) {
+    throw new Error(
+      `Cannot permanently delete ${documents[index].id} document. Please delete it and try again.`
+    );
+  }
```
tommoor, line 16:
```suggestion
  const activeDocument = documents.find((doc) => !doc.deletedAt);

  if (activeDocument) {
    throw new Error(
      `Cannot permanently delete ${activeDocument.id} document. Please delete it and try again.`
    );
  }
```

This should work and be a little clearer to follow

tommoor (approved):

PR #2194, opened 2021-06-08
fix: Two restore options when an archived document is deleted

iamsaumya:
This PR fixes the bug of two restore options showing in the document menu when an archived document is deleted.

### DIFF
--- a/app/menus/DocumentMenu.js
+++ b/app/menus/DocumentMenu.js
@@ -218,12 +218,7 @@ function DocumentMenu({
           items={[
             {
               title: t("Restore"),
-              visible: !!can.unarchive,
-              onClick: handleRestore,
-            },
-            {
-              title: t("Restore"),
-              visible: !!(collection && can.restore),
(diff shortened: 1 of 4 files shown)


iamsaumya:
There is another bug that will require one line of change. Should I create a separate PR or fix it here?

tommoor:
Feel free to add the fix on

iamsaumya:
In my DB, the parentDocumentId was not updating to null when we assigned with `undefined`

tommoor:
The test wasn't passing locally? 🤔

iamsaumya:
Ahh! I have been encountering this issue for the last week, I guess. Something is wrong with circleCI.

tommoor:
Sorry, I mean before you made this change was the test not passing with `undefined`? It should act the same.

This other test failure is sporadic on CI, it needs addressing :(

iamsaumya:
For that, I did not check at my localhost.

tommoor (approved):

PR #2195, opened 2021-06-08
feat: bambooHR integration

matbrgz:


matbrgz:
@tommoor could you please prioritize it? Already test and works well.

tommoor:
It seems like you're just iframing a website here, there's nothing embed-specific about it? Could you provide some screenshots of how it looks please.

matbrgz:
print tom
Yes this is a website, but the BambooHR recommends embed the website.

tommoor:
I'm inclined not to merge this one as it's just embedding the website with no special handling or styling.

Instead it seems like it's time to figure out a pattern for allowing any arbitrary website to be iframed, aka a "Frame" option in the block menu that works with any url. To achieve this we'd need a way in the Markdown of tagging a link as loading in an iframe – the title field could be abused for this purpose in the same way we're doing for image alignment 🤔

PR #2197, opened 2021-06-08
feat: twitter post integration

matbrgz:
- [x] Twitter Tweet
- [ ] Twitter HashTag Timeline


tommoor:
Closing this out as it's been in draft for a long time, twitter.com does not actually allow itself to be embedded in an iframe like this unfortunately – the embed will fail with a security error.

wvds:
> twitter.com does not actually allow itself to be embedded in an iframe

I actually got something working in [this codesandbox](https://codesandbox.io/s/affectionate-dew-izsmw?file=/src/index.js).
But when I try to apply the same in an Outline embed I have a problem with a request to Twitter that stays on the `pending` status, see screenshot:
Screenshot from 2021-08-25 00-32-18

@tommoor Maybe you've an idea why this is? Maybe a security restriction from the backend?

tommoor:
Hard to tell here but it’s likely to be the csp - policy for which third
party scripts can be loaded, on mobile or I would point you to the file

PR #2198, opened 2021-06-08
feat: google maps integration

matbrgz:
image


tommoor:
I'm inclined not to merge this one because it requires too much on behalf of the user. I think if a pasted Google Maps link worked it would be a different matter but the extra steps and having to copy a specific portion of an embed tag is just too finicky.

matbrgz:
I'm studying how to improve this using just the browser address.

PR #2210, opened 2021-06-11
feat: Dynamic site title and description

svallory:
With this PR it is possible to define SITE_NAME and SITE_DESCRIPTION env variables to override the default values.

I realized a second ago, when I was rebasing you guys just added `//inject-title//`, but since this was already done and also allow changing the description I thought I would send it anyways

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -136,3 +136,8 @@ SMTP_REPLY_EMAIL=
 # The default interface language. See translate.getoutline.com for a list of 
 # available language codes and their rough percentage translated.
 DEFAULT_LANGUAGE=en_US
+
+# Site name and description (optional)
+# these will be added to meta tags
+SITE_NAME=
+SITE_DESCRIPTION=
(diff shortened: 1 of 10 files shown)


tommoor:
Hi @svallory – thank you for taking the time to PR. Unfortunately this isn't a change we can accept right now, the description only really applies to search engines so it's not of important to self-hosting. The title change, as you mentioned, clashes with recent PR's in a way that is unmergeable.

I do think there's an opportunity to show the team name in the title automatically when self hosting but as this only applies to the login screen it's pretty low priority.

svallory:
@tommoor both the title and description appear when you share a link on whatsapp or twitter. Since I'm using it for a portuguese speaking audience it was very important for us to change it. I don't remember now how the dynamic title was initially implemented, but I guess it wasn't configurable, so you may want to reconsider that.

I rebased my branch on the latest changes so it would become mergeable. I can do that again if it's needed. I'm not trying to push for this PR to be accepted though. I'm justing raising some points that may have been missed.

tommoor:
Interesting, well the fact that the document title is included in the title now should solve for that a lot!

Issue #2175, opened 2021-06-04
Allow settings TLS Cipher for email configuration


tommoor:
See this discussion: https://github.com/outline/outline/discussions/2172

Some email servers require us to explicitly set the TLS mode used.

PR #2214, opened 2021-06-11
Use SMTP_SECURE environment variable to force secure parameter of smtp configuration

Nainterceptor:
Hello, 

Here is a small pull request to change strategy related to SMTP "secure" configuration, that currently rely on NODE_ENV.

I think it should be overrided by an environment variable, because SMTP configuration is not related to Outline configuration.

Best regards


tommoor:
related https://github.com/outline/outline/issues/2175

tommoor (approved):

tommoor:
Thanks!

PR #2215, opened 2021-06-11
Add Cloudron custom app support

svallory:
Not sure you've heard about [Cloudron](https://www.cloudron.io/), but it's a very easy way to deploy apps.

I've added in this PR all the configuration needed to package Outline as a Cloudron custom app so it can be installed. I've installed it on my server and it's running beautifully

I'm sending the PR because I thought it would be nice if this was officially supported. If you guys want to, the app can also be distributed via the Cloudron app store. If you guys don't wanna or can't take on that extra job we at Toki would love to take it, then all you'd need to do is tell people they can get a cloudron app at our repo

The custom app is already available at
https://hub.docker.com/repository/docker/tokilabs/outline-cloudron

### DIFF
--- a/CloudronManifest.json
+++ b/CloudronManifest.json
@@ -1,0 +1,21 @@
+{
+  "id": "com.tokilabs.outline",
+  "title": "Outline Knowledge Base",
+  "author": "Saulo Vallory <saulo@toki.life>",
+  "description": "file://cloudron/DESCRIPTION.md",
+  "tagline": "Your team’s knowledge base",
+  "version": "0.56.0",
+  "healthCheckPath": "/",
+  "httpPort": 3000,
+  "addons": {
+    "localstorage": {},
+    "redis": { "noPassword": true },
+    "postgresql": {}
+  },
+  "manifestVersion": 2,
+  "website": "https://www.getoutline.com",
+  "contactEmail": "saulo@toki.life",
+  "icon": "file://public/icon-1024.png",
+  "tags": [ "knowledge", "wiki", "team", "collaboration" ],
+  "mediaLinks": [ "https://user-images.githubusercontent.com/380914/110356468-26374600-7fef-11eb-9f6a-f2cc2c8c6590.png" ]
+}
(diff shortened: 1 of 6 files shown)


tommoor:
Hi @svallory – thanks for the PR, I think the main thing we don't want to support is yet another place to keep track of all the environment variables, it already feels like there is one too many and this adds another file for that.

I'd say you should keep it under tokilabs for now and perhaps in the future when Outline has more resources or if cloudron becomes more popular we can consider bringing into the main repository again.

svallory:
Hi @tommoor! Thanks for the answer.

There isn't actually another file where the environment variables are kept. Well, not exactly. If you are talking about [setup-app-env.sh](https://github.com/outline/outline/pull/2215/files#diff-cd532af47bc72060ab07eea88d109f7389f1ef92cfb30ccaf9e03e0f0633e3f8) that's a helper to setup the variables in the actual environment.

But I see your point. I can make that file import the values from the `.env` file and then set them in the actual app environment. The reason I had to do that is that Cloudron doesn't allow setting custom startup parameters for running the container, so the variables need to be set in the environment.

I don't mind keeping it at tokilabs. I'll implement your suggestion either way. But then I have two questions:

1. Would you merge if I keep all variable setting in the `.env`?
2. If not, am I authorized to publish the app from tokilabs repository to the Cloudron store? (I'm happy to pass control should you guys ever decide to take over)

svallory:
@tommoor I talked to the guys at Cloudron and I have a solution to the variables configuration issue. There's a default '/app/data' folder where a user can manually edit files so the only change would be that the `.env` would live there instead of the project root. Is that acceptable?

tommoor:
I don't think it would be acceptable to move the .env to a nested folder like that for everyone, however we could create a symbolic link in `/app/data` to the `.env` in the root?

svallory:
The root of the issue is that having the `.env` inside the image prevents the same image from being deployed in different environments. In Cloudron, `/app/data/` is a commonly mounted volume (you could mount any path though). In fact, we could create a script that gets the path to the `.env` file from an environment variable with `/app/data/.env` being the default.

As a side note, I think some of the settings in `.env` may pose a security risk if the file is embedded in the docker image.

Issue #1267, opened 2020-05-13
issue with google signin getting this error Invalid key length


prakash9680:
after google sign in it redirect to this error.
Error: Invalid key length
    at Cipheriv.createCipherBase (internal/crypto/cipher.js:103:19)
    at Cipheriv.createCipherWithIV (internal/crypto/cipher.js:121:20)
    at new Cipheriv (internal/crypto/cipher.js:225:22)
    at Object.createCipheriv (crypto.js:117:10)
    at model.set (/var/www/outline/node_modules/sequelize-encrypted/index.js:41:33)
    at model.set (/var/www/outline/node_modules/sequelize/lib/model.js:3655:32)
    at model.funcs.<computed> (/var/www/outline/node_modules/sequelize/lib/model.js:1103:25)
    at Function.setRandomJwtSecret (/var/www/outline/server/models/User.js:151:3)
    at /var/www/outline/node_modules/sequelize/lib/hooks.js:130:19
    at tryCatcher (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/util.js:16:23)
    at Object.gotValue (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/reduce.js:166:18)
    at Object.gotAccum (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/reduce.js:155:25)
    at Object.tryCatcher (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/promise.js:547:31)
    at Promise._settlePromise (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/promise.js:604:18)
    at Promise._settlePromise0 (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/promise.js:649:10)
    at Promise._settlePromises (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/promise.js:729:18)
    at _drainQueueStep (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/async.js:93:12)
    at _drainQueue (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/async.js:86:9)
    at Async._drainQueues (/var/www/outline/node_modules/sequelize/node_modules/bluebird/js/release/async.js:102:5)
[...]

PR #2229, opened 2021-06-16
Fix "secret key" variable description

g-psantos:
Updated the description of the `SECRET_KEY` variable in the `.env.sample` file to clarify that the key needs to be 32 bytes long and hex-encoded and make the suggested OpenSSL command slightly more prominent.

The previous description of "32 character hexadecimal" was confusing as it left open the possibility of a hex-encoded 16-byte key (see #1267 -- and I made the same mistake a year later).

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -8,8 +8,7 @@
 
 # –––––––––––––––– REQUIRED ––––––––––––––––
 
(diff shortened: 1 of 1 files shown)


tommoor:
Works for me, thanks 👍

PR #2252, opened 2021-06-25
fix: Email auth should allow same guest user on multiple subdomains

### LINKED ISSUE #2249: Email auth does not prioritize current subdomain
**To Reproduce**
1. Have multiple accounts with the same email
2. Go to one subdomain
3. Use the email
4. Forwarded to SSO for other subdomain

**Expected behavior**

- Current subdomain should be prioritized

tommoor:
It is possible to invite, say, a contractor `tom@example.com` to two teams – `one.getoutline.com` and `two.getoutline.com`, prior to these changes it was not possible to signin to the second created team via email, even when doing so via the `two.getoutline.com` subdomain.

Now the current domain or subdomain is given precedence.

closes #2249

### DIFF
--- a/server/migrations/20170712055148-non-unique-email.js
+++ b/server/migrations/20170712055148-non-unique-email.js
@@ -1,15 +1,7 @@
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    await queryInterface.changeColumn('users', 'email', {
-      type: Sequelize.STRING,
-      unique: false,
-      allowNull: false,
-    });
-    await queryInterface.changeColumn('users', 'username', {
-      type: Sequelize.STRING,
-      unique: false,
-      allowNull: false,
-    });
+    await queryInterface.removeConstraint('users', 'users_email_key', {})
+    await queryInterface.removeConstraint('users', 'users_username_key', {})
   },
 
   down: async (queryInterface, Sequelize) => {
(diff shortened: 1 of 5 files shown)


server/migrations/20170712055148-non-unique-email.js
```diff
@@ -1,15 +1,7 @@
...
-      type: Sequelize.STRING,
-      unique: false,
-      allowNull: false,
-    });
-    await queryInterface.changeColumn('users', 'username', {
-      type: Sequelize.STRING,
-      unique: false,
-      allowNull: false,
-    });
+    await queryInterface.removeConstraint('users', 'users_email_key', {})
```
tommoor, line 3: Fixes email column indexed as unique in tests.

server/test/support.js
```diff
@@ -3,17 +3,15 @@ import { v4 as uuidv4 } from "uuid";
...
 
-export function flushdb() {
-  const sql = sequelize.getQueryInterface();
-  const tables = Object.keys(sequelize.models).map((model) => {
-    const n = sequelize.models[model].getTableName();
-    return sql.queryGenerator.quoteTable(
-      typeof n === "string" ? n : n.tableName
-    );
-  });
+const sql = sequelize.getQueryInterface();
```
tommoor, line 6: Just trying to make tests run slightly faster by not doing these extra steps for every test run. Probably negligible compared to the actual truncation step though.

PR #2254, opened 2021-06-26
fix: Data loading state not reset when props change to `PaginatedList`

### LINKED ISSUE #2251: Document data loading offset not reset when navigating between collections
**To Reproduce**
1. Reload app
1. Go to a collection with more than 25 documents
2. Scroll down to second page
3. Navigate to a second collection

**Expected behavior**

- Documents from second collection load, but they don't.

tommoor:
closes #2251


tommoor:
@iamsaumya @thenanyu first enzyme/ui tests 💪

PR #2272, opened 2021-07-04
fix: nested link positions for RTL titles in sidebar

fshafiee:
Hey, @tommoor.

I wanted to patch outline regarding RTL titles and sidebars, then I noticed you already worked on it. Thanks!
However, I found some small issues with the side bar regarding nested positions and edit mode, which I tried to address in this PR.

### Before the fix
1. Nested documents appear flat.
before-read-only
2. Wrong input direction during edit.
before-editable

### With the fix
1. Even though the text are aligned to left, the direction is correct and nesting is preserved. Since direction of app (regardless of the content) is LTR, this is acceptable even if texts are RTL.
after-read-only

2. Input direction is correct during edit.
after-editable


tommoor:
Nice, thanks for catching that

PR #2275, opened 2021-07-05
feat: translations

falleng0d:
It is now possible to translate document creation placeholder, templates and some warnings that were not previously covered

This was made possible by i18n wrapper:

`export default withTranslation()<DocumentEditor>(DocumentEditor);`


falleng0d:
This is my first contribution. And I plan to keep doing more as we are going to use this internally. If moderators have any feedback to how I should proceed, I'm all ears

package.json
```diff
@@ -11,6 +11,7 @@
     "build": "yarn clean && yarn build:webpack && yarn build:i18n && yarn build:server",
     "start": "node ./build/server/index.js",
     "dev": "nodemon --exec \"yarn build:server && yarn build:i18n && node --inspect=0.0.0.0 build/server/index.js\" -e js --ignore build/ --ignore app/",
+    "dev:essential": "nodemon --exec \"yarn build:server && node --inspect=0.0.0.0 build/server/index.js\" -e js --ignore build/ --ignore app/",
```
tommoor, line 14: Please remove unrelated change

app/scenes/Document/components/Document.js
```diff
@@ -383,28 +381,36 @@ class DocumentScene extends React.Component<Props> {
...
                 </Notice>
               )}
               {document.archivedAt && !document.deletedAt && (
                 <Notice muted>
-                  Archived by {document.updatedBy.name}{" "}
-                  <Time dateTime={document.archivedAt} /> ago
+                  <Trans>
+                    Archived by {document.updatedBy.name}{" "}
+                    <Time dateTime={document.archivedAt} /> ago
+                  </Trans>
```
tommoor, line 397: For the translation framework to correctly interpolate dynamic values they must be in the format of `{{ thing }}` (with spaces for consistency). This comment applies to several of the changes in this file – see here as a working example:

https://github.com/outline/outline/blob/e845652cb8f9295c13bfe1c2d2324654b9b3dd77/app/menus/NewChildDocumentMenu.js#L34-L38

I think something like this would work:

```suggestion
                  <Trans
                    defaults="Archived by {{ userName }} <time> ago"
                    values={{ userName: document.updatedBy.name }}
                    components={{ time: <Time dateTime={document.archivedAt} /> }}
                  />
```

.gitignore
```diff
@@ -8,3 +8,4 @@ stats.json
 .DS_Store
 fakes3/*
 .idea
+.dev.dockerfile
```
tommoor, line 11: Please remove unrelated change

tommoor (changes requested):
This is great, thanks – and congrats on your first GitHub pull request too!

I have some suggested changes to enable merging

falleng0d:
I made the corrections as requested. 

About the reverted changes:
- chore: Ignored local development .dev.dockerfile files
- chore: Add faster build command that ignores translation file

These and other local changes I have made could prove useful to improving local development experience and making it easier to get started and productive. What are you thoughts on me sharing them in a separete PR?

tommoor:
**- chore: Ignored local development .dev.dockerfile files**
Considering dev.dockerfile isn't something that's in the codebase it doesn't make sense to include an ignore for it. The main dockerfile is considered to be for development.

**chore: Add faster build command that ignores translation file**
The full build is ran relatively rarely, most development is done with hot-reload on the clientside so I think having this extra command to save a second or so is confusing.

You're welcome to keep these changes on your local branch but they don't seem to be needed in the main project

tommoor:
Oh, it seems like the translation framework does not like items in the `components` array having attributes so this isn't working as I'd hoped, the timestamps are missing:
https://react.i18next.com/latest/trans-component#using-for-less-than-br-greater-than-and-other-simple-html-elements-in-translations-v-10-4-0

I think the way to deal with this is to separate the translated text from the timestamp and add the `addSuffix` prop as it is done here: 
https://github.com/outline/outline/commit/1285efc49a985bfe3f878e5030ea5d2240c56e41 – this should work for all except the last case, but for that one luckily there is no need to actual interpolate values (see comment)

app/scenes/Document/components/Document.js
```diff
@@ -383,28 +381,49 @@ class DocumentScene extends React.Component<Props> {
...
+                        values={{
+                          userName: document.updatedBy.name,
+                          documentName: document.noun,
+                        }}
+                        components={{
+                          time: (
+                            <Time dateTime={document.permanentlyDeletedAt} />
+                          ),
+                        }}
+                      />
```
tommoor, line 426: {document.template ? (
                        <Trans>
                          This template will be permanently deleted in Time
                          dateTime={document.permanentlyDeletedAt} /> unless
                          restored.
                        </Trans>
                      ) : (
                        <Trans>
                          This document will be permanently deleted in Time
                          dateTime={document.permanentlyDeletedAt} /> unless
                          restored.
                        </Trans>
                      )}

tommoor:
Let me know if you want to take this over the finish line?

falleng0d:
I will be working on it tomorrow. Sorry for the delay.

falleng0d:
ok. I believe the last change fixed it. 

Screenshot271
Screenshot272

tommoor (approved):
Great 👏

PR #2279, opened 2021-07-06
fix: Add access to document TOC on mobile

### LINKED ISSUE #2145: Add access to document TOC on mobile
Customer request through support, makes sense!

We can add the TOC button next to the menu button and have it trigger a context menu version of the table of contents on mobile which will slide up from the bottom.

iamsaumya:
Closes #2145


app/scenes/Document/components/Header.js
```diff
@@ -274,4 +282,19 @@ const Status = styled(Action)`
...
 
+const TocWrapper = styled(Action)`
+  padding-left: 0;
+  padding-right: 4px;
+  color: ${(props) => props.theme.slate};
+  position: fixed;
+  margin-left: 8px;
+  top: 12px;
+  left: 46px;
+  z-index: ${(props) => props.theme.depths.sidebar - 1};
```
tommoor, line 293: On mobile rn so I can't check - (screenshot?) but all of these positioning styles should be covered by using the context menu

tommoor (changes requested):

app/scenes/Document/components/Header.js
iamsaumya: Here is the screenshot. Please take a look.

Screen Shot 2021-07-08 at 09 09 06

> but all of these positioning styles should be covered by using the context menu

I couldn't do it. Let me know how I should go about it.

tommoor: Looks like if you move it inside `Header.actions` it becomes easier to position – can absolute position then:

```
position: absolute;
left: 52px;
```

tommoor: 99% sure it's unrelated, but do you see horizontal scrolling on docs on mobile when testing this?

iamsaumya: > Looks like if you move it inside `Header.actions` it becomes easier to position – can absolute position then:
> 
> ```
> position: absolute;
> left: 52px;
> ```

Worked like a charm. 

> you see horizontal scrolling on docs on mobile when testing this?

No, I didn't. Sometimes it was there when I was manually resizing the window, but a hard refresh fixes everything, and I couldn't find any overflowing on the predefined mobile resolutions of browsers.

app/components/ContextMenu/Template.js
```diff
@@ -130,7 +130,7 @@ function Template({ items, ...menu }: Props): React.Node {
           disabled={item.disabled}
           selected={item.selected}
           level={item.level}
-          target="_blank"
```
tommoor, line 131: Oh yes, I meant to mention this – I don't think this is the right coupling though, It could be possible to use the `level` prop and want an external link. The correct coupling would be to check if `item.href.startsWith("#")`

iamsaumya: Nice catch! I should have given more thought.

app/scenes/Document/components/Header.js
```diff
@@ -274,4 +282,9 @@ const Status = styled(Action)`
   color: ${(props) => props.theme.slate};
 `;
 
+const TocWrapper = styled(Action)`
+  position: absolute;
+  left: 52px;
```
tommoor, line 287:
```suggestion
  left: 42px;
```

With the existing padding on the left of those actions this actually needs reducing a bit

app/menus/TableOfContentsMenu.js
```diff
@@ -0,0 +1,60 @@
...
+            iconColor="currentColor"
+            borderOnHover
+            neutral
+          />
+        )}
+      </MenuButton>
+      <ContextMenu {...menu} aria-label={t("Table of contents")}>
+        <Template
+          {...menu}
+          items={headings.map((heading) => {
```
tommoor, line 47: One last thing – can we add a "Table of contents" header to this menu, it's not immediately obvious what it represents especially if you're not familiar with the document content. There's a menu item type for this:

https://github.com/outline/outline/blob/f4959f98eb94713aa2adf190a0d56b57ee7a3dd3/app/components/ContextMenu/Template.js#L52-L54

tommoor:
great stuff, thank you

app/menus/TableOfContentsMenu.js
iamsaumya: Yes Yes 😄

iamsaumya: It seems like this heading case is not handled in the code.

iamsaumya: This is how it looks now 

Screen Shot 2021-07-09 at 13 49 22

tommoor (approved):

PR #2283, opened 2021-07-07
fix: Make search query string user friendly

### LINKED ISSUE #2274: Make search query string user friendly
The query is included in the URL when its value is an empty string

**To Reproduce**
1. Go to Search
2. Search for any keyword
3. Change the "Active documents" filter and change it again to default
4.  Find the URL to have ```includeArchived=```

**Expected behavior**
If a query string has an empty string, do not include it in the URL.

iamsaumya:
Closes #2274

### DIFF
--- a/app/scenes/Drafts.js
+++ b/app/scenes/Drafts.js
@@ -50,10 +50,13 @@ class Drafts extends React.Component<Props> {
   }) => {
     this.props.history.replace({
       pathname: this.props.location.pathname,
-      search: queryString.stringify({
-        ...queryString.parse(this.props.location.search),
-        ...search,
-      }),
+      search: queryString.stringify(
+        {
+          ...queryString.parse(this.props.location.search),
+          ...search,
+        },
+        { skipEmptyString: true }
+      ),
     });
   };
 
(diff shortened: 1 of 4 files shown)


app/scenes/Drafts.js
```diff
@@ -50,10 +50,13 @@ class Drafts extends React.Component<Props> {
...
-      search: queryString.stringify({
-        ...queryString.parse(this.props.location.search),
-        ...search,
-      }),
+      search: queryString.stringify(
+        {
+          ...queryString.parse(this.props.location.search),
+          ...search,
+        },
+        { skipEmptyString: true }
```
tommoor, line 58: Nice, great catch on the option being available with an upgrade 👍

yarn.lock
```diff
@@ -10765,6 +10757,16 @@ query-string@^7.0.0:
     split-on-first "^1.0.0"
     strict-uri-encode "^2.0.0"
 
+query-string@^7.0.1:
```
tommoor, line 10760: So, `yarn` has this really annoying bug where it sometimes includes multiple versions of a dependency unnecessarily – see how here we have  `7.0.0` and `7.0.1` included in the bundle where just `7.0.1` would be fine as they have the same interface?

Thankfully there is a utility to fix this, I should automate it running after every yarn install to be honest – run this locally on the branch:

`npx yarn-deduplicate yarn.lock`

iamsaumya: Done

tommoor (approved):

Issue #2230, opened 2021-06-16
Document history sidebar should show move events


tommoor:
On more than one occasion it has confused folks that the history sidebar does not show the full history of a document, only snapshots.

It would be useful to show document move events and perhaps others in this sidebar, eg when it was shared publicly. Querying the events table might get everything we need here.

tommoor:
Currently we don't record revision creation in the events table, so this will need to change to be able to query the entire document history from the single table and paginate it correctly.

- [ ] Add recording of `revisions.create` events with revision creation
- [ ] Add `revisions.create` to those accessible through `events.list`
- [ ] Add script to backfill existing revisions with creation events
- [ ] Drive history sidebar from events table (branch started: https://github.com/outline/outline/tree/feat/issue-2230)

tommoor:
Rough sketch of timeline-style…

![image](https://user-images.githubusercontent.com/380914/125971303-cb2fbdfc-c756-4129-b277-5ae0fbfdd07b.png)

tommoor:
WIP: 

![image](https://user-images.githubusercontent.com/380914/125998115-b642eb34-1b9e-4b78-969e-744cddebb892.png)

PR #2287, opened 2021-07-07
feat: Add `documentId` filter to `events.list`

tommoor:
This is one of several enablers for https://github.com/outline/outline/issues/2230 so we can drive the history panel from the events API.

### DIFF
--- a/server/api/events.js
+++ b/server/api/events.js
@@ -16,6 +16,7 @@ router.post("events.list", auth(), pagination(), async (ctx) => {
   let {
     sort = "createdAt",
     actorId,
+    documentId,
     collectionId,
     direction,
     name,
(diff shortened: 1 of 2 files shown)


server/api/events.js
```diff
@@ -31,10 +32,12 @@ router.post("events.list", auth(), pagination(), async (ctx) => {
...
-    where = {
-      ...where,
-      actorId,
-    };
+    where = { ...where, actorId };
+  }
+
+  if (documentId) {
+    ctx.assertUuid(documentId, "documentId must be a UUID");
+    where = { ...where, documentId };
```
tommoor, line 40: No need to check permissions here as we're already filtering all results by accessible collections below

PR #2291, opened 2021-07-07
feat: Add Google Forms embed

matbrgz:


app/embeds/GoogleForms.js
```diff
@@ -0,0 +1,44 @@
...
+
+type Props = {|
+  attrs: {|
+    href: string,
+    matches: string[],
+  |},
+|};
+
+export default class GoogleForms extends React.Component<Props> {
+  static ENABLED = [URL_REGEX] || [URL_REGEX_MINIFIED];
```
tommoor, line 19: Similar comment to the other PR here, this syntax doesn't make sense – the second case will never be hit.

app/embeds/GoogleForms.js
```diff
@@ -0,0 +1,44 @@
...
+
+  render() {
+    return (
+      <Frame
+        {...this.props}
+        src={
+          URL_REGEX
```
tommoor, line 26: Second case in ternary will never be evaluated, the `URL_REGEX` always exists

app/embeds/index.js
```diff
@@ -148,6 +149,13 @@ export default [
     component: GoogleSlides,
     matcher: matcher(GoogleSlides),
   },
+  {
+    title: "Google Forms",
+    keywords: "survey",
+    icon: () => <Img src="/images/google-forms.png" />,
```
tommoor, line 155: Comment for icon itself – can you add a version that goes all the way to the top and bottom edges otherwise it will look smaller than the rest in the embed menu.

tommoor (changes requested):

matbrgz:
@iamsaumya could you please fix this implementation?

samuelbirocchi:
What needs to be done? I could take over this task.

matbrgz:
> What needs to be done? I could take over this task.

We need to fix the RegEx, they never catch the Forms.

PR #2292, opened 2021-07-07
feat: Add Google Calendar Embed

matbrgz:

PR #2300, opened 2021-07-09
feat: Add Google Calendar Embed

matbrgz:
Replaces #2292

@tommoor Now you can only use embed.


app/embeds/index.js
```diff
@@ -148,6 +149,13 @@ export default [
     component: GoogleSlides,
     matcher: matcher(GoogleSlides),
   },
+  {
+    title: "Google Calendar",
+    keywords: "calendar",
+    icon: () => <Img src="/images/google-calendar.png" />,
```
tommoor, line 155: The icon needs to go all the way to the edge of the image or it will appear the wrong size in the UI, see the existing icons as examples:

https://github.com/outline/outline/blob/main/public/images/google-docs.png

tommoor (changes requested):
Looks good apart from the icon 👍

PR #2303, opened 2021-07-09
fix: Add translation hooks on groups screen

### LINKED ISSUE #2117: Add translation hooks where missing
There are quite a few spots still where English strings are hardcoded, these need moving to use `react-i18n` and (ideally but not required) functional components. A lot of the remaining spots are in confirmation dialogs, modals, toast messages etc.

Fixing several screens at a time and doing multiple smaller PR's seems like the best way to get these tackled in a manageable manner. These are the components I'm currently aware:

- [x] `ErrorBoundary`
- [x] `InputRich`
- [x] `CollectionDelete`
- [x] `CollectionExport`
- [x] `DocumentTemplatize`
- [x] `DocumentMove`
- [x] `DocumentNew`
- [x] `GroupDelete`
- [x] `GroupEdit`
- [x] `GroupNew`
- [x] `GroupMembers`
- [x] `Invite`
- [x] `UserDelete` 
- [x] `Settings/Details`
- [x] `Settings/Notifications`
- [x] `Settings/Security`
- [x] `Settings/Tokens`
- [x] `Settings/Zapier`

iamsaumya:
towards #2117


app/scenes/GroupDelete.js
```diff
@@ -1,59 +1,57 @@
...
-
-    return (
-      <Flex column>
-        <form onSubmit={this.handleSubmit}>
-          <HelpText>
+  return (
+    <Flex column>
+      <form onSubmit={handleSubmit}>
+        <HelpText>
+          <Trans>
```
tommoor, line 43: Where there is interpolation in the translation string please use this style for the <Trans> api:

https://github.com/outline/outline/blob/a134773d4e79b2b1f9126c3a13e72479335c5d86/app/scenes/CollectionPermissions/index.js#L167-L171

The interpolation should be in the style `{{thingName}}` for the translation framework to correctly recognize as a dynamic value.

app/scenes/GroupDelete.js
```diff
@@ -1,59 +1,57 @@
...
-            {this.isDeleting ? "Deleting…" : "I’m sure – Delete"}
-          </Button>
-        </form>
-      </Flex>
-    );
-  }
+          </Trans>
+        </HelpText>
+        <Button type="submit" danger>
+          {isDeleting ? `${t("Deleting")}...` : t("I’m sure – Delete")}
```
tommoor, line 50:
```suggestion
          {isDeleting ? `${t("Deleting")}…` : t("I’m sure – Delete")}
```

Real ellipsis pls 😉

app/scenes/GroupEdit.js
```diff
@@ -1,70 +1,71 @@
...
+      </Flex>
 
-        <Button type="submit" disabled={this.isSaving || !this.name}>
-          {this.isSaving ? "Saving…" : "Save"}
-        </Button>
-      </form>
-    );
-  }
+      <Button type="submit" disabled={isSaving || !name}>
+        {isSaving ? `${t("Saving")}...` : t("Save")}
```
tommoor, line 65:
```suggestion
        {isSaving ? `${t("Saving")}…` : t("Save")}
```

app/scenes/GroupMembers/GroupMembers.js
```diff
@@ -20,112 +14,101 @@ import PaginatedList from "components/PaginatedList";
...
+  const [addModalOpen, setAddModalOpen] = React.useState();
+  const { users, groupMemberships, policies, ui } = useStores();
+  const { t } = useTranslation();
+  const can = policies.abilities(group.id);
 
-  handleAddModalOpen = () => {
-    this.addModalOpen = true;
-  };
+  // validate user is not null
+  useCurrentUser();
```
tommoor, line 31:
```suggestion
```

This line alone doesn't do anything unless you use the return value?

app/scenes/GroupMembers/GroupMembers.js
```diff
@@ -20,112 +14,101 @@ import PaginatedList from "components/PaginatedList";
...
-      <Flex column>
-        {can.update ? (
-          <>
-            <HelpText>
+  return (
+    <Flex column>
+      {can.update ? (
+        <>
+          <HelpText>
+            <Trans>
```
tommoor, line 57: Same comment on `Trans` usage

app/scenes/GroupMembers/GroupMembers.js
```diff
@@ -20,112 +14,101 @@ import PaginatedList from "components/PaginatedList";
...
+              icon={<PlusIcon />}
+              neutral
+            >
+              {t("Add people")}…
+            </Button>
+          </span>
+        </>
+      ) : (
+        <HelpText>
+          <Trans>
```
tommoor, line 76: Same comment on `Trans` usage

app/scenes/GroupNew.js
```diff
@@ -1,90 +1,90 @@
...
-          onRequestClose={this.props.onSubmit}
-          isOpen={!!this.group}
-        >
-          <GroupMembers group={this.group} />
-        </Modal>
-      </>
-    );
-  }
+        <Button type="submit" disabled={isSaving || !name}>
+          {isSaving ? `${t("Creating")}...` : t("Continue")}
```
tommoor, line 76:
```suggestion
          {isSaving ? `${t("Creating")}…` : t("Continue")}
```

tommoor (changes requested):

app/scenes/GroupMembers/GroupMembers.js
iamsaumya: There was this guard condition `if (!user) return null;`. It was the only use of `user` in the code. I thought there must be some bug or validation requirement, thus invoking this hook to ensure that the user is present.

tommoor: I think you're safe to remove as it's unused, must have been a copy/paste error

app/scenes/GroupNew.js
```diff
@@ -1,90 +1,90 @@
...
 import Flex from "components/Flex";
 import HelpText from "components/HelpText";
 import Input from "components/Input";
 import Modal from "components/Modal";
+import useStores from "hooks/useStores";
 
 type Props = {
-  history: RouterHistory,
-  ui: UiStore,
   groups: GroupsStore,
```
tommoor, line 16:
```suggestion
```

app/scenes/GroupNew.js
```diff
@@ -1,90 +1,90 @@
...
   onSubmit: () => void,
 };
 
-@observer
-class GroupNew extends React.Component<Props> {
-  @observable name: string = "";
-  @observable isSaving: boolean;
-  @observable group: Group;
+function GroupNew({ groups, onSubmit }: Props) {
+  const { ui } = useStores();
```
tommoor, line 21:
```suggestion
  const { ui, groups } = useStores();
```

tommoor: The functionality of this component is currently broken as groups is undefined

tommoor (changes requested):

app/scenes/GroupNew.js
iamsaumya: Thanks! I feel so STUPID.

tommoor: This is exactly the kind of thing that better typing should catch

shared/i18n/locales/en_US/translation.json
```diff
@@ -326,11 +327,20 @@
...
+  "You can edit the name of this group at any time, however doing so too often might confuse your team mates.": "You can edit the name of this group at any time, however doing so too often might confuse your team mates.",
   "{{userName}} was added to the group": "{{userName}} was added to the group",
   "Add team members below to give them access to the group. Need to add someone who’s not yet on the team yet?": "Add team members below to give them access to the group. Need to add someone who’s not yet on the team yet?",
   "Invite them to {{teamName}}": "Invite them to {{teamName}}",
   "{{userName}} was removed from the group": "{{userName}} was removed from the group",
+  "Add and remove team members in the <em>{{groupName}}</em>\n              group. Adding people to the group will give them access to any\n              collections this group has been added to.": "Add and remove team members in the <em>{{groupName}}</em>\n              group. Adding people to the group will give them access to any\n              collections this group has been added to.",
```
tommoor, line 336: One last thing, could we remove these newlines? While they are okay in the translation strings when necessary, we should avoid them when they have no semantic value and are just an artifact of the code.

iamsaumya: I will definitely. It has newlines because the length of the line in the file gets too long otherwise.

tommoor (approved):

PR #2309, opened 2021-07-11
feat: Personal collection

tommoor:
Provides a private space to write docs that is not visible to other team members by default. Another spike that was sat on a branch and could be completed


wvds:
@tommoor Is this something that is still being considered? Currently we're using drafts to have somewhat personal notes, but a private space for person would be a nice addition!

tommoor:
Yea, would still love to have something like this

JezzRDZ:
@tommoor is this feature still on the table?

tommoor:
Yea, we should still do this

PR #2317, opened 2021-07-13
fix: Show starred docs in sidebar

### LINKED ISSUE #1313: Show starred docs in sidebar
Currently the treatment for starred is the same as other docs, they basically just get added to another screen that acts the same as any other collection . While this is _okay_ it's barely better than exposing the database schema to the user – starred docs should be **easier to access** than other docs.

### Proposed changes: 

- Remove the existing "Starred" item at the top of the sidebar
- Add a "Starred" heading to the sidebar when the user has starred documents above collections
- Render upto X (10?) documents below it in a flat list and a "Show more…" link at the end
- The "show more" link should navigate to the existing starred page for when there are many starred items to filter through (Use similar styling as the e.g. "New collection")

iamsaumya:
Closes #1313 

- [x] Change starred to bookmarked on user-facing instances. 
- [x] Remove starred scene


iamsaumya:
@tommoor, please take a look.

tommoor:
Thank you – leave this with me for a few, having seen it in action might need to rethink a little 😆

iamsaumya:
Haha, Sure 👍🏻

tommoor:
Okay, here's what I think is needed to make this workable – I'd be interested in your feedback…

- "Show more…" should only appear when user has an exceptionally large amount of starred items, let's say >10.
- Given the above the `/starred` route becomes quite redundant and rarely used, I'm thinking we can convert this to a filter on the search so you can filter to starred docs. This could be tackled in a separate upfront PR.
- Repeating the ⭐️ icon in the sidebar is a bit much, It might be best to leave this space empty, or show a simple bullet point (will need to do some design explorations here).
- When a starred doc is clicked it should be highlighted in the starred section too
- If a starred doc has child documents it should be possible to expand within the starred section

iamsaumya:
> Given the above, the /starred route becomes quite redundant and rarely used. I'm thinking we can convert this to a filter on the search so you can filter to starred docs. This could be tackled in a separate upfront PR.

Given how the search works, we need to pass some queries to it with the filters. It will be weird to show docs without any queries. 

>Repeating the star icon in the sidebar is a bit much, It might be best to leave this space empty, or show a simple bullet point (will need to do some design explorations here).

A suggestion - Given that we will only show at most 10 docs, Is it possible to have some sort of color binding for each star? It will look lively in the sidebar.

tommoor:
> Is it possible to have some sort of color binding for each star?

I'm having trouble imagining this working well, if the colors are "random" then they have no meaning and if they're user choosable then that's a lot to ask of folks

iamsaumya:
Random in a sense, we will have a list of 10 colors, the first starred doc will always have that specific color, and so on

Or maybe we can pick the color from their collection's icon?

iamsaumya:
So I guess we will go without any icons. 

Any comments for the suggestions on the approach to the `search` page?

tommoor:
> Any comments for the suggestion on the approach to the search page?

I think it could make sense to add a "Starred documents" to this filter and expose it via a query string parameter, thoughts?

image

iamsaumya:
Adding a filter for `starred documents` is a good idea indeed. 

So when a user clicks on `show more...`, should they be redirected to the search page with the filter selected? 

If so, I am hesitant about the idea to show the `starred documents` without any `search query` on the search page as it feels like an anti-pattern.

tommoor:
Fair, I suppose I didn't think that one through thoroughly enough 😆 

In that case perhaps "Show more" just expands the display in the sidebar to load + show all of the users starred items

iamsaumya:
Sounds great! 🤜🏻🤛🏻

iamsaumya:
I made some progress here.

Screenshot from 2021-07-23 17-27-45

iamsaumya:
I will complete the above todos when we are happy with this implementation.

app/components/Sidebar/components/SidebarLink.js
```diff
@@ -156,20 +157,18 @@ const Link = styled(NavLink)`
...
+  }
 
-    &:hover {
-      color: ${(props) =>
-        props.$isActiveDrop ? props.theme.white : props.theme.text};
-    }
+  &:hover {
+    color: ${(props) =>
+      props.$isActiveDrop ? props.theme.white : props.theme.text};
   }
```
iamsaumya, line 172: Apparently, we had an extra curly bracket here.

tommoor: I think you just accidentally moved it above?

iamsaumya: Oh, so `&:hover` selection is inside the media query?

iamsaumya:
@tommoor, I think this is done. Please give it a review 😄

iamsaumya:
Hey, just bumping up to your notifications, in case you have forget about this 😅

tommoor:
Haven't forgotten – I think the functionality is basically spot-on, just pondering on how these are aligned/displayed still, there are three ways to align these items and each one looks out of alignment in a different way depending on if there is a disclosure or not 😢 

thinking through that…

iamsaumya:
I vouch for this.

I tried different designs and chose this implementation in the end, which looked best to me. 

Still, I feel that something isn't right about the presentation.

tommoor:
The problem as I'm sure you saw is that documents at the same level in hierarchy are not aligned horizontally because of the disclosure, but if you _do_ align them it leaves a gap that looks strange, for example:

Screen Shot 2021-07-26 at 3 16 42 PM

You could also push the disclosure into the gutter, but this looks wrong because the titles kind of _want_ to be aligned with all the other text:

Screen Shot 2021-07-26 at 3 17 50 PM

iamsaumya:
Yupp, that's why I aligned everything in a line, it looked the best of all 😅

tommoor:
telegram-cloud-photo-size-1-4904638037267425657-y

Possible way forward, discussed in chat.

iamsaumya:
LGTM 👍🏻

iamsaumya:
Do we want the bookmark section to be collapsible?

tommoor:
Would be nice – show a disclosure on hover of heading?

iamsaumya:
And when it is closed, should the disclosure be visible?

tommoor:
Yes!

iamsaumya:
Done ✅

tommoor:
We spoke privately on this and this design ended up being the front-runner 😅 

telegram-cloud-photo-size-1-4907198000984598886-y

iamsaumya:
Please review 😃

tommoor (changes requested):
Thanks! Some feedback – upon opening the branch my sidebar looks like this?

image

- [x] Everything should be expanded by default
- [x] The headings should use the same style as other items (see design above), perhaps we could make the entire thing clickable?
- [x] Lets keep the disclosure visible at all times with this design
- [x] "New collection…" item should be inset the same amount as collections

iamsaumya:
`depth` prop is one of the best implementations I have ever come across. No need to worry about `px`. Just pass an appropriate number 😍

tommoor:
Sorry, found a few more things – but after playing with it I'm confident this layout will work 🙏 

- [ ] If there are no starred docs then hide the heading, currently looks like:
      image

- [ ]  The top "drop cursor" in the collections section is positioned too high up, try dragging a collection to the top to see. This also makes the bottom half of the "Collections" header unclickable.
- [ ] Is it possible to inset the loading state under "starred" to match the content?
- [ ] Is it possible to include the "..." menu on starred items in the sidebar the same as in the regular collection?

iamsaumya:
> If there are no starred docs then hide the heading, currently looks like

Sure. Actually, I deliberately took the decision to show it. I thought users should be aware that this estate is where starred documents will appear in future.

tommoor:
>  I deliberately took the decision to show it. I thought users should be aware that this estate is where starred documents will appear

I can see the logic, but it's kind of 100px of prime real estate pushing down the main content :) Thankfully we have lots of ways to star a doc so it should be discoverable, it can also be in a future onboarding if needed.

app/components/Sidebar/components/StarredLink.js
```diff
@@ -88,6 +88,7 @@ function StarredLink({ depth, title, to, documentId, collectionId }: Props) {
                 document={document}
                 onOpen={handleMenuOpen}
                 onClose={handleMenuClose}
+                modal={false}
```
iamsaumya, line 91: Sometimes the menu is placed in the top left corner. This seems to remove it.

iamsaumya:
Ready for review 😄

tommoor:
Did a fair few final bits of polish here, lmk if anything stands out as problematic – otherwise I think this should be good to merge!

iamsaumya:
Thanks for the help here 😄 . Looks good to me!

tommoor (approved):
🙏  I do hope that this change is well received

iamsaumya:
It will be fun to have feedback from someone who has starred 100s of docs 😁

PR #2318, opened 2021-07-14
fix: Archive and trash a document by dropping in the sidebar

### LINKED ISSUE #2299: Archive and trash a document by dropping in the sidebar
### Discussed in https://github.com/outline/outline/discussions/2135

<sup>Originally posted by **iamsaumya** May 12, 2021</sup>
While testing for bugs, I find myself with a lot of documents at times. Now, I either have to delete or archive the documents. Both the option are taxing for a large number of documents. 

From the user's point of view, they too will find it tedious to archive each document through the menu. 

How about dropping the document on the archive menu item in the sidebar? It will be much faster in this case. 

Considering the fact we already have drag and drop implemented in the app, the users are already accustomed to the notion of drag and drop. </div>

iamsaumya:
Closes #2299


app/components/Sidebar/components/TrashLink.js
```diff
@@ -0,0 +1,66 @@
...
+        <Modal
+          title={t("Delete {{ documentName }}", {
+            documentName: document.noun,
+          })}
+          onRequestClose={handleModalClose}
+          isOpen={showDeleteModal}
+        >
+          <DocumentDelete document={document} onSubmit={handleModalClose} />
+        </Modal>
+      )}
```
iamsaumya, line 61: This logic to show modal while dropping works in chrome but not in firefox. Could you help me how to debug this kind of bug?

iamsaumya: @tommoor

tommoor: Will look at it, one note here is that I don't think we need state for the delete modal bool AND the document, we can probably just use the document being in state to know if to show the modal?

tommoor: It's actually working in Firefox for me 🤔

iamsaumya: It's not even throwing any error or warning.

tommoor: https://user-images.githubusercontent.com/380914/125659527-2c7f5a85-e877-4f14-8ce9-619fe6ffd32e.mp4

iamsaumya: Great! Now we can narrow down this to `firefox` on `ubuntu`. It's one of those "works on my machine" bugs xD

iamsaumya: I tried different changes, but nothing solved the issue. The lead, for now, is that the issue is with the `Modal` component, not the `DocumentDelete`.

tommoor: What version of Firefox? I'm on `89.0.2`

tommoor:
Looks great, the icon change is super cute.

app/components/Sidebar/components/TrashLink.js
iamsaumya: Me too 89.0.2 (64-bit)

tommoor: I think it's down to you to debug as I still can't repro, try putting logs in the relevant drag/drop callbacks to see where it isn't getting to

iamsaumya: No luck finding the cause. I used browser debugging tools, and it feels like firefox is gulping the change 😞

tommoor: Does the icon animate on firefox like it's _going_ to work? Or no feedback at all?

iamsaumya: It does animate.

app/components/Sidebar/components/TrashLink.js
```diff
@@ -0,0 +1,61 @@
...
+import SidebarLink from "./SidebarLink";
+
+function TrashLink({ documents }) {
+  const { policies } = useStores();
+  const { t } = useTranslation();
+  const [document, setDocument] = useState();
+
+  const [{ isDocumentDropping }, dropToTrashDocument] = useDrop({
+    accept: "document",
+    drop: async (item, monitor) => {
```
tommoor, line 20:
```suggestion
    drop: (item, monitor) => {
```

Try removing? Unnecessary anyway 🤷 – a debugger in here is triggered or no?

tommoor: My best guess is that maybe it's opening the modal but then the following `mouseup` is counting as a click "outside" the modal and it immediately closes. If that's the case wrapping the setDocument in a `setTimeout` might do the trick

iamsaumya: Bull's-eye! It worked!!

iamsaumya: This is the issue because override the `hover` instead of `drop` does the work.

iamsaumya: Hurray, we did it 🎉

iamsaumya:
Please consider merging #2339 before this.

tommoor (approved):

app/components/Sidebar/Main.js
```diff
@@ -142,7 +142,21 @@ function MainSidebar() {
             </Section>
             <Section>
               <ArchiveLink documents={documents} />
-              <TrashLink documents={documents} />
+              {can.createDocument && (
```
tommoor, line 145: Bad merge here, ArchiveLink appears twice

app/components/Sidebar/components/ArchiveLink.js
```diff
@@ -8,15 +8,15 @@ import useStores from "../../../hooks/useStores";
...
+  const { policies, toasts } = useStores();
   const { t } = useTranslation();
 
   const [{ isDocumentDropping }, dropToArchiveDocument] = useDrop({
     accept: "document",
     drop: async (item, monitor) => {
       const document = documents.get(item.id);
       await document.archive();
-      ui.showToast(t("Document archived"), { type: "success" });
+      toasts.showToast(t("Document archived"), { type: "success" });
```
tommoor, line 19: Don't want to use the hook?

PR #2322, opened 2021-07-14
chore: Loading placeholders

tommoor:
You should rarely ever see a loading screen, but if you do it had better be damn good looking.

- Normalized naming and file locations for loading placeholders
- Rounded, and improved animation
- Fixed a small css bug


app/components/Flex.js
```diff
@@ -45,7 +45,7 @@ const Container = styled.div`
   align-items: ${({ align }) => align};
   justify-content: ${({ justify }) => justify};
   flex-shrink: ${({ shrink }) => (shrink ? 1 : "initial")};
-  gap: ${({ gap }) => `${gap}px` || "initial"};
+  gap: ${({ gap }) => (gap ? `${gap}px` : "initial")};
```
tommoor, line 48: Incorrect use of `||`

PR #2326, opened 2021-07-16
fix: Allow searching of previous document titles 

### LINKED ISSUE #2211: Allow searching of previous document titles
It would be nice if it was possible to find a document by searching any of its previous titles. This would need the following:

- A new `previousTitles` column as an array of strings
- New column should be search indexed in the same way as title column
- Write previous title to new column whenever it changes in `api/documents.update`
- Include new column in search methods `searchForUser` and `searchForTeam`
- Include tests for above behavior

iamsaumya:
closes #2211


server/migrations/20210716071454-search-index-previousTitles.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const searchDocument = `
+    CREATE OR REPLACE FUNCTION documents_search_trigger() RETURNS trigger AS $$
+    begin
+      new."searchVector" :=
+        setweight(to_tsvector('english', coalesce(new.title, '')),'A') ||
+        setweight(to_tsvector('english', coalesce(array_to_string(new."previousTitles", ' , '),'')),'A') ||
```
iamsaumya, line 10: @tommoor, I had to google search to figure this. It works, but do let me know if this is not the way you would have done.

tommoor: Will do a little more research but gut tells me this is the right way – We should probably set the weights like so:

- Current title: `A`
- Current text: `B`
- Previous titles: `C`

server/models/Document.test.js
```diff
@@ -221,6 +221,41 @@ describe("#searchForTeam", () => {
...
+      teamId: team.id,
+      collectionId: collection.id,
+      title: "test number 1",
+    });
+
+    document.title = "change";
+    await document.save();
+
+    const { totalCount } = await Document.searchForTeam(team, "test number");
+    expect(totalCount).toBe("1");
```
tommoor, line 238: We can also test that documents that _currently_ have the title come before documents that _previously_ had the title in results ordering

server/migrations/20210716071454-search-index-previousTitles.js
iamsaumya: Works like a charm. 

Screenshot from 2021-07-17 11-35-18

app/components/DocumentListItem.js
```diff
@@ -63,6 +64,10 @@ function DocumentListItem(props: Props, ref) {
   const queryIsInTitle =
     !!highlight &&
     !!document.title.toLowerCase().includes(highlight.toLowerCase());
+  const queryIsInContext =
+    !!highlight && !!context?.toLowerCase().includes(highlight?.toLowerCase());
+  const showPreviousTitleTip =
+    highlight && !queryIsInTitle && !queryIsInContext;
```
iamsaumya, line 70: Works without changing the API. What do you think about this?

tommoor: I think it is pragmatic but deserving of a code comment as it's a loose coupling that may become an unreliable assumption in the future. For this PR it makes sense!

tommoor:
Pretty awesome how little extra code this ended up being

tommoor (changes requested):

app/components/DocumentListItem.js
```diff
@@ -63,6 +64,12 @@ function DocumentListItem(props: Props, ref) {
   const queryIsInTitle =
     !!highlight &&
     !!document.title.toLowerCase().includes(highlight.toLowerCase());
+  const queryIsInContext =
+    !!highlight && !!context?.toLowerCase().includes(highlight?.toLowerCase());
+
+  // Is used to show tooltip when the highlight isn't present in the title and context.
+  const showPreviousTitleTip =
```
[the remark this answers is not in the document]
iamsaumya: > original search term which isn't actually guaranteed to appear anywhere in the results given the word stemming that Postgres does

Oh!

> Given this i'm tempted to say we just drop this additional treatment for now, thoughts?

Sure, no problem! Just give me a sec

app/components/DocumentListItem.js
```diff
@@ -88,6 +95,17 @@ function DocumentListItem(props: Props, ref) {
           {document.isNew && document.createdBy.id !== currentUser.id && (
             <Badge yellow>{t("New")}</Badge>
           )}
+          {showPreviousTitleTip && (
+            <Tooltip
+              tooltip={t("A previous title matches your search")}
+              delay={500}
+              placement="bottom"
+            >
+              <IconButton>
```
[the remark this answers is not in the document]
iamsaumya: Icon as a children was throwing an error, so I thought IconButton must be the way to go.

PR #2337, opened 2021-07-17
feat: Optimize Dockerfile

lolPants:
## Dockerfile Improvements
* Use new BuildKit dockerfile syntaxes
* Leverage multi-stage build steps for faster builds
* Strip Yarn install cache from image, reducing final size
* Use a stricter version of `yarn install` such that it will fail with an invalid lockfile
* Run as a non-root user for security

## Notes
* `yarn-deduplicate` was moved to a required dependency, otherwise running `yarn install --production` will fail on fresh installs.
* In my testing the image sizes were reduced from ~1.4GB to ~450MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,43 @@
-FROM node:14-alpine
+# syntax=docker/dockerfile:1.2
+FROM node:14-alpine AS deps-common
 
-ENV APP_PATH /opt/outline
-RUN mkdir -p $APP_PATH
+ARG APP_PATH /opt/outline
+WORKDIR $APP_PATH
+COPY ./package.json ./yarn.lock ./
 
+# ---
+FROM deps-common AS deps-dev
+RUN yarn install --no-optional --frozen-lockfile && \
+  yarn cache clean
+
+# ---
+FROM deps-common AS deps-prod
+RUN yarn install --production=true --frozen-lockfile && \
+  yarn cache clean
+
+# ---
+FROM node:14-alpine AS builder
 WORKDIR $APP_PATH
 
-COPY package.json ./
-COPY yarn.lock ./
-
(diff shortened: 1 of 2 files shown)


tommoor:
> yarn-deduplicate was moved to a required dependency, otherwise running yarn install --production will fail on fresh installs.

Hmm, okay this makes sense – we probably need to change this so that deduplicate only runs in dev if possible. Running it after a frozen lockfile defeats the purpose a little.

> In my testing the image sizes were reduced from ~1.4GB to ~450MB

Very cool

lolPants:
> Hmm, okay this makes sense – we probably need to change this so that deduplicate only runs in dev if possible. Running it after a frozen lockfile defeats the purpose a little.

I believe the `prepare` script does exactly this, running only when `--production` is not specified. Alternatively you could craft a script leveraging `NODE_ENV` and only running deduplication when set to `production`.

tommoor:
Great idea on `prepare`, I've updated this on `main`

lolPants:
Updated the PR to revert the change to `yarn-deduplicate`, double-checking the image builds as expected and then this should be good to merge.

lolPants:
All looks good on my end!

tommoor:
Testing this now, did most of the size saving come from the cache removal as far as you can tell?

lolPants:
That certainly helps a lot, but Docker build layers are additive, so installing dev-deps and then removing them in a later step will still retain the full size of said deps. Leveraging multi-stage builds you can use dev-deps to build in a different step and copy in the prod deps to the final image.

tommoor:
This pre-release includes these Dockerfile changes if you wanted to try:

`docker pull outlinewiki/outline:version-0.58.0-4`

PR #2339, opened 2021-07-19
fix: Separate toasts storage to own MobX store

### LINKED ISSUE #2329: Separate toasts storage to own `MobX` store
It feels like 90% of the time the `UiStore` is imported into a component it is so that we can create a toast. Suggesting the following improvements for maintainability and performance:

- Add a hook for creating toasts (see API suggestion below).
- Use hook instead of access ui store in functional components. Note: there are still 7 class components where toasts are accessed, we can avoid converting to functional as part of this.
- Move storage of toasts to their own store, separate from `ui` – this allows ui store to be for interface settings only. This should be easy once the hook has abstracted away `ui.showToast` access

```javascript

import useToasts from "hooks/useToasts";

// within component
const { showToast, hideToast } = useToasts();

// usage as today
showToast(
  t("Content goest here"),
  { type: "info" }
);

```

iamsaumya:
closes #2329


iamsaumya:
This is already too much to review. I will convert class components in some other PR 😄

app/hooks/useToasts.js
```diff
@@ -0,0 +1,26 @@
...
+    },
+    [toasts]
+  );
+
+  const hideToast = useCallback(
+    (toastId: string) => {
+      toasts.removeToast(toastId);
+    },
+    [toasts]
+  );
```
tommoor, line 23: It doesn't seem like these memo's are needed, can pass the store methods directly back?

app/stores/ToastsStore.js
```diff
@@ -0,0 +1,52 @@
...
+    return id;
+  };
+
+  @action
+  removeToast = (id: string) => {
+    this.toasts.delete(id);
+  };
+
+  @computed
+  get orderedToasts(): Toast[] {
```
tommoor, line 49:
```suggestion
  get orderedData(): Toast[] {
```

Lets rename to match other stores now that it's single-purpose

app/stores/ToastsStore.js
```diff
@@ -0,0 +1,52 @@
...
+      type: options.type,
+      timeout: options.timeout,
+      action: options.action,
+    });
+    this.lastToastId = id;
+    return id;
+  };
+
+  @action
+  removeToast = (id: string) => {
```
tommoor, line 44: I'd be open to renaming this to `hideToast`

app/scenes/Document/components/Document.js
```diff
@@ -91,7 +92,7 @@ class DocumentScene extends React.Component<Props> {
     } else if (prevProps.document.revision !== this.lastRevision) {
       if (auth.user && document.updatedBy.id !== auth.user.id) {
         this.props.toasts.showToast(
-          `Document updated by ${document.updatedBy.name}`,
+          t(`Document updated by ${document.updatedBy.name}`),
```
tommoor, line 95:
```suggestion
          t(`Document updated by {{userName}}`, { userName: document.updatedBy.name }),
```

Good catch, but gotta do the interpolation right 😉

iamsaumya: Ohh, sorry about that!

tommoor: I wrote it in on GitHub so sorry about the linting

tommoor (approved):

PR #2342, opened 2021-07-20
fix: Show tasks completion on document list items

### LINKED ISSUE #2320: Show tasks completion on document list items
It would be neat to expose a preview of the completed checklist items on documents, see how it's done in a couple of other products:

#### Linear
image

#### GitHub
image

iamsaumya:
Closes #2320


server/models/Document.js
```diff
@@ -100,6 +102,20 @@ const Document = sequelize.define(
...
+
+        let countCheckbox = (this.text.match(CHECKBOX_REGEX) || []).length;
+        if (!countCheckbox) {
+          return [];
+        }
+
+        let countTicked = (this.text.match(TICKED_CHECKBOX_REGEX) || []).length;
+
+        return [countTicked, countCheckbox];
+      },
```
tommoor, line 118:
```js
tasks: {
  completed: 1,
  total: 4
}
```

Thoughts on this interface?

iamsaumya: Sure! It is more descriptive and maintainble.

app/scenes/Document/components/Document.js
```diff
@@ -214,6 +215,8 @@ class DocumentScene extends React.Component<Props> {
     this.isSaving = true;
     this.isPublishing = !!options.publish;
 
+    document.tasks = getTasks(document.text);
```
iamsaumya, line 218: To make the experience more fluid

shared/utils/getTasks.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+const TICKED_CHECKBOX_REGEX = /\[(x)\]\s/g;
+const CHECKBOX_REGEX = /\[(x|\s|)\]\s/g;
```
tommoor, line 4: Could we use this regex to match what the editor uses?

https://github.com/outline/rich-markdown-editor/blob/main/src/lib/markdown/checkboxes.ts#L4

That way we can run the regex only once and then look at the contents of the first capture group to determine if the checkbox is checked or not

iamsaumya: I was conscious of the head ups, but I could not understand the editor's regex.

Is it possible to have `-` and `_` inside the brackets? Also, it has capital X instead of x, whereas the document `text` attribute has `x` inside the brackets.

tommoor: It's possible although granted unlikely in imported markdown. Just thinking for consistency's sake if they appear as checkboxes in the editor they should _always_ show up in the tasks list too.

The capitalization doesn't matter when the case insensitive flag is used

app/components/DocumentMeta.js
```diff
@@ -61,8 +62,20 @@ function DocumentMeta({
     deletedAt,
     isDraft,
     lastViewedAt,
+    isTasks,
+    tasksPercentage,
+    tasksMessage,
   } = document;
 
+  const progressBar = React.useMemo(() => {
```
tommoor, line 70: Could also just pull this out to a top level functional component

app/components/DocumentMeta.js
```diff
@@ -61,8 +62,20 @@ function DocumentMeta({
...
+    return (
+      <>
+        <CircularProgressBar percentage={tasksPercentage} />
+        &nbsp;{tasksMessage}
```
tommoor, line 74: Let's move `tasksMessage` inline here so that it can be translated, I prefer the english language version used on GitHub as an inspiration as 1/3 might not be immediately clear in the context of Outline.

### Suggested variants

- `complete=0`: "X tasks"
- `complete>0 && complete<total`: "X of X tasks"
- `complete===total`: "X tasks done"

app/components/DocumentMeta.js
```diff
@@ -154,6 +167,7 @@ function DocumentMeta({
         </span>
       )}
       &nbsp;{timeSinceNow()}
+      {isTasks && <>&nbsp;•&nbsp;{progressBar}</>}
```
tommoor, line 170: Can we add a check for `!isTemplate` here, seems weird to show the progress bar on templates as they'll never be filled out 👍

app/components/CircularProgressBar.js
```diff
@@ -0,0 +1,50 @@
...
+import React from "react";
+import { useTheme } from "styled-components";
+
+const cleanPercentage = (percentage) => {
+  const tooLow = !Number.isFinite(+percentage) || percentage < 0;
+  const tooHigh = percentage > 100;
+  return tooLow ? 0 : tooHigh ? 100 : +percentage;
+};
+
+const Circle = ({ colour, pct }: { colour: string, pct?: number }) => {
```
tommoor, line 11: - US spelling for `color` pls, I'm English so it kills me a little on the inside, but it's better to be consistent. 
- `pct` => `percentage`

app/components/CircularProgressBar.js
```diff
@@ -0,0 +1,50 @@
...
+      style={{ transition: "stroke-dashoffset 0.6s ease 0s" }}
+    ></circle>
+  );
+};
+
+const CircularProgressBar = ({ percentage }: { percentage: number }) => {
+  const theme = useTheme();
+  const pct = cleanPercentage(percentage);
+  return (
+    <svg width={18} height={18}>
```
tommoor, line 41: Is it possible to add a `size` prop here in px and reduce the default to `16`?

tommoor (changes requested):
Dude, fantastic first pass 👏

server/models/Document.test.js
```diff
@@ -430,3 +430,79 @@ describe("#findByPk", () => {
...
 });
+
+describe("tasks", () => {
+  test("should consider all the possible checkTtems", async () => {
+    const document = await buildDocument({
+      text: `- [x] test
+      - [X] test
+      - [ ] test
+      - [-] test
+      - [_] test`,
```
iamsaumya, line 441: They are matched through the regex, but I am not sure about their behaviors. Are they considered checked or unchecked?

tommoor: ✅  checked

iamsaumya: My guess was right 😄

tommoor:
Hope you don't mind me making some minor edits 😄

iamsaumya:
> Hope you don't mind me making some minor edits

Oh, why would I? I will get to learn something.

tommoor (approved):
Excited to see how this is received 😆

Issue #2336, opened 2021-07-17
`auth.info` endpoint should check authentication provider access token validity


tommoor:
- When requesting `/api/auth.info` the endpoint should make a request to load the users profile data from the SSO provider (if any)
- If the request fails with access denied then a `401` response should be returned to the client
- This will cause the local session to be cleared and forward the browser to the login screen. (This functionality already exists on `401` response)

tommoor:
The tricky thing with this one is how to abstract the SSO provider request, I'd imagine the best thing to do would be to expose a method within all of the `auth/provider/x` files, `validateToken`, or similar.

iamsaumya:
For slack the function is already there. For Google I have found an endpoint, but it doesn't seem to work for now.

iamsaumya:
Do we have token rotation enabled for slack? 

For google - we need to handle generating the access token in `auth.info` because they are short-lived, and the user will be logged out every time if we don't handle this case.

tommoor:
We store refresh tokens for Slack and it seems like _most_ Google accounts, for some reason they don't all have it – perhaps users are able to not allow offline access and that results in it not being returned 🤔

iamsaumya:
I am not sure, I am using my college id, and It's even not returning for me.

iamsaumya:
It seems we were not passing all the required fields to get offline access. 

This will work 

```js  
passport.authenticate(providerName, {
      accessType: "offline",
      prompt: "consent",
    })
```

PR #2365, opened 2021-07-24
fix: auth.info endpoint should check authentication provider access token validity

iamsaumya:
WIP for #2336

### DIFF
--- a/server/api/auth.js
+++ b/server/api/auth.js
@@ -3,9 +3,11 @@ import Router from "koa-router";
 import { find } from "lodash";
 import { parseDomain, isCustomSubdomain } from "../../shared/utils/domains";
 import providers from "../auth/providers";
+import { AuthenticationError } from "../errors";
 import auth from "../middlewares/authentication";
-import { Team } from "../models";
+import { AuthenticationProvider, Team, UserAuthentication } from "../models";
 import { presentUser, presentTeam, presentPolicies } from "../presenters";
+import * as Slack from "../slack";
 import { isCustomDomain } from "../utils/domains";
 
 const router = new Router();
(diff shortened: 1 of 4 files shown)


iamsaumya:
Tom, Could you skim through the files, please? I want to verify this is how you want this issue to be handled. 

For the most part, slack is done, and google is left.

server/api/auth.js
```diff
@@ -105,6 +107,47 @@ router.post("auth.info", auth(), async (ctx) => {
...
+      const data = await Slack.postUrlEncodedForm("oauth.v2.access", {
+        client_id: process.env.SLACK_KEY,
+        client_secret: process.env.SLACK_SECRET,
+        grant_type: "refresh_token",
+        refresh_token: userAuthentication.refreshToken,
+      });
+
+      if (!data.ok) {
+        throw new AuthenticationError(data.error);
+      }
```
iamsaumya, line 142: Here is particular, these are the [possible errors](https://api.slack.com/methods/oauth.v2.access#errors), and I am not sure how to go about handling them. 

Right now, Giving `404` will do no good because the user will always get stuck here when they retry.

Should we send something like `contact the team`?

tommoor:
I think you've made a solid discovery with the lack of refresh token/offline access request for Google OAuth provider – it seems that we should PR just this part first and get it in the codebase, we can then wait ~a month or so before rolling out any validation code (next release?), that would give enough time for many users to have reauthenticated and gotten a refresh token. 

After all, we don't want to be in a situation where 50% of users are suddenly logged out :)

iamsaumya:
Sure 👍🏻

PR #2369, opened 2021-07-25
Generic OAuth2 Authentication

EternalDeiwos:
A very simple implementation of most of the things suggested in #2205. My original intention was to verify, validate, and consume an ID Token containing appropriate properties to authenticate and populate the user properties, however there were some difficulties with this:

1. Not all passport strategies are compatible with your `koa-passport` implementation, which unfortunately includes `openid-client`. It would be possible to make it work, but that would either require forking it or upstream changes in the original, neither of which are likely to be feasible. Therefore I used `passport-oauth2` which doesn't interact with JWTs at all and instead just resolves the `userinfo`. In practice this works fine, although it may limit the number of IdPs that can be used with this easily.
2. To enable PKCE I needed to add `koa-session`, which is a standardized way to manage state. PKCE is necessary to protect the application against CSRF attacks, however it is optional in the OAuth2 specification. It is highly recommended, although I am vaguely aware of some `StateStore` implementation on the server which is now duplicate code. Additionally, the configuration used for the sessions are currently all defaults (see `server/app.js`); some further configuration there is likely necessary.
3. I also could not include code to resolve the majority of the configuration (`authorization_endpoint`, `token_endpoint`, etc.) from a discovery service due to the fact that the providers are loaded and routers mounted asynchronously. If this is refactored in the future, it would make this auth provider more convenient to use.
4. There are some cosmetic items that could be improved, such as the auth provider icon, however given the provider is generic, the icon would also need to be generic. I figured someone else would have a better idea for how to deal with this.

#### Additions

The provider provisions users and the team from the contents of `userinfo`. Specifically, the mapping is as follows:
- `team_name` (non-standard claim name) &#8594; `team.name`
- `name` (standard OIDC claim name) &#8594; `user.name`
- `preferred_username` (standard OIDC claim name) &#8594; `user.username`
- `email` (standard OIDC claim name) &#8594; `user.email`
- `picture` (standard OIDC claim name &mdash; optional) &#8594; `user.avatarUrl`
- `sub` (standard OIDC claim name) &#8594; `authentication.providerId`

#### Configuration

All additional env vars have been added to `.env.sample` but here is a summary:
- `COOKIE_SECRET` &mdash; This is for `koa-session`.
- `OIDC_DISPLAY_NAME` &mdash; Cosmetic display name for the provider configuration. Mostly to customize the "Continue with..." button. Defaults to `OAuth2`.
- `OIDC_CLIENT_ID`
- `OIDC_CLIENT_SECRET`
- `OIDC_AUTH_URI`
- `OIDC_TOKEN_URI`
- `OIDC_USERINFO_URI`
- ~~`OIDC_USERNAME_CLAIM` &mdash; Override which claim to check for the user's username. Defaults to `preferred_username` which is the standard OIDC claim name.~~ I removed this one due to it being a violation of the flow object specification for `accountProvisioner`. 
- `OIDC_TEAM_CLAIM` &mdash; Override which claim to check for the name of the team. Defaults to `team_name`.
- `OIDC_SCOPES` &mdash; Space separated list of scopes. Must at least contain `openid` to grant access to `userinfo`. Defaults to `openid profile email`.


tommoor:
Hey Greg – a ​few questions:

> **Not all passport strategies are compatible with your koa-passport implementation**

In what sense are you seeing incompatibility? There's nothing particularly unique about the implementation that I'm aware of.

> **To enable PKCE I needed to add koa-session**

PKCE is only necessary for public clients, our client is server side so PKCE is unnecessary and doesn't add anything security-wise to the OAuth process as I understand. I'd really like to avoid the complexity of maintaining _another_ session variable so this part is a bit of a blocker for me.

> **due to the fact that the providers are loaded and routers mounted asynchronously**

Any more details on this? Not sure why it's a limitation but I trust that it is.

> **There are some cosmetic items that could be improved, such as the auth provider icon**

Design is supposed to work okay without an icon, particularly if it's the only auth option then it should look fine.

EternalDeiwos:
Hi Tom,

First a disclaimer; I am not extremely familiar with Koa but I have used Express so I stand under correction.

> In what sense are you seeing incompatibility? There's nothing particularly unique about the implementation that I'm aware of.

I have not tested widely, so I cannot say with certainty, however while building this I encountered two problems:

1. Either due to a bug or an incompatibility with the `openid-client` strategy in particular, the query string on the redirect from the IdP would not be parsed. I traced the problem from [here](https://github.com/panva/node-openid-client/blob/main/lib/passport_strategy.js#L86) to [here](https://github.com/panva/node-openid-client/blob/main/lib/client.js#L341) before deciding on a path of lesser resistance, which was `passport-oauth2`. Given the amount of use and traffic `openid-client` and my observation of the differences with Koa, I made the assumption that the problem was one of compatibility. Again I may be corrected on this.
2. Both `openid-client`, `passport-oauth2`, and many other passport strategies rely on the existence of `req.session` which conforms to the [interface](https://github.com/expressjs/session#store) defined by `express-session` and is mirrored by `koa-session`. While many strategies themselves allow you to pass in a `store` variable, many do not, and many that do _still_ require `req.session` to be defined and conformant in addition... for some reason. Again, I did not spend a huge amount of time looking at this, as I proceeded with the path of least resistance, which leads me to the next point.

> I'd really like to avoid the complexity of maintaining another session variable so this part is a bit of a blocker for me.

I understand and agree with your hesitation, which is why I drew attention to this in the first place. Maintaining multiple session mechanisms in the app would almost certainly create problems down the line so I do not recommend that. I would offer that using `koa-session` would be the healthier solution for Outline for the following reasons:

1. As mentioned previously, many existing libraries that are widely used in the passport ecosystem rely in the existence of the `req.session` interface. I don't particularly agree with how it is set up; but it was established as a convention around Node 0.10 and has stuck since. If you want to make use of those libraries, including `passport-oauth2`, then the optimal way to do so is to support that interface. You would probably know better than me which other libraries use a similar mechanism that would benefit from a more standard interface. This is an assumption that underpins the following points, and I have a clear bias in making it.
2. One could extend your existing `StateStore` to conform to the interface, and therefore may result in the two mechanisms being combined, however I would not recommend this as it would result in a large amount of additional code and "hacking it together". Given the neatness of your code base and your level of organization, I imagine this would make your skin crawl just as much as it would mine.
3. One could embrace `koa-session` and factor out your existing code to use that instead. Doing so, you would gain the benefit of external updates that are heavily tested by a large community and supported by the developers of Koa themselves; better compatibility with anything that relies on that mechanism; access to a mature, pluggable session management mechanism that would require minimal modifications to use external session management in the future; and would certainly simplify and reduce the amount of code you have to maintain.

> PKCE is only necessary for public clients, our client is server side so PKCE is unnecessary and doesn't add anything security-wise to the OAuth process as I understand.

I disagree. While PKCE was originally designed to secure specifically the implicit flow against CSRF attacks which were a major hindrance to its use in specifically SPAs and native apps, the same benefits seen there are relevant and applicable to the authorization code flow as well. PKCE created a mitigation for authorization code theft that otherwise wouldn't have existed, and is distinct from the authentication of secure clients. It is not strictly required _yet_, however it is [currently a best practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-02#section-9.8) and I believe they intend to make it a requirement. Additionally, we basically get the benefit "for free" so long as we enable it.

> > due to the fact that the providers are loaded and routers mounted asynchronously
>
> Any more details on this? Not sure why it's a limitation but I trust that it is.

I made an error here; I meant to say that the providers and routers are mounted _synchronously_. This prevents me from making requests for external discovery documents, and most importantly, public key material to verify signatures.

EternalDeiwos:
My above comments aside; perhaps a separate discussion is necessary around the use of `koa-session` that does not belong in this PR.

tommoor:
> It is not strictly required yet, however it is currently a best practice and I believe they intend to make it a requirement

I see, thank you for the context link there

> perhaps a separate discussion is necessary around the use of koa-session that does not belong in this PR.

Yes, I think if we can remove it and use `store: new StateStore()` as with all of the other providers that will enable a much faster merge. You shouldn't need to fork or anything, `store` option is part of oauth2 provider too.

EternalDeiwos:
@tommoor please see #2388 for those changes.

PR #2373, opened 2021-07-26
fix: Add icons to menu items

### LINKED ISSUE #2345: Add icons to menu items
I'm excited for this to make menus much more easily scannable, it will also enable the [quick menu](https://github.com/outline/outline/issues/2137) to be more visual for the same reasons and is considered a blocker for the completion of that PR.

### Rough steps

- [x] Add missing icons to `outline-icons` (design done, new version coming soon)
- [x] Publish new version of `outline-icons` (`outline-icons@1.28.0`)
- [x] Add `icon` as first class property on `MenuItem` rendered on the left hand side of the text
- [x] Convert the places where custom icons are currently used to use the `icon` prop, eg `NewDocumentMenu` shows collection icons and Appearance in `AccountMenu` has icons too
- [x] Add icons to existing menu items where missing

iamsaumya:
Closes #2345


tommoor:
Sweet, okay first feedback that I hope you agree with after seeing this IRL:

- After viewing I think we should only add icons on "..." context menus, and of those most but not all
- `AccountMenu` looks worse with icons, in the long run it's probably worth making this a totally custom menu design to be honest. Maybe just use the new `icon` prop for "Appearance" but drop the rest?
- Also drop the icons on the sorting menu
- The icon has made the menu items a touch larger, I think we should go to 14px font size and the following padding on desktop:

image

iamsaumya:
> - AccountMenu looks worse with icons, in the long run it's probably worth making this a totally custom menu design to be honest. Maybe just use the new icon prop for "Appearance" but drop the rest?
> - Also drop the icons on the sorting menu
> - The icon has made the menu items a touch larger, I think we should go to 14px font size and the following padding on desktop:

I couldn't agree more 💯 

> After viewing I think we should only add icons on "..." context menus, and of those most but not all

However, I really like the documents and collection menu with the icons. Having only "..." context menus with icons will make others feel orphaned.

Well, this is from my naive perspective. I will love to hear your reasoning on this approach.

iamsaumya:
If we can group these menu items with icons in a section, I think that will look more polished. What do you suggest? 

Screenshot from 2021-07-27 17-07-10

tommoor:
Ah sorry, misunderstanding, I meant the menus that are accessed through a "..." icon.
I too like the document and collection menus with icons on all items :)

iamsaumya:
What's the icon for embeds menu item?

Screenshot from 2021-07-27 23-19-13

tommoor:
Perhaps `BuildingBlocksIcon` – `AlignImageCenterIcon` might work too, looks a lot like an embed

iamsaumya:
`CodeIcon` comes as the icon when searched for embeds icon on google. What do you say?

tommoor:
Yea, that's understandable – I'm a bit wary of it because we actually do have code blocks that use this icon and those _arent_ embeds.

iamsaumya:
Just two more icons are required.

1. Restore - It can be similar to `HistoryIcon` without the clock.
2. Permanent Delete - No gap between the lid and can.

tommoor:
image

Incoming

tommoor:
`v1.29.0` -> https://github.com/outline/outline-icons/commit/915efedf627504d16786be02e02efa2429c51c79

iamsaumya:
Great!!

Where do you want me to use the `crossIcon`? Is it permanent delete?

tommoor:
pls

iamsaumya:
Ready for review

app/components/ContextMenu/MenuItem.js
```diff
@@ -133,6 +137,11 @@ export const MenuAnchor = styled.a`
     padding: ${(props) => (props.$toggleable ? "4px 12px" : "6px 12px")};
     font-size: 15px;
   `};
+
+  ${breakpoint("desktop")`
```
tommoor, line 141: Tablet === desktop in terms of styling, can move this into the above clause

tommoor (commented):
I have a slightly awkward request… if there are any icons in the menu template then we should adjust the padding of all items in the menu so that the text aligns, something like this:

image

It's okay for this to only work for <Template> style menus as it'll be a forcing function to move over for remaining spots.

app/menus/NewDocumentMenu.js
```diff
@@ -45,7 +44,7 @@ function NewDocumentMenu() {
       <MenuButton {...menu}>
         {(props) => (
           <Button icon={<PlusIcon />} {...props} small>
-            {`${t("New doc")}…`}
+            {`${t("New doc")}`}
```
tommoor, line 47:
```suggestion
            {`${t("New doc")}…`}
```

This was actually purposeful to signify that there is another menu/choice

app/menus/AccountMenu.js
```diff
@@ -91,6 +43,65 @@ function AccountMenu(props: Props) {
     }
   }, [menu, ui.theme, previousTheme]);
 
+  const items = () =>
+    filterTemplateItems([
```
tommoor, line 47: Do we need to call `filterTemplateItems` here? It's the first thing that's called inside of `<Template>`

iamsaumya: No need.

This is also done in CollectionMenu. Will fix.

app/components/MenuIconWrapper.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import styled from "styled-components";
+
+const MenuIconWrapper = styled.span`
+  position: relative;
+  width: 24px;
+  height: 24px;
+  right: 4px;
+  color: ${(props) => props.theme.textSecondary};
```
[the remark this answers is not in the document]
iamsaumya: _copy_ _paste_ error 🤦🏻‍♂️

tommoor (approved):

PR #2388, opened 2021-07-28
feat: Generic OAuth2 Authentication

EternalDeiwos:
As per discussion on #2369; recreated as a separate PR to compare to a separate branch.


.env.sample
```diff
@@ -106,21 +115,33 @@ DEBUG=cache,presenters,events,emails,mailer,utils,http,server,services
...
+OIDC_DISPLAY_NAME=OAuth2
+
+# Space separated auth scopes.
+OIDC_SCOPES=openid profile email
+
+# Override the claim used to assign users' usernames.
+OIDC_USERNAME_CLAIM=preferred_username
+
+# Override the claim used to assign the name of the team.
+OIDC_TEAM_CLAIM=team_name
```
tommoor, line 136: Could we move these up to L87 alongside the other `OIDC_`?

`OIDC_USERNAME_CLAIM` and `OIDC_TEAM_CLAIM` are unused in this PR, did you want to include the logic for that or remove these?

server/auth/providers/oauth2.js
```diff
@@ -0,0 +1,123 @@
...
+      },
+    });
+
+    try {
+      return done(null, await response.json());
+    } catch (ex) {
+      return done(new Error("Failed to parse user profile"));
+    }
+  } catch (err) {
+    return done(new Error("Failed to fetch user profile"));
```
tommoor, line 49: Should probably log the error here at the very least otherwise debugging this will be very difficult

.env.sample
```diff
@@ -69,24 +69,33 @@ SLACK_SECRET=get_the_secret_of_above_key
...
 # the guide for details on setting up your Azure App:
 # => https://wiki.generaloutline.com/share/dfa77e56-d4d2-4b51-8ff8-84ea6608faa4
-AZURE_CLIENT_ID= 
+AZURE_CLIENT_ID=
 AZURE_CLIENT_SECRET=
 AZURE_RESOURCE_APP_ID=
 
+# To configure generic OIDC auth, you'll need some kind of identity provider.
+# See documentation for whichever IdP you use to acquire the following info:
+# Redirect URI is https://<URL>/auth/oidc.callback
```
tommoor, line 81:
```suggestion
# Redirect URI is https://<URL>/auth/oauth2.callback
```

Judging by the provider name this should be changed, I think that leaving this as `oidc` might be a better choice though and change the provider name, thoughts?

server/auth/providers/oauth2.js
```diff
@@ -0,0 +1,123 @@
...
+              `Domain ${domain} is not on the whitelist`
+            );
+          }
+
+          const subdomain = domain.split(".")[0];
+
+          const result = await accountProvisioner({
+            ip: req.ip,
+            team: {
+              name: profile.team_name,
```
tommoor, line 89: I'm not super familiar with `OIDC`, could we link to the spec that outlines all the available fields on `profile` as a comment? That would be really useful for future maintenance.

.env.sample
EternalDeiwos: > `OIDC_USERNAME_CLAIM` and `OIDC_TEAM_CLAIM` are unused in this PR, did you want to include the logic for that or remove these?

This is a good catch. I'm not actually pulling `username` from profile at all as the `accountProvisioner` is not expecting it. I have removed it.

On the other hand I did want to include `OIDC_TEAM_NAME` because it is not a standard claim, so other users will likely want to configure it to be whatever claim their IdP puts out.

.env.sample
EternalDeiwos: @tommoor I'm inclined to agree... the passport strategy is a bit of a misnomer given the `userinfo` extension, which is part of the `oidc` spec 🤷

Either way, I agree it needs to be consistent. Good catch.

server/auth/providers/oauth2.js
JaneJeon: Hey, I know I’m not the PR author but OIDC’s “standard-ish” claims are actually quite wide: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

Probably best to just link this and not actually list everything :)

.env.sample
JaneJeon: Thoughts on using `openid-client`’s strategy rather than `passport-oauth2` which is rather outdated? I have an implementation I can share should you need, and I feel it’d be safer relying on OIDC-spec compliant libs

.env.sample
JaneJeon: And can we get some example `OIDC_TEAM_NAME` values for the popular OIDC providers like Auth0/Keycloak/MS/etc? It’s definitely not standard and it would be nice to know which fields would “count” as a team name and how that relates to the authorization decisions :)

server/auth/providers/oidc.js
```diff
@@ -0,0 +1,129 @@
...
+      // OpenID Connect standard profile claims can be found in the official
+      // specification.
+      // https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+
+      // Non-standard claims may be configured by individual identity providers.
+      // Any claim supplied in response to the userinfo request will be
+      // available on the `profile` parameter
+      async function (req, accessToken, refreshToken, profile, done) {
+        try {
+          const parts = profile.email.split("@");
```
JaneJeon, line 77: I don’t know about using user-provided email to “verify” domain ownership (something like `email_verified` would be more appropriate for it) as it’d be a security risk - you can pass whatever email you’d like at the point of signing up… Probably better to check `aud` or `issuer` of the JWT claim as they’re standard by the OIDC spec!

server/auth/providers/oidc.js
```diff
@@ -0,0 +1,129 @@
...
+              `Domain ${domain} is not on the whitelist`
+            );
+          }
+
+          const subdomain = domain.split(".")[0];
+
+          const result = await accountProvisioner({
+            ip: req.ip,
+            team: {
+              name: profile[OIDC_TEAM_CLAIM],
```
JaneJeon, line 95: Might want to use `lodash.get` here - you never know which provider nests this in their JWT (as it isn’t a standard claim)

.env.sample
JaneJeon: Plus using OIDC means being able to utilize super kEwL features like PKCE/client-credentials flow (which means you literally don’t even have to supply OIDC client secret!!)

JaneJeon: Plus there’s really no “standard” way of getting user info in OAuth2 - it all depends on the identity provider’s implementation (so for facebook/github/twitter/etc you’d all need different ways of getting the user, *plus* there’s no “standardized” way of returning users, whereas with OIDC there is a strict, documented standard) - e.g. see `passport-facebook`, `passport-twitter`, `passport-$literallyHalfTheAuth` which have to extend `passport-oauth2` with bespoke “get user info” code

.env.sample
tommoor: Alternatively, because this authentication provider is only for use by self hosted installations where they will be a single team we could hard code the team name to something which can be changed later￼￼ by the user.

That removes this complication from the config entirely￼

EternalDeiwos: @tommoor That sounds feasible. Would that be something one would change through the Outline interface?

@JaneJeon that would be a bit difficult as there are no standing examples of it. It depends on the precise requirements of individual users. For example, with keycloak, one is able to define mappers attached to scopes or directly to clients that modify the contents of tokens and user info. Every identity provider probably does this differently, if they support it at all.

server/auth/providers/oauth2.js
EternalDeiwos: Solved this by passing the existing errors forward.

.env.sample
EternalDeiwos: @JaneJeon I actually started with `openid-client` before moving to `passport-oauth2` because I was blocked due to issues arising. See #2369 for the discussion, including some discussion of PKCE. In general `oidc-client` is the more popular and better maintained library. Perhaps we can refactor to it once the standing issues have been dealt with.

Regarding "standard" userinfo; the way I have done it in this PR is the standard way according to the OIDC spec, which is to use the supplied userinfo endpoint, and the acquired access token to resolve a user info document containing claims.

Additionally, even if we did have PKCE, the authorization code flow is still superior to the implicit flow from a security perspective, however that does not mean that authorization code flow does not benefit from PKCE.

server/auth/providers/oidc.js
EternalDeiwos: This is a good idea

server/auth/providers/oidc.js
EternalDeiwos: In this case it is more for domain membership (I think). There's another env variable that allows you to limit access to users that have email addresses on a particular domain.

.env.sample
JaneJeon: as I've commented elsewhere in the PR, I believe the "proper" method would be to check against `aud`/`issuer`, not "team name", which are *guaranteed* to be there if you're using OIDC

JaneJeon: https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

server/auth/providers/oidc.js
JaneJeon: If a user completes a OIDC flow w/ your given client, they are already "in the system", meaning they already belong to the domain. If one wishes to *limit* the "pool" of people who can sign into outline, it would be better done by using the *client* to do so - after all, identity providers can configure clients to have access only to certain people/pools/etc, so by authenticating with the locked-down client you can limit who can login to outline.

.env.sample
JaneJeon: And again, id tokens (`sub`, `aud`, `issuer`, etc) - which is what you use to verify the identity of the user in a standardized manner - are *only* available for OIDC.

server/auth/providers/oauth2.js
EternalDeiwos: I've linked to the appropriate location in the spec in comments (same as the link above), however I would note here that part of my use case is specifically that one can manipulate your identity provider to provide whatever claims you want. This functionality will help later when looking at something like #2206. The goal eventually (for me) is authorization that is handled entirely by an external IdP.

.env.sample
EternalDeiwos: I disagree that the "proper" method would be to check against any of the existing claims because that would overload their purpose.

The fact is that "team name" refers to an attribute that is explicitly used by Outline and serves a specific purpose. Additionally we cannot predict the contents of userinfo documents provided by all the identity providers this functionality will be used with in the future; despite being standardized there are plenty inconsistencies between implementations. Therefore the solution is to either specify some other way; perhaps as a separate configuration variable entirely; or we must make it configurable as to where the information comes from.

Furthermore, I would also disagree with setting the default claim name for specifying the team name to something like `iss` or `aud` because, while that may be applicable for some, it may be a source of misconfiguration error for others. It should be possible to set `OIDC_TEAM_CLAIM=aud`, but if that is the behavior the user wants then they should make an explicit decision to do so.

server/auth/providers/oidc.js
EternalDeiwos: Outline uses many identity providers, some of which are social logins in their nature. This is simply to provide compatibility to an already existing access control mechanism; one doesn't _need_ to specify the `ALLOWED_DOMAINS` variable. That being said, for my own use case I am looking to control user authorization directly from the IdP as you describe, and thus would not be using this mechanism.

.env.sample
EternalDeiwos: One thing to consider here is that perhaps there shouldn't be a default at all... possibly the best behavior would be to remove the default and add an error prompting the user to configure it.

tommoor: > Would that be something one would change through the Outline interface?

Yes, it's already configurable. Honestly 99% of the reason this field exists is because the project is designed to be multi-tenant for cloud purposes. But as I mentioned before this is just not applicable in self-hosted (in fact, it's against the license terms) so it's fine to hard code this to "Wiki" or similar.

server/auth/providers/oidc.js
tommoor: I think if `issuer` is guaranteed to be a domain it would be okay to use that instead of splitting the email

EternalDeiwos: Issuer is a URI

.env.sample
EternalDeiwos: Aha. This is fair. Changes incoming.

JokerQyou:
Is there anything blocking this PR from being merged?

EternalDeiwos:
@JokerQyou I forgot to hit the "request re-review" button for about a week... 🤦‍♂️

EternalDeiwos:
It should show as long as `OIDC_CLIENT_ID` is specified. If it does not then there's a bug that needs fixing.

tommoor:
This is a limitation of the open source addition that needs addressing, if you already have a team in the database then the signin methods associated with that team will be used. It needs to be changed so that for OSS any configured signin method for the instance can be used

PR #2390, opened 2021-07-29
feat: Default to "recently viewed"

tommoor:
- When joining a team for the first time send the user to the default Collection directly. This improves the smoothness of onboarding experience by automatically expanding the sidebar too
- "Recently viewed" is a much more useful default for the Home screen

### DIFF
--- a/app/components/Tab.js
+++ b/app/components/Tab.js
@@ -13,7 +13,7 @@ type Props = {
 const NavLinkWithChildrenFunc = ({ to, exact = false, children, ...rest }) => (
   <Route path={to} exact={exact}>
     {({ match }) => (
-      <NavLink to={to} {...rest}>
+      <NavLink to={to} exact={exact} {...rest}>
         {children(match)}
       </NavLink>
     )}
--- a/app/scenes/Home.js
+++ b/app/scenes/Home.js
@@ -5,6 +5,7 @@ import * as React from "react";
 import { useTranslation } from "react-i18next";
 import { Switch, Route } from "react-router-dom";
 import { Action } from "components/Actions";
+import Empty from "components/Empty";
 import Heading from "components/Heading";
 import InputSearchPage from "components/InputSearchPage";
 import LanguagePrompt from "components/LanguagePrompt";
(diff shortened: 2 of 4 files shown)


iamsaumya (approved):
Nice feature!

iamsaumya:
Recently viewed stays active when we switch to other tabs. Is it the expected behaviour? 

https://user-images.githubusercontent.com/40148448/127549354-d1d12d1c-ea43-4f57-a888-08ef429bb10c.mp4

tommoor:
I think in this case it just happens that the same doc applies in all tabs because of the nature of the test data

iamsaumya:
I think that is because the first tab matches to `/home`?

tommoor:
I'm confident it's not broken as mechanism is the same as before 😬

iamsaumya:
Of course, It's not broken because of this PR 😉 . It caught my eyes while I was trying this locally (also reproducible on the starred page). 

I think passing the exact prop to `Navlink` will fix this bug.

https://github.com/outline/outline/blob/63eb8aadafcabd94015151eec7c2ff3c27ee2c47/app/components/Tab.js#L16

```<NavLink to={to} exact={exact} {...rest}>```

If you wish to handle this in separate PR, I'm up for it 😄

tommoor:
https://user-images.githubusercontent.com/380914/127648022-be448a5b-6fd7-494d-8eaa-fdedb6e5a133.mp4

The way that `Switch` works it matches the first route it comes across, so as long as the generic route is at the end it's fine – I haven't changed anything and Starred works. I may end up eating my words but I'm still convinced this is just your local of varied local data making it look broken

iamsaumya:
Sorry to push this further. The communication gap is at my end because I can see the issue in your video as well. 

When we navigate to the starred page, the `Alphabetical` is colored `#9BA6B2`, which signifies it is inactive. 

But when we click on the `Alphabetical` tab, the `Recently updated` stays at `#4E5C6E`, which is meant to be for the active tab. I think it should go to `#9BA6B2` (a lighter tone), signifying inactiveness.

tommoor:
I finally see what you mean, I thought you were referring to the content 🙈

tommoor:
Addressed, looks like it regressed when I introduced the border animation

PR #2392, opened 2021-07-29
fix: Enable offline access to google accounts

iamsaumya:
towards #2336

### DIFF
--- a/server/auth/providers/google.js
+++ b/server/auth/providers/google.js
@@ -87,7 +87,13 @@ if (GOOGLE_CLIENT_ID) {
     )
   );
 
-  router.get("google", passport.authenticate(providerName));
+  router.get(
+    "google",
+    passport.authenticate(providerName, {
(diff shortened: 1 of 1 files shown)


server/auth/providers/google.js
```diff
@@ -87,7 +87,13 @@ if (GOOGLE_CLIENT_ID) {
     )
   );
 
-  router.get("google", passport.authenticate(providerName));
+  router.get(
+    "google",
+    passport.authenticate(providerName, {
+      accessType: "offline",
+      prompt: "consent",
```
tommoor, line 94: Does this override L39?

https://github.com/outline/outline/blob/09623b4050e2fd51b6a914fbca3f90e2ae28ac91/server/auth/providers/google.js#L39

iamsaumya: It does override.

tommoor (approved):
✅  Tested

PR #2394, opened 2021-07-30
perf: Separate slow joins

tommoor:
Move from a LEFT OUTER JOIN to a separate query


tommoor:
This worked a treat:

image

PR #2399, opened 2021-07-30
feat: Support for github OAuth

rllola:
This PR add support for github OAuth and verify that your are member of a specific organization before letting you in.

In order to have it working you will need to create an OAuth app in github (best to create a organization OAuth app) and get the `GITHUB_CLIENT_ID` and the `GITHUB_CLIENT_SECRET`. `GITHUB_ORGANIZATION` is also required to restrict members access.


rllola:
It looks like I need to retrieve a `subdomain` info. I am not sure what exactly it refers to. More information on how to improve would be much appreciated.

tommoor:
Hey, I believe this functionality will be covered by the work in https://github.com/outline/outline/pull/2388 as GitHub appears to be OpenID compatible

rllola:
> Hey, I believe this functionality will be covered by the work in #2388 as GitHub appears to be OpenID compatible

Nice! I saw it but didn't realize that it could be used for Github.

I haven't tested it yet and will proceed to. I do have a doubt on the email part. Github will only share the public email unless you specified scope `user:email` **and** call a specific `/user/emails` (see https://docs.github.com/en/rest/reference/users#list-email-addresses-for-the-authenticated-user). I kind of learned that the hard way in another project.

rllola:
@tommoor 
Any thoughts ?

rllola:
Up ?

rllola:
I have tested the OpenID Connect feature using Github OAuth process. I can confirm that it is not working. The email is not found because my email is not public on my github profile so it return`null`.

The only way to have it working is to specify the in scopes field `user:email` **and** to call the dedicated email endpoint to get it. There is unfortunately no way around it.

tommoor:
Returning to this – the amount of code here seems quite manageable, I'm not sure what kind of team would be using GitHub but not any of the other authentication methods though. I suppose it is sometimes used as a login method for development tools so in that way it could make sense.

rllola:
> I'm not sure what kind of team would be using GitHub but not any of the other authentication methods though.
Well mine... We are only using Github for our project.

Since I have setup Authelia to connect to Outline via OpenID (but it is a lot of work). Github Oauth is broadly used across teams.

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
...
+            subdomain: GITHUB_ORGANIZATION,
+          },
+          user: {
+            name: profile.displayName,
+            email,
+            avatarUrl: profile.photos[0].value,
+          },
+          authenticationProvider: {
+            name: providerName,
+            providerId: profile.id,
```
tommoor, line 92: In your installation are you using it on your own? This looks like it would cause issues – the `providerId` here should be a unique id that represents the team on the authentication provider, so in this case it should be the GitHub organization id.

rllola: You are right! I can fix that this week end.

In the end, I am using OpenID with Authelia and not this feature because I wanted the PR to be reviewed. I just quickly tested it.

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
...
+          },
+          authentication: {
+            providerId: profile.id,
```
rllola, line 95: @tommoor I guess here it is the same ?

server/routes/auth/providers/github.js
rllola: Ref : https://docs.github.com/en/rest/reference/orgs

server/routes/auth/providers/github.js
tommoor: Nope, this is the user specific id so it should be the github user id

app/components/AuthLogo/GithubLogo.tsx
```diff
@@ -0,0 +1,26 @@
+// @flow
+import * as React from "react";
+
+type Props = {
+  size?: number,
+  fill?: string,
+  className?: string,
```
tommoor, line 7:
```suggestion
  size?: number;
  fill?: string;
  className?: string;
```

app/components/AuthLogo/GithubLogo.tsx
```diff
@@ -0,0 +1,26 @@
+// @flow
```
tommoor, line 1:
```suggestion
```

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
+// @flow
+import passport from "@outlinewiki/koa-passport";
+import fetch from "fetch-with-proxy";
+import Router from "koa-router";
+import { Strategy as GithubStrategy } from "passport-github2";
+import accountProvisioner from "../../../commands/accountProvisioner";
+import env from "../../../env";
+import passportMiddleware from "../../../middlewares/passport";
+import { StateStore } from "../../../utils/passport";
```
tommoor, line 9:
```suggestion
import accountProvisioner from "@server/commands/accountProvisioner";
import env from "@server/env";
import passportMiddleware from "@server/middlewares/passport";
import { StateStore } from "@server/utils/passport";
```

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
+// @flow
```
tommoor, line 1:
```suggestion
```

Also note this file needs a `.ts` extension

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export const config = {
+  name: "Github",
+  enabled: !!GITHUB_CLIENT_ID,
+};
+
+if (GITHUB_CLIENT_ID) {
+  if (GITHUB_ORGANIZATION === "") {
+    throw new Error("`GITHUB_ORGANIZATION` required when using Github.");
```
tommoor, line 42:
```suggestion
    throw new Error("`GITHUB_ORGANIZATION` required when using GitHub.");
```

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
...
+
+  if (response.status === 204) {
+    return {};
+  }
+
+  return response.json();
+}
+
+export const config = {
+  name: "Github",
```
tommoor, line 36:
```suggestion
  name: "GitHub",
```

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
...
+
+        if (isMember.message === "Not Found") {
+          throw new Error("User not a member of the organization.");
+        }
+
+        const result = await accountProvisioner({
+          ip: req.ip,
+          team: {
+            name: orgInfo.name,
+            domain: orgInfo.blog,
```
tommoor, line 81:
```suggestion
```

Best to leave this out as any GitHub org can set their blog to anything, this should be a trusted parameter

server/routes/auth/providers/github.js
```diff
@@ -0,0 +1,115 @@
...
+        );
+
+        if (isMember.message === "Not Found") {
+          throw new Error("User not a member of the organization.");
+        }
+
+        const result = await accountProvisioner({
+          ip: req.ip,
+          team: {
+            name: orgInfo.name,
```
tommoor, line 80:
```suggestion
            name: orgInfo.company || orgInfo.name,
```

Use the proper company name if available?

tommoor (changes requested):

razor950:
will this be reviewed?

razor950:
@tommoor will this be reviewed soon?

alexanderadam:
> Since I have setup Authelia to connect to Outline via OpenID (but it is a lot of work).

Did you have your Authelia + Outline setup documented somewhere?

rllola:
Hi @alexanderadam 

I actually did! Here the link : https://medium.com/@Lola_Dam/guide-for-installing-outline-with-authelia-as-an-openid-provider-129a141f6090

Feel free to ask questions here : https://github.com/outline/outline/discussions/2768

AprilNEA:
Is there anything blocking this PR from being merged?

PR #2401, opened 2021-07-31
fix: Refactor collection exports to not send email attachment

### LINKED ISSUE #2125: Refactor collection exports to not send email attachment
Sending exports as an email attachment was perhaps a neat idea in the early days but we've reached the limit of what this can offer. Collections and their images regularly exceed the size that can be reliably sent over email. 

### User flow

- Generate the file on the event queue
- Upload the file to private storage bucket
- Email the user a link to download the export (The link should require a session)

### Considerations

As a security measure the emailed link should not be directly to the file, it should be to the web app which requires a session and forwards to a link generated to access storage.

In the future we may want to create a table to record exports, in this way we could immediately return an `export id` from the API when an export is requested and allow the end user to query the export progress

iamsaumya:
closes #2125


app/scenes/Settings/ImportExport.js
```diff
@@ -107,6 +264,11 @@ function ImportExport() {
     : false;
   const isImportable = hasCollections && hasDocuments;
 
+  const data = orderBy(exports.allData, sort, direction.toLowerCase()).slice(
```
iamsaumya, line 267: In the case of `sort = size` and `direction = asc`, the results are actually sorted in desc order. I have spent hours figuring out today, IDK what's wrong :/

iamsaumya:
Almost there 🛩️

server/exporter.js
```diff
@@ -15,44 +19,112 @@ const queueOptions = {
   },
 };
 
-async function exportAndEmailCollections(teamId: string, email: string) {
+async function eventCreate(teamId, userId, exportData) {
+  await Event.create({
```
tommoor, line 23: We seem to be overloading the `collections.export_all` events here – it's a cool idea to send the progress of the export to the frontend in realtime but I think we can think of those events as just updates of the export/fileoperation model. e.g. `exports.update` / `operation.update` 

Also of note, every `Event.create` is stored in the database and accessible through the audit log, so all of these individual status updates would become an audit log. That's probably overkill, we only need one for the initial export and the rest can be sent over the message bus without being persisted. For that you can use `Event.add` which is the non-persisting alternative.

server/mailer.js
```diff
@@ -108,14 +112,22 @@ export class Mailer {
...
+    to: string,
+    attachments?: Object[],
+    id: string,
+    state: string,
+  }) => {
     this.sendMail({
       to: opts.to,
       attachments: opts.attachments,
       title: "Your requested export",
       previewText: "Here's your request data export from Outline",
```
tommoor, line 125: These don't make much sense if it failed, we might need to split the mailer into two to be honest.

server/emails/ExportEmail.js
```diff
@@ -8,25 +8,35 @@ import Footer from "./components/Footer";
...
-export const exportEmailText = `
+export const exportEmailSuccessText = `
 Your Data Export
 
-Your requested data export is attached as a zip file to this email.
+Your requested data export is completed, please visit the export section to find the link to zip file.
 `;
 
-export const ExportEmail = () => {
+export const exportEmailErrorText = `
```
tommoor, line 17: Lets split this into two templates,`ExportSuccessEmail` and `ExportFailureEmail` these are pretty different

server/emails/ExportEmail.js
```diff
@@ -8,25 +8,35 @@ import Footer from "./components/Footer";
...
+Your Data Export
+
+Your requested data export is failed, please visit the export section to  try again.
```
tommoor, line 20:
```suggestion
Sorry, your requested data export has failed, please visit the export section to  try again – if the problem persists please contact support.
```

server/api/utils.js
```diff
@@ -34,6 +35,27 @@ router.post("utils.gc", async (ctx) => {
 
   log(`Destroyed ${countDeletedDocument} documents`);
 
+  log(`Expiring all the collection export older than 30 days…`);
```
tommoor, line 38: This is great, sorry to be a pain but we'll need a test for this addition – wouldn't want current exports being deleted.

server/api/collections.js
```diff
@@ -454,12 +455,23 @@ router.post("collections.export", auth(), async (ctx) => {
...
     method: ["withMembership", user.id],
   }).findByPk(id);
   authorize(user, "export", collection);
 
-  const filePath = await archiveCollection(collection);
+  const filePath = await archiveCollections([collection]);
+
+  const stat = await fs.promises.stat(filePath);
+
+  await Export.create({
```
tommoor, line 468: This shows up a little weirdly in the table with no url – would you be willing to move this to the same async system?

app/components/Table.js
```diff
@@ -225,7 +250,13 @@ const Cell = styled.td`
   }
 `;
 
+const CellWrapper = styled(Flex)`
+  margin: 4px;
```
tommoor, line 254: This doesn't seem to be an improvement tbh 😬

app/models/Export.js
```diff
@@ -0,0 +1,27 @@
...
+  name: string,
+  id: string,
+  avatarUrl: string,
+};
+
+type CollectionData = {
+  name: string,
+  id: string,
+  url: string,
+};
```
tommoor, line 14: If we use the standard presenters on the server these aren't needed

app/components/Table.js
```diff
@@ -62,32 +67,52 @@ function Table({
...
+  const prevSortBy = React.useRef(sortBy);
 
   React.useEffect(() => {
-    onChangePage(0);
-    onChangeSort(
-      sortBy.length ? sortBy[0].id : undefined,
-      sortBy.length && sortBy[0].desc ? "DESC" : "ASC"
-    );
-  }, [sortBy]);
+    if (!isEqual(sortBy, prevSortBy.current)) {
```
tommoor, line 93: Would be good to get some context for these 3 hook changes

app/components/SocketProvider.js
```diff
@@ -287,6 +290,16 @@ class SocketProvider extends React.Component<Props> {
       }
     });
 
+    this.socket.on("collections.export_all", (event) => {
+      const user = auth.user;
+      if (user) {
+        exports.data.set(event.id, {
```
tommoor, line 296: Could we just use the eventPresenter here?

tommoor (changes requested):
Really great so far, thanks for the patience with my review time

server/api/collections.js
iamsaumya: Yeah sure 👍🏻

server/api/utils.js
iamsaumya: testing is good for saving headaches later 💪🏻

app/components/Table.js
iamsaumya: Not necessarily the best approach, but made the tables readable on mobile. 

<table>
	<tr> 
		<th> Before on mobile </th>
		<th> After on mobile </th>
	</tr>
<tr>
<td> 
</td>
<td> 
</td> 
</tr>
</table>

app/components/Table.js
iamsaumya: Hooks had missing dependencies, and If I pass all the dependencies, there were too many renders.

It was also not working if we direct navigate `page=2` through a URL. It was resetting to `page=0`.

There were a few things I can't recall exactly. I read through the docs and found that we could update the table state using the reducer.

app/components/Table.js
tommoor: Gotcha, I still think the `ListItem` approach might be more appropriate for the volume of items we'll have in this list  anyway, it's very unlikely to ever reach more than 30 (1 per day for a month), which is two pages worth of results.

iamsaumya: Ah, sure. Let's do it with `ListItem` 😄

server/models/Export.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+import { DataTypes, sequelize } from "../sequelize";
+
+const Export = sequelize.define("export", {
```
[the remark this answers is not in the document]
iamsaumya: We will do it generic with type `export` for now.  Is that what you want? Leaving import implementation for some other day?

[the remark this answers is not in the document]
tommoor: I'm open to push back on this if you don't think it's a good idea, but yea – definitely not suggesting to do anything to import at this time.

app/components/Table.js
tommoor: Could pull the `Table` improvements into a separate PR potentially

app/components/SocketProvider.js
iamsaumya: I'm afraid Idk what do you mean by `exportPresenter`?

server/api/collections.js
iamsaumya: Hey, I reflected on this. Isn't the export table only visible to admins? This is also something that the users have the habit of downloading right away.

tommoor: It's true today but large collections or collections with lots of images lead to the same issues - the file is too big to email and too large to generate sync on the http server.

I think it would be a reasonable compromise to make this functionality only available to admins to make it reliable

app/components/SocketProvider.js
tommoor: presenters/export

tommoor: Hmm true, I misread this – but then I don't really understand why we're spreading the current authenticated user here, adding the user data that triggered the export would be a server-side concern. That's what threw me 😄

tommoor: I have learnt the hard way in the past to not have different places serializing the same backend model, as our typing gets better it'll also be harder to type these payloads that are half built.

app/scenes/CollectionExport.js
```diff
@@ -33,7 +33,7 @@ function CollectionExport({ collection, onSubmit }: Props) {
       <form onSubmit={handleSubmit}>
         <HelpText>
           <Trans
-            defaults="Exporting the collection <em>{{collectionName}}</em> may take a few seconds. Your documents will be downloaded as a zip of folders with files in Markdown format."
+            defaults="Exporting the collection <em>{{collectionName}}</em> may take a few seconds. Your documents will be a zip of folders with files in Markdown format. Please visit the Export section on settings to get the zip."
```
iamsaumya, line 36: @tommoor, you will certainly have to change the wording here 😆

iamsaumya:
PR going big than I expected 😮

app/components/SocketProvider.js
iamsaumya: I will keep it in mind. Not doing the spreading anymore.

iamsaumya:
@tommoor, live implementation has a possible privacy breach. If a user has a private collection, the admins can view it by exporting the team data. It has private collections that the admin doesn't have access to inside the app.

tommoor:
That's not ideal, but thankfully not too bad - admins are presumed to have access to all data in most SaaS products.

server/services/websockets.js
```diff
@@ -327,6 +327,22 @@ export default class Websockets {
         }
         return;
       }
+
+      case "fileOperations.update": {
+        return socketio
+          .to(`user-${event.actorId}`)
+          .emit("fileOperations.update", {
```
tommoor, line 334: Same question as before, this can use `presenters/fileOperation` can't it?

server/policies/collection.js
```diff
@@ -71,6 +71,31 @@ allow(User, ["share", "export"], Collection, (user, collection) => {
   return true;
 });
 
+allow(User, "export", Collection, (user, collection) => {
+  if (user.isViewer) return false;
+  if (!collection || user.teamId !== collection.teamId) return false;
+  if (!collection.sharing) return false;
```
tommoor, line 77: Doesn't seem applicable? I guess this was a mistake before 🙈

server/api/utils.js
```diff
@@ -34,6 +35,28 @@ router.post("utils.gc", async (ctx) => {
...
+      },
+      state: {
+        [Op.ne]: "expired",
+      },
+    },
+  });
+
+  await Promise.all(
+    exports.map(async (e) => {
+      e.state = "expired";
```
tommoor, line 54: It might be nice to make this a model method, so we can just do `model.expire()`

server/migrations/20210730210120-add-fileOperations.js
```diff
@@ -0,0 +1,63 @@
...
+      createdAt: {
+        type: Sequelize.DATE,
+        allowNull: false,
+      },
+      updatedAt: {
+        type: Sequelize.DATE,
+        allowNull: false,
+      },
+    })
+  },
```
tommoor, line 58: We'll need to an index on the `state` field here as it's queried in `utils.gc`

server/exporter.js
```diff
@@ -15,45 +20,170 @@ const queueOptions = {
...
+  email: string,
+  collection?: Collection
+) => {
   exporterQueue.add(
     {
       type: "export-collections",
       teamId,
+      userId,
       email,
+      collection,
```
tommoor, line 186: I think this should be collectionId otherwise we're serializing the entire collection into redis here

server/exporter.js
```diff
@@ -15,45 +20,170 @@ const queueOptions = {
...
 }
 
 exporterQueue.process(async (job) => {
   log("Process", job.data);
 
   switch (job.data.type) {
     case "export-collections":
-      return await exportAndEmailCollections(job.data.teamId, job.data.email);
+      return await exportAndEmailCollections(
+        job.data.teamId,
```
tommoor, line 165: At this many options I'd consider converting to an object, less bug-prone in the long run

server/emails/ExportSucessEmail.js
```diff
@@ -8,25 +8,28 @@ import Footer from "./components/Footer";
...
   return (
     <EmailTemplate>
       <Header />
 
       <Body>
         <Heading>Your Data Export</Heading>
         <p>
-          Your requested data export is attached as a zip file to this email.
+          Your requested data export is completed, please visit the export
+          section to find the link to zip file.
```
tommoor, line 26:
```suggestion
          Your requested data export is complete, the exported files are now available in the admin section. 
```

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+
+const FileOperationListItem = ({ item }: FileOperationListItemProps) => {
+  const user = useCurrentUser();
+  const { t } = useTranslation();
+  const isMobile = useMobile();
+  const params = useQuery();
+  const key = params.get("key") || "";
+
+  return (
+    <ImportExportListItemWrapper highlight={key === item.id}>
```
tommoor, line 31: Can we use `components/List/Item` to cut down on some of the custom components here?

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+        <AvatarWrapper src={item.user.avatarUrl} size={32} />
+      </Flex>
+      <Content>
+        <Heading>
+          {item.collection ? (
+            <Title>
+              <Link to={item.collection.url}>{item.collection.name}</Link>
+            </Title>
+          ) : (
+            <Title>All collections</Title>
```
tommoor, line 42: Seems like translation hooks are missing 😄

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+  const user = useCurrentUser();
+  const { t } = useTranslation();
+  const isMobile = useMobile();
+  const params = useQuery();
+  const key = params.get("key") || "";
+
+  return (
+    <ImportExportListItemWrapper highlight={key === item.id}>
+      <Flex align="center" column>
+        <AvatarWrapper src={item.user.avatarUrl} size={32} />
```
tommoor, line 33: A big avatar seems overkill for this type of item, we could go with an icon or maybe nothing at all is needed.

Sidenote: I know there was no design given 😉

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+          )}
+          <Badge>{item.state}</Badge>
+        </Heading>
+        <Container>
+          <Flex>
+            {user.id === item.user.id ? t("You") : item.user.name} created&nbsp;
+            <Time dateTime={item.createdAt} addSuffix shorten />
+          </Flex>
+          {!isMobile && <>&nbsp;•&nbsp;</>}
+          <Flex>{(item.size / (1024 * 1024)).toPrecision(2)}MB</Flex>
```
tommoor, line 52: Can this be a computed property on the frontend model?

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+      </Content>
+
+      {item.url && (
+        <Actions>
+          <Button
+            onClick={() => window.open(item.url, "_blank")}
+            icon={<LinkIcon fill="currentColor" />}
+            neutral
+          >
+            {t("View")}
```
tommoor, line 63:
```suggestion
            {t("Download")}
```

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+          {!isMobile && <>&nbsp;•&nbsp;</>}
+          <Flex>{(item.size / (1024 * 1024)).toPrecision(2)}MB</Flex>
+        </Container>
+      </Content>
+
+      {item.url && (
+        <Actions>
+          <Button
+            onClick={() => window.open(item.url, "_blank")}
+            icon={<LinkIcon fill="currentColor" />}
```
tommoor, line 60: No icon needed here

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+          </Flex>
+          {!isMobile && <>&nbsp;•&nbsp;</>}
+          <Flex>{(item.size / (1024 * 1024)).toPrecision(2)}MB</Flex>
+        </Container>
+      </Content>
+
+      {item.url && (
+        <Actions>
+          <Button
+            onClick={() => window.open(item.url, "_blank")}
```
tommoor, line 59: The `url` is a direct link to the storage without any authentication so this isn't going to work in production setups – all of the uploaded files require a presigned url (the same as attachments).

I think we'll need to use the same approach here as attachments where the is an api url, eg `/api/file_operations.redirect?id=<uuid>` that would calculate and redirect to the presigned url on click.

This would also actually allow us to embed a link in the email which would work as long as the user is signed in – seems fine? 🤔

app/scenes/Settings/ImportExport.js
```diff
@@ -199,6 +200,17 @@ function ImportExport() {
           ? `${t("Requesting Export")}…`
           : t("Export Data")}
       </Button>
+      <Heading>
+        <Trans>Exports</Trans>
```
tommoor, line 204: Lets use `components/Subheading` – maybe "Recent exports" for the title?

app/components/FileOperationListItem.js
```diff
@@ -0,0 +1,158 @@
...
+      <Content>
+        <Heading>
+          {item.collection ? (
+            <Title>
+              <Link to={item.collection.url}>{item.collection.name}</Link>
+            </Title>
+          ) : (
+            <Title>All collections</Title>
+          )}
+          <Badge>{item.state}</Badge>
```
tommoor, line 44: We'll want to translate this, a mapping from state -> t("string"). I also think that "uploading" is kind of an odd one to show to the end user, it doesn't make much sense from their perspective so probably can translate that state to "Processing" ?

tommoor (changes requested):
Took a good look through, functionality is almost there once the signed url's are done (see attachments for a good example of that)

app/components/FileOperationListItem.js
iamsaumya: `creating` and `uploading` both could be mapped to `processing`. What do you think?

server/services/websockets.js
iamsaumya: I'm worried about passing the whole `user` object to the frontend.

app/components/FileOperationListItem.js
iamsaumya: Done 👍🏻

server/migrations/20210730210120-add-fileOperations.js
iamsaumya: Added the index on `state` and `type` because we use both the field while querying.

app/components/FileOperationListItem.js
tommoor: sounds good!

app/components/FileOperationListItem.js
iamsaumya: Not sure if it will allow us to embed a link because we have to hit Outline API using `POST`, Isn't it?

tommoor: You can use `GET`, there is some api middleware that handles this 👍

app/stores/FileOperationsStore.js
```diff
@@ -0,0 +1,37 @@
...
+  get orderedDataExports(): FileOperation[] {
+    return orderBy(this.exports, "createdAt", "desc");
+  }
+
+  async viewExport(fileOperationId: string) {
+    const res = await client.post("/fileOperations.redirect", {
+      id: fileOperationId,
+    });
+    invariant(res && res.data, "Data should be available");
+    window.open(res.data, "_blank");
```
iamsaumya, line 35: While redirecting, AWS was rejecting the request because the browser included a few extra headers and bearer tokens (`Only one type of authentication should be present` was the error). 

How would you circumvent this issue?

server/api/fileOperations.js
```diff
@@ -59,4 +61,24 @@ router.post("fileOperations.list", auth(), pagination(), async (ctx) => {
...
+  const team = await Team.findByPk(user.teamId);
+
+  const fileOp = await FileOperation.findByPk(id);
+  if (!fileOp) {
+    throw new NotFoundError();
+  }
+
+  authorize(user, fileOp.type, team);
+
+  const accessUrl = await getSignedUrl(fileOp.key);
```
iamsaumya, line 78: Now that we are using `key` to access the URL. I think we should remove the `url` attribute from the model. What do you think?

server/api/collections.js
```diff
@@ -454,11 +454,15 @@ router.post("collections.export", auth(), async (ctx) => {
...
+  const team = await Team.findByPk(user.teamId);
+  authorize(user, "export", team);
+
   const collection = await Collection.scope({
     method: ["withMembership", user.id],
   }).findByPk(id);
 
-  authorize(user, "export", collection);
+  ctx.assertPresent(collection, "Collection should be present");
+  authorize(user, "read", collection);
```
iamsaumya, line 465: Should at least have the read permission of collection.

server/api/fileOperations.js
tommoor: Agreed to remove from public responses

app/stores/FileOperationsStore.js
tommoor: You should be able to just make this a plain link something like `a href="/fileOperations.redirect?id=<>"`, perhaps that will resolve?

server/utils/s3.js
```diff
@@ -167,7 +167,7 @@ export const deleteFromS3 = (key: string) => {
     .promise();
 };
 
-export const getSignedImageUrl = async (key: string) => {
+export const getSignedUrl = async (key: string) => {
```
tommoor, line 170: 👍

server/api/fileOperations.js
```diff
@@ -59,4 +61,24 @@ router.post("fileOperations.list", auth(), pagination(), async (ctx) => {
...
+  ctx.body = {
+    data: accessUrl,
+  };
+});
```
tommoor, line 82: https://github.com/outline/outline/blob/a6d4d4ea3679c2159302a5bb8b08403b5a4a50ce/server/api/attachments.js#L150

app/stores/FileOperationsStore.js
iamsaumya: Works ✔️

server/services/websockets.js
iamsaumya: Should we emit `user` and `collection` objects? 

Observing the code in the file, my goal was to pass as little data as possible to the frontend.

tommoor: That's understandable, there are two reasons that we generally don't include the object in websocket messages –

1. scalability, when your team has 1000's of users that's just a lot more data over the wire
2. permissions – every team member has separate permissions and it's very easy to accidentally send them a websocket message with information they shouldn't have access to.

In this case with it only being sent down the single user channel I'm less wary of two cases, although the collection object in particular is still likely to be quite big. Which properties are we actually using at this point, is it just the `id` and `name` of each?

iamsaumya: We were using `name` and `url`. Because of the recent changes we only use `name` for now.

tommoor: In this scenario `collectionName` is totally okay rather than a partial `collection` object, sorry for the back and forth

tommoor:
@iamsaumya why the merge without final approval? 😢

iamsaumya:
Shit! I didn't know how the hell this happened.

iamsaumya:
Apologies for the incident. I don't know when this got merged. Perhaps it could be when my laptop was lagging, and I was trying to move the mouse. 😢

PR #2420, opened 2021-08-04
fix: Sidebar links highlighting issue when a template is deleted or archived.

### LINKED ISSUE #2419: Sidebar Link highlighting issue when a template is deleted or archived.
**To Reproduce**
1. Delete a template
2. Open the template 
3. Trash and template are highlighted in the sidebar

iamsaumya:
closes #2419

### DIFF
--- a/app/components/Sidebar/Main.js
+++ b/app/components/Sidebar/Main.js
@@ -149,7 +149,11 @@ function MainSidebar() {
                     exact={false}
                     label={t("Templates")}
                     active={
-                      documents.active ? documents.active.template : undefined
+                      documents.active
+                        ? documents.active.isTemplate &&
+                          !documents.active.isDeleted &&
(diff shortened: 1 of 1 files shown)


iamsaumya:
These additional of `isTemplate`, `isArchived` and all are the cause of these minor bugs, and we cannot always have all the edge cases in mind. 

I am wondering if I can come up with something around this. Something that gives a single point of truth.

tommoor:
Typically you'd have these active states driven off the route automatically, but for that to work you'd have to change the document route every time it changed state and we don't really want that. 

We could wrap up `!isArchived && i!sDeleted` into a single `isActive`, perhaps?

tommoor (approved):

iamsaumya:
Let's keep it as it is for now. If in the future we reencounter a bug because of this, we'll take a diligent look.

PR #2429, opened 2021-08-06
feat: Allow drafts to have no collection

### LINKED ISSUE #2122: Allow drafts to have no collection
Currently when creating a document you are first asked which collection it should go in. In some cases the choice is clear, but in others you don't want to make that decision immediately. Instead it would be better if creating a new document placed it in Drafts, and once you're ready to Publish a collection and location is chosen.

### Flow

- New document
- Document is created in drafts without collection
- "Publish" button is pressed
- New UI asks where to publish (collection and nesting). These options may be preset in the case of creating the document from somewhere like an existing document, but this would be a good chance to confirm.
 
### Considerations

- All permissions are built around collections and collection memberships right now, so the biggest concern would be changing these to allow for drafts without a collection and updating appropriate tests
- We will need to make `collectionId` on documents optional
- The breadcrumb for drafts and elsewhere in the UI will need updating to show collection optionally
- We will need to make `collectionId` on publish required as an API validation

iamsaumya:
Closes #2122


iamsaumya:
@tommoor, please give this a review. Especially to check if the approach and server-side changes are in the right direction.

iamsaumya:
Which endpoint should I send the `collectionId`? 

I didn't change `documents.update` because there is already a lot happening, so I took the easy path with `documents.restore` for now.

tommoor:
> I didn't change documents.update because there is already a lot happening, 

`documents.move` seems like it might be the right one?

iamsaumya:
I will do it through `documents.update` 👍🏻 . I will extract it to command as well.

server/api/documents.test.js
```diff
@@ -812,15 +812,15 @@ describe("#documents.drafts", () => {
...
+    const { document, collection, admin } = await seed();
     document.publishedAt = null;
     await document.save();
 
     collection.permission = null;
     await collection.save();
 
     const res = await server.post("/api/documents.drafts", {
-      body: { token: user.getJwtToken() },
+      body: { token: admin.getJwtToken() },
```
iamsaumya, line 823: admin is the user not present in the collection.

tommoor:
> Do we want the archived docs to move to trash when their collection is deleted?

I don't think so, it's unexpected behavior as once something is archived it no longer "feels" associated with the collection.

iamsaumya:
The current flow to publish a draft doc is different from the mock. Would you please give this a try? If it doesn't feel right, I'll update the modal to look like mock.

tommoor:
Sorry, I still haven't gone through this entirely yet – but I can say that it should be more similar to the mock in that I think it's important to make it a two step process where you select the location and then click "Publish" separately. 

I would still like to maintain the sharing of the selection component so it would probably be an improvement to make the "move" modal work the same way where you select the local and then there is a confirmation "Move document" button below to finish the process.

iamsaumya:
Understood 👍🏻

iamsaumya:
Would you please give this a review? 

- The modal pops up from the bottom in the mobile view because it's inside the `Header` component. 
The modal's width is bigger than the mock; otherwise, it was ellipsizing after two nested docs.

app/components/DocumentBreadcrumb.js
```diff
@@ -58,18 +58,9 @@ function useCategory(document) {
 
 const DocumentBreadcrumb = ({ document, children, onlyText }: Props) => {
   const { collections } = useStores();
-  const { t } = useTranslation();
   const category = useCategory(document);
 
-  let collection = collections.get(document.collectionId);
```
tommoor, line 64: This is interesting, I think that the Deleted Collection logic is still needed, but needs to be modified to check `document.collectionId` exists – after all docs in trash with deleted collections are still a thing.

app/components/DocumentBreadcrumb.js
```diff
@@ -111,7 +102,7 @@ const DocumentBreadcrumb = ({ document, children, onlyText }: Props) => {
   if (onlyText === true) {
     return (
       <>
-        {collection.name}
+        {collection && collection.name}
```
tommoor, line 105:
```suggestion
        {collection?.name}
```

I believe we can start to embrace the future

app/components/DocumentListItem.js
```diff
@@ -63,10 +63,9 @@ function DocumentListItem(props: Props, ref) {
   const queryIsInTitle =
     !!highlight &&
     !!document.title.toLowerCase().includes(highlight.toLowerCase());
-  const canStar =
-    !document.isDraft && !document.isArchived && !document.isTemplate;
+  const canStar = policies.abilities(document.id).star;
   const can = policies.abilities(currentTeam.id);
-  const canCollection = policies.abilities(document.collectionId);
+  const canCollection = policies.abilities(document.computedCollectionId);
```
tommoor, line 68:
```suggestion
  const canCollection = document.collectionId ? policies.abilities(document.collectionId) : {};
```

It's odd here that `canStar` is a boolean, but `canCollection` is an object 🤔

app/components/DocumentListItem.js
```diff
@@ -132,7 +131,7 @@ function DocumentListItem(props: Props, ref) {
             <>
               <Button
                 as={Link}
-                to={newDocumentUrl(document.collectionId, {
+                to={newDocumentUrl(document.computedCollectionId, {
```
tommoor, line 134:
```suggestion
                to={newDocumentUrl(document.collectionId, {
```

As mentioned in chat, lets update `newDocumentUrl` so that undefined is valid first argument

app/components/DocumentMeta.js
```diff
@@ -75,7 +75,7 @@ function DocumentMeta({
     return null;
   }
 
-  const collection = collections.get(document.collectionId);
+  const collection = collections.get(document.computedCollectionId);
```
tommoor, line 78:
```suggestion
  const collection = document.collectionId ? collections.get(document.collectionId) : undefined;
```

app/components/NewDocumentButton.js
```diff
@@ -0,0 +1,18 @@
...
+import { useTranslation } from "react-i18next";
+import { Link } from "react-router-dom";
+import Button from "components/Button";
+
+const NewDocumentButton = () => {
+  const { t } = useTranslation();
+
+  return (
+    <Button as={Link} to={"/doc/new"} icon={<PlusIcon />}>
+      {t("New draft doc")}
```
tommoor, line 13:
```suggestion
      {t("New doc")}
```

It's always been a draft first, no need for the addition here. I do like small reusable components like this, more please!

app/components/NewDocumentButton.js
```diff
@@ -0,0 +1,18 @@
...
+import * as React from "react";
+import { useTranslation } from "react-i18next";
+import { Link } from "react-router-dom";
+import Button from "components/Button";
+
+const NewDocumentButton = () => {
+  const { t } = useTranslation();
+
+  return (
+    <Button as={Link} to={"/doc/new"} icon={<PlusIcon />}>
```
tommoor, line 12: Could use `newDocumentUrl()` here when modified

app/components/PathToDocument.js
```diff
@@ -6,36 +6,30 @@ import styled from "styled-components";
...
+import Checkbox from "components/Checkbox";
 import CollectionIcon from "components/CollectionIcon";
 import Flex from "components/Flex";
 
 type Props = {
   result: DocumentPath,
   document?: ?Document,
   collection: ?Collection,
-  onSuccess?: () => void,
+  checked?: boolean,
```
tommoor, line 17:
```suggestion
  selected?: boolean,
```

Suggest selected - checked makes presumptions about the UI display

app/components/PathToDocument.js
```diff
@@ -6,36 +6,30 @@ import styled from "styled-components";
...
+  onSelect?: (DocumentPath) => void,
   style?: Object,
   ref?: (?React.ElementRef<"div">) => void,
 };
 
 @observer
 class PathToDocument extends React.Component<Props> {
-  handleClick = async (ev: SyntheticEvent<>) => {
+  handleClick = (ev: SyntheticEvent<>) => {
```
tommoor, line 25: I don't think this method is needed at all, could just attach `this.props.onSelect` directly – afterall we're passing in the `result` prop to just pass it back out again.

This component is super old, feel free to tear it apart of rebuild as you see fit.

app/components/PathToDocument.js
```diff
@@ -49,16 +43,23 @@ class PathToDocument extends React.Component<Props> {
...
+            .map((doc) => <Title key={doc.id}>{doc.title}</Title>)
+            .reduce((prev, curr) => [prev, <StyledGoToIcon />, curr])}
+          {document && (
+            <DocumentTitle>
+              {" "}
+              <StyledGoToIcon /> <Title>{document.title}</Title>
+            </DocumentTitle>
+          )}
+        </Flex>
+        {checked && (
```
tommoor, line 59: As discussed in chat, "selected" treatment can be a primary background state.

app/menus/DocumentMenu.js
```diff
@@ -156,7 +156,8 @@ function DocumentMenu({
     [document]
   );
 
-  const collection = collections.get(document.collectionId);
+  const collectionId = document.computedCollectionId;
+  const collection = collections.get(collectionId);
```
tommoor, line 160:
```suggestion
  const collection = document.collectionId ? collections.get(document.collectionId) : undefined;
```

app/menus/NewChildDocumentMenu.js
```diff
@@ -18,9 +18,13 @@ function NewChildDocumentMenu({ document, label }: Props) {
   const menu = useMenuState({ modal: true });
   const { collections } = useStores();
   const { t } = useTranslation();
-  const collection = collections.get(document.collectionId);
+  const collection = collections.get(document.computedCollectionId);
```
tommoor, line 21: As elsewhere

app/scenes/Document/components/Document.js
```diff
@@ -220,6 +222,12 @@ class DocumentScene extends React.Component<Props> {
     document.title = title;
     document.text = text;
 
+    if (options.collectionId) {
```
tommoor, line 225: We need to check authorization to publish into the collection and parent document here, `create` endpoint should have the needed logic given it was responsible for this before

app/scenes/Document/components/Header.js
```diff
@@ -65,17 +77,139 @@ function DocumentHeader({
...
+    }
     onSave({ done: true, publish: true });
-  }, [onSave]);
+  }, [dialog, hasCollection, onSave]);
+
+  const handleFilter = (ev) => {
+    setSearchTerm(ev.target.value);
+  };
+
+  const searchIndex = React.useMemo(() => {
```
tommoor, line 124: Way too much going on this component, please refactor the popover into it's own component 😄

app/scenes/Document/components/Header.js
```diff
@@ -170,7 +304,7 @@ function DocumentHeader({
                 <TemplatesMenu document={document} />
               </Action>
             )}
-            {!isEditing && (!isMobile || !isTemplate) && (
+            {!isEditing && !isMobile && !isTemplate && can.share && (
```
tommoor, line 307: Not sure what happened here, doesn't seem like it should be edited at all

app/scenes/Document/components/Header.js
```diff
@@ -286,4 +482,84 @@ const TocWrapper = styled(Action)`
...
+  padding: 10px;
+  box-shadow: ${(props) => props.theme.menuShadow};
+
+  ${breakpoint("mobile", "tablet")`
+    right: -2vh;
+    width: 90vw;
+`};
+`;
+
+const Input = styled.input`
```
tommoor, line 505: I think we can use a regular `InputSearch` here

app/scenes/DocumentMove.js
```diff
@@ -160,6 +196,9 @@ function DocumentMove({ document, onRequestClose }: Props) {
             </AutoSizer>
           </Results>
         </NewLocation>
+        <Flex justify="flex-end">
+          <Button onClick={handlePublish}>Publish</Button>
```
tommoor, line 200: It's "Move", not "Publish" here… also missing translation hooks 😉

server/api/documents.js
```diff
@@ -915,7 +910,7 @@ router.post("documents.star", auth(), async (ctx) => {
 
   const user = ctx.state.user;
   const document = await Document.findByPk(id, { userId: user.id });
-  authorize(user, "read", document);
+  authorize(user, "star", document);
```
tommoor, line 913: 🙈

server/api/documents.js
```diff
@@ -1030,84 +1027,81 @@ router.post("documents.update", auth(), async (ctx) => {
...
+    autosave,
+    done,
+    templateId,
+    append,
+    collectionId,
+    editorVersion,
+    parentDocumentId,
+  });
+
   if (publish) {
```
tommoor, line 1061: Move all of this event creation into the command

tommoor (changes requested):

tommoor:
Random other thing, definitely like that you changed the star policy – but as it's now correctly showing star on drafts the ui is the wrong way around here (star should be to right of label):

image

iamsaumya:
Thanks for the review! I will wrap this up swiftly. Please expect a request for 2nd round of review tomorrow 😄 .

app/scenes/Document/components/Header.js
iamsaumya: Previous logic will show `Share` on the mobile if the doc isn't a template.

It was collapsing with the TOC icon. What I inferred is that we don't want to show the `Share` on mobile.

tommoor: Okay, just the `&& can.share` that isn't needed – we always show the share button now, even if public sharing is disabled

app/components/CollectionIcon.js
```diff
@@ -14,12 +13,10 @@ type Props = {
 };
 
 function ResolvedCollectionIcon({ collection, expanded, size }: Props) {
-  const { ui } = useStores();
-
   // If the chosen icon color is very dark then we invert it in dark mode
   // otherwise it will be impossible to see against the dark background.
   const color =
-    ui.resolvedTheme === "dark" && collection.color !== "currentColor"
```
iamsaumya, line 22: Is it necessary to only have a color correction in dark mode?

app/stores/UsersStore.js
```diff
@@ -219,7 +219,7 @@ export default class UsersStore extends BaseStore<User> {
     return queriedUsers(users, query);
   };
 
-  inGroup = (groupId: string, query: string) => {
+  inGroup = (groupId: string, query: string): User[] => {
```
iamsaumya, line 222: There were warnings on return types is missing.

server/commands/documentUpdater.js
```diff
@@ -0,0 +1,100 @@
...
+  document.lastModifiedById = user.id;
+
+  let transaction;
+  let updatedDocument;
+  try {
+    transaction = await sequelize.transaction();
+
+    if (publish) {
+      if (!document.publishedAt && !document.collection && collectionId) {
+        document.collectionId = collectionId;
```
iamsaumya, line 40: Could we do policies checks inside the commands? I checked the other commands, and they didn't have the checks. Is it by design?

iamsaumya:
Read for review

server/commands/documentUpdater.js
tommoor: It's by design right now to do policy checks in controllers. It makes the commands more reusable – I'm happy to reconsider this more holisitically.

app/components/CollectionIcon.js
tommoor: There is an assumption here that light mode is the default I suppose, but I don't see a reason to remove this check in this PR

iamsaumya: If the collection icon has a default color, it remains black when selecting it in the light mode.
That's the reason for the update. Is there a better way to do this?

tommoor: Typically we've just target the svg in css with an override, a better approach might be to add a color prop to CollectionIcon that we can use to pass white in when selected

iamsaumya:
Not resuing the computation for showing rows was itching me. 

This is how I solved it for now. There is a slight change in the UI of `DocumentMove`, and everything works for now. Another possible solution to retain the UI is to use the `render props` design pattern. Let me know what you think. As always - Open to better suggestions.

iamsaumya:
Ready again for review 😄

iamsaumya:
Tom, Could you please resolve the merge conflict? I fear creating subtle bugs as I don't have an understanding of the implementation of Collab editing.

tommoor:
will do

iamsaumya:
I'm aware you are extremely busy. Thank you so much 🙏🏻

iamsaumya:
Let's move forward with this 😄. Please let me know if there's anything that needs polishing

almereyda:
This branch currently has conflicts. Could you rebase on main again, and offer that for review?

Since the original issue #2122 has been authored by @tommoor, I feel there is a chance this contribution can be accepted and incorporated into the Outline, now that the big structural changes of last year (Passport, TypeScript, Collaborative Editing) are in place and make the software more maintainable for the foreseeable future.

iamsaumya:
Ahh, a lot of conflicts to resolve. The branch is outdated, but the feat was almost there functionality-wise. Tom, Could you checkout the branch and point out any feat/flow oddities, please? I just need the green light on the look and feel. Resolving the conflict is the easiest part.

tommoor:
Not quite there yet, one last Typescript conversion branch to land – but yes we still want to land this, just a couple of unresolved issues over the Publish UI that need design work.

iamsaumya:
open

PR #2431, opened 2021-08-08
fix: prevent everyone in the team from viewing docs of private collections

iamsaumya:
Closes https://github.com/outline/outline/issues/2122#issuecomment-894682277

### DIFF
--- a/server/api/documents.js
+++ b/server/api/documents.js
@@ -575,6 +575,18 @@ async function loadDocument({
     }
 
     if (document.deletedAt) {
+      // if the doc collection is deleted, we check if the current user was present in the collection.
+      if (document.collectionId && !document.collection) {
+        const collectionIds = await user.collectionIds({ paranoid: false });
+
+        if (
+          !collectionIds.find(
+            (collectionId) => collectionId === document.collectionId
+          )
+        ) {
+          throw new AuthorizationError();
+        }
+      }
       authorize(user, "restore", document);
     } else {
       authorize(user, "read", document);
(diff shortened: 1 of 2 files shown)


server/api/documents.js
```diff
@@ -575,6 +575,18 @@ async function loadDocument({
     }
 
     if (document.deletedAt) {
+      // if the doc collection is deleted, we check if the current user was present in the collection.
```
tommoor, line 578: This is policy logic that has "escaped" into the controller layer, perhaps we could change the `withCollection` scope to include deleted collections? And that would achieve this 🤔

tommoor: We'll need test coverage for this leak too

iamsaumya: We can use the collection `update` policy which will help us keep the controller and policy logic separate. 

Suppose the team didn't want a user to view/edit a document while the collection was there. I guess they certainly wouldn't want the user to restore the doc just because of presence.

tommoor: Just checking, you're agreeing with me here to use the policy instead?

iamsaumya: I'm confused. Should we change withCollection scope or use Collection update policy?

iamsaumya: I'm using `update` policy here.

tommoor: This works and might be the way to go, but I'm wondering now that you have the test if it might make sense to achieve the same thing by including deleted collections in the `withCollection` scope…

https://github.com/outline/outline/blob/ed5320507d15d512a4869188c344c957f025df02/server/models/Document.js#L166

PR #2438, opened 2021-08-10
fix: Redirect to collection on self-hosted

iamsaumya:
I realised that I was not redirected to the collection when launching the app for the first time 😄

### DIFF
--- a/server/utils/authentication.js
+++ b/server/utils/authentication.js
@@ -4,7 +4,7 @@ import * as Sentry from "@sentry/node";
 import { addMonths } from "date-fns";
 import { type Context } from "koa";
 import { pick } from "lodash";
-import { User, Event, Team } from "../models";
+import { User, Event, Team, Collection, View } from "../models";
 import { getCookieDomain } from "../utils/domains";
 
 export function getAllowedDomains(): string[] {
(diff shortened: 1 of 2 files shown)


server/utils/authentication.js
```diff
@@ -99,6 +99,23 @@ export async function signIn(
       httpOnly: false,
       expires,
     });
-    ctx.redirect(`${team.url}/home${isNewUser ? "?welcome" : ""}`);
+
+    const [team, collection, view] = await Promise.all([
+      Team.findByPk(user.teamId),
+      Collection.findOne({
+        where: { teamId: user.teamId },
+        order: [["index", "ASC"]],
```
iamsaumya, line 107: Assures that we fetch the top-most collection.

tommoor:
Yea, I knew this was a difference between the two but honestly I just didn't want to replicate the logic in two places. First-run self-hosted experience isn't something I care too much about 🙈

server/utils/authentication.js
iamsaumya: Ohh! Copy paste mistake it is.

iamsaumya:
I see. So, would you like to go ahead with PR, or should I close it?

tommoor:
It's fine, you fixed my laziness

tommoor (approved):

PR #2450, opened 2021-08-14
fix: Improvements in table component

iamsaumya:


app/components/Table.js
```diff
@@ -24,6 +25,7 @@ export type Props = {|
   onChangePage: (index: number) => void,
   onChangeSort: (sort: ?string, direction: "ASC" | "DESC") => void,
   columns: any,
+  defaultSortDirection: "ASC" | "DESC",
```
tommoor, line 28: Because the sort direction is lowercase in url parameters it might

app/components/Table.js
```diff
@@ -124,7 +147,7 @@ function Table({
                       },
                     ])}
                   >
-                    {cell.render("Cell")}
+                    <CellWrapper>{cell.render("Cell")}</CellWrapper>
```
tommoor, line 150: The additional padding here causes the content to be misaligned with the headers:

image

I think the aim of improved spacing on mobile is good though, my suggestion would be:

- Apply padding to both the table `td` / `th` elements directly
- Remove the padding on first-child/last-child so that the left and right edges still align

tommoor (approved):

PR #2458, opened 2021-08-17
feat: Add hosted domain hint when signing in through Google SSO from subdomain

### LINKED ISSUE #2454: Restrict Google SSO from subdomains
When triggering SSO from a known subdomain it's possible to set the hosted domain that is displayed on the SSO prompt, this will better set expectations and stop users from logging into one team from the subdomain of another.

tommoor:
> The `hd` parameter streamlines the login process for G Suite hosted accounts. By including the domain of the G Suite user (for example, mycollege.edu), you can indicate that the account selection UI should be optimized for accounts at that domain

To be very clear – it is just a hint and can be bypassed or modified by the client, it provides no security – just a push in the right direction and some reassurance that they are at the right Outline subdomain.

closes #2454

ref: https://developers.google.com/identity/protocols/oauth2/openid-connect#hd-param

### DIFF
--- a/server/auth/providers/google.js
+++ b/server/auth/providers/google.js
@@ -10,6 +10,7 @@ import {
   GoogleWorkspaceInvalidError,
 } from "../../errors";
 import passportMiddleware from "../../middlewares/passport";
+import { AuthenticationProvider } from "../../models";
 import { getAllowedDomains } from "../../utils/authentication";
 import { StateStore } from "../../utils/passport";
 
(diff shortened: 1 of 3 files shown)


server/auth/providers/google.js
```diff
@@ -86,12 +86,12 @@ if (GOOGLE_CLIENT_ID) {
...
   );
 
-  router.get(
-    "google",
+  router.get("google", (ctx) =>
     passport.authenticate(providerName, {
       accessType: "offline",
       prompt: "select_account consent",
-    })
+      hd: ctx.request.query?.hd,
```
tommoor, line 93: I don't have a distinct reason why it's a problem, but it feels risky to forward this through – could be an avenue for modifying query values such as `state` potentially?

It's probably safer to pass an `id` to `/auth/google` and then re-query the db to get the hosted domain.

PR #2460, opened 2021-08-18
fix: Refactor collection exports to not send email attachment

### LINKED ISSUE #2125: Refactor collection exports to not send email attachment

iamsaumya:
closes #2125 

I accidentally merged PR #2401.


iamsaumya:
Ready for review 😄

tommoor:
UX thoughts after testing these changes…

- [ ] "Recent exports" header is shown when there are no exports, lets just hide the section in that scenario
- [ ] Paragraph of text above "Export Data" button mentions emailing the zip, lets tweak that text to say that we'll email when the export is complete.
- [ ] Spacing here is tight, maybe throw a `<br />` between
    image
- [ ] This size is strange, I think the method needs a little more attention at low file sizes…
      image
- [ ] The "Download" button on exports appears before the operation is complete, clicking it results in an error. hide?
- [ ] Listing "complete" here is actually a bit confusing when you didn't see the export process – there isn't a lot of context, I'm not sure if that means it should be removed so we only show status when it's incomplete, maybe? image
- [ ] When triggering an individual collection export there is no feedback, maybe we should forward the user to `/settings/import-export`?

app/models/FileOperation.js
```diff
@@ -0,0 +1,23 @@
...
+  state: string;
+  collection: ?Collection;
+  key: ?string;
+  size: number;
+  type: string;
+  user: User;
+  createdAt: string;
+
+  @computed
+  get sizeInMB(): string {
```
tommoor, line 18: As mentioned in comment, I think we need something slightly more sophisticated here that can deal with a wider range of sizes.

app/scenes/Settings/ImportExport.js
```diff
@@ -199,6 +211,46 @@ function ImportExport() {
...
+            key={item.id}
+            title={
+              item.collection ? item.collection.name : t("All collections")
+            }
+            subtitle={
+              <>
+                <Flex>
+                  {stateMapping[item.state]}&nbsp;•&nbsp;
+                  {user.id === item.user.id ? t("You") : item.user.name}{" "}
+                  created&nbsp;
```
tommoor, line 232: Missing translation hook here, we should make sure to include the subject to get a proper translation – something like:

```javascript
t(`{{userName}} requested`, { userName: user.id === item.user.id ? t("You") : item.user.name})
```

app/scenes/Settings/ImportExport.js
```diff
@@ -199,6 +211,46 @@ function ImportExport() {
...
+                onClick={() =>
+                  window.open(
+                    `/api/fileOperations.redirect?id=${item.id}`,
+                    "_blank"
+                  )
+                }
+                neutral
+              >
+                {t("Download")}
+              </Button>
```
tommoor, line 249: I _believe_ this will work

```suggestion
              <Button
                as="a"
                href={`/api/fileOperations.redirect?id=${item.id}`}
                neutral
              >
                {t("Download")}
              </Button>
```

server/api/collections.js
```diff
@@ -454,25 +454,21 @@ router.post("collections.export", auth(), async (ctx) => {
...
-    data: { title: collection.title },
-    ip: ctx.request.ip,
-  });
+  exportCollections(user.teamId, user.id, user.email, id);
 
-  ctx.attachment(`${collection.name}.zip`);
-  ctx.set("Content-Type", "application/force-download");
-  ctx.body = fs.createReadStream(filePath);
+  ctx.body = {
+    success: true,
```
tommoor, line 470: Could we return the file operation as `data` here? It would allow programmatically downloading the export once it's complete if the API consumer has the `id`

server/api/collections.js
```diff
@@ -482,26 +478,26 @@ router.post("collections.export_all", auth(), async (ctx) => {
...
   authorize(user, "export", team);
 
-  await Event.create({
-    name: "collections.export",
-    teamId: user.teamId,
-    actorId: user.id,
-    ip: ctx.request.ip,
-  });
-
   if (download) {
```
tommoor, line 481: Not sure it needs to be now but the plan will be to remove this flag entirely

server/api/fileOperations.js
```diff
@@ -0,0 +1,83 @@
...
+import { FileOperation, User, Collection, Team } from "../models";
+import policy from "../policies";
+import { presentFileOperation } from "../presenters";
+import { getSignedUrl } from "../utils/s3";
+import pagination from "./middlewares/pagination";
+
+const { authorize } = policy;
+const router = new Router();
+
+router.post("fileOperations.list", auth(), pagination(), async (ctx) => {
```
tommoor, line 14: If we add a `fileOperations.info` it will allow polling the state of the export via the API, automating data export is actually a surprisingly common usecase so I do believe it would be used.

server/api/fileOperations.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const user = ctx.state.user;
+
+  const where = {
+    teamId: user.teamId,
+    type,
+  };
+
+  const team = await Team.findByPk(user.teamId);
+  authorize(user, type, team);
```
tommoor, line 29: `type` is coming directly from user input here, we're allowing them to choose what authorization they like :P 

Can use `ctx.assertIn(type, ["export", "import"], ...)` to ensure it's one of those two values instead of `assertPresent`

server/api/fileOperations.js
```diff
@@ -0,0 +1,83 @@
...
+      ...ctx.state.pagination,
+      total,
+    },
+    data: exports.map(presentFileOperation),
+  };
+});
+
+router.post("fileOperations.redirect", auth(), async (ctx) => {
+  const { id } = ctx.body;
+  ctx.assertPresent(id, "id is required");
```
tommoor, line 66:
```suggestion
  ctx.assertUuid(id, "id is required");
```

Even more specific

server/api/fileOperations.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const user = ctx.state.user;
+  const team = await Team.findByPk(user.teamId);
+
+  const fileOp = await FileOperation.findByPk(id);
+  if (!fileOp) {
+    throw new NotFoundError();
+  }
+
```
tommoor, line 75: I think we can check if the fileOp.status !== complete and throw an error here

server/services/websockets.js
```diff
@@ -327,6 +327,15 @@ export default class Websockets {
         }
         return;
       }
+
+      case "fileOperations.update": {
+        return socketio
+          .to(`user-${event.actorId}`)
+          .emit("fileOperations.update", {
+            ...event.data,
+          });
```
tommoor, line 336:
```suggestion
          .emit("fileOperations.update", event.data);
```

server/presenters/fileOperation.js
```diff
@@ -0,0 +1,16 @@
...
+import { FileOperation } from "../models";
+import { presentCollection, presentUser } from ".";
+
+export default function present(data: FileOperation) {
+  return {
+    id: data.id,
+    type: data.type,
+    state: data.state,
+    collection: data.collection ? presentCollection(data.collection) : null,
+    key: data.key,
```
tommoor, line 11: Do we use `key`?  I think we can keep this private too

server/models/Event.js
```diff
@@ -94,6 +95,7 @@ Event.AUDIT_EVENTS = [
   "documents.delete",
   "documents.permanent_delete",
   "documents.restore",
+  "fileOperations.update",
```
tommoor, line 98:
```suggestion
```

No need, these won't be saved

server/emails/ExportSucessEmail.js
```diff
@@ -8,25 +8,30 @@ import Footer from "./components/Footer";
...
+          Your requested data export is complete, the exported files are also
+          available in the admin section.
         </p>
         <EmptySpace height={10} />
         <p>
-          <Button href={`${process.env.URL}/home`}>Go to dashboard</Button>
+          <Button
+            href={`${process.env.URL}/api/fileOperations.redirect?id=${id}`}
+          >
+            View my export
```
tommoor, line 33:
```suggestion
            Download
```

tommoor (changes requested):
Getting late, I might have a couple more things but this is the bulk of my feedback~

As you mentioned this feature was a lot bigger than we both expected, glad you're able to power through it.

iamsaumya:
Thanks for the review. I agree with all the points 👍🏻 .

> When triggering an individual collection export there is no feedback, maybe we should forward the user to /settings/import-export?

We send an email to the user when the export is ready. Redirecting them could lead to a scenario where the server already has a big queue to process, and the user will be staring at the progress of their export. At least, this is why I didn't propose to redirect them. What do you think?

tommoor:
Sounds like a toast message is the way to go then, something like "Export started, you will receive an email when it’s complete"

iamsaumya:
> * [ ]  Paragraph of text above "Export Data" button mentions emailing the zip, lets tweak that text to say that we'll email when the export is complete.

A full export might take some time, consider exporting a single document or collection if possible. The export data will be a zip of all your documents in Markdown format, and we’ll email a link to {{email}} when the export is complete.

Is this good?

tommoor:
> A full export might take some time, consider exporting a single document or collection. The exported data is a zip of your documents in Markdown format. You may leave this page once the export has started – we will email a link to {{email}} when complete.

app/models/FileOperation.js
iamsaumya: Noted. Earlier I neglected this, thinking collections would generally be greater than MB in production.

server/api/collections.js
iamsaumya: We are not using this in the app. I'm up for removing this. Is there something stopping us?

tommoor: It might be in use by API consumers, but honestly the whole reason this exercise is happening is to avoid running large exports on the API process so lets remove :)

server/api/collections.js
```diff
@@ -454,59 +455,64 @@ router.post("collections.export", auth(), async (ctx) => {
...
+      },
+    ],
   });
 
-  ctx.attachment(`${collection.name}.zip`);
-  ctx.set("Content-Type", "application/force-download");
-  ctx.body = fs.createReadStream(filePath);
+  ctx.body = {
+    success: true,
+    data: { fileOperation: presentFileOperation(exportData) },
```
iamsaumya, line 501: Using fileOperation as key instead of not presenting it directly because the response is from a collection endpoint, and fileOperation is different model.

iamsaumya:
> Listing "complete" here is actually a bit confusing when you didn't see the export process – there isn't a lot of context, I'm not sure if that means it should be removed so we only show status when it's incomplete, maybe?

I was giving this some thought because we mention a `complete` state of export in the messages stating we will send an email when it is reached. However, we are not showing the `Download` button until the complete state is reached. I think we can remove this tag when we reach `complete`.

iamsaumya:
Checklist complete. Please review 😄

server/api/collections.js
```diff
@@ -454,59 +455,64 @@ router.post("collections.export", auth(), async (ctx) => {
...
-  } else {
-    // async operation to create zip archive and email user
-    exportCollections(user.teamId, user.email);
-
-    ctx.body = {
-      success: true,
-    };
-  }
+  ctx.body = {
+    success: true,
```
tommoor, line 514: Return `fileOperation` here the same as above?

server/api/fileOperations.js
```diff
@@ -0,0 +1,123 @@
...
+  authorize(user, fileOperation.type, team);
+
+  if (!fileOperation) {
+    throw new NotFoundError();
+  }
+
+  ctx.body = {
+    data: presentFileOperation(fileOperation),
+  };
+});
```
tommoor, line 44:
```suggestion
});

```

server/api/fileOperations.js
```diff
@@ -0,0 +1,123 @@
...
+
+  if (!fileOp) {
+    throw new NotFoundError();
+  }
+
+  if (fileOp.state !== "complete") {
+    throw new ValidationError("file operation is not complete yet");
+  }
+
+  authorize(user, fileOp.type, team);
```
tommoor, line 116: The `authorize` call here should come before the state check above otherwise we're leaking the state of internal system to an unauthorized user

server/api/fileOperations.js
```diff
@@ -0,0 +1,123 @@
...
+        model: User,
+        as: "user",
+        paranoid: false,
+      },
+      {
+        model: Collection,
+        as: "collection",
+        paranoid: false,
+      },
+    ],
```
tommoor, line 32: Seems like this is repeated a lot, we can use sequelize's `defaultScope` in the model to make sure they are always included and remove this include statement in several places

server/migrations/20210730210120-add-fileOperations.js
```diff
@@ -0,0 +1,65 @@
...
+      createdAt: {
+        type: Sequelize.DATE,
+        allowNull: false,
+      },
+      updatedAt: {
+        type: Sequelize.DATE,
+        allowNull: false,
+      },
+    })
+    await queryInterface.addIndex('file_operations', ["state","type"]);
```
tommoor, line 58: This adds a single compound-index which would require "state" to be in the query, I don't think that's what you meant to do?

```suggestion
    await queryInterface.addIndex('file_operations', ["state"]);
    await queryInterface.addIndex('file_operations', ["type"]);
```

If not `removeIndex` below will also need changing

server/migrations/20210730210120-add-fileOperations.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeIndex('file_operations', ["state","type"]);
```
tommoor, line 62:
```suggestion
    await queryInterface.removeIndex('file_operations', ["state"]);
    await queryInterface.removeIndex('file_operations', ["type"]);
```

server/emails/ExportFailureEmail.js
```diff
@@ -0,0 +1,40 @@
...
+
+      <Body>
+        <Heading>Your Data Export</Heading>
+        <p>
+          Sorry, your requested data export has failed, please visit the export
+          section to try again – if the problem persists please contact support.
+        </p>
+        <EmptySpace height={10} />
+        <p>
+          <Button href={`${process.env.URL}/settings/import-export`}>
```
tommoor, line 31: We need to pass `teamUrl` and use it here (same as welcome email) instead of the `process.env.URL` otherwise on cloud hosted it will send them to `https://app.getoutline.com/settings/ where they won't be signed in.

server/emails/ExportSucessEmail.js
```diff
@@ -8,25 +8,30 @@ import Footer from "./components/Footer";
...
         <Heading>Your Data Export</Heading>
         <p>
-          Your requested data export is attached as a zip file to this email.
+          Your requested data export is complete, the exported files are also
+          available in the admin section.
         </p>
         <EmptySpace height={10} />
         <p>
-          <Button href={`${process.env.URL}/home`}>Go to dashboard</Button>
+          <Button
```
tommoor, line 30: We need to pass `teamUrl` and use it here (same as welcome email)  instead of the `process.env.URL` otherwise on cloud hosted it will send them to `https://app.getoutline.com/api/... where they won't be signed in.

server/mailer.js
```diff
@@ -14,7 +14,15 @@ import {
...
-import { ExportEmail, exportEmailText } from "./emails/ExportEmail";
+import {
+  ExportFailureEmail,
+  exportEmailFailureText,
+} from "./emails/ExportFailureEmail";
+
+import {
+  ExportSuccessEmail,
+  exportEmailSuccessText,
+} from "./emails/ExportSucessEmail";
```
tommoor, line 25: "Success" is spelt wrong in this filename, seems okay elsewhere

server/emails/ExportSucessEmail.js
```diff
@@ -8,25 +8,30 @@ import Footer from "./components/Footer";
...
   return (
     <EmailTemplate>
       <Header />
 
       <Body>
         <Heading>Your Data Export</Heading>
         <p>
-          Your requested data export is attached as a zip file to this email.
+          Your requested data export is complete, the exported files are also
+          available in the admin section.
```
tommoor, line 26: We should link "admin section"?

server/emails/ExportSucessEmail.js
```diff
@@ -8,25 +8,30 @@ import Footer from "./components/Footer";
 import Header from "./components/Header";
 import Heading from "./components/Heading";
 
-export const exportEmailText = `
+export const exportEmailSuccessText = `
 Your Data Export
 
-Your requested data export is attached as a zip file to this email.
+Your requested data export is complete, the exported files are also available in the admin section.
```
tommoor, line 14: Add a link to "admin section" below otherwise this text email is a dead end.

server/emails/ExportFailureEmail.js
```diff
@@ -0,0 +1,40 @@
...
+import EmptySpace from "./components/EmptySpace";
+import Footer from "./components/Footer";
+import Header from "./components/Header";
+import Heading from "./components/Heading";
+
+export const exportEmailFailureText = `
+Your Data Export
+
+Sorry, your requested data export has failed, please visit the export
+section to try again – if the problem persists please contact support.
```
tommoor, line 15: Add a link to "admin section" below otherwise this text email is a dead end.

server/emails/ExportFailureEmail.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const ExportFailureEmail = () => {
+  return (
+    <EmailTemplate>
+      <Header />
+
+      <Body>
+        <Heading>Your Data Export</Heading>
+        <p>
+          Sorry, your requested data export has failed, please visit the export
```
tommoor, line 26:
```suggestion
          Sorry, your requested data export has failed, please visit the admin
```

server/emails/ExportFailureEmail.js
```diff
@@ -0,0 +1,40 @@
...
+import EmailTemplate from "./components/EmailLayout";
+import EmptySpace from "./components/EmptySpace";
+import Footer from "./components/Footer";
+import Header from "./components/Header";
+import Heading from "./components/Heading";
+
+export const exportEmailFailureText = `
+Your Data Export
+
+Sorry, your requested data export has failed, please visit the export
```
tommoor, line 14:
```suggestion
Sorry, your requested data export has failed, please visit the admin
```

server/exporter.js
```diff
@@ -15,45 +19,188 @@ const queueOptions = {
...
+    collections = [await Collection.findByPk(collectionId)];
+  }
+
+  const acl = process.env.AWS_S3_ACL || "private";
+
+  let exportData;
+  let state;
+  let key;
+
+  if (fileOperationId) {
```
tommoor, line 83: I'm trying to follow why in one case the fileOperationId is created outside and passed in, but in the other it's inside of this method, are you able to explain?

Overall everything works well but this method feels "messy"

server/exporter.js
```diff
@@ -15,45 +19,188 @@ const queueOptions = {
...
   const team = await Team.findByPk(teamId);
-  const collections = await Collection.findAll({
-    where: { teamId },
-    order: [["name", "ASC"]],
-  });
+  const user = await User.findByPk(userId);
+
+  let collections;
+  if (!collectionId) {
+    collections = await Collection.scope({
```
tommoor, line 60: There is a convenience method for this, `await user.collectionIds();`, that calculates all of the permissions. So could grab the id's and then query collections with id in that list.

tommoor (changes requested):
I'm sorry the feedback is never ending on this one

app/stores/FileOperationsStore.js
```diff
@@ -0,0 +1,27 @@
...
+  @computed
+  get exports(): FileOperation[] {
+    return Array.from(this.data.values()).reduce(
+      (acc, fileOp) => (fileOp.type === "export" ? [...acc, fileOp] : acc),
+      []
+    );
+  }
+
+  @computed
+  get orderedDataExports(): FileOperation[] {
```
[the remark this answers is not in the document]
iamsaumya: I named it so believing in future we will need an `orderedData` for `import` as well.

server/api/collections.js
```diff
@@ -454,59 +455,64 @@ router.post("collections.export", auth(), async (ctx) => {
...
-    collectionId: collection.id,
-    teamId: user.teamId,
-    actorId: user.id,
-    data: { title: collection.title },
-    ip: ctx.request.ip,
+  const name = collection ? collection.name : team.name;
+  const acl = process.env.AWS_S3_ACL || "private";
+  const key = getAWSKeyForFileOp(team.id, name, acl);
+
+  let exportData;
```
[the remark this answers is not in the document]
iamsaumya: the `fileOperation` is of export type, maybe `exportFileOperation`?

server/exporter.js
iamsaumya: Modified both the endpoints to create fileOperation before putting tasks on the queue

server/migrations/20210730210120-add-fileOperations.js
iamsaumya: We use `type` and `state` attributes while querying. I thought a compound index would be must faster than having a single index.

Suppose we are not going with compound indexes. I think the `type` index is all that we need.  

```    
where: {
  type: "export",
  createdAt: {
    [Op.lt]: subDays(new Date(), 30),
  },
  state: {
    [Op.ne]: "expired",
  },
 },
 ```

iamsaumya: Noted.

tommoor (approved):

tommoor:
I think there is a follow up here to refactor the exporter to use the command pattern, it should clean things up quite a lot. But this was needed prior to this work so it's in no way a requirement for the PR.

iamsaumya:
Doing the merge 🤞🏻

PR #2469, opened 2021-08-21
feat: Always show share button

### LINKED ISSUE #2444: Show "Share" button when 'Public document sharing' is disabled for internal team sharing
### Discussed in https://github.com/outline/outline/discussions/2443

<sup>Originally posted by **victorhuberta** August 12, 2021</sup>
The sharing feature is very intuitive and easy to use when 'Public document sharing' for a collection is enabled. But when it's disabled, the 'Share' button disappears and the user has to manually copy the browser URL and send it to their teammate. Afterwards, the teammate opens the URL and is greeted with the same page. The problem is that unlike with the usual share link, this page displays the document with the sidebar and looks busy.

I think it would be great if the 'Share' button doesn't disappear when 'Public document sharing' is disabled, and the user can share the link in precisely the same way (but they cannot enable public view, so only for teammates). Then when their teammates click on the link, they will be greeted by the document in a clean manner, without the sidebar. What do you guys think?</div>

wvds:
Show the share button even when public sharing is disabled in the security settings.
It basically checks if the person who opens the share link is logged in and has read access. If that's the case it will display the document.

Let me know if there's anything that need to be changed or if there's other feedback!

closes #2443 
closes #2444

### DIFF
--- a/app/scenes/Document/components/Header.js
+++ b/app/scenes/Document/components/Header.js
@@ -79,7 +79,6 @@ function DocumentHeader({
   const isNew = document.isNewDocument;
   const isTemplate = document.isTemplate;
   const can = policies.abilities(document.id);
-  const canShareDocument = auth.team && auth.team.sharing && can.share;
   const canToggleEmbeds = auth.team && auth.team.documentEmbeds;
   const canEdit = can.update && !isEditing;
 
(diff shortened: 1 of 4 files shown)


tommoor (approved):
Fantastic 🙏

PR #2470, opened 2021-08-22
feat: Add support for oauth authentication with Discord.

FeralMib:
This PR add support for Discord oauth2 authentication and optionally checks if the user is in one of the allowed teams.

Not sure I've handled the teams / guild mapping (iterate over all?) appropriately however it works for my self-hosted use case.

Thanks for a fantastic wiki platform!


tommoor:
Hi @FeralMib, thanks for the contribution but we're not accepting any more authentication methods at this time

PR #2473, opened 2021-08-23
fix: Add ability to choose user permission level when inviting

### LINKED ISSUE #2113: Add ability to choose user permission level when inviting
It would be nice if we offered a dropdown when inviting a user to choose if they should have "view-only" permissions now this is a feature, the server model should support this without issue however the API will need to get the parameter added for when creating invites.

iamsaumya:
Closes #2113


app/scenes/Invite.js
```diff
@@ -26,15 +27,16 @@ type Props = {|
 type InviteRequest = {
   email: string,
   name: string,
+  role: string,
```
tommoor, line 30:
```suggestion
  role: "member" | "viewer" | "admin",
```

lets type this a little more strictly, there's a few spots in this PR

app/scenes/Settings/components/UserInviteMenu.js
```diff
@@ -0,0 +1,53 @@
...
+      />
+    </Wrapper>
+  );
+};
+
+const Wrapper = styled(Flex)`
+  margin: 0px;
+  position: relative;
+  flex-shrink: unset;
+  top: ${(props) => (props.setTop ? "28px" : "0")};
```
tommoor, line 50: I think using the select input will help avoid this odd api

app/scenes/Settings/components/UserInviteMenu.js
```diff
@@ -0,0 +1,53 @@
...
+        key: "admin",
+        label: t("Admin"),
+      },
+    ],
+    [t]
+  );
+
+  return (
+    <Wrapper setTop={setTop}>
+      <FilterOptions
```
tommoor, line 36: Can we use the select input component for this?

server/commands/userInviter.js
```diff
@@ -52,6 +52,8 @@ export default async function userInviter({
       name: invite.name,
       email: invite.email,
       service: null,
+      isAdmin: invite.role === "admin",
```
tommoor, line 55: We need to check the `inviteUser` policy in this api method to ensure only admins can invite.

Looks like this was an existing bug which was bad, but with the ability to create admins it would become a privilege escalation issue.

app/scenes/Settings/components/UserInviteMenu.js
```diff
@@ -0,0 +1,53 @@
...
+import FilterOptions from "components/FilterOptions";
+import Flex from "components/Flex";
+
+type Props = {|
+  activeKey: string,
+  onSelect: (key: ?string) => void,
+  setTop: boolean,
+|};
+
+const UserInviteMenu = ({ activeKey, onSelect, setTop }: Props) => {
```
tommoor, line 14: Maybe `InputSelectRole`, see `InputSelectPermission` as an example

tommoor (changes requested):
Great start, it works

app/scenes/Invite.js
```diff
@@ -182,7 +192,13 @@ function Invite({ onSubmit }: Props) {
...
             value={invite.name}
             required={!!invite.email}
-            flex
+          />
+          &nbsp;
+          <InputSelectRole
+            onChange={(ev) => handleRoleChange(ev, index)}
+            value={invite.role}
+            labelHidden={index !== 0}
+            short
```
iamsaumya, line 201: `Send Invites` button is not aligned with the `Select`. Either we will have to increase the modal size, or we can absolutely position the button below the select.

How would you go about this?

tommoor: Lets create an issue to implement the custom select with aim of fixing this

PR #2485, opened 2021-08-24
fix: Draft doc are readable by collection member through URL

iamsaumya:


tommoor:
This is actually by design 😬 – so you can share an in-progress draft with team members. It's possible we could lock it down further when per-document sharing is built

iamsaumya:
Ohh 🤦🏻‍♂️

iamsaumya:
Should they be able to update and delete it too?

tommoor:
Update is okay, for collaboration – delete maybe not? This issue is going to disappear with your other PR anyway

PR #2488, opened 2021-08-25
Add support for authentication with Telegram.

FeralMib:
This PR add support for Telegram authentication, automatically enrolling the user in the first of the "allowed" teams.

Telegram auth integration was much harder / messier than Discord (https://github.com/outline/outline/pull/2470) as they don't provide a real oauth2 workflow. Typically you need to use their js library which creates a button automatically in an iframe (in their theme, white background, specific size) which will run your callback on auth, similar to oauth2 but not quite fully compatible.

Thankfully their library has a function to handle the same auth workflow with your own button, but getting the right data for it took a bit more plumbing in.

There's probably a cleaner way this could have been done, but it's my first time using React so had to figure it out as I went. Either way it's working pretty well for me as it stands!

### DIFF
--- a/app/components/AuthLogo/TelegramLogo.js
+++ b/app/components/AuthLogo/TelegramLogo.js
@@ -1,0 +1,31 @@
+// @flow
+import * as React from "react";
+
+type Props = {
+  size?: number,
+  fill?: string,
+  className?: string,
+};
+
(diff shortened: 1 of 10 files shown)


tommoor:
Thanks @FeralMib – this is here for anyone that wants it but we're not going to maintain within the main codebase for now.

PR #2496, opened 2021-08-27
fix: Move "public document sharing" to "Permissions"

### LINKED ISSUE #2483: Move "public document sharing" to "Permissions"
- Currently the setting for "Public document sharing" in a collection is inside the "Edit…" modal, it would be better placed in the "Permissions" modal below the default visibility and before individual member access.

image

### LINKED ISSUE #2487: Add `collections.permission_changed` event
Currently in the event log it's not possible to tell the difference between someone editing a collection title/description or completely changing the access permission, we need to separate these two actions into different events.

iamsaumya:
closes #2483 
closes #2487


server/events.js
```diff
@@ -145,6 +145,18 @@ export type CollectionEvent =
...
       data: { index: string },
       ip: string,
+    }
+  | {
+      name: "collections.permission_changed",
+      collectionId: string,
+      teamId: string,
+      actorId: string,
+      data: {
+        index: string,
```
tommoor, line 155:
```suggestion
```

server/api/collections.js
```diff
@@ -562,9 +563,17 @@ router.post("collections.update", auth(), async (ctx) => {
     collection.color = color;
   }
   if (permission !== undefined) {
+    // frontend sends empty string
+    ctx.assertIn(
+      permission,
+      ["read_write", "read", "", null],
```
tommoor, line 569:
```suggestion
      ["read_write", "read", ""],
```

Can it really be `null`?

iamsaumya: Some tests pass `null` I didn't update them with `""` thinking it must be some legacy feat that we are taking care of.

tommoor (approved):

PR #2502, opened 2021-08-27
fix: Improvements in share feat

iamsaumya:
- No `shares.create` request made when a doc is viewed
- Use `read` policy in `shares.update` to allow users to create links for team sharing 
- Update `allow(User, "update", Share)` policy to check document sharing ability of user.

### DIFF
--- a/server/api/shares.js
+++ b/server/api/shares.js
@@ -3,7 +3,7 @@ import Router from "koa-router";
 import Sequelize from "sequelize";
 import { NotFoundError } from "../errors";
 import auth from "../middlewares/authentication";
-import { Document, User, Event, Share, Team } from "../models";
+import { Document, User, Event, Share, Team, Collection } from "../models";
 import policy from "../policies";
 import { presentShare, presentPolicies } from "../presenters";
 import pagination from "./middlewares/pagination";
@@ -18,7 +18,9 @@ router.post("shares.info", auth(), async (ctx) => {
 
   const user = ctx.state.user;
   let shares = [];
-  let share = await Share.findOne({
+  let share = await Share.scope({
+    method: ["withCollection", user.id],
+  }).findOne({
     where: id
       ? {
           id,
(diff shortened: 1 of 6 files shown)


server/api/shares.js
```diff
@@ -190,7 +209,8 @@ router.post("shares.create", auth(), async (ctx) => {
   const user = ctx.state.user;
   const document = await Document.findByPk(documentId, { userId: user.id });
   const team = await Team.findByPk(user.teamId);
-  authorize(user, "share", document);
+  // user could be creating the share link to share with team members
+  authorize(user, "read", document);
   authorize(user, "share", team);
```
tommoor, line 214: I think we can lose this line as it again refers to public sharing

server/api/shares.js
```diff
@@ -147,7 +147,26 @@ router.post("shares.update", auth(), async (ctx) => {
   ctx.assertUuid(id, "id is required");
 
   const { user } = ctx.state;
-  const share = await Share.findByPk(id);
+
+  // fetch the share with document and collection.
+  const share = await Share.findByPk(id, {
```
tommoor, line 152: Consider moving this into the model as a named scope to keep the controllers clean. Interestingly the default scope includes the user and team too…

server/models/Share.js
```diff
@@ -44,6 +44,27 @@ Share.associate = (models) => {
...
+            {
+              model: models.Collection.scope({
+                method: ["withMembership", userId],
+              }),
+              as: "collection",
+            },
+          ],
+        },
+        { association: "user", paranoid: false },
+        { association: "team" },
```
iamsaumya, line 64: I was confused about whether to include these associations or not. In the end, took the decision to include otherwise we will have to scope `defaultScope` and `withCollection` scope in each query.

There is another possible solution to include collection memberships in the default scope but the fetched memberships will be left unused.

tommoor (approved):

PR #2514, opened 2021-08-30
perf: Move exports to worker service

### LINKED ISSUE #2509: Move exporter heavy lifting to worker service
Follow on to #2495 and #2125 – the heavy lifting for export operations should happen in the worker service, currently it's happening async via the queue but in the web service as the exporter queue is processed there.

I believe it should be refactored like so:

- API controller should call a command
- The command should create a `FileOperation` and an `Event`, `collections.export`/`collections.export_all` with the file operation.id.
- There should be an `ExportProcessor` that listens for these events
- The old "exporter" queue should be removed

tommoor:
Moves heavy lifting of queries and zip creation to the new worker service. Also in this nice refactor: less code, using the command pattern, and removal of a custom queue.

closes #2509


tommoor:
@iamsaumya don't hate me for this refactor so soon after you landed code, I really needed to make sure that the zip etc is built on worker.

iamsaumya:
Not at all 😄. I loved the way you have re-written the code, especially the use of `fileOperation.update()` and fetching the `collectionIds`👌🏻 

LGTM ✅

iamsaumya (approved):
If you rerun the app, we can add the updated translations of PR.

PR #2516, opened 2021-08-30
fix: Warning when dragging document between collections with different user permissions

### LINKED ISSUE #2150: Warning when dragging document between collections with different user permissions
When moving a document today it just happens immediately, It would be nice if when moving a document from a collection with limited access to another with different access we would show a confirmation dialog that the user intended to change the permission level.

iamsaumya:
closes #2150


app/components/Sidebar/components/CollectionLink.js
```diff
@@ -67,9 +77,19 @@ function CollectionLink({
...
     drop: (item, monitor) => {
+      const { id, permission } = item;
       if (monitor.didDrop()) return;
       if (!collection) return;
-      documents.move(item.id, collection.id);
+      if (collection.id === item.collectionId) return;
+      if (
+        permission !== collection.permission &&
+        (permission === null || collection.permission === null)
+      ) {
```
tommoor, line 87: I think we need to make this so the modal is only triggered when the new permission is more open than the old permission, eg. only in the case of moving from `no default access` -> `read default access`.

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -203,6 +223,17 @@ function CollectionLink({
             index={index}
           />
         ))}
+      <Modal
+        title={t("Document Move")}
```
tommoor, line 227:
```suggestion
        title={t("Move document")}
```

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -203,6 +223,17 @@ function CollectionLink({
...
+        onRequestClose={() => handlePermissionClose()}
```
tommoor, line 228:
```suggestion
        onRequestClose={handlePermissionClose}
```

app/components/Sidebar/components/CollectionLink.js
```diff
@@ -203,6 +223,17 @@ function CollectionLink({
...
+        isOpen={permissionOpen}
+      >
+        <DocumentReparent
+          item={itemRef.current}
+          collection={collection}
+          onSubmit={() => handlePermissionClose()}
```
tommoor, line 234:
```suggestion
          onSubmit={handlePermissionClose}
```

app/components/Sidebar/components/DocumentLink.js
```diff
@@ -128,7 +128,14 @@ function DocumentLink(
...
   const [{ isDragging }, drag] = useDrag({
     type: "document",
-    item: () => ({ ...node, depth, active: isActiveDocument }),
+    item: () => ({
+      ...node,
+      depth,
+      active: isActiveDocument,
+      permission: collection?.permission,
+      collectionId: collection?.id,
+      collectionName: collection?.name,
```
tommoor, line 137: We should store only the `id` here, if the name and other details are needed in the modal then they can be loaded from the collections store?

tommoor (commented):
Thank you

iamsaumya:
Ohh, I didn't know it is a conflict here 😅

tommoor (approved):

PR #2531, opened 2021-09-05
Allow users to customize the OIDC team name

soulteary:
The latest [merged code](https://github.com/outline/outline/pull/2388) I looked at included OIDC support, but the team name was hardly written in the code.

Perhaps, you can use environment variables to allow users to set the team name when customizing OIDC.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -84,6 +84,8 @@ OIDC_CLIENT_SECRET=
 OIDC_AUTH_URI=
 OIDC_TOKEN_URI=
 OIDC_USERINFO_URI=
+# optional, used to set the team that the user belongs to, which will be displayed on the interface.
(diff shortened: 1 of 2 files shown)


tommoor:
Hi @soulteary – it's not necessary as you can already customize the team name in the settings area, the hardcoded value for OIDC is just an initial default.

soulteary:
I tested it again and it did. In addition, I have made an example of using outline quickly with traefik. Maybe I will adjust it later. Can I replace the development steps in the README file as a contributor to quickly get started with outline debugging?

https://github.com/soulteary/docker-outline


And, during the test, I found that the outline occasionally appeared and the first user was not set as the administrator. If it reappears later, I will give feedback again.

@tommoor Thanks for your correction

Issue #2548, opened 2021-09-10
Local development environment issues


soulteary:
**To Reproduce**

0. I've test it on macos 11.4 with node 14 / node 16, docker node14-alpine, node14
1. git clone a clean copy.
2. run `npm run build` or `npm run build:webpack`
3. trigger build error, regardless of whether you set `NODE_ENV=production`

```bash
yarn build:webpack
yarn run v1.22.11
$ webpack --config webpack.config.prod.js
Hash: f9489632113e5b6e87c9
Version: webpack 4.44.1
Time: 635ms
Built at: 2021/09/11 上午1:33:54
 4 assets
Entrypoint main = runtime.329aa36197b1b7ec6528.js runtime.329aa36197b1b7ec6528.js.map main.061a07767b3c1a3c197e.js main.061a07767b3c1a3c197e.js.map
[0] multi ./app/index 28 bytes {main} [built]
[xATA] ./app/index.js 143 bytes {main} [built] [failed] [1 error]

ERROR in ./app/index.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: EISDIR: illegal operation on a directory, read
 @ multi ./app/index main[0]
Child html-webpack-plugin for "index.html":
     1 asset
    Entrypoint undefined = index.html
    [YuTi] (webpack)/buildin/module.js 497 bytes {0} [built]
    [la58] ./node_modules/html-webpack-plugin/lib/loader.js!./server/static/index.html 1.87 KiB {0} [built]
    [yLpj] (webpack)/buildin/global.js 472 bytes {0} [built]
        + 1 hidden module
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Outline code branch: `main`, https://github.com/outline/outline/commit/231ab2da03b81c13110e8b09aaaea632c06430fc

soulteary:
In fact, I originally planned to provide a clean and quick-start local environment for problems like [#2422](https://github.com/outline/outline/issues/2422), and then discovered this problem.

But I see that the CI environment is normal. After repeated testing, I think it is necessary to inform the development team about this, because it may prevent the intervention of general developers unless they directly conduct secondary development based on the ancient docker image.

soulteary:
I found the problem. The outline in the docker-compose.yml configuration used in the project is because the way of mapping local and docker files is a bit outdated.

After the docker container is started, `npm run dev` will create a package.json folder by mistake, causing the above problems.

[...]

PR #1660, opened 2020-11-17
Collaborative editing

tommoor:
This branch was recently (Aug 2021) force-pushed to reflect a new approach to this feature.

- [ ] Alpha release (merge to `main`)

- [ ] Add ability to flag feature for admin in API
- [ ] Reduce `documents.info` fetches
- [ ] Add features UI in admin
- [ ] Image upload decorations

- [ ] Restore document functionality
- [ ] Redis scaling to multi-server
- [ ] Handle permissions changing while multiplayer is connected
- [ ] Handle `documents.update` API request should update CRDT state

### Done

- [x] Documentation for running multiplayer server
- [x] Sync failure after edit while offline and reconnect (https://github.com/ueberdosis/hocuspocus/issues/183)
- [x] Embeds not created when transforming to CRDT 
- [x] Content duplication issue (if converted to CRDT and not saved on server)
- [x] Move authentication from headers to websocket message (https://github.com/ueberdosis/hocuspocus/pull/163)
- [x] Refactor service architecture
- [x] Loading state when server takes >300ms to respond
- [x] Events added in multiplayer process not propagating to processors
- [x] Permission denied state/handling
- [x] Refactor so revision history, archived, deleted do not interact with multiplayer service
- [x] Tasks not updating with remote edit
- [x] Updating of todo tasks locally
- [x] Block restore document during beta
- [x] Add client storage
- [x] Show UI when websocket is disconnected to indicate lack of sync
- [x] Hook up awareness presence to avatar display
- [x] Add features validation
- [x] Templates functionality
- [x] read-only permissions
- [x] Bug: Documents being marked as "updated" when converting to CRDT
- [x] Move back to main `y-prosemirror` trunk away from personal fork
- [x] Improve "reconnecting…" state a little
- [x] Multi-process tracking
- [x] Bug: `url` not returned for new documents
- [x] Bug: "Publish" button for drafts leaves edit but doesn't publish
- [x] "New" badge appears erroneously on drafts
- [x] Bug: Content duplicated after sleep (may already be fixed, need to triple test)
- [x] Refactor so dhared docs do not interact with multiplayer service

related #811

PR #2549, opened 2021-09-11
Add method for local development

soulteary:
As reported in the [previous issue](https://github.com/outline/outline/issues/2548), the compose and docker file files in the current project cannot be implemented in the local environment, and all use docker for rapid development.

Developers need to install too many dependencies, configure additional authorization services, and distract their efforts outside the outline.

Considering that the current configuration file is still used by other processes, I added two additional files as a new solution for local development. I don't know if it can help the project.

@tommoor

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,12 @@
+FROM node:14-alpine
+
+WORKDIR /opt/outline
+
+COPY ./package.json ./yarn.lock ./
+
+RUN yarn install --no-optional --frozen-lockfile && yarn cache clean
+
+COPY . .
+
+USER root
+
--- a/README.md
+++ b/README.md
@@ -82,7 +82,11 @@ yarn run upgrade
 
 ## Local Development
 
-For contributing features and fixes you can quickly get an environment running using Docker by following these steps:
+For contributing features and fixes you can quickly get an environment running using Docker by using two ways.
+
+### Method A
+
+Following these steps:
 
 1. Install these dependencies if you don't already have them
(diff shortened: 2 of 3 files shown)


soulteary:
Due to the https://github.com/outline/outline/pull/1660 change of the main code, a file merge conflict occurred, and the code was rebase and [force pushed](https://github.com/outline/outline/compare/5e39c9e1dc4f196ac00dec68c9b39ba291e08b6a..5ebd40e64cef3f7cd388d35b820da2bc9f00b2e9).

tommoor:
Hi, I definitely agree that the setup needs to be better and more "one click" but I don't think the answer is to add an additional development flow - 

I'd rather keep a single dev setup for everyone, it makes things simple.

soulteary:
I agree with you, so do you consider replacing the original method with a simpler method?

Another PR takes less than 10 minutes to debug and develop based on this method.

In fact, there is a very troublesome issue that affects whether developers participate in contributing code, whether the project development environment is too complicated, or whether the installed global dependencies will affect the cleanliness of the system.

After all, many people need work in addition to participating in open source projects.

tommoor:
> I agree with you, so do you consider replacing the original method with a simpler method?

I don't want to run NodeJS within docker at this time, it's significantly slower – we moved away from that method. I don't think having Node.JS installed on the local machine is a huge ask.

tommoor:
We have begun a revamp of the docs using Outline to host them, it's getting much better now:
https://github.com/outline/outline#installation

PR #2559, opened 2021-09-13
fix: Cleanup forking model

tommoor:
Because `index.js` is called multiple times when `WEB_CONCURRENCY > 1` it results in multiple logs that are confusing.


server/index.js
```diff
@@ -37,7 +31,28 @@ const serviceNames = uniq(
...
 
-async function start(id, disconnect) {
+// The number of processes to run, defaults to the number of CPU's available
+// for the web service, and 1 for collaboration during the beta period.
+const processCount = serviceNames.includes("collaboration")
+  ? 1
+  : env.WEB_CONCURRENCY || undefined;
+
+// This function will only be called once in the original process
+function master() {
```
tommoor, line 41: Check env, migrations, and setup update timer only once

server/index.js
```diff
@@ -37,7 +31,28 @@ const serviceNames = uniq(
     .map((service) => service.trim())
 );
 
-async function start(id, disconnect) {
+// The number of processes to run, defaults to the number of CPU's available
+// for the web service, and 1 for collaboration during the beta period.
+const processCount = serviceNames.includes("collaboration")
```
tommoor, line 36: Respect `WEB_CONCURRENCY` for workers too

PR #2562, opened 2021-09-13
feat: Quick (cmd+k) menu

### LINKED ISSUE #2137: Quick (cmd+k) menu
The time has come, Outline must have an extensible `cmd+k` menu to make navigating and searching a breeze. We can then convert the existing search page into something more akin to an advanced-search with improved filtering functionality.

Items that should be included:

### Global 

- All sidebar menu items
- All collections
- All documents (titles are searchable from the cache in collections)

### Contextual

Could we allow an existing rendered context menu to specify that it's items should be available in cmd+k? This would be an easy and DRY way to implement contextual actions that already exist.

iamsaumya:
wip closes #2137


iamsaumya:
Hey @tommoor, PR is open for your feedback. I will love to know if the approach feels good to you and any accessibility measures you would want me to take.

PR #2569, opened 2021-09-14
feat: map preferred_username claim to user record

EternalDeiwos:
This is functionality described in the initial OIDC PR (#2369), but was removed as it was easier to make the PR smaller. I am unsure how much use the username property sees, however I do note that it exists and that most IdPs will provide some kind of `username` claim.

If this is useful then I will add polish as necessary.

#### Additions

Add an additional field to the authenticated user from the contents of `userinfo`. Specifically, the mapping is as follows:

- `preferred_username` (standard OIDC claim) &#8594; `user.username`

#### Configuration

- `OIDC_USERNAME_CLAIM` &mdash; Override which claim to check for the user's username. Defaults to `preferred_username` which is the standard OIDC claim name.

### DIFF
--- a/server/routes/auth/providers/oidc.js
+++ b/server/routes/auth/providers/oidc.js
@@ -23,7 +23,8 @@ const OIDC_AUTH_URI = process.env.OIDC_AUTH_URI;
 const OIDC_TOKEN_URI = process.env.OIDC_TOKEN_URI;
 const OIDC_USERINFO_URI = process.env.OIDC_USERINFO_URI;
 const OIDC_SCOPES = process.env.OIDC_SCOPES || "";
-const OIDC_USERNAME_CLAIM = process.env.OIDC_USERNAME_CLAIM;
+const OIDC_USERNAME_CLAIM =
+  process.env.OIDC_USERNAME_CLAIM || "preferred_username";
 const allowedDomains = getAllowedDomains();
 
 export const config = {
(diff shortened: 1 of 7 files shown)


tommoor (commented):
> I am unsure how much use the username property sees

Honestly, nothing right now – we might never use it, or it might become great for @mentions, In the coming months it's difficult to say. I'm happy to add support for bringing that data in considering the column is there.

> If this is useful then I will add polish as necessary.

Please re-request when complete

EternalDeiwos:
- Tests updated &mdash; I didn't think that adding any was necessary
- Instructions and pertinent information added inline and in `.env.sample`

Let me know if there's anything missing.

server/routes/auth/providers/oidc.js
```diff
@@ -105,7 +106,9 @@ if (OIDC_CLIENT_ID) {
               name: profile.name,
               email: profile.email,
               avatarUrl: profile.picture,
-              username: profile[OIDC_USERNAME_CLAIM],
+              // Claim name can be overriden using an env variable.
+              // Default is 'preferred_username' as per OIDC spec.
+              username: get(profile, OIDC_USERNAME_CLAIM),
```
tommoor, line 111: Should we include the default as third argument here?

EternalDeiwos: This makes sense but would mean that the env defaults will start to be scattered around the code file. Current placement was intended to keep them all together.

tommoor: Oh, I missed it at the top – ignore.

EternalDeiwos: Oh 🤣. Rolling back.

tommoor (approved):

Issue #2492, opened 2021-08-26
Context menus should not extend off-screen


tommoor:
![image](https://user-images.githubusercontent.com/380914/130856111-cd04806b-e64c-4b8d-82aa-7c4ec595c58f.png)

The menu should never extend off the screen edge, it should be sized so there is a small buffer of 8px and the contents should be scrollable if it won't fit.

PR #2571, opened 2021-09-14
fix: Implement custom Select Input

### LINKED ISSUE #2510: Implement custom Select Input
There have been a few instances which pointed towards the need of having a custom Select Input. 
For example, We have custom FilterOptions. 

Useful link: https://github.com/timelessco/renderlesskit-react

iamsaumya:
wip closes #2510

### DIFF
--- a/app/components/InputSelectPermission.js
+++ b/app/components/InputSelectPermission.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from "react";
 import { useTranslation } from "react-i18next";
-import InputSelect, { type Props, type Option } from "./InputSelect";
+import SelectInput, { type Props, type Option } from "./SelectInput";
 
 export default function InputSelectPermission(
   props: $Rest<Props, { options: Array<Option> }>
@@ -9,7 +9,7 @@ export default function InputSelectPermission(
   const { t } = useTranslation();
 
   return (
-    <InputSelect
+    <SelectInput
       label={t("Default access")}
       options={[
         { label: t("View and edit"), value: "read_write" },
--- a/app/scenes/CollectionPermissions/index.js
+++ b/app/scenes/CollectionPermissions/index.js
@@ -321,7 +321,6 @@ const Empty = styled(HelpText)`
 `;
 
 const PermissionExplainer = styled(HelpText)`
-  margin-top: -8px;
   margin-bottom: 24px;
 `;
 
--- a/app/components/SelectInput.js
+++ b/app/components/SelectInput.js
@@ -1,0 +1,191 @@
+// @flow
+import {
+  Select,
+  SelectOption,
+  SelectPopover,
+  useSelectState,
+} from "@renderlesskit/react";
+import * as React from "react";
+import { VisuallyHidden } from "reakit/VisuallyHidden";
+import styled from "styled-components";
+import breakpoint from "styled-components-breakpoint";
+import Button, { Inner } from "components/Button";
+import Flex from "components/Flex";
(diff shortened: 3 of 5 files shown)


iamsaumya:
I'm moving forward with this. I guess approx 0.54% is okay?

tommoor:
> I guess approx 0.54% is okay?

Yes, seems okay

tommoor:
Nice! A couple of things I noticed:

- When accessing the menu via keyboard the active state is not visible. eg `tab` and then `up` and `down` arrows
- The menu seems to be stuck to the bottom of the trigger, even when doing so pushes it off screen – best place to see this is on the language selection select. Not sure if anything in the lib helps here?
- The "checked" state is missing showing what the currently selected item is (see native menus)

iamsaumya:
> The menu seems to be stuck to the bottom of the trigger, even when doing so pushes it off screen – best place to see this is on the language selection select. Not sure if anything in the lib helps here?

In my testing, the menu does position itself at the top if there is not enough space at the bottom

https://user-images.githubusercontent.com/40148448/134821571-05ab2e1c-8bae-4c1a-904a-3ef44fe29ec6.mov

However, there is an edge case where the height of the menu just doesn't fit on either side. It feels that the menu is being pushed off the screen but with a click outside the menu solves the issue. 

https://user-images.githubusercontent.com/40148448/134821682-d348ad9a-ed79-4707-a7d6-0999f5e48cd2.mov

iamsaumya:
Does any thought cross your mind on why this must be happening?

tommoor:
I think it is the same issue as this then… https://github.com/outline/outline/issues/2492

app/components/InputSelectPermission.js
```diff
@@ -17,7 +17,7 @@ export default function InputSelectPermission(
       options={[
         { label: t("View and edit"), value: "read_write" },
         { label: t("View only"), value: "read" },
-        { label: t("No access"), value: "" },
+        { label: t("No access"), value: "no_access" },
```
tommoor, line 20: oh?

app/scenes/CollectionPermissions/index.js
```diff
@@ -178,6 +179,7 @@ function CollectionPermissions({ collection }: Props) {
   const sharing = collection.sharing;
   const teamSharingEnabled = !!auth.team && auth.team.sharing;
 
+  console.log("collection permission forpm index", collection.permission);
```
tommoor, line 182: debugging

app/components/InputSelectPermission.js
iamsaumya: The package wasn't letting us select the item

iamsaumya:
I'm extracting the maxHeight logic you wrote in a hook to use in InputSelect; I hope that's okay!

app/components/InputSelect.js
```diff
@@ -1,99 +1,179 @@
...
   label?: string,
+  ariaLabelPlural: string,
+  ariaLabel: string,
   short?: boolean,
   className?: string,
   labelHidden?: boolean,
   options: Option[],
-  onBlur?: () => void,
-  onFocus?: () => void,
+  onChange: (string) => Promise<void> | void,
```
iamsaumya, line 31: I'm curious. Is this _the_ way to handle cases where onChange could be `async` or `sync` function?

tommoor:
Made some updates, lmk if you spot any oddities!

iamsaumya:
Aye aye captain

iamsaumya:
Will do

PR #2594, opened 2021-09-19
fix: Move request helper function

### LINKED ISSUE #2558: DRY pattern for authentication providers : generalise and refactor request helper function
**To Reproduce**
1. Check the proposed `request` helper functions in
  - https://github.com/outline/outline/pull/2555/files#diff-f0e746a85f6e88d33814c261dd6c9600d11c869bca5c218965df9e57a547d584R22
  - https://github.com/outline/outline/pull/2470/files#diff-62df1778b1eb07f08a197bb3e13edb936d2feb585c6805ae635439c6d1447a5cR23
  - https://github.com/outline/outline/pull/2399/files#diff-f97adb7e5dc222e5b4d1d38efef3dce5b2e3a0edae071acc4028537932eb832fR19
2. Compare them with the Azure helper function in

https://github.com/outline/outline/blob/1e6769b0143aa32dbb1dc351537c22ce8aea4473/server/auth/providers/azure.js#L20-L29

3. See the duplication of code

**Expected behavior**

The "Don't Repeat Yourself" (DRY) pattern is respected where possible and similar code is abstracted away into a higher level, preferably `../../utils/passport`, which is imported in all of these modules. At least all of those who reuse the uniform example from Azure authentication with `Bearer` tokens could make use of this step.

iamsaumya:
closes #2558


server/routes/auth/providers/azure.js
```diff
@@ -18,17 +17,6 @@ const AZURE_RESOURCE_APP_ID = process.env.AZURE_RESOURCE_APP_ID;
 
 const scopes = [];
 
-export async function request(endpoint: string, accessToken: string) {
```
iamsaumya, line 21: I couldn't find an instance where we are using this export.

iamsaumya:
I couldn't think of any generalization apart from just returning `response` without calling `.json()`. Is there any generalization that you want?

tommoor:
I think the intention of original issue was to use the same function for all of the passport strategies as it was essentially copy and pasted in 3 places

tommoor:
The same function is in all of the providers, not just Azure 😉

iamsaumya:
I went through the files, and it's not :( 

He pointed out that contributors were copy-pasting the `request` method from azure in each of their PR. If we merge this, they will use this `request` method instead of duplicating. Apologies in advance if I have missed something.

PR #2622, opened 2021-10-02
fix: Per-group visibility settings

### LINKED ISSUE #2616: Per-group visibility settings
### Discussed in https://github.com/outline/outline/discussions/2614

<sup>Originally posted by **dmezh** September 27, 2021</sup>
Currently, Outline only lets non-admins see groups that they are themselves members of (commit 43493fcbbdf89d38b37cb0eb82b49ef1027c193b).

Per-group visibility settings would be great, because then users could use groups to more effectively share private collections. A toggle that makes the group 'private' or 'public' would accomplish this.</div>

### LINKED ISSUE #2621: Don't show create a group option to member
In addition, if there are no groups to show to the user - we should not show search input and show a "No groups found to add" message. 

**To Reproduce**

1. Be a member
1. Go to collection permission
2. Click on Add groups
3. See create a group option

Screenshot 2021-10-02 at 3 06 03 AM

iamsaumya:
closes #2616 
closes #2621


thenanyu:
@iamsaumya — Thanks for your work on this!

I'm putting in a patch to make groups visible to all users, instead of making it configurable on a group-by-group basis like you've done here. See https://github.com/outline/outline/pull/2729. We're going to start with this simpler approach and see if that solves the main issue for our users.

However, I noticed some refactoring work and the changes to address https://github.com/outline/outline/issues/2621 in this PR; if you want to reduce the scope of this PR to just those pieces, I'm happy to work with you to review and get those parts merged

iamsaumya:
That's nice. It solves the problem from the root. 

Let's close this for now. I think I will follow up with a new PR addressing #2621, if needed.

tommoor:
Thank you for the iteration 🙏

PR #2632, opened 2021-10-06
Add changing appearance for guests

### LINKED ISSUE #2579: A Guest User should be able to switch color modes on a shared page
### Discussed in https://github.com/outline/outline/discussions/2577

<sup>Originally posted by **PhilReinking** September 15, 2021</sup>
I suggest adding an option to the shared page view so that a user can toggle between light and dark mode. The current situation is that Outline uses the system default. But that might not be always the user's preferred setting for the application or the contents he is reading.

As an outline user, I already have a color mode option in the main dropdown.
I think that a simple icon/button (something subtle, like the "show contents" toggle) would be sufficient, to switch between modes on the shared page view.</div>

akrantz01:
resolves #2579 

Adds a button located on the right side of the screen for shared documents that allows changing the appearance for guest users.

### DIFF
--- a/app/scenes/Document/components/Header.js
+++ b/app/scenes/Document/components/Header.js
@@ -4,7 +4,9 @@ import {
   TableOfContentsIcon,
   EditIcon,
   PlusIcon,
+  MoonIcon,
   MoreIcon,
+  SunIcon,
 } from "outline-icons";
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks for the contribution 🥳

Issue #2542, opened 2021-09-09
Unable to login by using OIDC provider, KeyCloak and VMI


mzch:
Refer to #2535

I run OIDC provider, [KeyCloak](https://github.com/keycloak/keycloak) and [Verify My Identity](https://github.com/videntity/vmi) and try to login to Outline by using them, but fail.
When using KeyCloak, KeyCloak shows login form and I can login successfully but Outline shows login screen again, never show app home. KeyCloak and Outline has no trouble log.
When using VMI, it reported an error as below in getting OIDC token:
```
TypeError: Object of type 'bytes' is not JSON serializable
```
The VMI developer told me that Outline would probably have an encoding trouble. I googled it, and found this issue seemed to be around utf-8 encoding.

When using Google OAuth or Slack OAuth, I can login successfully and see app home.

**To Reproduce**
1. Set OIDC provider into `.env`
2. `npm start`
3. Go to login screen
4. Click `Contine with OpenID Connect` and  login

**Expected behavior**
Login successfully and show app home.

**Screenshots**
When using VMI, an error message shows as below:
<img width="1497" alt="ScreenShot" src="https://user-images.githubusercontent.com/621178/132623861-f1c6cc85-7dc7-4a19-99c2-beb6d7cc35cf.png">

**Outline (please complete the following information):**
- Install: self hosted
- Version: `0.59.0` (`commit 2f5cf90cb7b340fa1e8d7758fea1e3bf702b7fa5`)

tommoor:
There are no error messages in the server logs? You're the first one to try it with KeyCloak, so it's not overly surprising there might be teething issues. For Google OIDC I had to set the following:

```bash
OIDC_CLIENT_ID=REDACTED
OIDC_CLIENT_SECRET=REDACTED
OIDC_AUTH_URI=https://accounts.google.com/o/oauth2/v2/auth
OIDC_TOKEN_URI=https://oauth2.googleapis.com/token
OIDC_USERINFO_URI=https://openidconnect.googleapis.com/v1/userinfo
OIDC_DISPLAY_NAME=OpenID Connect
OIDC_SCOPES=openid email profile
```

mzch:
Yes. KeyCloak outputs no log. My config is as below:

```
OIDC_CLIENT_ID=[client_id]
OIDC_CLIENT_SECRET=[secret]
OIDC_AUTH_URI=https://oidcc.com/auth/realms/letlive/protocol/openid-connect/auth
[...]

PR #2659, opened 2021-10-13
feat: OIDC now conditionally overwrites the domain for the user

indietyp:
Fixes https://github.com/outline/outline/issues/2542#issuecomment-917324993 and #2678

The problem is that the subdomain of the email is used, which is not always desirable, especially in self-hosted circumstances where the same team might use different email domains (e.g., example.com, extern.example.com, intern.example.com). The current behavior puts everyone into different teams even tho they have the same top-level domain.

The changes here overwrite the domain used, **after** all checks are done on the email domain, ensuring that `allowedDomains` are still checked. This way, domains can still be restricted if needed, but everyone is implicitly put into the same team.


indietyp:
Hi @tommoor,

I would be willing to implement such behavior, but when setting `ALLOWED_DOMAINS`, which domain should be used as the team name? I used the new OICD to circumvent this problem.

One solution would be to take the first domain from `ALLOWED_DOMAINS` and use that one, but that is very implicit.

Should this functionality be to the `DEPLOYMENT` env so that it won't happen for the cloud installation?

tommoor:
> which domain should be used as the team name?

Which team name are you thinking of? It could probably use the first team that signs up, it's only a default and is all configurable in the in-app settings anyway.

> Should this functionality be to the DEPLOYMENT env so that it won't happen for the cloud installation?

Yes, should be flagged behind `DEPLOYMENT !== "hosted"`

indietyp:
I moved from the authentication to the `accountProvisioner`, is this better?

server/commands/accountProvisioner.js
```diff
@@ -53,13 +53,15 @@ export default async function accountProvisioner({
...
   try {
-    result = await teamCreator({
-      name: teamParams.name,
-      domain: teamParams.domain,
-      subdomain: teamParams.subdomain,
-      avatarUrl: teamParams.avatarUrl,
-      authenticationProvider: authenticationProviderParams,
-    });
+    result =
+      (await findExistingTeam(authenticationProviderParams)) ||
```
tommoor, line 57: Move this to the top of `teamCreator` rather than exporting and using here and I think we're good – are you able to add some basic tests around these expectations?

indietyp: sure can do!

kristianvld: @indietyp How is the progress on these changes?

indietyp: Sorry, I completely forgot to implement these changes. Our company moved on from evaluating Outline so this was subsequently shifted out of my focus and into my freetime. I am hoping to be able to finish the review next Tuesday.

almereyda:
Hi @indietyp !

We have been testing this branch with @tg-x, where we both have a different email domain set in our OIDC IDP, and it did not allow them to log in to an instance configured for "my" email domain, the first one used to initialise the Wiki team.

Is there anything else we can test, before this is refactored as requested above and subsequently merged?

How can we make sure that this feature branch solves #2678?

Thank you for your effort and contribution here.

indietyp:
Hi, I am terribly sorry for the delay. Thank you for looking into it. 

I will try to rewrite and test the pull request this weekend. With university and work, I have been a bit overloaded and didn't really have enough free time to focus on open-source much.

almereyda:
Bilal,

thank you very much for your effort to help us self-hosters with opening
Ouline to a wider user base.
And thanks to Tom and the rest of the team for releasing their core
business asset to the world as open source, which allows us to use and
improve it.

Please also take your time and don't feel pressed by us, especially in this
last, shorter month of the year.

Take it easy,

Jon

decentral1se:
Chiming in to say, very much need this to make outline work for my generic openid connect login approach.

Thanks for this work @indietyp :heart:

server/commands/accountProvisioner.test.ts
```diff
@@ -193,4 +194,69 @@ describe("accountProvisioner", () => {
     const collectionCount = await Collection.count();
     expect(collectionCount).toEqual(1);
   });
+
+  it("should skip findExistingTeam in hosted context", async () => {
+    expect(false).toBeTruthy(); // fail on purpose
```
tommoor, line 199: ?

server/commands/accountProvisioner.test.ts
```diff
@@ -193,4 +194,69 @@ describe("accountProvisioner", () => {
...
+    process.env = env; // reset
+  });
+
+  it("should return existing team, in self-hosted context", async () => {
+    const env = process.env;
+    process.env = { ...env, DEPLOYMENT: "xxx" }; // self to something that isn't hosted
+
+    const team = await buildTeam();
+    const authenticationProvider = (await team.getAuthenticationProviders())[0];
+    const existingTeam = await findExistingTeam({
```
tommoor, line 221: This file should be tests for accountProvisioner, we seem to be testing `findExistingTeam` here instead

server/commands/accountProvisioner.test.ts
```diff
@@ -193,4 +194,69 @@ describe("accountProvisioner", () => {
...
+    process.env = env; // reset
+  });
+
+  it("should create new authentication provider on existing team if in self-hosted context", async () => {
+    const env = process.env;
+    process.env = { ...env, DEPLOYMENT: "xxx" }; // self to something that isn't hosted
+
+    const team = await buildTeam(); // initial team accountProvider is slack
+    const providerId = uuidv4();
+    const existingTeam = await findExistingTeam({ name: "OICD", providerId });
```
tommoor, line 241: Similarly to above comment

server/commands/accountProvisioner.ts
```diff
@@ -45,6 +45,44 @@ export type AccountProvisionerResult = {
   isNewUser: boolean;
 };
 
+export async function findExistingTeam(
+  authenticationProvider: Props["authenticationProvider"]
+): Promise<TeamCreatorResult | null> {
+  // Should outline deployed in a multi-tenant environment, skip searching
+  // for an existing team.
+  console.log(process.env.DEVELOPMENT);
```
tommoor, line 53: Logging

server/commands/accountProvisioner.ts
```diff
@@ -45,6 +45,44 @@ export type AccountProvisionerResult = {
...
+  if (process.env.DEPLOYMENT === "hosted") return null;
+
+  // get the first team that exists, ordered by createdAt
+  const team = await Team.findOne({ limit: 1, order: ["createdAt"] });
+  if (team === null) {
+    return null;
+  }
+
+  // query if a corresponding authenticationProvider already exists
+  const authenticationProviders = await team.getAuthenticationProviders({
```
tommoor, line 63: This logic doesn't seem to be restricted to OIDC in any way, which means if someone tries to signin with another Google account then it would just remap them to an existing one and allow signin?

server/commands/accountProvisioner.ts
```diff
@@ -55,13 +93,15 @@ export default async function accountProvisioner({
...
   try {
-    result = await teamCreator({
-      name: teamParams.name,
-      domain: teamParams.domain,
-      subdomain: teamParams.subdomain,
-      avatarUrl: teamParams.avatarUrl,
-      authenticationProvider: authenticationProviderParams,
-    });
+    result =
+      (await findExistingTeam(authenticationProviderParams)) ||
```
tommoor, line 97: This should be part of teamCreator instead of it's own thing

package.json
```diff
@@ -238,7 +238,7 @@
     "@types/tmp": "^0.2.2",
     "@types/turndown": "^5.0.1",
     "@types/utf8": "^3.0.0",
-    "@typescript-eslint/eslint-plugin": "^5.3.1",
+    "@typescript-eslint/eslint-plugin": "^5.6.0",
```
tommoor, line 241: Unrelated package changes seem to have introduced a lot of yarn.lock changes.

tommoor (changes requested):
I appreciate the work here however as it stands the PR is still far off from what would be landable, it also seems to introduce a large security hole in the form of re-mapping auth providers if an existing auth provider of the same name exists.

It's a shame that the domain was mapped from the email in the original OIDC implementation – we should probably have just had it as a static string, but changing that now would break existing installations.

I'm going to try and tackle this as part of  https://github.com/outline/outline/pull/2986

PR #2665, opened 2021-10-15
fix: Add default role option for new users

### LINKED ISSUE #2547: Add default role option for new users
The ability to set a "default role" in the team security settings that would be applied to newly created users is something that we'd accept a PR for. We have read-only users currently, there's just no way to say that all new users should get this role by default.

ref: https://github.com/outline/outline/discussions/2130#discussioncomment-721451

iamsaumya:
closes #2547

### DIFF
--- a/app/scenes/Settings/Security.js
+++ b/app/scenes/Settings/Security.js
@@ -28,7 +28,6 @@ function Security() {
   });
 
   const notes = {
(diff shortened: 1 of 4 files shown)


iamsaumya:
Agree. Admin is too much power given by default for every user.

PR #2701, opened 2021-10-29
feat: Seamless Edit

### LINKED ISSUE #1526: Create "Always Edit" Mode
Document editors like Google Docs and Dropbox Paper are effectively always in edit mode. Many users are accustomed to this convention, so we might want to explore the approach as well.

Practically, there are numerous complications that need to be dealt with. This might be best hidden behind a "labs" type of feature flag so that early versions can be dog-fooded.

tommoor:
Removes the "Edit" button and auto-collapsing of the sidebar when collaborative editing is enabled for a team. This might prove to be controversial with some existing users? But It's definitely the best for the long term health and adoption of this project.

Sidebar can still be optionally collapsed with button or `cmd+.`, however it is disconnected from editing.

cc @thenanyu – is this... the dream?

closes #1526
closes https://github.com/outline/outline/discussions/2670
related https://github.com/outline/outline/discussions/2694


tommoor:
Noticed that we're missing a button to copy code blocks as this was previously read-only

tommoor:
It's also not easy to visit links, this is a bigger deal.

PR #2713, opened 2021-10-30
chore: Remove `react-keydown`

### LINKED ISSUE #1856: Update hotkeys to use a hooks-based solution

tommoor:
Avoided converting the components to functional all in one go as part of this, instead opting for a wrapper to enable hook usage in class components temporarily.

In the medium-term almost all of these key registrations will move to [action definitions](https://github.com/outline/outline/tree/main/app/actions/definitions) but at least we can now convert these components to functional without this being a blocker + one less dependency.

@iamsaumya pinging you as you previously [took a pass](https://github.com/outline/outline/issues/1856) at this :) 

related https://github.com/timc1/kbar/issues/107
closes #1856


app/hooks/useKeyDown.js
```diff
@@ -0,0 +1,69 @@
...
+  }, deps);
+
+  React.useEffect(() => {
+    callbacks.push(handler);
+
+    return () => {
+      callbacks = callbacks.filter((cb) => cb !== handler);
+    };
+  });
+}
```
iamsaumya, line 46: I have a doubt here. Why do we have deps props? My best guess is it's because we want to memorize the handler but it doesn't seem to work because we are anyway pushing `handler` to `callbacks` every render.  

It feels kind of hard to read. If `fn` is memorized then why do we need `deps`? if not, then fn should always be one of the deps. 

We can move all the handler logic inside `useEffect`. What do you think of this?

```js
  React.useEffect(() => {
    const predicate = createKeyPredicate(key);
    const handler = (event: KeyboardEvent) => {
      if (predicate(event)) fn(event);
    };
    callbacks.push(handler);

    return () => {
      callbacks = callbacks.filter((cb) => cb !== handler);
    };
  });
```

tommoor: I haven't actually needed `deps` yet, so could potentially just remove that complexity until it's needed for something…

PR #2729, opened 2021-11-04
fix: visible groups

thenanyu:
alternative solution to https://github.com/outline/outline/issues/2616

Groups and their members are not secret information generally, so we should try to take a more permissive and less complex approach to begin with.

This PR simply allows the entire team to see every group.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "build:webpack": "webpack --config webpack.config.prod.js",
     "build": "yarn clean && yarn build:webpack && yarn build:i18n && yarn build:server",
     "start": "node ./build/server/index.js",
-    "dev": "yarn concurrently -n api,collaboration  -c \"blue,magenta\" \"node --inspect=0.0.0.0 build/server/index.js --services=collaboration,websockets,admin,web,worker\"",
+    "dev": "NODE_ENV=development yarn concurrently -n api,collaboration  -c \"blue,magenta\" \"node --inspect=0.0.0.0 build/server/index.js --services=collaboration,websockets,admin,web,worker\"",
     "dev:watch": "nodemon --exec \"yarn build:server && yarn build:i18n && yarn dev\" -e js --ignore build/ --ignore app/ --ignore flow-typed/",
     "lint": "eslint app server shared",
     "deploy": "git push heroku master",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "build:webpack": "webpack --config webpack.config.prod.js",
     "build": "yarn clean && yarn build:webpack && yarn build:i18n && yarn build:server",
     "start": "node ./build/server/index.js",
-    "dev": "yarn concurrently -n api,collaboration  -c \"blue,magenta\" \"node --inspect=0.0.0.0 build/server/index.js --services=collaboration,websockets,admin,web,worker\"",
+    "dev": "NODE_ENV=development yarn concurrently -n api,collaboration  -c \"blue,magenta\" \"node --inspect=0.0.0.0 build/server/index.js --services=collaboration,websockets,admin,web,worker\"",
```
thenanyu, line 13: This was necessary to get email login working locally to test group behavior; I think it's worth leaving it in this part of the script, as opposed to expecting developers to add it to their `.env` files

server/mailer.js
```diff
@@ -204,12 +204,18 @@ export class Mailer {
   };
 
   signin = async (opts: { to: string, token: string, teamUrl: string }) => {
+    const signInLink = signinEmailText(opts);
+
+    if (process.env.NODE_ENV === "development") {
+      Logger.info(`Sign-In link: ${signInLink}`);
+    }
```
thenanyu, line 211: I can go either way with leaving this in, but it's useful when the dev mail service is unavailable or if the developer doesn't have a network connection

server/routes/api/documents.test.js
```diff
@@ -1230,7 +1230,11 @@ describe("#documents.search", () => {
   });
 
   it("should return documents for a specific private collection", async () => {
-    const { user, collection } = await seed();
+    const user = await buildUser();
+    const collection = await buildCollection({
+      teamId: user.teamId,
+    });
+
```
thenanyu, line 1237: tried to move towards model-specific builders for tests to possibly solve deadlock problems; leaving them in as it's a general improvement

tommoor: Yea, I've been trying to remove `seed` gradually to make things faster too

tommoor (approved):

tommoor:
Are you wanting to change anything else here or go to merge?

thenanyu:
I'll merge it. Did it on the plane but internet failed

Issue #2486, opened 2021-08-25
Sidebar links should not reserve space for "..." menu


tommoor:
Currently a large space is reserved to the right of every link in the sidebar for the "..." menu to appear when you hover. Instead it should overlay ontop of the content to allow for another ~24px of horizontal space before the text wraps.

![image](https://user-images.githubusercontent.com/380914/130713378-a5ce4f28-98e8-465a-a3ae-c4f5b99c21b6.png)

charlesharries:
I'd like to take a swing at this one, if that's alright? I've spotted #2695 as well so I can take that into account while I build this one out.

It looks like Notion appends the `[...] [+]` buttons to the document title on `mouseover`:

![ellipsis-over](https://user-images.githubusercontent.com/20623195/139528237-63b3c75a-b72b-479b-99ac-dd757e76b0a2.gif)

But I don't see why toggling a `display: none` on CSS `:hover` wouldn't work just as well. I'll have a test across browsers and see what I can come up with.

tommoor:
By all means, give it a go 👍

PR #2741, opened 2021-11-06
feat: DocumentMenu button overlay style

charlesharries:
Following on from #2486, I've added a style to the action buttons that display alongside the sidebar links to allow them to 'overlay' the text of the sidebar link—rather than reserving 24px of space to the right of the link for the button (see screenshots below).

These styles should only apply to the sidebar links—not to any other document link actions across the application.

I've also added some type annotations for the `OverflowMenuButton`'s props. They're not complete but they cover all of the props used by `OverflowMenuButton` directly.

## Testing
1. Spin up the test site and create a child document under **Welcome**. Give it a long title, preferably one that wraps onto multiple lines.
2. Hover over that document's name in the sidebar; the 'actions' button should overlay the text on hover.
3. Click into the **Welcome** document; your new document should appear in the list of child documents.
4. Hovering over the title of your document here should show the original style—the three `MoreInfo` dots with no background.

#### Before:
The sidebar actions button pushes the text over, even when not displayed

#### After:
The sidebar actions button overlays the text when the text is hovered

The sidebar actions button overlays the text when the text is hovered in dark mode, as well

### DIFF
--- a/app/components/ContextMenu/OverflowMenuButton.js
+++ b/app/components/ContextMenu/OverflowMenuButton.js
@@ -4,11 +4,17 @@ import * as React from "react";
 import { MenuButton } from "reakit/Menu";
 import NudeButton from "components/NudeButton";
 
+type Props = {
+  iconColor?: string,
+  className?: string,
+  isOverlay?: boolean,
+};
+
 export default function OverflowMenuButton({
   iconColor,
   className,
   ...rest
-}: any) {
+}: Props) {
   return (
     <MenuButton {...rest}>
       {(props) => (
--- a/app/components/NudeButton.js
+++ b/app/components/NudeButton.js
@@ -5,7 +5,12 @@ import styled from "styled-components";
 const Button = styled.button`
   width: ${(props) => props.width || props.size}px;
   height: ${(props) => props.height || props.size}px;
-  background: none;
+  background: ${(props) =>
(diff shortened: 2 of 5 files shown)


tommoor:
Hey @charlesharries thanks for taking this on. The issue was mainly about positioning rather than styling, I think ideally the button would look the same as today so that it's not distracting on hover, i.e. the same color background as the sidebar.

It seems that in reality nothing for positioning is needed? I'll also note that the "same" button exists in `StarredLink`

charlesharries:
Hi @tommoor thanks for the response, and for the heads-up on the `StarredLink` component! That all makes sense—I've removed the `isOverlay` logic so that the button overlays the text without any new styling:

The DocumentLink action button overlays the DocumentLink text

tommoor:
The gif demo is useful, we just need to get the background to match what's behind on the sidebar link so that the "..." isn't clashing with the text.

charlesharries:
Riteo, I've re-added that `isOverlay` logic with the more pared-down styles for both the `DocumentLink` and `StarredLink` components in the sidebar:

overlay-redux

Hopefully the third time's the charm haha!

tommoor:
I had to make a few more fixes to get this to work perfectly but the fix is here and I added you as a co-author credit

https://github.com/outline/outline/commit/d13254b5436f8ad6286f4c04823fdb6183e179fe

PR #2760, opened 2021-11-12
feat: Highlight active ListItem in outline.

dkkb:
Hi, I think it would be fantastic if we could highlight the active item in the outline.

Dark theme:

Light theme:


tommoor:
Yes, I think this is clearer

tommoor (approved):

dkkb:
> Yes, I think this is clearer in this case! I'd encourage you in the future to file issues before creating PR's so that we can discuss to avoid doing work that isn't mergable.

Sure.

PR #2761, opened 2021-11-12
fix: fix issue where the title can be modified in read-only mode

dkkb:
The title can be changed (but not saved) when the document is in read only mode.

Before:

After:

----

By the way, I don't see any use of `disabled` props, perhaps `disabled` might be renamed to `readOnly`?


tommoor:
`disabled` and `readOnly` are there to try and mirror the interface of `<input />`. Thanks for the fix!

PR #2776, opened 2021-11-16
Let redis, pgsql and s3 listen on localhost only. (docker-compose)

splitline:
If you open such services publicly, that might cause some security issues (i.e. every one can access those services). No matter it's on production or development.


tommoor:
Hey @splitline – the docker compose file is for development, it really doesn't matter – I think being more flexible here is actually a positive.

Issue #2302, opened 2021-07-09
Move editor into codebase


tommoor:
See: https://github.com/outline/rich-markdown-editor/discussions/481

I plan to archive the editor project as unmaintained and move the code into the Outline codebase. This will allow much tighter integration, sharing of UI components, and speed up development. The editor is core to the Outline experience, it's important that it's recognizable and excellent.

Blocked by #1282

soulteary:
Perhaps an protocol should be added to allow users to freely choose **editor** components, such as switching to tinymce, etc.

This can help outline team save a lot of development time, focus on legacy features and experience issues of current products.

tommoor:
A highly integrated editor experience is key to Outline, we won't be making it swappable by the user 😄 . As a general rule this project is highly opinionated and doesn't try to please everyone like many OSS projects.

soulteary:
> A highly integrated editor experience is key to Outline, we won't be making it swappable by the user 😄 . As a general rule this project is highly opinionated and doesn't try to please everyone like many OSS projects.

Maybe you are right.

Perhaps this is also one of the competitiveness of this project. 😄

svallory:
Would it be acceptable to isolate the editor behind an interface so forks could be created replacing it? The reason I'm asking is that right now editing is kinda hard. I've encountered many small annoyances while editing :/

Also, MHO is that Outline offers a lot more than the editor and while the editor is in fact at the center of the outline experience, there are way more advanced opensource editors that could be used to save up time for other feature developments like smart tables, sharing of collections.

tommoor:
Not really interested in this, we're proud of the editor and would rather use that time and effort to fix the small annoyances - please file issues for those.

The upcoming collab editing also makes switching editors much more difficult

PR #2783, opened 2021-11-17
chore: Move to Typescript

### LINKED ISSUE #1282: Move to Typescript
We already moved the editor repo over: https://github.com/outline/rich-markdown-editor

Some further discussion in this thread:
https://github.com/outline/outline/discussions/1269

tommoor:
This PR moves the entire project to Typescript. Due to the ~1000 ignores this will lead to a messy codebase for a while, but the churn is worth it – all of those ignore comments are places that were never type-safe previously.

- Retains compilation via webpack/babel for now
- Type checking via `tsc`
- Used `@khanacademy/flow-to-ts` to convert as many types as possible
- Removed usage of mobx `inject`
- `npx ts-migrate reignore` to ignore places that were not typed and now caught by TS
- Does not yet type the backend models, we can do that separately as it was never type-safe before
- Lots of `ts-expect-error` comments where conversion could not be performed, these will be removed over time

closes #1282 
unblocks #2302


tommoor:
@thenanyu feeling close to deployable at this point, just got to merge `main` back in and do another pass over as much functionality as possible

tommoor:
@iamsaumya if you do get a little time would love for you to test out this branch too, few know the app as well as you 😄

tommoor:
This will merge tomorrow a.m.

PR #2804, opened 2021-11-30
fix: Remove star option on publicly shared docs

### LINKED ISSUE #2803: Remove star from document headings on shared documents
Currently the option to star appears in headings on shared documents when signed in with permission. This is a bit strange, we should keep the separation between public/private here and not show the star.

tommoor:
closes #2803

### DIFF
--- a/app/scenes/Document/components/EditableTitle.tsx
+++ b/app/scenes/Document/components/EditableTitle.tsx
@@ -15,9 +15,15 @@ import { isModKey } from "~/utils/keyboard";
 type Props = {
   value: string;
   document: Document;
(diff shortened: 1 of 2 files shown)


app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -15,9 +15,15 @@ import { isModKey } from "~/utils/keyboard";
 type Props = {
   value: string;
   document: Document;
+  /** Should the title be editable, policies will also be considered separately */
```
tommoor, line 18: These comments show up when you use the component, so that's pretty nice. We should make more use of this going forward!

PR #2807, opened 2021-11-30
fix: No reserving space in sidebar for overflow

### LINKED ISSUE #2486: Sidebar links should not reserve space for "..." menu

tommoor:
closes #2486

based on #2741 – this PR was half way there but it drilled an `isOverlay` prop down through several components that don't need it and also didn't handle the background color correctly when the sidebar item is highlighted.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -25,5 +25,17 @@
     "@babel/plugin-transform-destructuring",
     "@babel/plugin-transform-regenerator",
     "transform-class-properties"
-  ]
+  ],
+  "env": {
+    "production": {
+      "plugins": [
+        [
+          "styled-components",
+          {
+            "displayName": false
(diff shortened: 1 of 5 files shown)


.babelrc
```diff
@@ -25,5 +25,17 @@
...
     "transform-class-properties"
-  ]
+  ],
+  "env": {
+    "production": {
+      "plugins": [
+        [
+          "styled-components",
+          {
+            "displayName": false
```
tommoor, line 35: Unrelated, but nice file size improvement for css in production

PR #2809, opened 2021-11-30
fix: Clickable area for docs in sidebar is obstructed

### LINKED ISSUE #2808: Clickable area for docs in sidebar is obstructed
Looks like the `DropCursor` component is in the way

tommoor:
Also simplifies the logic for collapsing collections when dragging by avoiding sending state out to the parent component and relying on the `useDrop` `monitor` instead.

This is so much nicer now, no "dead" areas in the sidebar when hovering over docs in manually sorted collections.

closes #2808

### DIFF
--- a/app/components/Sidebar/components/DocumentLink.tsx
+++ b/app/components/Sidebar/components/DocumentLink.tsx
@@ -197,7 +197,7 @@ function DocumentLink(
   });
 
   // Drop to reorder
-  const [{ isOverReorder }, dropToReorder] = useDrop({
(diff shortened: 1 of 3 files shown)


app/components/Sidebar/components/DocumentLink.tsx
```diff
@@ -275,7 +276,7 @@ function DocumentLink(
             </DropToImport>
           </div>
         </Draggable>
-        {manualSort && (
+        {manualSort && isDraggingAnyDocument && (
```
tommoor, line 279: Don't render the cursor unless we're drag-n-dropping'

Note: We were already doing this for collections!

Issue #2810, opened 2021-11-30
ReferenceListItem document icon overflow.


dkkb:
The document icon will overflow if the title of the document in ReferenceListItem is too long, and the 'text-overflow' property will not work.

**Screenshots**

Now:
<img src="https://user-images.githubusercontent.com/82504881/143984896-080bad62-f623-49b7-9b3b-b8854c6c4dde.png" width=400>

Fixed:

<img src="https://user-images.githubusercontent.com/82504881/143988386-1e4cfb6e-a53b-4e95-9c7b-9287aec76a25.png" width=400>

**Outline (please complete the following information):**
- Install: [getoutline.com or self hosted]
- Version: [commit sha if self hosted]

PR #2811, opened 2021-11-30
fix: Document icon overflow in ReferenceListItem

dkkb:
Related to https://github.com/outline/outline/issues/2810.

Now:

Fixed:

-------

We can deduce the cause of the overflow problem from this stackoverflow answer. We can do some tricks to achieve `text-overflow: ellipsis`. The solution is `width: calc(100%)`. We also need to change the display to inline block and add the line`vertical-align: middle` to align the svg and h3 vertically.

Reference:
https://stackoverflow.com/questions/17779293/css-text-overflow-ellipsis-not-working
https://stackoverflow.com/questions/9670469/css-vertical-alignment-of-inline-inline-block-elements

### DIFF
--- a/app/scenes/Document/components/ReferenceListItem.tsx
+++ b/app/scenes/Document/components/ReferenceListItem.tsx
@@ -31,14 +31,13 @@ const DocumentLink = styled(Link)`
   }
 `;
 
-const Title = styled.h3`
-  display: flex;
-  align-items: center;
+const Title = styled.div`
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


tommoor:
Hi @dkkb – note that the vertical spacing has increased on your branch but should be retained as-is.

dkkb:
> Hi @dkkb – note that the vertical spacing has increased on your branch but should be retained as-is.

What about now?

image

PR #2812, opened 2021-11-30
fix: document publishing

thenanyu:
adds `done` and `publish` options to document.update - these got removed during the switchover to ts

### DIFF
--- a/app/models/Document.ts
+++ b/app/models/Document.ts
@@ -305,6 +305,8 @@ export default class Document extends BaseModel {
           },
           {
             lastRevision: options.lastRevision,
+            publish: options?.publish,
+            done: options?.done,
           }
         );
       }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "build": "yarn clean && yarn build:webpack && yarn build:i18n && yarn build:server",
     "start": "node ./build/server/index.js",
     "dev": "NODE_ENV=development yarn concurrently -n api,collaboration  -c \"blue,magenta\" \"node --inspect=0.0.0.0 build/server/index.js --services=collaboration,websockets,admin,web,worker\"",
-    "dev:watch": "nodemon --exec \"yarn build:server && yarn build:i18n && yarn dev\" -e js --ignore build/ --ignore app/",
+    "dev:watch": "nodemon --exec \"yarn build:server && yarn build:i18n && yarn dev\" -e js,ts --ignore build/ --ignore app/",
```
thenanyu, line 14: needed to add ts to filetypes under watch now that we've switched over

app/models/Document.ts
```diff
@@ -305,6 +305,8 @@ export default class Document extends BaseModel {
           },
           {
             lastRevision: options.lastRevision,
+            publish: options?.publish,
+            done: options?.done,
```
thenanyu, line 309: admittedly, I'm not 100% sure what "done" does system wide, but this brings it back in line with the previous implementation

tommoor: `done` is almost a relic now, it was originally a signal to the backend that editing session is complete. It makes sense to include it here and remove everywhere later.

tommoor (approved):

PR #2814, opened 2021-12-02
fix: focus input when page refresh

iamsaumya:


tommoor:
Could you describe what this is fixing? I forget

iamsaumya:
The input wasn't focused on the search page when navigating through kbar or refreshing the search page.

PR #2817, opened 2021-12-04
fix: Catch error in promise and fix typo

l2dy:
In try/catch blocks, `return await` is necessary to reach the `catch` clause.


tommoor (approved):
Good spot!

PR #2820, opened 2021-12-04
feat: Show drafts in sidebar when viewing

### LINKED ISSUE #2800: Show draft documents in sidebar navigation when viewing
It is currently unclear where draft documents live, we can improve this experience somewhat by showing drafts in the sidebar in a slightly "ghosted" state when viewing, something like:

image

- The breadcrumb can also be changed to reflect the location where the draft is and will be published, rather than showing as being a child of "Drafts". 
- Drafts will always be the top-most child in parent
- Ensure that drag and drop is disabled for these drafts for now

Related #2512

tommoor:
closes #2800


tommoor:
@thenanyu probably wouldn't try to review the code too much, it's kind of a big ball of changes – but would appreciate if you can checkout branch and give it a go.

server/routes/api/documents.ts
```diff
@@ -91,7 +91,7 @@ router.post("documents.list", auth(), pagination(), async (ctx) => {
     // index sort is special because it uses the order of the documents in the
     // collection.documentStructure rather than a database column
     if (sort === "index") {
-      documentIds = collection.documentStructure
+      documentIds = (collection.documentStructure || [])
```
thenanyu, line 94: did something change that allows document structure to come back as null or undefined?

tommoor: An existing bug I could pull into it's own PR, if you try and do `documents.list` for a new and empty collection you'll get an internal server error today

app/models/BaseModel.ts
```diff
@@ -49,7 +63,43 @@ export default class BaseModel {
...
+   * Returns a boolean indicating whether the model has been persisted to db
+   *
+   * @returns boolean true if the model has never been persisted
+   */
+  @computed
+  get isNew(): boolean {
+    return !this.id;
+  }
+
+  protected persistedAttributes: Partial<BaseModel> = {};
```
thenanyu, line 104: this represents a dirty state of updates right?

app/models/BaseModel.ts
```diff
@@ -49,7 +63,43 @@ export default class BaseModel {
...
 
+  /**
+   * Returns a plain object representation of the model
+   *
+   * @returns {Record<string, any>}
+   */
   toJS = (): Record<string, any> => {
-    return { ...this };
+    const fields = getFieldsForModel(this);
+    return pick(this, fields) || [];
```
thenanyu, line 73: What kind of issues were you seeing with the previous serialization style such that you went with explicit field defs? was a bunch of cruft making it into the json?

tommoor: The main thing is that I wanted a centralized store of which fields on a model _should_ be tracked which this did not provide, decorators felt like a nice DX to provide for this

app/models/BaseModel.ts
tommoor: Represents the tracked `@Field` attributes as they were when last successfully persisted to the API.

PR #2824, opened 2021-12-06
chore: Remove outline service in docker-compose.yml

l2dy:
Commit 18cf148bd176d06d760555542bd8784eab5a36d9 made Yarn run locally outside of Docker, so there is no need to mount files into the container. This commit allows hosts without Node.js environment to run Outline with docker-compose.

Depends on #2823.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,19 +18,3 @@ services:
       - "127.0.0.1:4569:4569"
     volumes:
       - ./fakes3:/fakes3_root
-  outline:
-    image: outline:v001
-    command: yarn dev
-    build:
-      context: .
-      dockerfile: Dockerfile
-      args:
-        pull: 1
-    ports:
-      - "3000:3000"
-    volumes:
-      - .:/opt/outline
-    depends_on:
-      - postgres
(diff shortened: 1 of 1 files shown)


tommoor:
Honestly the entire `outline:` section of this yaml should probably be removed. It's not even used any more.

l2dy:
It's not used for development, but I still find it convenient to bootstrap a server with just `docker-compose up`.

Removing the `outline:` section would make it clear that Yarn should run locally, which is acceptable to me too.

routerino:
A core feature of docker is that the OS is supposed to be abstracted. Having dependencies inside docker containers that rely on assumptions about the hosts is a bad idea (if I'm reading that right).

For smaller clients I don't use kubernetes at all, just docker hosts running 20-30 containers. If I had to manage outside dependencies for 20-30 containers, I may as well just not use containers at all.

tommoor:
Yea, I think it's worth keeping it as clear as possible that this file is not expected to be for production use, that hasn't really been the case up until now

l2dy:
OK, PR updated.

PR #2830, opened 2021-12-07
feat: Support IAM role authentication for S3

### LINKED ISSUE #2829: AWS IAM role support for S3
### Discussed in https://github.com/outline/outline/discussions/2386

<sup>Originally posted by **scott-dunt** July 28, 2021</sup>
When running applications on 'EC2' hosts in Amazon Web Services, AWS, we prefer to use 'IAM Roles' to enable access to AWS resources like S3 buckets. This way we don't have security ID's and password - keys stored in files on the hosts, the access is allowed and assigned to the HOST system internally by AWs: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html

My testing seems to indicate that the Outline app does not work with IAM roles.  I have tried using no and blank credentials in the .env file and neither works:

AWS_ACCESS_KEY_ID=
AWS_SECRET_ACCESS_KEY=

Any chance to add this functionality?
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials</div>

l2dy:
With an asynchronous credential provider like EC2 IAM roles, `createPresignedPost()` should always be called with an asynchronous callback.

closes #2829

Note: This implementation has not been tested against S3-compatible storage.

### DIFF
--- a/server/routes/api/attachments.ts
+++ b/server/routes/api/attachments.ts
@@ -1,4 +1,3 @@
-import { format } from "date-fns";
 import Router from "koa-router";
 import { v4 as uuidv4 } from "uuid";
 import { NotFoundError } from "@server/errors";
@@ -6,10 +5,8 @@ import auth from "@server/middlewares/authentication";
 import { Attachment, Document, Event } from "@server/models";
 import policy from "@server/policies";
 import {
-  makePolicy,
-  getSignature,
+  getPresignedPost,
   publicS3Endpoint,
-  makeCredential,
   getSignedUrl,
 } from "@server/utils/s3";
 import { assertPresent } from "@server/validation";
@@ -34,9 +31,7 @@ router.post("attachments.create", auth(), async (ctx) => {
       : "private";
   const bucket = acl === "public-read" ? "public" : "uploads";
   const key = `${bucket}/${user.id}/${s3Key}/${name}`;
-  const credential = makeCredential();
-  const longDate = format(new Date(), "yyyyMMdd'T'HHmmss'Z'");
-  const policy = makePolicy(credential, longDate, acl, contentType);
+  const presignedPost = await getPresignedPost(key, acl, contentType);
   const endpoint = publicS3Endpoint();
   const url = `${endpoint}/${key}`;
 
@@ -73,13 +68,7 @@ router.post("attachments.create", auth(), async (ctx) => {
       form: {
         "Cache-Control": "max-age=31557600",
         "Content-Type": contentType,
-        acl,
-        key,
-        policy,
-        "x-amz-algorithm": "AWS4-HMAC-SHA256",
-        "x-amz-credential": credential,
-        "x-amz-date": longDate,
-        "x-amz-signature": getSignature(policy),
+        ...presignedPost.fields,
       },
       attachment: {
         documentId,
--- a/server/utils/s3.ts
+++ b/server/utils/s3.ts
@@ -1,4 +1,5 @@
 import crypto from "crypto";
+import util from "util";
 import AWS from "aws-sdk";
 import { addHours, format } from "date-fns";
 import fetch from "fetch-with-proxy";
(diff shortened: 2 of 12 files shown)


server/utils/s3.ts
```diff
@@ -93,6 +93,37 @@ export const getSignature = (policy: string) => {
...
+
+  return new Promise<AWS.S3.PresignedPost>((resolve, reject) => {
+    s3.createPresignedPost(params, (err, data) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(data);
+      }
+    });
+  });
```
tommoor, line 124: We can use [promisify](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original) here to tidy this up a bit

server/routes/api/attachments.ts
```diff
@@ -71,15 +67,11 @@ router.post("attachments.create", auth(), async (ctx) => {
...
-        key,
-        policy,
-        "x-amz-algorithm": "AWS4-HMAC-SHA256",
-        "x-amz-credential": credential,
-        "x-amz-date": longDate,
-        "x-amz-signature": getSignature(policy),
+        ...{
+          "Cache-Control": "max-age=31557600",
+          "Content-Type": contentType,
+        },
```
tommoor, line 73:
```suggestion
        "Cache-Control": "max-age=31557600",
        "Content-Type": contentType,
```

No need for the object spread here, right?

server/routes/api/attachments.ts
```diff
@@ -34,9 +32,7 @@ router.post("attachments.create", auth(), async (ctx) => {
       : "private";
   const bucket = acl === "public-read" ? "public" : "uploads";
   const key = `${bucket}/${user.id}/${s3Key}/${name}`;
-  const credential = makeCredential();
-  const longDate = format(new Date(), "yyyyMMdd'T'HHmmss'Z'");
-  const policy = makePolicy(credential, longDate, acl, contentType);
+  const presignedPost = await getPresignedPost(key, acl, contentType);
   const endpoint = publicS3Endpoint();
   const url = `${endpoint}/${key}`;
```
tommoor, line 37: I wonder if `presignedPost.url` could be used now? 🤔

server/utils/s3.ts
l2dy: It looks great, but I don't know how to fix this error.

```
 FAIL  server/routes/api/fileOperations.****.ts
  ● Test suite failed to run

    TypeError [ERR_INVALID_ARG_TYPE]: The "original" argument must be of type function. Received undefined

      26 |   signatureVersion: "v4",
      27 | });
    > 28 | const createPresignedPost = util.promisify<
         |                                  ^
      29 |   AWS.S3.PresignedPost.Params,
      30 |   AWS.S3.PresignedPost
      31 | >(s3.createPresignedPost);

      at Object.<anonymous> (server/utils/s3.ts:28:34)
      at Object.<anonymous> (server/models/Attachment.ts:2:1)
```

l2dy: @tommoor Any suggestions? I'd rather not spend too much time on fighting TypeScript's type inference limitations.

tommoor: Sorry, wanted to check it out and take a look – it's not actually so bad, it's just that in a couple of the tests `aws-s3` lib is mocked, so the `createPresignedPost` needs mocking too. Search for `jest.mock("aws-sdk"` to see the spots.

l2dy: Thanks! I was wondering why it claims to have received `undefined`.

l2dy:
How come another test fails after adding the mock function? It passed without mock changes.

tommoor:
A flaky test 😢 , I re-ran.

tommoor (approved):

tommoor:
`X-Amz-Algorithm` is the more correct header name but I can't imagine that the tencent implementation is case sensitive – that would be quite a strange decision. The SDK also bumped a few versions as part of this PR.

It might help if you're able to provide the errors in question

Coderhypo:
@tommoor  Thanks for your reply.

Yes, this is an incredible issue, the new headers are indeed more spec-compliant. I also submitted a work order to Tencent Cloud. From the feedback of the work order, they are also more inclined to the headers problem, which is unbelievable.

I'm still setting up a local environment, and I'll do more verification on this situation, determine the root cause, and then see how to fix the problem. 😢

Coderhypo:
Update: through communication with TencentCloud, they said that this problem will be fixed in the future😂.

PR #2833, opened 2021-12-08
feat: Support custom CDN in Dockerfile

l2dy:
`process.env.CDN_URL` is evaluated during compile time in `webpack.config.prod.js`, so it has to be passed to `yarn build` to work.


tommoor:
I think this went unnoticed because it was 99% added for the hosted version of the app, typically self hosters have no need for a CDN :) Anyway, thanks for the fix!

l2dy:
CloudFront has a free tier, so why not ;)

PR #2844, opened 2021-12-11
fix: Bind context to util.promisify

l2dy:
`.bind()` returns `any`, so it no longer benefits from type checking.


tommoor:
Could you detail the issue this change is addressing?

l2dy:
`createPresignedPost` from node_modules/aws-sdk/lib/services/s3.js tries to access `this.config.params`, but with util.promisify, the context is lost, so `this.config` becomes `undefined` and accessing `this.config.params` throws a TypeError. See `https://github.com/nodejs/node/issues/13338`.

```js
var boundParams = this.config.params || {};
```

tommoor:
Makes sense, but testing this locally and on the staging environment never hit this issue curiously

PR #2845, opened 2021-12-11
feat: Add searching of sub actions in command menu

### LINKED ISSUE #2688: Command bar (beta)
Some bugs:

- [x] If document is renamed then command bar results are not rebuilt (@timc1 I think I might need to use the slug as the `id` so that it changes if the title does instead of the uuid)
- [x] Keys left in input after used in sequence (https://github.com/timc1/kbar/issues/95)
- [x] Command bar could get a little wider on large screens
- [x] Result items should not wrap to next line
- [x] Search child commands (https://github.com/timc1/kbar/issues/5)

tommoor:
closes #2688 cc @timc1


timc1 (approved):
🤩

PR #2856, opened 2021-12-13
fix: various bugs on draft sidebar items

thenanyu:
- show children when the only child is the active draft
- disable drop onto draft items
- fix danger button hover color

### DIFF
--- a/app/components/Button.tsx
+++ b/app/components/Button.tsx
@@ -72,7 +72,7 @@ const RealButton = styled.button<{
       fill: ${props.iconColor || props.theme.buttonNeutralText};
     }`
     }
-    
+
 
     &:hover:not(:disabled) {
       background: ${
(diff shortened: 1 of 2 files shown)


app/components/Button.tsx
```diff
@@ -100,7 +100,7 @@ const RealButton = styled.button<{
       background: ${props.theme.danger};
       color: ${props.theme.white};
 
-      &:hover {
+      &:hover:not(:disabled) {
```
thenanyu, line 103: danger buttons were hovering blue due to insufficient specificity

app/components/Sidebar/components/DocumentLink.tsx
```diff
@@ -47,7 +47,8 @@ function DocumentLink(
   const { documents, policies } = useStores();
   const { t } = useTranslation();
   const isActiveDocument = activeDocument && activeDocument.id === node.id;
-  const hasChildDocuments = !!node.children.length;
+  const hasChildDocuments =
+    !!node.children.length || activeDocument?.parentDocumentId === node.id;
```
thenanyu, line 51: Items were not displaying at all when there were no children, but we need to display the list when there's an active draft

tommoor (approved):
perf 🙏

PR #2857, opened 2021-12-14
feat: Add event selection to Slack post integration

### LINKED ISSUE #2757: Duplicate slack notification when creating new document
Whenever I *create* a new document in a collection that has Slack notification turned on, I get both the "published a new document" and "updated a document" notification in the channel.

Thanks in advance!

**To Reproduce**
1. Turn on slack notifications for a collection
2. Create a new document in that collection
3. See two notifications in slack instead of one

**Expected behavior**
I think only the "published" notification should be sent to slack.

**Screenshots**
image

tommoor:
### Done

- feat: Add event selection to Slack post integration

### Todo

- [x] Touch up design
- [x] endpoint tests
- [x] Migrate existing integrations to add `events`


app/models/Integration.ts
```diff
@@ -18,24 +17,11 @@ class Integration extends BaseModel {
...
-      ...data,
-    });
-    extendObservable(this, data);
-    return true;
-  };
-
-  @action
-  delete = () => {
-    return this.store.delete(this);
-  };
```
tommoor, line 38: Not sure why this was duplicated from BaseModel here

app/scenes/Settings/Slack.tsx
```diff
@@ -40,6 +42,16 @@ function Slack() {
     (i) => i.type === "command"
   );
 
+  const groupedCollections = collections.orderedData
```
tommoor, line 45: Sort connected channels to the top

PR #2864, opened 2021-12-17
feat: Add '+' button to create new doc in sidebar

### LINKED ISSUE #2695: Add "new child doc" button in sidebar
### Discussed in https://github.com/outline/outline/discussions/2694

<sup>Originally posted by **dkkb** October 26, 2021</sup>
I used to use Notion and I really like this project, here's some advice:

1. Creating new doc.
When I want to make a new document, I go to the 'Article' folder, move the mouse to the right side, and make a new document in two steps; if I want to make a nested document, I need to take another step.
Let's take a look at Notion, it displays the 'Create doc' menu button, it's simple and elegant.
Maybe it's a good idea to add `Create new doc` menu button. Because adding new doc is popular function, so put this button on the right is ideal.

2. Edit mode
People modify their documents frequently, therefore changing the view mode to edit mode is a crucial design decision. Actually, I appreciate the design in Notion, to change a document in Notion, simply click the doc area,  when you see the vertical cursor. OK, you wan work on it, that's simple. I really like this design.
In outline, we must find the `Edit` button and click, as we all know, the area of doc is large, so it's so easy to enter edit mode. Maybe many guys like writing in a single page, but I like the content sidebar, it displays an outline that helps me think clearly. So I think it's a good idea to edit the doc with the sidebar.

What are your thoughts about it?

Best wishes.</div>

tommoor:
closes #2695


dkkb:
Cheers.

PR #2886, opened 2021-12-29
chore: Typescript database models

### LINKED ISSUE #2798: Type backend models
Follow on from https://github.com/outline/outline/pull/2783 – this is the main thing that will remove hundreds of `@ts-expect-error` comments on the backend.

In the spirit of fast iteration I think to "get things working" we should just add the type annotations to the existing functional model declarations as described here: https://sequelize.org/master/manual/typescript.html#usage-of--code-sequelize-define--code-

However, the new class style looks nice and allows the use of decorators, another future enhancement!

tommoor:
Follow on to the conversion #1282 that left a large amount of suppressions in the backend as the database models were all typed as `any`. This moves the model definitions to [sequelize-typescript](https://www.npmjs.com/package/sequelize-typescript) with all of the decorator goodies that come with this conversion.

In the process we were able to go from 461 -> ~100 `@ts-expect-error` suppressions left in the backend code. There are quite a lot of gotchas with this conversion which I'll leave comments on in the PR.

closes #2798


app/typings/index.d.ts
```diff
@@ -4,63 +4,8 @@ declare module "boundless-arrow-key-navigation";
...
 
-declare module "sequelize-encrypted";
-
 declare module "styled-components-breakpoint";
 
-declare module "formidable/lib/file";
-
-declare module "socket.io-client";
-
-declare module "socket.io-redis" {
```
tommoor, line 15: Moved server and shared typings to the correct folders

server/.babelrc
```diff
@@ -1,24 +1,19 @@
 {
   "presets": [
     "@babel/preset-react",
-    "@babel/preset-typescript",
     [
       "@babel/preset-env",
       {
-        "corejs": {
```
tommoor, line 8: corejs polyfills for reflect metadata were conflicting with `reflect-metadata` so the easiest thing to do here was remove.

server/.babelrc
```diff
@@ -1,24 +1,19 @@
...
+        }
       }
-    ]
+    ],
+    "@babel/preset-typescript"
   ],
   "plugins": [
-    "transform-class-properties",
-    "tsconfig-paths-module-resolver",
+    "babel-plugin-transform-typescript-metadata",
```
tommoor, line 15: To support outputting of decorator metadata to Reflect, this is key to how sequelize-typescript is able to generate database models from type definitions

server/commands/accountProvisioner.test.ts
```diff
@@ -48,7 +49,7 @@ describe("accountProvisioner", () => {
         scopes: ["read"],
       },
     });
-    const authentications = await user.getAuthentications();
+    const authentications = await user.$get("authentications");
```
tommoor, line 52: The syntax changed as part of moving lib

server/models/index.ts
```diff
@@ -1,88 +1,49 @@
...
-  SearchQuery,
-  Share,
-  Star,
-  Team,
-  User,
-  UserAuthentication,
-  View,
-  FileOperation,
-};
+export { default as ApiKey } from "./ApiKey";
```
tommoor, line 1: Nice to reduce the boilerplate in this file down to a single line

server/models/decorators/Fix.ts
```diff
@@ -0,0 +1,50 @@
+/**
```
tommoor, line 1: This is the most unfortunate thing about this PR, all very hacky but hopefully it's somewhat temporary and a better solution will be available as the tooling matures.

server/presenters/__snapshots__/user.test.ts.snap
```diff
@@ -2,27 +2,27 @@
 
 exports[`presents a user 1`] = `
 Object {
-  "avatarUrl": undefined,
```
tommoor, line 5: Hmm, these snapshots weren't great before huh!?

PR #2895, opened 2022-01-03
docs: Update instructions

max-mykhailenko:
Additional details for using redis or postgres installed outside docker

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,17 +47,18 @@ For a manual self-hosted production installation these are the recommended steps
 1. Setup the database with `yarn db:migrate`. Production assumes an SSL connection to the database by default, if
    Postgres is on the same machine and is not SSL you can migrate with `yarn db:migrate --env=production-ssl-disabled`, for example:
 
-   `docker run --rm outlinewiki/outline yarn db:migrate`
+   `docker run --rm --env-file=.env outlinewiki/outline yarn db:migrate`
 
 1. Start the container:
 
-   `docker run outlinewiki/outline`
+   `docker run --env-file=.env outlinewiki/outline`
 
 1. Visit http://you_server_ip:3000 and you should be able to see Outline page
 
    > Port number can be changed using the `PORT` environment variable
 
 1. (Optional) You can add an `nginx` or other reverse proxy to serve your instance of Outline for a clean URL without the port number, support SSL, etc.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,17 +47,18 @@ For a manual self-hosted production installation these are the recommended steps
 1. Setup the database with `yarn db:migrate`. Production assumes an SSL connection to the database by default, if
    Postgres is on the same machine and is not SSL you can migrate with `yarn db:migrate --env=production-ssl-disabled`, for example:
 
-   `docker run --rm outlinewiki/outline yarn db:migrate`
+   `docker run --rm --env-file=.env outlinewiki/outline yarn db:migrate`
```
tommoor, line 50: Would prefer to revert these lines, it makes the assumption that your environment variables are stored in a file. We describe this in the first step.

README.md
```diff
@@ -47,17 +47,18 @@ For a manual self-hosted production installation these are the recommended steps
...
 
-   `docker run outlinewiki/outline`
+   `docker run --env-file=.env outlinewiki/outline`
 
 1. Visit http://you_server_ip:3000 and you should be able to see Outline page
 
    > Port number can be changed using the `PORT` environment variable
 
 1. (Optional) You can add an `nginx` or other reverse proxy to serve your instance of Outline for a clean URL without the port number, support SSL, etc.
+2. (Optional) If you setup your redis or postgres outside docker container you should use additional flag `--net="host"` for docker. In this case you will be able to use `localhost` as host in your `.env` file for redis and postgres.
```
tommoor, line 61:
```suggestion
2. (Optional) If you setup your redis or postgres outside docker container you could use additional flag `--net="host"` for docker. In this case you will be able to use `localhost` as host in your `.env` file for redis and postgres.
```

README.md
max-mykhailenko: Recently I was confused about 5 minutes why command doesn't work. Copy paste instruction is what we need. For now I need to pass `--env` parameter to every docker command, is it right?

tommoor: Well, _it depends_ – what's copy and paste for your setup is not for another. Many environments will inject environment variables through other mechanisms

max-mykhailenko: Ok, I'll remove that. Thank you. Please share your thoughts how to pass env variables in another way. Maybe my workaround not so perfect for production mode.

Issue #2802, opened 2021-11-29
Public parent documents lose content when navigating back


tommoor:
**To Reproduce**
1. Go to https://wiki.generaloutline.com/share/13dec265-c002-4c73-8025-75981f29b0d0/doc/aws-NfelJ2WRAj
2. Click on a subdocument
3. Go back

tommoor:
Ah, interestingly seems like you have to be signed in for this to happen.

iamsaumya:
By content - Do you mean the subtree and breadcrumbs?

PR #2899, opened 2022-01-06
feat: Add navigation sidebar to shared documents

thenanyu:
closes https://github.com/outline/outline/issues/2898
closes https://github.com/outline/outline/issues/2802


app/components/Sidebar/components/SharedDocumentLink.tsx
```diff
@@ -0,0 +1,139 @@
...
+          </>
+        }
+        depth={depth}
+        exact={false}
+        scrollIntoViewIfNeeded={!document?.isStarred}
+        isDraft={isDraft}
+        ref={ref}
+        isActive={() => {
+          return !!isActiveDocument;
+        }}
```
thenanyu, line 116: this definition of `isActive` seems straightforward enough, but a little suspicious. what do you think @tommoor

tommoor: A little suspicious in that the `to` prop should match the location and mark as active automatically, you know the case in which that isn't happening?

thenanyu: when you're at the root of a shared doc, there's no doc slug, so it doesn't match the `to` in the sidebar link. I started by overriding only that one case, but this code is simpler, otherwise it's 

```
return !!(isShareRoot ? isActiveDocument : match)
```

thenanyu:
functionality is done at this point, need to factor out the copy pasta and clean up

app/scenes/Document/components/References.tsx
```diff
@@ -21,19 +21,19 @@ function References({ document }: Props) {
...
-  const isBacklinksTab = location.hash === "#backlinks" || !showParentDocuments;
+  const showChildDocuments = !!children.length;
+  const isBacklinksTab = location.hash === "#backlinks" || !showChildDocuments;
 
-  return showBacklinks || showParentDocuments ? (
+  return showBacklinks || showChildDocuments ? (
     <Fade>
       <Tabs>
-        {showParentDocuments && (
+        {showChildDocuments && (
```
thenanyu, line 36: I renamed this variable because it seemed backwards

app/stores/DocumentsStore.ts
```diff
@@ -37,6 +37,8 @@ type ImportOptions = {
 };
 
 export default class DocumentsStore extends BaseStore<Document> {
+  sharedTreeCache: Map<string, NavigationNode | undefined> = new Map();
```
thenanyu, line 40: 🥦 moving shared tree cache into the store

app/components/Layout.tsx
```diff
@@ -1,149 +1,80 @@
...
-            <Switch>
-              <Route
-                path={`/doc/${slug}/history/:revisionId?`}
-                component={DocumentHistory}
-              />
-            </Switch>
-          </React.Suspense>
-        </Container>
-        <CommandBar />
+
```
thenanyu, line 15: 🍖 This change is effectively a new file with a rename. The old app layout is now `AuthenticatedLayout`.

app/scenes/Document/Shared.tsx
```diff
@@ -31,6 +36,7 @@ export default function SharedDocumentScene(props: Props) {
   const { shareId, documentSlug } = props.match.params;
 
   // ensure the wider page color always matches the theme
+  // TODO: I don't know what this does
   React.useEffect(() => {
     window.document.body.style.background = theme.background;
   }, [theme]);
```
thenanyu, line 42: @tommoor what does this useeffect do here? it doesn't seem to make a difference whether is commented out or not

tommoor: What the comment says :P – I think it can be seen in the area around the page eg if you scroll past the page edges

thenanyu: you'll have to show me sometime, I cannot see any difference, when inspecting `window.body` styles

app/components/Button.tsx
```diff
@@ -158,7 +158,7 @@ const Button = <T extends React.ElementType = "button">(
   const hasIcon = icon !== undefined;
 
   return (
-    <RealButton type={type || "button"} ref={ref} neutral={neutral} {...rest}>
+    <RealButton type={type || "button"} ref={ref} $neutral={neutral} {...rest}>
```
thenanyu, line 161: this was throwing react warnings

app/components/Layout.tsx
thenanyu: Here we factor out all the display elements of the layout to use in both the authenticated context and the sharing context

app/scenes/Document/components/References.tsx
tommoor: Agreed, this was weird

app/stores/DocumentsStore.ts
```diff
@@ -468,9 +474,15 @@ export default class DocumentsStore extends BaseStore<Document> {
...
-        };
+        if (!options.shareId)
+          return {
+            document: doc,
+          };
+        else if (this.sharedTreeCache.has(options.shareId))
+          return {
+            document: doc,
+            sharedTree: this.sharedTreeCache.get(options.shareId),
+          };
```
tommoor, line 485:
```suggestion
        if (!options.shareId) {
          return {
            document: doc,
          };
        } else if (this.sharedTreeCache.has(options.shareId))
          return {
            document: doc,
            sharedTree: this.sharedTreeCache.get(options.shareId),
          };
        }
```

app/stores/DocumentsStore.ts
```diff
@@ -486,9 +498,16 @@ export default class DocumentsStore extends BaseStore<Document> {
       const document = this.data.get(res.data.document.id);
       invariant(document, "Document not available");
 
+      if (options.shareId) {
```
tommoor, line 501: This worked out quite nicely, we could probably combine `fetch` and `fetchWithSharedTree` methods now if we wanted. Perhaps in future

tommoor (commented):
This is great – agreed it's ready to go once tests are passing

app/stores/DocumentsStore.ts
thenanyu: might do as part of the collections work, since collections will need similar treatment. we can make it nice and parallel

thenanyu:
I think this is good to go now 🤞🏼

tommoor:
image

Just one issue – we've gained a 16px margin on the left of publicly shared docs which makes them slightly off-center, it's noticeable by the alignment of the contents icon in the top left mostly

thenanyu:
should be good now I think

tommoor (approved):

Issue #2917, opened 2022-01-09
Add ability to collapse and expand collections that are not active


tommoor:
The same behavior as documents essentially, it is currently too restrictive. The only difference is that collections have icons – we can probably try showing the disclosure when you hover over the sidebar item

tommoor:
I believe the layout should stay the same, but on hover the disclosure should show to the left of the existing collection icon in the gutter and allow toggling of expanded/collapsed.

PR #2931, opened 2022-01-12
Hotfix/0.61.1

anastasia-muraya:
Hello! While using Wiki our team find it really uncomfortable when we open some collection and we can not collapse it to hide all articles. Can you add this function ? https://prnt.sc/269r760


tommoor:
The request is tracked here https://github.com/outline/outline/issues/2917 please stop submitting the same question, many thanks.

PR #2859, opened 2021-12-15
feat: Attachment of Files to Documents using S3

FeralMib:
This is an extension of https://github.com/outline/outline/pull/1588 including the original commit there, rebased and extended on current master.

For further details see the original PR.

The bulk of the required changes are contained in https://github.com/outline/rich-markdown-editor/pull/603

PR #2936, opened 2022-01-13
feat: Attachment of files as block to documents using s3

warnus:
This commit is compatible with [outline/rich-markdown-editor#624](https://github.com/outline/rich-markdown-editor/pull/624)

Same work with #2859

### DIFF
--- a/app/components/Editor.tsx
+++ b/app/components/Editor.tsx
@@ -53,6 +53,16 @@ function Editor(props: Props, ref: React.Ref<any>) {
     },
     [id]
   );
+ 
+  const onUpload = React.useCallback(
+    async (file: File) => {
+      const result = await uploadFile(file, {
+        documentId: id,
+      });
+      return result.url;
+    },
+    [id]
+  );
 
   const onClickLink = React.useCallback(
     (href: string, event: MouseEvent) => {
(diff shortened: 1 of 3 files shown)


tommoor:
Hi @warnus – thank you for the effort here, unfortunately this work is incompatible with the ongoing effort to move the editor code into this codebase (https://github.com/outline/outline/pull/2930) and as such, won't be able to be merged. One of the reasons we're doing that is to make it much faster and easier to develop features like this though!

warnus:
Hi @tommoor . Thanks for the reply. I willl update this again to be compatible. 

By the way, I have a question. Would you check this pr? (https://github.com/outline/rich-markdown-editor/pull/624)

I want to know how to set attribute variable. I'm not familiar with prosemirror, so I'd appreciate it if you could tell me.

What I want to do is pass the file url and file title from markdown to dom.

tommoor:
I will look, but it might not be until next week. I do not plan on maintaining the editor repo any longer – see:
https://github.com/outline/rich-markdown-editor/discussions/481

warnus:
Oh, I' sorry. I didn't see it. Thanks for the letting me know.
So the editor is merged with the Outline project?  Or will it be a private project?
Because I really need the file upload function, so I want to develop it. 🥲

PR #2950, opened 2022-01-19
fix: changing the title and body content in quick succession would trigger unsaved changes warning prompt

### LINKED ISSUE #2949: "unsaved changes" alert upon document creation
Upon the creation of a document, if the title is set, followed by any operations in the content area prior to the initial save, an error alert with "you have unsaved changes" pops up when the initial save is triggered. Regardless of which option is chosen, no changes to the new document are made. Additionally, if the document is manually saved as a draft or published prior to the initial save, the issue still occurs.

**To Reproduce**
1. Create a new document.
2. Add a title and add content within ~5 seconds of the document creation.
3. See error.

**Expected behavior**
Document should save without the alert.

**Screenshots**
Screen Shot 2022-01-18 at 8 43 02 AM

- Install: self hosted
- Version: d0e7f2de65b63f94bdccb99e0d5234af3426799f

 - OS: macOS
 - Browser: Chrome
 - Version 96.0.4664.110

thenanyu:
fixes #2949

### DIFF
--- a/app/scenes/Document/components/Document.tsx
+++ b/app/scenes/Document/components/Document.tsx
@@ -38,6 +38,7 @@ import { isCustomDomain } from "~/utils/domains";
 import { emojiToUrl } from "~/utils/emoji";
 import { isModKey } from "~/utils/keyboard";
 import {
+  editDocumentUrl,
   documentMoveUrl,
   documentHistoryUrl,
   editDocumentUrl,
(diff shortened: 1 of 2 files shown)


shared/i18n/locales/en_US/translation.json
```diff
@@ -40,6 +40,7 @@
   "Dark": "Dark",
   "Light": "Light",
   "System": "System",
+  "Appearance": "Appearance",
```
thenanyu, line 43: I have no idea what this is lol

tommoor: auto-generated – it should be on master already but nbd

app/scenes/Document/components/Document.tsx
```diff
@@ -466,9 +466,20 @@ class DocumentScene extends React.Component<Props> {
...
                   }
-                  message={t(
-                    `You have unsaved changes.\nAre you sure you want to discard them?`
-                  )}
+                  message={(location, action) => {
+                    if (
+                      // a URL replace matching the current document indicates a title change
+                      // no guard is needed for this transition
+                      action === "REPLACE" &&
+                      location.pathname.includes(document.url)
```
tommoor: Can we use just the edit path?

thenanyu:
@tommoor I think this should be good now, just tested

tommoor (approved):

PR #2953, opened 2022-01-19
feat: Add reordering to starred documents

### LINKED ISSUE #2484: Allow drag to reorder of starred documents
This seems like it was inevitable 😃 

We can add `index` field to `stars` the same way we have on collections, new starred items should go to the top of the list

tommoor:
The main work here is converting the `Star` model to being a first-class citizen on the frontend like `Pin` instead of being a property on documents. You can think of `Star` as basically a private `Pin`.

This means a bit more boilerplate but will allow for more efficient queries with less joins for document lists on the backend once the old code paths are removed in a followup PR (removing the `withStarred` scope and `document.starred` attribute)

closes #2484


tommoor:
@thenanyu optional review material, If you don't look I'll merge in a few days

server/commands/starDestroyer.ts
```diff
@@ -0,0 +1,54 @@
...
+};
+
+/**
+ * This command destroys a document star. This just removes the star itself and
+ * does not touch the document
+ *
+ * @param Props The properties of the star to destroy
+ * @returns void
+ */
+export default async function starDestroyer({
```
thenanyu, line 23: 😆

tommoor: ⭐ 💥

PR #2957, opened 2022-01-19
fix: issue-2956 

### LINKED ISSUE #2956: TypeError: Cannot read properties of undefined (reading 'children')
Sentry Issue: [OUTLINE-CLOUD-1HT](https://sentry.io/organizations/outline-jl/issues/2943054431/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'children')
  File "/app/build/server/models/Collection.js", line 213, in Object.getDocumentTree
    children: (0, _collections.sortNavigationNodes)(result.children, sort)
  File "/app/build/server/routes/api/documents.js", line 579, in null.<anonymous>
    sharedTree: share && share.includeChildDocuments ? collection.getDocumentTree(share.documentId) : undefined
  ?, in runMicrotasks
```

thenanyu:
fixes #2956 

- when a draft sharing is updated, the auth check failed because it could not locate the draft document when performing an auth check against collections. fixed here

- also uncovers and fixes a related issue: check the "include children" box when sharing a draft document causes an authorization error

- there's an open question on whether we should hide this option from draft shares. this patch make it a safe state but the UX is confusing enough that I think we should remove the option for drafts, so I also hide the secondary checkbox from draft documents

### DIFF
--- a/app/scenes/Document/components/SharePopover.tsx
+++ b/app/scenes/Document/components/SharePopover.tsx
@@ -165,7 +165,7 @@ function SharePopover({
         <HelpText>{t("Only team members with permission can view")}</HelpText>
       )}
 
-      {canPublish && share?.published && (
+      {canPublish && share?.published && !document.isDraft && (
         <SwitchWrapper>
           <Switch
             id="includeChildDocuments"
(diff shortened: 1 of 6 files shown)


tommoor (approved):
Perfect

PR #2961, opened 2022-01-20
fix: Missing space character

rossmeissl:
Interpose a space in the templates scene empty state.


tommoor (commented):
Good find, thanks! We need to move the space outside of the translation string – will probably need to be `{" "}` otherwise it will invalidate all of the existing translations unnecessarily

Issue #2966, opened 2022-01-21
feat: move the star button and adjust the editable title area.


dkkb:
**Screenshots**
If applicable, add screenshots or videos to help explain your problem.

**Expected behavior**

Here's the situation, when I want to edit the title, it's difficult to click the tail of title, because the `star` button is so close to the title. I only star the very important documents, therefore I don't use the star buton too much, but editing title is a frequently used function.

Here's what I'd recommend:
1. Move the `star` button to the top area, near the `New doc` button. It's very useful to fix a button to a certain location just like `Share` and `New doc`, but now if we wanna `star` the document, we need to find the tail of the title, which takes some time because document lengths vary. This will also allow us to edit the title without accidentally pressing the `star` button.

By the way, I find that we can `star` the document while editing it, is this the rationale for placing this button near the title so that we may utilize it while editing documents?

2. Extend the editable title's width. It's incredibly user-friendly. As I previously stated, if the area is not fixed, we must move the mouse to the area of the title text. If the width of the editable title can be increased (maybe we can use `display: block`), we may simply click someplace behind the text and begin writing, that's simple.

<img src="https://user-images.githubusercontent.com/82504881/150457259-163fba0a-e791-4629-8d04-b5aa6f299955.png" width=300>

tommoor:
It is a little finicky, but I'm not convinced that the solution is just to remove it. We can reduce the clickable area on the left of the star considerably, if that fails then moving to the header might work.

PR #2967, opened 2022-01-21
feat: Adjust editable title area.

dkkb:
Related to  https://github.com/outline/outline/issues/2966

Before

After


tommoor:
Thanks for pushing on this – I went a [different way](https://github.com/outline/outline/commit/37a29934ec88cb50db4f1497aa7d5f046f4a2888) with the solution but I'm quite sure you will find it works much better than before while retaining the star.

PR #2968, opened 2022-01-21
fix: Disabling editor embeds should work with collaborative editing

### LINKED ISSUE #2919: "Rich service embeds" not respected when collaboration enabled
With the new seamless editor, the "Rich service embeds" setting is not respected. Even if it is disabled in settings, embeds will still be shown in the editor. Tested on both self-hosted and getoutline.com.

_Originally posted by @Snuupy in https://github.com/outline/outline/issues/811#issuecomment-1008228837_

tommoor:
Displays supported embeds as a clickable widget when rich embeds are disabled, this is a change from how it used to be where they would revert to a plain link – but that method of switching between block and inline nodes is not compatible with the collaborative editing structure. In the long run this actually seems like a nicer approach anyway.

image

This also supports one user having embeds disabled for the doc while another does not and they can still collabo without issue.

closes #2919


app/components/Editor.tsx
```diff
@@ -94,7 +91,7 @@ function Editor(props: Props, ref: React.Ref<any>) {
         ref={ref}
         uploadImage={onUploadImage}
         onShowToast={onShowToast}
-        embeds={props.disableEmbeds ? EMPTY_ARRAY : embeds}
```
tommoor, line 97: It's nice to be able to remove this hack

app/menus/DocumentMenu.tsx
```diff
@@ -410,20 +410,14 @@ function DocumentMenu({
               type: "button",
               title: t("Enable embeds"),
               onClick: document.enableEmbeds,
-              visible:
-                !!showToggleEmbeds &&
-                document.embedsDisabled &&
-                !team.collaborativeEditing,
+              visible: !!showToggleEmbeds && document.embedsDisabled,
```
tommoor, line 413: Restore this functionality

shared/editor/nodes/Embed.tsx
```diff
@@ -57,22 +58,24 @@ export default class Embed extends Node {
...
+    const { embeds, embedsDisabled } = this.editor.props;
 
     // matches are cached in module state to avoid re running loops and regex
-    // here. Unfortuantely this function is not compatible with React.memo or
+    // here. Unfortunately this function is not compatible with React.memo or
     // we would use that instead.
     const hit = cache[node.attrs.href];
     let Component = hit ? hit.Component : undefined;
     let matches = hit ? hit.matches : undefined;
+    let embed = hit ? hit.embed : undefined;
```
tommoor, line 69: This is due a refactor tbh, might do it on this branch yet

Issue #2974, opened 2022-01-22
Cannot Unnest document


ChuckJonas:
**To Reproduce**
1. Create a collection called `testing`
2. Create 2 documents called `A` and `B` under the `testing` collection
2. Drag `B` on top of `A`.  `B` is now a child of `A`
3. Drag `B` back to the `testing` collection
4. `B` will remain nested under `A`

**Expected behavior**
I would expect this would put it back in the collection root.  I also tried dragging it "along side" (in between) other top level documents, but it's not possible)

**Outline (please complete the following information):**
- Install: getoutline.com

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser brave
 - Version V1.34.81

ChuckJonas:
The only way I can figure out how to unnest the document is to drag it to a different collection, then drag it back

tommoor:
Interesting, I think I can imagine what is happening here – I think it sounds like alphabetic sorting is turned on? Change this option to manual and then you will be able to drag between top level items

![image](https://user-images.githubusercontent.com/380914/150626805-f7a6ca54-7788-4a9f-b4d3-27972491f995.png)

ChuckJonas:
ya that was it!  Seems like you should still be able to drag in between items even with alphabetical search on (it would just auto-sort after).

As a new user, I didn't even know I had Alphabetical search was on.

ChuckJonas:
Can you confirm this is the line which results in no change when dropping the nested item on its own collection?

https://github.com/outline/outline/blob/main/app/components/Sidebar/components/CollectionLink.tsx#L84

I'm evaluating this library for our team.  Only spent about 4m scanning the codebase but it seems very easy to follow 👍 .

tommoor:
That seems like that spot, yep!

ChuckJonas:
@tommoor I spent a minute to see if I could get the app running and fix the issue.  Not sure if simply checking the depth here is actually sufficient, but it seems to work as expected.

Didn't have a chance to look at the test suite yet either, but wanted to confirm:

A: this change is desired
B: this approach is correct

before I continue any further.  

If tests are required and you'd like me to move forward, it would be great if you could point me to the approach spec.

Thanks!

ChuckJonas:
PS: I noticed a `no-op` error when testing the following scenario:

```
collection 1
   A
       B
           C
collection 2
```
-->
```
collection 1
   A
       
collection 2
   B
       C
```

Confirmed it exists before my change.

```
[...]

PR #2975, opened 2022-01-22
fix: issue-2974 reparent nested document when dropped on current collection

ChuckJonas:
#2974

WIP:  Need to confirm this is the correct approach & if new unit tests are appropriate

### DIFF
--- a/app/components/Sidebar/components/CollectionLink.tsx
+++ b/app/components/Sidebar/components/CollectionLink.tsx
@@ -81,7 +81,12 @@ function CollectionLink({
       const { id, collectionId } = item;
       if (monitor.didDrop()) return;
       if (!collection) return;
-      if (collection.id === collectionId) return;
+
+      const document = documents.get(id);
+      if (collection.id === collectionId && !document?.parentDocumentId) {
+        return;
+      }
+
       const prevCollection = collections.get(collectionId);
 
       if (


app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -81,7 +81,15 @@ function CollectionLink({
       const { id, collectionId } = item;
       if (monitor.didDrop()) return;
       if (!collection) return;
-      if (collection.id === collectionId) return;
+
+      const document = documents.get(id);
+      if (
+        collection.id === collectionId &&
+        document?.parentDocumentId?.length === undefined
```
ChuckJonas, line 88: FYI: `parentDocumentId` is types as `string | undefined` but this was actually returning `null` when a parent is not set. Checking `length` seemed like the safest approach.

tommoor: The types are very new – probably needs tweaking :) Either way an empty check should be fine `!document?.parentDocumentId`

ChuckJonas: 👍  updated

I think I've been conditioned from getting burned by `0 == false` too many times in the past...  Obviously not a concern here (unless `0` is a valid parent id).

tommoor (approved):

PR #2981, opened 2022-01-23
fix: Flash of content when selecting text on RHS of document

### LINKED ISSUE #2977: Occasional flash of text content when selecting
Sometimes when you're selecting within the editor the text disappears for an individual frame, need to get to the bottom of this jank – it makes the editor feel much less "solid".

tommoor:
closes #2977

### DIFF
--- a/app/editor/components/FloatingToolbar.tsx
+++ b/app/editor/components/FloatingToolbar.tsx
@@ -37,7 +37,13 @@ function usePosition({
   const viewportHeight = useViewportHeight();
   const isTouchDevice = useMediaQuery("(hover: none) and (pointer: coarse)");
 
-  if (!active || !menuWidth || !menuHeight || isSelectingText) {
+  if (
+    !active ||
+    !menuWidth ||
+    !menuHeight ||
+    !menuRef.current ||
+    isSelectingText
+  ) {
     return defaultPosition;
   }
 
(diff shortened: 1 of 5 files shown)


app/editor/components/FloatingToolbar.tsx
```diff
@@ -146,54 +152,54 @@ function usePosition({
...
+        <Wrapper
+          active={props.active && position.visible}
+          ref={menuRef}
+          offset={position.offset}
+          style={{
+            top: `${position.top}px`,
+            left: `${position.left}px`,
+          }}
+        >
+          {props.children}
```
tommoor, line 197: Key is to always render the children otherwise the first render has the wrong width and it results in the toolbar being mispositioned off the right of the screen

app/editor/components/SelectionToolbar.tsx
```diff
@@ -229,27 +229,25 @@ export default class SelectionToolbar extends React.Component<Props> {
     }
 
     return (
-      <Portal>
```
tommoor, line 232: Double portal – `FloatingToolbar` is already portalled

app/editor/components/ToolbarMenu.tsx
```diff
@@ -14,12 +14,12 @@ type Props = {
 };
 
 const FlexibleWrapper = styled.div`
+  color: ${(props) => props.theme.toolbarItem};
   display: flex;
   gap: 8px;
 `;
 
 function ToolbarMenu(props: Props) {
-  const theme = useTheme();
```
tommoor, line 22: Just nice to avoid `useTheme` if not really needed

app/hooks/useComponentSize.ts
```diff
@@ -4,8 +4,8 @@ export default function useComponentSize(
   ref: React.RefObject<HTMLElement>
 ): { width: number; height: number } {
   const [size, setSize] = useState({
-    width: 0,
-    height: 0,
+    width: ref.current?.clientWidth || 0,
```
tommoor, line 7: Should at least attempt to set the right size on mount

PR #2985, opened 2022-01-23
Fix: consistently check allowed domains

eug-vs:
### Context
Currently the way allowed domains are checked is (just an example): https://github.com/outline/outline/blob/390a1343b788af56ad52f9bb267edf82e65920bf/server/routes/auth/providers/oidc.ts#L87-L92
Notice that empty `ALLOWED_DOMAINS` will result into this error never being thrown thus allowing all domains. This is desired behavior. 

### Problem
In this piece of code introduced in #1954 we forgot to check the length of `ALLOWED_DOMAINS`. As a result empty `ALLOWED_DOMAINS` in self-hosted installations will be equivalent to all domains being invalid. As a workaround new users' domains have to be explicitly whitelisted (which is a big pain as we can't even wildcard it or smth).
https://github.com/outline/outline/blob/390a1343b788af56ad52f9bb267edf82e65920bf/server/commands/teamCreator.ts#L59-L75

### Solution
Add function that checks if domain is allowed and consistently use it across application. The original function is preserved in case we ever need it, but I guess we could just put it into new one as well.

### DIFF
--- a/server/commands/teamCreator.ts
+++ b/server/commands/teamCreator.ts
@@ -1,7 +1,7 @@
 import invariant from "invariant";
 import Logger from "@server/logging/logger";
 import { Team, AuthenticationProvider } from "@server/models";
-import { getAllowedDomains } from "@server/utils/authentication";
+import { isDomainAllowed } from "@server/utils/authentication";
 import { generateAvatarUrl } from "@server/utils/avatars";
 import { MaximumTeamsError } from "../errors";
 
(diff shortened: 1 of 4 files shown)


eug-vs:
I believe the same problem was also discussed here: https://github.com/outline/outline/discussions/2641

tommoor:
Thanks @eug-vs – solid improve.

PR #2986, opened 2022-01-24
fix: Self-hosted should show signin options for all configured authentication methods

### LINKED ISSUE #2678: OIDC login fails for multiple users with different e-mail domains
When you self host Outline and use the OIDC auth provider only people with the same domain in their email is able to log in.

Due to the restriction of only having one team when selfhosting, and you are automatically assigned a team based upon your email host name, two people with different domains in their emails are not able to log in to the same outline instance.

The first time you log in a team is created, if person had an @gmail.com address then other users that authenticate through the same OIDC provider will not be able to log in if they have anything else than @gmail.com, e.g. if they have an @outlook.com address etc.

The following console error is thrown:
```
ClientError [UnauthorizedError]: The maximum number of teams has been reached
    at new AuthenticationError (/opt/outline/build/server/errors.js:32:34)
    at accountProvisioner (/opt/outline/build/server/commands/accountProvisioner.js:42:11)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async OAuth2Strategy._verify (/opt/outline/build/server/routes/auth/providers/oidc.js:98:22) {
  redirectUrl: 'https://wiki.itemize.no',
  id: 'authentication_required'
}
```

I suspect that maybe this piece of code here is to blame. In my mind it would make more sense if we only looked up the team based upon the name, which is static when self-hosting, instead of looking up using the (name, domain, subdomain) tuple.
https://github.com/outline/outline/blob/51971d2c9af0ac3db6206e53b0a38062db308a81/server/routes/auth/providers/oidc.js#L93-L98

tommoor:
### Problem

Currently when self hosting once the first team has been registered the list of authentication options is loaded from that team, even if you configure more in the environment (eg Azure + Google).

### Solution

- All configured auth options are shown
- Users can signin as long as the domain is in the list of allowed domains

closes #2678
closes #2659


server/routes/api/auth.ts
```diff
@@ -10,18 +10,19 @@ import providers from "../auth/providers";
...
       // guest sign-in is an exception as it does not have an authentication
       // provider using passport, instead it exists as a boolean option on the team
       if (provider.id === "email") {
-        return team && team.guestSignin;
+        return team?.guestSignin;
       }
 
       return (
         !team ||
+        process.env.DEPLOYMENT !== "hosted" ||
```
tommoor, line 25: This is 🍖 – don't filter available auth providers on login screen

server/commands/teamCreator.ts
```diff
@@ -54,27 +53,24 @@ export default async function teamCreator({
...
-        authenticationProvider: authP,
-        team,
-        isNewTeam: false,
-      };
-    }
+    if (team) {
+      // If the domain is allowed then we want to assign to the existing team,
+      // otherwise we prevent the creation of another team on self-hosted
+      // instances.
+      if (domain && isDomainAllowed(domain)) {
```
tommoor, line 62: isDomainAllowed returns true when no ALLOWED_DOMAINS value is set, this is going to be a change in behavior for existing installations. I don't yet have the best answer for this but it's not quite mergable as it stands

Limezy: Couldn't you add a flag in the .env like MULTI_AUTH=true or false, that defaults to false ?
Existing installation behavior wouldn't change as long as MULTI_AUTH is not set or set to false ?
I understand it's not ideal to add new .env variables but just trying to understand what's blocking that PR

tommoor: I definitely don't want to split authentication codepaths like this, it will lead to security bugs in the long run and more unnecessary config should always be a last resort.

If we get to this (https://github.com/outline/outline/issues/3412) within the next month then it might be a nice middleground – and at this point we could easily advise that all self-hosted installations add in allowed domains in team settings after upgrading.

Limezy: Many thanks for your answer, and very nice to see that you already have a solution in mind for the "multi-auth" case !

JacobBlana:
We are currently having google auth on our self-hosted solution. I am trying to add gitlab login option all afternoon . Do i undestand correctly that it is not possible at the moment? Perhaps will it ever be?

tommoor:
It will be possible when this lands, it's blocked by https://github.com/outline/outline/issues/3412 which is under develoment.

In the meantime you could get it working by adding all the env variables and then adding an extra row to `authentication_providers` table manually. `name=oidc`, `providerId=<yourdomain.com>`

JacobBlana:
Omg man.. Sweet.. Actually works 🤯 Any idea when will the feature hit production release ? And mainly will it cause any issues that i added the db row myself ? Thanks.

smartygus:
Really glad to stumble across this info.

I'm kind of doing it the other way around: I started with Gitlab, and now our organisation has Azure AD and I wanted to add that, but wasn't seeing any effect after adding the AZURE ENV variables. It took a bit of digging through the code to figure out what the value of `providerId` for the `azure` authentication provider should be, but as far as I can tell it's the organisation's Tenant ID (which can be found in the Azure Admin Portal directly on the overview page).

I manually added the row to the table in Postgres and it seems to work okay, but like @JacobBlana I'm also unsure if this modification will just seamlessly fit in with future releases. Any advice here @tommoor?

I'm also wanting to actually drop Gitlab and use Azure exclusively, but unsure what happens to existing users. When I logged in via Azure with my own account, it ended up creating a new account, because the email address returned from Azure was Michael.Smart@ whereas via Gitlab it was michael.smart@. It seems like all our email addresses in Azure are capitalised like this, so all existing users will also likely end up with additional accounts. We don't have that many users because I haven't rolled Outline out fully yet, so in this situation I can probably work around it, by just letting user's re-login and then re-assigning them to collections and groups. But it seems like it would be advantageous for emails to be handles case-insensitively and that an existing user could have multiple authentication methods. I can create a separate issue for this if that's better.

almereyda:
To chime in quickly on a small point, it appears #1945 will allow to link multiple IdPs to a single user account in Outline.

tommoor:
> I manually added the row to the table in Postgres and it seems to work okay, but like @JacobBlana I'm also unsure if this modification will just seamlessly fit in with future releases. Any advice here @tommoor?

It will be fine as long as the values in PG are correct.

> it ended up creating a new account, because the email address returned from Azure was Michael.Smart@

Currently accounts are never matched by email address, so it's not to do with capitalization. Each account is unique based on the vendor id.

rockman88v:
> It will be possible when this lands, it's blocked by #3412 which is under develoment.
> 
> In the meantime you could get it working by adding all the env variables and then adding an extra row to `authentication_providers` table manually to link it to your existing team. `name=oidc`, `providerId=<yourdomain.com>`

@tommoor Can you hint how to add a row in authentication_providers for slack?

tommoor:
`name=slack`, `providerId=<SlackTeamID>` (The Slack team id begins with "T")

rockman88v:
> viderId=<yourd
What I have now is getoutline selfhost with Google authen. But I could not find any record in both tables: **authentication_providers**  and **teams**
        table_name        | column_name |        data_type         | is_nullable
--------------------------+-------------+--------------------------+-------------
 authentication_providers | id          | uuid                     | NO
 authentication_providers | enabled     | boolean                  | NO
 authentication_providers | teamId      | uuid                     | NO
 authentication_providers | createdAt   | timestamp with time zone | NO
 authentication_providers | name        | character varying        | NO
 authentication_providers | providerId  | character varying        | NO
 
Now I want to change authen from Google to Slack. I comment all Google env in docker.env file and reinstall from begining but Slack authen is still not appear. 
Can you help me to solve this? 

PS: This is how i tried to add manual record in postgres and it failed:
outline=# INSERT INTO authentication_providers(name,providerId,enabled) VALUES ('slack', 'T03HJNDBXXX', 'true');
ERROR:  column "providerid" of relation "authentication_providers" does not exist

tommoor:
You need to quote camelCase fields in your query. "providerId"

rockman88v:
> providerId

yep, double quote the colum name with camelCase fix the issue of colum name not found. But it require teamId, how can Iget this for slack? 

outline=# INSERT INTO authentication_providers(id,name, "providerId", enabled) VALUES (gen_random_uuid(),'slack', 'T03HJNDBPUP', true);
ERROR:  **null value** in column "**teamId**" of relation "authentication_providers" violates not-null constraint
DETAIL:  Failing row contains (452e3fce-98bd-4917-88c5-bdafbaedf93c, slack, T03HJNDBPUP, t, null, null).

Issue #2980, opened 2022-01-22
Sidebar on shared documents needs some space at the bottom


tommoor:
Otherwise the logo overlaps

![image](https://user-images.githubusercontent.com/380914/150657779-bcba85e1-3b7c-4bc4-bd86-431b76d09e37.png)

PR #2997, opened 2022-01-25
fix: add toc to mobile views and account for branding on shared view layouts

### LINKED ISSUE #2801: Breadcrumbs should be visible on shared docs on mobile
Currently, they are not

image

thenanyu:
addresses the tertiary issues identified in https://github.com/outline/outline/issues/2801

- makes the app branding more visually pleasing on shared pages
- shows the table of contents button on shared views

I opted *not* to display the full breadcrumb at this time because we just aded the navigation sidebar, which serves much the same purpose in most circumstances. Given that real estate is limited, I think this is fine for now.

closes https://github.com/outline/outline/issues/2980
closes #2801


tommoor:
image

Thinking on mobile that the branding should just act like a standard footer underneath the page content – it's still pretty easy to have it overlaid on content with these changes

thenanyu:
@tommoor I put the branding inline on mobile, I think this is good to go

tommoor:
Are you also seeing the toc button inset?

image

thenanyu:
latest updates restructure headers across the app to include the mobile menu button. This is a more cohesive design and allows us to move away from positioning hacks

tommoor (approved):

PR #3008, opened 2022-01-26
fix: remove scrollbar on x-axis in toc

iamsaumya:


iamsaumya:
Any particular reason why we are debouncing the showToast?

tommoor:
It was a copy/paste from `Settings -> Security`, I think it felt more natural on that screen as it has multiple checkboxes – currently a non-issue here and could be removed

shared/utils/parseDocumentSlug.ts
```diff
@@ -11,5 +11,7 @@ export default function parseDocumentSlug(url: string) {
     }
   }
 
-  return parsed.replace(/^\/doc\//, "");
+  return parsed.lastIndexOf("/doc/") === 0
```
tommoor, line 14: gonna need context/tests for this change 😄

iamsaumya: `- [ ] Setup the [Slack integration](/settings/integrations/slack)`

In the Getting started doc, hovering over this led to API request with `/settings/integrations/slack` as Id.

PR #3014, opened 2022-01-27
hide image caption when blank

ChuckJonas:
Helps address #3013

B16z7HYPWf

### DIFF
--- a/shared/editor/nodes/Image.tsx
+++ b/shared/editor/nodes/Image.tsx
@@ -548,7 +548,7 @@ const Caption = styled.p`
   cursor: text;
 
   &:empty:not(:focus) {
-    visibility: hidden;
+    display: none;
   }
 
   &:empty:before {
@@ -570,6 +570,6 @@ const ImageWrapper = styled.span`
   }
 
   &.ProseMirror-selectednode + ${Caption} {
-    visibility: visible;
+    display: inherit;
   }
 `;


tommoor:
Hi @ChuckJonas – this is purposeful as the caption input/placeholder is shown when you select the image – without it selecting images causes the document below the image to reflow

ChuckJonas:
@tommoor I had a feeling it might have been done on purpose.  I guess I would just challenge this statement a bit: 

> without it selecting images causes the document below the image to move down

Is this actually undesired behavior?

I think this would only happen in "edit" mode.  And an element changing container size while you are potentially mutating it seems reasonable.  IMO, it felt natural and intuitive enough.

Regardless, this PR would only slightly improves the whitespacing issue.  I guess if the "trailing line" was fixed, this might not be needed.

No hard feelings if you reject this.

tommoor:
I think if the trailing whitespace is fixed then the caption can be made tight enough that it's not considered an issue

ChuckJonas:
Just for reference, here's a side by side comparison of before and after: 

Communication_Policy_-_Callaway_Cloud_Consulting_png_and_Communication_Policy_-_Callaway_Cloud_Consulting

tommoor:
@ChuckJonas I've given it more thought and have changed my opinion on this, I incorporated your suggestions into some other image positioning fixes (^above)

thanks!

PR #3029, opened 2022-01-29
feat: add the ability to choose default collection

### LINKED ISSUE #2912: Add ability to choose "Home" collection

iamsaumya:
closes #2912 

- [x] Unit tests


iamsaumya:
Github copilot wrote tests so accurately 😲

app/components/InputSelect.tsx
```diff
@@ -35,6 +36,8 @@ export type Props = {
   options: Option[];
   note?: React.ReactNode;
   onChange: (value: string | null) => void;
+  renderOption?: (option: Option, select: SelectStateReturn) => React.ReactNode;
```
tommoor, line 39: I don't understand why these changes are needed, the idea is that you can pass the rendered items directly into the `options[]` prop as elsewhere so custom render methods aren't required.

One pattern we've used is to just create a parent component per-select-type, so in this case there might be a `DefaultCollectionSelectInput` for example that is capable of loading the data it needs and passing the correct values to the `options` prop.

Does that make sense?

app/components/Sidebar/components/Version.tsx
```diff
@@ -9,18 +9,17 @@ export default function Version() {
...
       );
       const releases = await res.json();
 
-      for (const release of releases) {
-        if (release.tag_name === `v${version}`) {
-          return setReleasesBehind(out);
-        } else {
-          out++;
-        }
+      if (Array.isArray(releases)) {
```
tommoor, line 17: This change should be removed/separate PR with explanation

app/components/SocketProvider.tsx
```diff
@@ -186,14 +187,17 @@ class SocketProvider extends React.Component<Props> {
...
-          const collection = collections.get(collectionId) || {};
+          const collection = collections.get(collectionId);
 
           if (event.event === "collections.delete") {
-            const collection = collections.get(collectionId);
-
             if (collection) {
               collection.deletedAt = collectionDescriptor.updatedAt;
             }
+            updatePreferredCollection(
```
tommoor, line 196: This change should come down automatically as it's own `teams.update` websocket event when the team is updated on the server, that would mean a lot of cleanup here.

Have you been following the commands refactor? I think the logic in the teams update endpoint needs pulling into a `teamUpdater` command that you can then use everywhere that you're currently calling `await team.update` directly and it will take care of this.

If that's not clear I could probably commit this work separately for you to pull in.

app/components/SocketProvider.tsx
```diff
@@ -209,17 +213,31 @@ class SocketProvider extends React.Component<Props> {
 
           // if we already have the latest version (it was us that performed
           // the change) then we don't need to update anything either.
-          // @ts-expect-error ts-migrate(2339) FIXME: Property 'updatedAt' does not exist on type '{}'.
-          const { updatedAt } = collection;
 
-          if (updatedAt === collectionDescriptor.updatedAt) {
+          if (
```
tommoor, line 217: Similarly re: should come down in it's own event already

app/models/Team.ts
```diff
@@ -27,6 +27,10 @@ class Team extends BaseModel {
   @observable
   documentEmbeds: boolean;
 
+  @Field
+  @observable
+  preferredCollectionId: string | null;
```
tommoor, line 32: Rename might be a bit of a pain, but better now than later. "Preferred" has some extra connotations in English, "default" is also language we use elsewhere

```suggestion
  defaultCollectionId: string | null;
```

app/scenes/Settings/Features.tsx
```diff
@@ -3,34 +3,94 @@ import { BeakerIcon } from "outline-icons";
...
   const [data, setData] = useState({
     collaborativeEditing: team.collaborativeEditing,
+    preferredCollectionId: team.preferredCollectionId,
   });
 
-  const handleChange = React.useCallback(
-    async (ev: React.ChangeEvent<HTMLInputElement>) => {
-      const newData = { ...data, [ev.target.name]: ev.target.checked };
-      setData(newData);
+  React.useEffect(() => {
```
tommoor, line 30: Would prefer this option on the `Team -> Details` screen, `Features` is more for enabling/disabling of entire beta features.

Maybe should change "Details" to "General" soon 🤔

app/scenes/Settings/Features.tsx
```diff
@@ -56,6 +116,12 @@ function Features() {
           </Trans>
         }
       />
+      <PreferredCollection
```
tommoor, line 119: As mentioned in `InputSelect` comment, I think this component can just fetch it's own data rather than having a parent component responsible.

app/scenes/Settings/components/PreferredCollection.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { IconWrapper } from "~/components/Sidebar/components/SidebarLink";
+
+type PreferredCollectionProps = {
+  collections: Collection[];
+  fetching: boolean;
+  onPreferredCollectionChange: (value: string) => void;
+  preferredCollectionId: string | null;
+};
+
+const PreferredCollection = ({
```
tommoor, line 19:
```suggestion
const DefaultCollectionInputSelect = ({
```

app/stores/CollectionsStore.ts
```diff
@@ -162,6 +165,31 @@ export default class CollectionsStore extends BaseStore<Collection> {
     }
   }
 
+  @action
+  fetchPublicCollections = async (
```
tommoor, line 169: Good news – this extra method isn't needed. All collections are guaranteed to be available on the client. But you could create a getter for public collections, something like:

```javascript
@computed
get publicCollections() {
  return this.orderedData.filter(collection => ["read", "read_write"].includes(collection.permission));
}
```

app/utils/updatePreferredCollection.ts
```diff
@@ -0,0 +1,27 @@
+import Collection from "~/models/Collection";
+import Team from "~/models/Team";
+
+const updatePreferredCollection = (
```
tommoor, line 4: I think you'll find this file isn't needed after above changes

server/migrations/20220129092607-add-preferredCollectionId.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn("teams", "preferredCollectionId", {
```
tommoor, line 5:
```suggestion
    await queryInterface.addColumn("teams", "defaultCollectionId", {
```

server/migrations/20220129092607-add-preferredCollectionId.js
```diff
@@ -0,0 +1,15 @@
...
+      type: Sequelize.UUID,
+      defaultValue: null,
+      allowNull: true,
+    });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeColumn("teams", "preferredCollectionId");
```
tommoor, line 13:
```suggestion
    await queryInterface.removeColumn("teams", "defaultCollectionId");
```

tommoor (changes requested):
Thank you for taking this on, hope you don't mind the extensive feedback

app/components/Sidebar/components/Version.tsx
iamsaumya: Explanation - I triggered the API rate-limiting for my IP. It was a pain when the request failed, and I had to close the error screen every time to get to the `features` page.

app/scenes/Settings/Features.tsx
iamsaumya: I didn't add it on the `Details` page because it felt more like a page to view the details of the team than to use features to change the Outline default flow.

app/stores/CollectionsStore.ts
iamsaumya: It was intended for the case when we open the details page directly. Let's make the usual `collections.list` request and then use the `publicCollection` to get the collections. What do you think?

tommoor: Agreed

app/scenes/Settings/Features.tsx
tommoor: Yea it's understandable it's not clear yet - I think adding more settings under details is okay for now

iamsaumya:
Thoughts on showing a confirm dialog when the default collection is being made private?

tommoor:
We could add a warning in the collection permissions dialog before saving, this would be a nice touch but I wouldn't make it an entirely separate dialog

iamsaumya:
One more point I want to make 😅  Should we present the default collection behaviour in UI as the "Landing Page" rather than "Default Collection"? Sounds more friendly and apt.

app/components/DefaultCollectionInputSelect.tsx
```diff
@@ -0,0 +1,116 @@
...
+          setFetchError(error);
+        } finally {
+          setFetching(false);
+        }
+      }
+    }
+    load();
+  }, [showToast, fetchError, t, fetching, collections]);
+
+  const onSelectCollection = React.useCallback(
```
tommoor, line 44: Can we move this handling upto the parent form like a standard input? Just expose `onChange`. It's just a bit weird that it immediately triggers the "Saving…" and toast on the button as soon as you change it.

app/components/DefaultCollectionInputSelect.tsx
```diff
@@ -0,0 +1,116 @@
...
+  );
+
+  if (fetching) {
+    return null;
+  }
+
+  return (
+    <InputSelect
+      value={team.defaultCollectionId ?? "home"}
+      label={t("Collection")}
```
tommoor, line 104:
```suggestion
      label={t("Start view")}
```

I like your suggestion of making this language friendlier… perhaps?

app/components/DefaultCollectionInputSelect.tsx
```diff
@@ -0,0 +1,116 @@
...
+      options={options}
+      onChange={onSelectCollection}
+      ariaLabel={t("Default collection")}
+      note={t(
+        "We will redirect users to the selected collection when they sign in."
```
tommoor, line 109:
```suggestion
        "This is the screen that team members will first see when they sign in."
```

app/scenes/CollectionDelete.tsx
```diff
@@ -53,6 +55,19 @@ function CollectionDelete({ collection, onSubmit }: Props) {
             }}
           />
         </HelpText>
+        {team.defaultCollectionId === collection.id ? (
+          <HelpText>
+            <Trans
+              defaults="Also, <em>{{collectionName}}</em> is being used as a Default Collection. Deleting it will reset the Default Collection to Home page."
```
tommoor, line 61:
```suggestion
              defaults="Also, <em>{{collectionName}}</em> is being used as the start view – deleting it will reset the start view to the Home page."
```

server/commands/teamUpdater.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (collaborativeEditing !== undefined) {
+    team.collaborativeEditing = collaborativeEditing;
+  }
+  if (defaultUserRole !== undefined) {
+    team.defaultUserRole = defaultUserRole;
+  }
+
+  const changes = team.changed();
+
+  const savedTeam = await team.save();
```
tommoor, line 54: Could you wrap this `save` and `Event.create` in a transaction? (see other commands) – trying to make sure that we never have a model update without the associated event log in commands – gradual best practice…

server/commands/teamUpdater.ts
```diff
@@ -0,0 +1,73 @@
+import { Event, Team, User } from "@server/models";
+
+type TeamUpdaterProps = {
+  params: Partial<Team>;
+  ip?: string;
+  user: User;
+  team: Team;
+};
+
+const teamUpdater = async ({ params, user, team, ip }: TeamUpdaterProps) => {
```
tommoor, line 10: Great 👍

tommoor (commented):
Yes, this is almost perfect. A couple of small comments and the only other thing is to make sure that if you visit the outline url directly when signed in it will forward you to the default collection instead of home.

tommoor:
btw, added a new linting rule on `main` so I merged that into your branch and fixed confilicts to make life easier

app/scenes/Login/index.tsx
```diff
@@ -66,6 +67,20 @@ function Login() {
     auth.fetchConfig();
   }, [auth]);
 
+  React.useEffect(() => {
+    async function load() {
+      if (auth.authenticated) {
+        await auth.fetch();
```
tommoor, line 73: Do you remember the need for this? Definitely a red flag when already authenticated

iamsaumya: `auth.team` was  `undefined` even if we are authenticated.

iamsaumya: Screenshot 2022-02-09 at 3 06 01 AM
The Auth Store is empty in the local storage. It should have the team and user when logged in, right?

tommoor: It should yea, looks like there is a bigger bug here 😢  will take a look

tommoor: https://github.com/outline/outline/pull/3078

tommoor: Okay, fix merged - lets see if that helps tidy this up – this turned out to be a great find!

tommoor (approved):

tommoor:
Most excellent 😄

PR #3031, opened 2022-01-30
feat: File attachments

### LINKED ISSUE #859: File Attachments

tommoor:
### Todo

- [x] Handle files without a known mimetype
- [x] `FileExtension` component handle files with no and long extension
- [x] Investigate 10% increase in bundle size (crypto module?) 😬 
- [x] Add and gracefully handle file size limit
- [x] Allow images to be uploaded as file attachments (useful for e.g. SVG resources)
- [x] Hook into `onProgress` event to render upload progress
- [x] Improve styling of placeholder
- [x] Fix bug with theme not updating in PM ComponentViews (This has always been an issue, but it's more pronounced with attachments)
- [x] Fix upload placeholder appears ontop of empty block placeholder
- [x] Handle underlying file has been deleted
- [x] Bump editor version
- [x] Handle upload failure in refactor
- [x] Markdown parse/serialize
- [x] Add block menu option
- [x] Handle paste/drag and drop
- [x] Refactor uploading logic, shared with images
- [x] Base styling, shared with disabled embeds
- [x] Shared pages

closes #859 
closes #2859


tommoor:
I'm going to refactor this to not use a placeholder in favor of a full editor node while the upload is happening. The main benefit of this is:

- we can avoid style duplication
- we can avoid a bug in yjs with placeholders
- we can show the uploading progress to other users in the same doc at the same time

The downside is more complexity when/if the upload fails.

tommoor:
This is ready to go – before it can be merged we need to perform some maintenance on the storage for Outline's cloud hosted customers.

tommoor:
Planning to have maintenance complete this week

Bence98007:
I see that this is merged, is it already in outline? Because I cannot find it in any menus, and Im using v.0.62.0

tommoor:
The feature will be part of the next major release, `v0.63.0` – these are monthly.

Bence98007:
Sounds good! :) Sorry for the unnecessary comment.

Issue #3030, opened 2022-01-29
Note should wrap as usual in the filteritem.


iamsaumya:
<img width="446" alt="Screenshot 2022-01-29 at 6 21 00 PM" src="https://user-images.githubusercontent.com/40148448/151661687-218ae00e-51ca-4f7e-bc33-2809650dc4ab.png">

tommoor:
arghhhh

louisphn:
May I help with this?

tommoor:
All yours

PR #3034, opened 2022-01-31
fix: issue 3030 HelpText layout

louisphn:
## What I have done:
- [x] add work-break and whitespace in HelpText component style
- [x] roughly check if there is any regression in the application layout (I would be grateful if you can help me double check this too) 

## Result screenshots:
PC version (1440x789 screen size)
image

Tablet version (768x1024 screen size)
image

Smartphone version
image

Please take a look and let me know if there is any other issues.
Thanks

### DIFF
--- a/app/components/HelpText.ts
+++ b/app/components/HelpText.ts
@@ -4,6 +4,8 @@ const HelpText = styled.p<{ small?: boolean }>`
   margin-top: 0;
   color: ${(props) => props.theme.textSecondary};
   font-size: ${(props) => (props.small ? "14px" : "inherit")};
+  word-break: break-all;
+  white-space: normal;
 `;
 
 export default HelpText;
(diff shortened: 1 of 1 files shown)


app/components/HelpText.ts
```diff
@@ -4,6 +4,8 @@ const HelpText = styled.p<{ small?: boolean }>`
   margin-top: 0;
   color: ${(props) => props.theme.textSecondary};
   font-size: ${(props) => (props.small ? "14px" : "inherit")};
+  word-break: break-all;
```
tommoor, line 7: Is word break required? I'm not sure we want a situation where words are split across lines:

image

louisphn:
@tommoor Thanks for the feedback. Turns out it's not necessary to implement `word-break: break-all` at all. Here is the result:
Samsung Galaxy S8+:
image

Iphone SE:
image

Iphone XR:
image

tommoor:
Great, thanks! I know it's only one line but we can't accept the submission until CLI is signed.

louisphn:
I have already signed but the status of CLI is signed here did not changed, do I need to take any other step to make it change?
image

tommoor:
Okay. You probably need to add the email address that's on the commit to your GitHub account. We'll pass on the requirement here, but something to consider for future

tommoor:
closes #3030

PR #3041, opened 2022-02-01
fix: list editing regression

### LINKED ISSUE #3039: Difficult to "exit" bulleted list, when writing bulleted list with multiple levels
If you're editing a document and using a bulleted list with more than one level (i.e. indentation), it's difficult to end the list because the bullets continue (get added) upon hitting return key. Also seemingly difficult to delete those additional bullets via the keyboard.

**To Reproduce**
1. Go to edit a post
2. Begin a bulleted list
3. Indent one of the bullets, e.g.
```
- Here is a primary bullet
  - Here is a secondary bullet
 ```
4. After typing the secondary bullet, hit return a few times (to try to end the bulleted list and return to typing normal text).
5. The bulleted list continues, and does not end upon hitting backspace either.

**Expected behavior**
After hitting return a couple of time at the end of the bulleted list, you'd expect the bulleted list to end and go back to typing normal paragraph text.

**Screenshots**
Hard to show something not being deleted, but this is the result of the bug:
Screen Shot 2022-01-31 at 2 04 21 PM

- Install: getoutline.com
- Version: n/a

 - OS:  macOS 12.1
 - Browser: Firefox
 - Version:  96.0.2

### PR DESCRIPTION
**Pull request by tommoor**
When the editor was moved into the codebase this dependency got upgraded and it looks to not be compatible. For now the fix is to downgrade until we can upgrade `prosemirror-transform` to match

closes #3039

### CODE REVIEW

File: yarn.lock [hunk 1]
```diff
@@ -11823,22 +11823,22 @@ prosemirror-markdown@^1.5.2:
...
-prosemirror-model@^1.0.0, prosemirror-model@^1.13.3, prosemirror-model@^1.14.3, prosemirror-model@^1.8.1:
+prosemirror-model@1.16.1, prosemirror-model@^1.0.0, prosemirror-model@^1.1.0, prosemirror-model@^1.8.1:
   version "1.16.1"
   resolved "https://registry.yarnpkg.com/prosemirror-model/-/prosemirror-model-1.16.1.tgz#fb388270bc9609b66298d6a7e15d0cc1d6c61253"
   integrity sha512-r1/w0HDU40TtkXp0DyKBnFPYwd8FSlUSJmGCGFv4DeynfeSlyQF2FD0RQbVEMOe6P3PpUSXM6LZBV7W/YNZ4mA==
   dependencies:
     orderedmap "^1.1.0"
 
-prosemirror-schema-list@^1.1.2:
-  version "1.1.6"
```
tommoor, line 11834: After testing – this upgrade from `1.1.4` -> `1.1.6` seems to be the issue

app/editor/components/Styles.ts
```diff
@@ -620,6 +620,7 @@ const EditorStyles = styled.div<{
     position: relative;
     top: 1px;
     transition: transform 100ms ease-in-out;
+    opacity: .8;
```
tommoor, line 623: Unrelated tweak – once you're messing with lists you SEE THINGS

PR #3078, opened 2022-02-09
fix: Auth persistence to `localStorage`

tommoor:
I'm not sure exactly when this regression occurred, but at some point serializing these stores to localStorage stopped working due to a serialization loop. It may have been with the TS conversion (likely).

### DIFF
--- a/app/models/BaseModel.ts
+++ b/app/models/BaseModel.ts
@@ -22,7 +22,7 @@ export default class BaseModel {
     try {
       // ensure that the id is passed if the document has one
       if (!params) {
-        params = this.toJS();
+        params = this.toAPI();
       }
 
       const model = await this.store.save({ ...params, id: this.id });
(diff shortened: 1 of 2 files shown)


app/stores/AuthStore.ts
```diff
@@ -68,19 +68,21 @@ export default class AuthStore {
...
     }
 
     this.rehydrate(data);
+
     // persists this entire store to localstorage whenever any keys are changed
     autorun(() => {
       try {
         localStorage.setItem(AUTH_STORE, this.asJson);
-      } catch (_) {
-        // no-op Safari private mode
```
tommoor, line 81: This used to be the case, but I just tested and it's no longer an issue on desktop. Will capture some exceptions in production and see if we can ignore something more specific as this catch-all is what resulted in the regression going unnoticed here

app/models/BaseModel.ts
```diff
@@ -64,23 +64,46 @@ export default class BaseModel {
...
 
   /**
-   * Returns a plain object representation of the model
+   * Returns a plain object representation of fields on the model for
+   * persistence to the server API
    *
    * @returns {Record<string, any>}
    */
-  toJS = (): Record<string, any> => {
+  toAPI = (): Record<string, any> => {
```
tommoor, line 72: Just renamed to make this less ambiguous

PR #3102, opened 2022-02-12
fix: Add ability to collapse and expand collections that are not active

### LINKED ISSUE #2917: Add ability to collapse and expand collections that are not active

iamsaumya:
closes #2917


iamsaumya:
If you approve of the transition, I will also add it to the starred section.

thenanyu:
functional comments here
https://www.loom.com/share/07dc4fd10781418dae50bdda15c262a5

thenanyu (changes requested):
Functional testing here
https://www.loom.com/share/07dc4fd10781418dae50bdda15c262a5

1. the currently selected collection should be collapsible
2. the timing of the collapse animation seems off in some subtle ways
3. changing collections should not auto-collapse other collections

iamsaumya:
Thanks Nan, I have fixed the functional issues and made the transition smoother. 

> changing collections should not auto-collapse other collections

The behaviour I opted for while changing the collection is to collapse uiActiveCollection, irrespective of whether it was expanded or collapsed. Users are accustomed to it being collapsed on change. Let me know if you have done it differently.

app/components/Sidebar/components/Transition.tsx
```diff
@@ -0,0 +1,10 @@
+import styled from "styled-components";
```
thenanyu, line 1: I like the effect here, but it's causing a lot of small display problems due to the `overflow: hidden`. Let's ship this PR without the animated transition aspect and we can handle adding a transition in a separate change set

I can detail the issues there

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -242,7 +255,7 @@ function CollectionLink({
                 />
               }
               exact={false}
-              depth={0.5}
+              depth={0.6}
```
thenanyu, line 258: keep this at `0.5`. It's a little tight on the display but it's not too bad because it only shows on hover

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -197,19 +199,21 @@ function CollectionLink({
   const isDraggingAnyCollection =
     isDraggingAnotherCollection || isCollectionDragging;
 
+  React.useEffect(() => {
+    if (isDraggingAnyCollection) {
+      setExpanded(false);
```
thenanyu, line 204: this will close any open collection because of a drag without re-opening. collections that were previously open should become open again once the drag finishes

thenanyu (changes requested):

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -197,19 +198,28 @@ function CollectionLink({
   const isDraggingAnyCollection =
     isDraggingAnotherCollection || isCollectionDragging;
 
+  const expandedStateBeforeDragging = React.useRef<boolean | null>(null);
```
thenanyu, line 201: This approach using refs works but I don't prefer it stylistically. It overloads the meaning of `expanded` from "the user chose this to be expanded" to "the user may have chosen this, or it may be for other reasons". Better to make this explicit and not layer it on with refs.

Note that we already have all of the information with `isDraggingAnyCollection` and `expanded`

Try something like this pseudocode
```
expanded = useState(...)
isDraggingAnyCollection = ...

displayExpanded = expanded & !isDraggingAnyCollection
```

thenanyu (changes requested):

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -264,17 +275,17 @@ function CollectionLink({
             />
           </DropToImport>
         </Draggable>
-        {expanded && manualSort && (
+        {displayDocumentLinks && manualSort && (
           <DropCursor isActiveDrop={isOverReorder} innerRef={dropToReorder} />
         )}
-        {isDraggingAnyCollection && (
+        {!displayDocumentLinks && (
```
thenanyu, line 281: I haven't tried it yet but on first glance, this seems strange to me - are you sure this is right? This appears to show drop cursor when there is no active drag happening. Should this just remain `isDraggingAnyCollection`?

app/components/Sidebar/components/SidebarLink.tsx
```diff
@@ -158,6 +158,10 @@ const Link = styled(NavLink)<{ $isActiveDrop?: boolean; $isDraft?: boolean }>`
     transition: fill 50ms;
   }
 
+  &:hover svg {
+    display: inline-block;
```
thenanyu, line 162: very minor comment: does "display: inline" work? because I believe that's the default here

app/components/Sidebar/components/CollectionLink.tsx
iamsaumya: Yeah, you are right.

tommoor:
I dunno about you guys but I feel like I'm going to need a button or keyboard shortcut to collapse everything in the sidebar back down with this change. VSCode has a button for it at the top right of folder structure, I'm sure there _is_ a shortcut but I can't find it.

thenanyu:
> I dunno about you guys but I feel like I'm going to need a button or keyboard shortcut to collapse everything in the sidebar back down with this change

@tommoor the need isn't as great here as it is in VS Code because the folders will all collapse down on page refresh, but something to add later, especially if we start keeping the open state of the collections past the page refresh

thenanyu:
@iamsaumya — I make some changes to this PR

1. close only the dragged collection when dragging, instead of all collections. this is a judgement call for simplicity and less movement and jerkiness during the drag process (plus it wasn't quite working correctly before)
2. fixes a bug where all of the open states for documents got lost after closing/opening an enclosing document or collection
3. fixes a bug where the react-dnd library would frequently throw errors of the type `cannot find target with ID=...`
4. fixes a bug where the drop cursor for collections was in the wrong position when it was expanded

thenanyu (approved):

thenanyu:
@tommoor can you click-test before you merge

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -278,27 +273,30 @@ function CollectionLink({
...
-            prefetchDocument={prefetchDocument}
-            canUpdate={canUpdate}
-            isDraft={node.isDraft}
-            depth={2}
-            index={index}
-          />
-        ))}
+      </Relative>
+      <Relative>
+        <Folder $expanded={displayDocumentLinks}>
```
tommoor, line 278: This change worries me a bit because there are teams with thousands of documents, this will now render them hidden wherase before it didn't at all.

thenanyu: good point, can probably be solved by making a distinction between "collapsed because user said so" and "collapsed because in the middle of a drag"

it's still not perfect behavior that way but performance will be optimized

PR #3126, opened 2022-02-16
feat: Search shared documents

thenanyu:
This PR enables search for shared document trees
image

The underlying engine is the same as the internal search functionality, which results in similar performance, previews, and highlights.

**Major changes**

- adds `SearchPopover` and `SearchListItem`: main UI components powering the inline search feature
- changes `documents.search` API to accept a `shareId` parameter to authorize search queries for logged out users
  - adds `shareId` field to search queries table for searches without authenticated users
- fixes many small issues in `PaginatedList` component
- improves keyboard navigation and usability of existing search interfaces
- improves dark mode styling for various list items and sidebar navigation

**to do**

The PR should be shippable as is. There are a couple of tasks that are candidates for immediate follow-up, or for inclusion into this PR if @tommoor thinks it's important

1. the general styling of the search popover and result items is only ok, and could be improved aesthetically
2. the loading state behavior of the search popover is correct and straightforward, but can be changed to never display loading states. this is a little complicated and I wanted to get this version vetted first


app/components/Input.tsx
```diff
@@ -119,14 +119,17 @@ export type Props = React.HTMLAttributes<HTMLInputElement> & {
   onChange?: (
     ev: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => unknown;
+  innerRef?: React.RefObject<HTMLInputElement | HTMLTextAreaElement>;
```
thenanyu, line 122: I'm having a bit of trouble with this typing, could use a set of eyes and suggestions

server/models/Document.ts
```diff
@@ -436,32 +440,53 @@ class Document extends ParanoidModel {
...
+    ${documentClause}
     "deletedAt" IS NULL AND
     "publishedAt" IS NOT NULL
   `;
     const selectSql = `
     SELECT
       id,
       ts_rank(documents."searchVector", to_tsquery('english', :query)) as "searchRanking",
-      ts_headline('english', "text", to_tsquery('english', :query), 'MaxFragments=1, MinWords=20, MaxWords=30') as "searchContext"
+      ts_headline('english', "text", to_tsquery('english', :query), 'MaxFragments=1, MinWords=${minWords}, MaxWords=${maxWords}') as "searchContext"
```
tommoor, line 489:
```suggestion
      ts_headline('english', "text", to_tsquery('english', :query), 'MaxFragments=1, MinWords=:minWords, MaxWords=:maxWords') as "searchContext"
```

These should use query replacement

server/models/Document.ts
```diff
@@ -580,7 +605,7 @@ class Document extends ParanoidModel {
   SELECT
     id,
     ts_rank(documents."searchVector", to_tsquery('english', :query)) as "searchRanking",
-    ts_headline('english', "text", to_tsquery('english', :query), 'MaxFragments=1, MinWords=20, MaxWords=30') as "searchContext"
+    ts_headline('english', "text", to_tsquery('english', :query), 'MaxFragments=1, MinWords=${minWords}, MaxWords=${maxWords}') as "searchContext"
```
tommoor, line 608:
```suggestion
    ts_headline('english', "text", to_tsquery('english', :query), 'MaxFragments=1, MinWords=:minWords, MaxWords=:maxWords') as "searchContext"
```

server/models/Document.ts
```diff
@@ -436,32 +440,53 @@ class Document extends ParanoidModel {
...
       };
     }
 
-    // Build the SQL query to get documentIds, ranking, and search term context
+    // restrict to documents in the tree of a shared document when one is provided
+    let documentIds;
+
+    if (options.share?.includeChildDocuments) {
+      const sharedDocument = await options.share.$get("document");
+      if (sharedDocument) {
```
tommoor, line 468: This seems like a potential way to leak data, if the share document is not able to load for some reason, eg it's deleted then this `if` clause isn't hit and no restricting `documentIds` are added to the query.

Probably just throw here if `sharedDocument` not found?

server/routes/api/documents.ts
```diff
@@ -795,82 +801,132 @@ router.post("documents.search_titles", auth(), pagination(), async (ctx) => {
...
+      shareId,
+      minWords,
+      maxWords,
+    } = ctx.body;
+    assertNotEmpty(query, "query is required");
+
+    const { offset, limit } = ctx.state.pagination;
+    // this typing is a bit ugly, would be better to have the auth middleware
+    // define the data type for state.user
+    const { user }: { user: User | undefined } = ctx.state;
```
tommoor, line 827: It's available – type as `ctx: ContextWithState` to get what you need

server/routes/api/documents.ts
```diff
@@ -795,82 +801,132 @@ router.post("documents.search_titles", auth(), pagination(), async (ctx) => {
...
+    );
+
+    // When requesting subsequent pages of search results we don't want to record
+    // duplicate search query records
+    if (offset === 0) {
+      SearchQuery.create({
+        userId: user?.id,
+        teamId,
+        shareId,
+        source: ctx.state.authType || "app", // we need to write a valid enum member into here
```
tommoor, line 917: Is there not a valid `authType` because we're potentially using a shareId. If that's the case it might be better to fix this in the auth middleware with a new enum, `unauthenticated` or similar?

server/models/Document.ts
```diff
@@ -58,10 +59,13 @@ type SearchOptions = {
   limit?: number;
   offset?: number;
   collectionId?: string;
+  share?: Share;
   dateFilter?: DateFilter;
   collaboratorIds?: string[];
   includeArchived?: boolean;
   includeDrafts?: boolean;
+  minWords?: number;
```
tommoor, line 67: Would consider renaming at this level, it's too ambiguous – eg.

```suggestion
  snippetMinWords?: number;
```

app/components/SearchPopover.tsx
```diff
@@ -0,0 +1,164 @@
...
+  };
+
+  return (
+    <>
+      <PopoverDisclosure {...popover}>
+        {(props) => {
+          // props assumes the disclosure is a button, but we want a type-ahead
+          // so we take the aria props, and ref and ignore the event handlers
+          return (
+            <InputSearch
```
tommoor, line 110: image

Can we align vertically with the doc header?

app/components/SearchPopover.tsx
```diff
@@ -0,0 +1,164 @@
+import debounce from "lodash/debounce";
```
tommoor, line 1:
```suggestion
import { debounce } from "lodash";
```

is okay

app/components/SearchPopover.tsx
```diff
@@ -0,0 +1,164 @@
...
+      >
+        <PaginatedList
+          options={{ query, minWords: 10, maxWords: 11 }}
+          items={searchResults}
+          fetch={performSearch}
+          onEscape={handleEscapeList}
+          empty={
+            <NoResults>{t("No results for {{query}}", { query })}</NoResults>
+          }
+          loading={<PlaceholderList count={3} header={{ height: 20 }} />}
```
tommoor, line 137: Is this amount of flashing expected? We need to improve it somehow as it's crazy distracting

https://user-images.githubusercontent.com/380914/161211773-f3d4a2f8-c1ed-4fd0-aba4-59aa106a29ee.mp4

tommoor (changes requested):

tommoor:
Oh, i forgot to mention the biggest bug I found – for some reason publicly shared pages are no longer scrollable!

app/components/SearchPopover.tsx
thenanyu: Yeah it's a lot — one of the remaining extra todos is to remove the intermittent loading states. The implementation now is "out of the box" but clearly the update is necessary for this to ship.

app/components/Input.tsx
```diff
@@ -147,10 +148,6 @@ class Input extends React.Component<Props> {
     }
   };
 
-  focus() {
```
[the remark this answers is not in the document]
thenanyu: the way we're doing ref forwarding on this changed to the innerRef style and allowing a fully ref to be passed in— whereas before we were basically only exposing the `focus` method — see https://github.com/outline/outline/pull/3126/files#diff-29694077e2da16c9552d22067354d104efaf8d6d13effae2eebc5570bdd2ef19L70

app/components/Popover.tsx
```diff
@@ -1,40 +1,50 @@
...
+import { Popover as ReakitPopover, PopoverProps } from "reakit/Popover";
 import styled from "styled-components";
 import breakpoint from "styled-components-breakpoint";
 import useMobile from "~/hooks/useMobile";
 import { fadeAndScaleIn } from "~/styles/animations";
 
-type Props = {
+type Props = PopoverProps & {
+  children: React.ReactNode;
+  // TODO what is tabindex for
```
thenanyu, line 11: @tommoor question for you — this seems odd. It looks like tabindex is getting trapped here and not passed down to children on purpose?

server/models/Document.ts
thenanyu: Fixed in the few lines below. I tried initially to do

```
const sharedDocument = await ...
invariant(sharedDocument, "some error");

// but TS thinks sharedDocument can still be null here
const childDocumentIds = await sharedDocument.getChildDocumentIds();
```

am I misunderstanding how invariant works or is it just temperamental?

server/routes/api/documents.ts
thenanyu: it's not quite right 

it defines state as
```typescript
  state: {
    user: User;
    token: string;
    authType: "app" | "api";
  };
```
User is not optional here, it works well for auth required but not for auth optional. It also wants to become a generic I think, because body is "unknown" and creates a lot of ts errors when trying to access something on it. I think it wants to be something like 

```typescript
  state: {
    user?: User;
    token?: string;
    authType?: "app" | "api";
  };
  body: any;
;
```
but that will blow the scope of this PR up a bit farther than we want.

server/routes/api/documents.ts
thenanyu: I think defaulting to "app" is correct, because if we're actually interested in the "source" of the search query. If it's not via the API, then we should assume it's through the app, as no api calls are possible without an api token.

That it's unauthenticated matters to conceptually to authtype but that's not what's being stored semantically. It just happens to line up

app/components/Popover.tsx
tommoor: It's being passed in the `...rest` no?

server/models/Document.ts
tommoor: That should work, are you sure you imported the right invariant from the `invariant` package? I've noticed that sometimes VSCode will pull in a different one from elsewhere

server/routes/api/documents.ts
tommoor: 😅  so. much. tech. debt.

app/components/Popover.tsx
thenanyu: yeah I got confused >_<

My original question was because I took away the typing and tsc doesn't complain. It seems unnecessary to me, but leaving it in as it's harmless for now

server/models/Document.ts
thenanyu: yeah that was the issue, changing it to that style now

thenanyu:
> Oh, i forgot to mention the biggest bug I found – for some reason publicly shared pages are no longer scrollable!

I couldn't reproduce this at all, dunno if that's because it got fixed under the covers or if there was some edge case I couldn't find. Do you still see this behavior?

tommoor:
Weird, no longer repro 🤷

thenanyu:
Updated the style of the search bar to be round in 2 locations

Shared Search
image

Global Search
image

But leaves in several other places

Member filter (could also change it here if desired)
image

Move Document (this is completely custom styling, including input height)
image

tommoor:
Agreed with those decisions, they are both more filters than search – move document should probably say "Filter" instead of "Search" actually in hindsight

tommoor (approved):

tommoor:
One UX thing I wonder – shouldn't the focus jump to the page when you navigate to a search result? It feels weird that the cursor stays in the search box.

thenanyu:
The normal thing to do seems to be: treat it as a full page load: clear the search box and don't focus anything.

Either this mean clear the input or the input is modal in the first place. This feels more natural I think

PR #3139, opened 2022-02-17
(feat.) Wekan Integreation (+ Trello Tests)

kuhnchris:
Adding Embed for Wekan to enable embedding Wekan boards and/or cards:
image


kuhnchris:
:point_down: i noticed the tests failing (locally aswell) - but not due to my changes, but due to some other unrelated TS files - probabbly something on master that is work-in-progress.

Limezy:
Great project !

tommoor:
Hi Chris, thanks for the contribution. Unfortunately we can't accept this at this time for a couple of main reasons:

- We're not accepting embeds that just include an entire webpage, it must have an embed specific view – it looks from your screenshot like you could navigate the entire site within the iframe?
- The regex match is far too broad, it will match many other links than just wekan. It seems like this is a self-hosted project, which is great but we'd need to have some config within Outline to specify where your Wekan instance is hosted.

PR #3144, opened 2022-02-18
fix: create a generic notice component

### LINKED ISSUE #3111: Add icons to archived and trashed banners
image

iamsaumya:
closes #3111


iamsaumya:
<table>
<tr>
<td> 
Screenshot 2022-02-19 at 1 44 09 AM
</td>
<td>
Screenshot 2022-02-19 at 1 44 15 AM
</td>
</tr>
</table>

tommoor:
I cherry-picked the share modal fix into main. For the icon I'm trying to always think in terms of systems with new work, in this case instead of making it a one-off lets add an icon property to `Notice` that takes a react component and positions it on the left so that any notice can have an icon! It looks like there needs to be another 4px or so of padding between icon and text too.

tommoor:
Did a 'lil refactor here – anything to get more logic out of `Document` which is far too big :)

iamsaumya:
Agree. We should split the Document and a few other files into smaller components. I think we have a few big files.

PR #3166, opened 2022-02-24
fix: sync the correct collection with edit action

iamsaumya:
The edit collection action was always happening with the previous collection.

### DIFF
--- a/app/hooks/useCommandBarActions.ts
+++ b/app/hooks/useCommandBarActions.ts
@@ -11,7 +11,10 @@ import useActionContext from "./useActionContext";
  *
  * @param actions actions to make available
  */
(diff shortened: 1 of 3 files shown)


app/hooks/useCommandBarActions.ts
```diff
@@ -11,7 +11,10 @@ import useActionContext from "./useActionContext";
  *
  * @param actions actions to make available
  */
-export default function useCommandBarActions(actions: Action[]) {
+export default function useCommandBarActions(
+  actions: Action[],
+  additionalDeps: string[] = []
```
tommoor, line 16:
```suggestion
  additionalDeps: React.DependencyList = []
```

No functional difference, but it's nice

tommoor (approved):

PR #3168, opened 2022-02-25
Inset icon in collection headers, minor ContentEditable refactor

tommoor:
More keylining, aligns the text of the collection name with the content below rather than the icon.

### Before

image

### After
image

### DIFF
--- a/app/components/ContentEditable.tsx
+++ b/app/components/ContentEditable.tsx
@@ -81,6 +81,17 @@ const ContentEditable = React.forwardRef(
       }
     }, [value, ref]);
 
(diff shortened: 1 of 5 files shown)


app/components/ContentEditable.tsx
```diff
@@ -81,6 +81,17 @@ const ContentEditable = React.forwardRef(
       }
     }, [value, ref]);
 
+    // Ensure only plain text can be pasted into title when pasting from another
```
tommoor, line 84: Moving standard handling and styling down into this component from `EditableTitle` for re-use.

PR #3171, opened 2022-02-26
perf: Navigation of shared trees feels slow

tommoor:
Removes the requirement for the `documents.info` request to complete before updating the state of the shared sidebar, this makes navigation feel much more responsive on shared links with nested docs.

### DIFF
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -87,7 +87,6 @@ function SharedDocumentScene(props: Props) {
           shareId,
         });
         setResponse(response);
-        ui.setActiveDocument(response.document);
       } catch (err) {
         setError(err);
       }
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -16,6 +16,11 @@ import Loading from "./components/Loading";
 
 const EMPTY_OBJECT = {};
 
+type Response = {
+  document: DocumentModel;
+  sharedTree?: NavigationNode | undefined;
+};
+
 type Props = RouteComponentProps<{
   shareId: string;
   documentSlug: string;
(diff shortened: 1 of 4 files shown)


app/scenes/Document/Shared.tsx
```diff
@@ -16,30 +16,70 @@ import Loading from "./components/Loading";
...
   const theme = useTheme();
-  const [response, setResponse] = React.useState<{
-    document: DocumentModel;
-    sharedTree?: NavigationNode | undefined;
-  }>();
+  const [response, setResponse] = React.useState<Response>();
   const [error, setError] = React.useState<Error | null | undefined>();
   const { documents } = useStores();
   const { shareId, documentSlug } = props.match.params;
+  const documentId = useDocumentId(documentSlug, response);
```
tommoor, line 70: TLDR: If the `sharedTree` is loaded in memory then we know all of the documentIds in advance, so we can make sure the correct sidebar item is selected while the rest of the document is loaded.

app/scenes/Document/Shared.tsx
```diff
@@ -16,38 +16,77 @@ import Loading from "./components/Loading";
...
+  }, [ui, documentId]);
+
   React.useEffect(() => {
     async function fetchData() {
       try {
         const response = await documents.fetchWithSharedTree(documentSlug, {
           shareId,
         });
         setResponse(response);
-        ui.setActiveDocument(response.document);
```
thenanyu, line 50: I made this change — I think it's correct to have only one call to setActiveDocument. Otherwise there's some strange looking behavior when there's a bit of latency and two successive navigation clicks.

tommoor:
> I think this kind of very optimistic client state change is a little dangerous 

It's not really optimistic in the traditional sense, it's just updating the UI to reflect your click rather than the state of some http request. I think ideally we would move the content to a loader after maybe ~500ms if it hasn't loaded but I'm going to land this as a good improvement and we can come back to that

PR #3187, opened 2022-02-28
fix: editing collections should not forward to collection on save

### LINKED ISSUE #3183: Editing collections should not forward to collection on save
Title says it all.

iamsaumya:
closes #3183

### DIFF
--- a/app/scenes/Collection.tsx
+++ b/app/scenes/Collection.tsx
@@ -48,14 +48,14 @@ function CollectionScene() {
   const can = usePolicy(collection?.id || "");
 
   React.useEffect(() => {
-    if (collection) {
+    if (collection?.name) {
       const canonicalUrl = updateCollectionUrl(match.url, collection);
 
       if (match.url !== canonicalUrl) {
         history.replace(canonicalUrl);
       }
     }
-  }, [collection, history, id, match.url]);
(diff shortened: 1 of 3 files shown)


app/scenes/CollectionEdit.tsx
```diff
@@ -40,14 +40,17 @@ const CollectionEdit = ({ collectionId, onSubmit }: Props) => {
...
+        const redirect = location.pathname === collection.url;
         await collection.save({
           name,
           icon,
           color,
           sort,
         });
-        history.push(collection.url);
         onSubmit();
+        if (redirect) {
```
tommoor, line 51: I'm pretty sure this can be entirely removed as there is already a `useEffect` in the CollectionScene to redirect to the canonical url if it changes

app/scenes/Collection.tsx
```diff
@@ -48,14 +48,14 @@ function CollectionScene() {
   const can = usePolicy(collection?.id || "");
 
   React.useEffect(() => {
-    if (collection) {
+    if (collection?.name) {
```
tommoor, line 51: Why the change?

iamsaumya: the useEffect wasn't running after the update.

tommoor (approved):

PR #3203, opened 2022-03-04
fix: don't hide sidebar when menu is open

iamsaumya:
Bug

https://user-images.githubusercontent.com/40148448/156844519-1b23ecde-bc1a-42e6-a047-712c9dba8754.mov

### DIFF
--- a/app/components/ContextMenu/index.tsx
+++ b/app/components/ContextMenu/index.tsx
@@ -56,15 +56,12 @@ export default function ContextMenu({
   const setIsMenuOpen = useSidebar();
 
   React.useEffect(() => {
-    if (ui.sidebarCollapsed && setIsMenuOpen) {
-      rest.visible ? setIsMenuOpen(true) : setIsMenuOpen(false);
-    }
-  }, [rest.visible, setIsMenuOpen, ui]);
-
-  React.useEffect(() => {
     if (rest.visible && !previousVisible) {
       if (onOpen) {
         onOpen();
+      }
+      if (ui.sidebarCollapsed && setIsMenuOpen) {
+        setIsMenuOpen((prev) => prev + 1);
       }
     }
 
@@ -72,8 +69,18 @@ export default function ContextMenu({
       if (onClose) {
         onClose();
       }
+      if (ui.sidebarCollapsed && setIsMenuOpen) {
+        setIsMenuOpen((prev) => prev - 1);
+      }
     }
-  }, [onOpen, onClose, previousVisible, rest.visible]);
+  }, [
+    onOpen,
+    onClose,
+    previousVisible,
+    rest.visible,
+    ui.sidebarCollapsed,
+    setIsMenuOpen,
+  ]);
 
   // Perf win – don't render anything until the menu has been opened
   if (!rest.visible && !previousVisible) {
--- a/app/components/ContextMenu/index.tsx
+++ b/app/components/ContextMenu/index.tsx
@@ -5,6 +5,8 @@ import styled from "styled-components";
 import breakpoint from "styled-components-breakpoint";
 import useMenuHeight from "~/hooks/useMenuHeight";
 import usePrevious from "~/hooks/usePrevious";
+import useSidebar from "~/hooks/useSidebar";
+import useStores from "~/hooks/useStores";
 import {
   fadeIn,
   fadeAndSlideUp,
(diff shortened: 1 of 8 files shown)


app/components/Sidebar/Sidebar.tsx
```diff
@@ -145,41 +147,43 @@ const Sidebar = React.forwardRef<HTMLDivElement, Props>(
...
+              style={toggleStyle}
               onClick={ui.toggleCollapsedSidebar}
-              direction={"right"}
-              aria-label={t("Expand")}
+              direction={ui.sidebarCollapsed ? "right" : "left"}
+              aria-label={ui.sidebarCollapsed ? t("Expand") : t("Collapse")}
             />
           )}
-        </Container>
-        {!ui.isEditing && (
```
tommoor, line 175: I feel like this was outside of the container for a very good reason because the refactor is too obvious, but for the life of me I'm not sure what.

app/components/ContextMenu/index.tsx
```diff
@@ -50,20 +52,35 @@ export default function ContextMenu({
...
   const backgroundRef = React.useRef<HTMLDivElement>(null);
+  const { ui } = useStores();
+  const setIsMenuOpen = useSidebar();
 
   React.useEffect(() => {
     if (rest.visible && !previousVisible) {
       if (onOpen) {
         onOpen();
       }
+      if (ui.sidebarCollapsed && setIsMenuOpen) {
```
tommoor, line 63: I'm not sure why `ui.sidebarCollapsed` is considered here, the menu is open either way.

app/components/ContextMenu/index.tsx
```diff
@@ -50,20 +52,35 @@ export default function ContextMenu({
   const previousVisible = usePrevious(rest.visible);
   const maxHeight = useMenuHeight(rest.visible, rest.unstable_disclosureRef);
   const backgroundRef = React.useRef<HTMLDivElement>(null);
+  const { ui } = useStores();
+  const setIsMenuOpen = useSidebar();
```
tommoor, line 56:
```suggestion
  const setOpenMenuCount = useSidebar();
```

Need to name this better, perhaps.

app/hooks/useSidebar.ts
```diff
@@ -0,0 +1,9 @@
+import React from "react";
```
tommoor, line 1: Need to rename this hook, it doesn't need to have anything to do with the sidebar 😆 – just tracking how many context menus are open.

app/hooks/useSidebar.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export const SidebarContext = React.createContext<React.Dispatch<
+  React.SetStateAction<number>
+> | null>(null);
+
+const useSidebar = () => React.useContext(SidebarContext);
```
tommoor, line 7: Using context is clearly the way to go but this particular implementation feels a little backward as the state of menus is stored inside of the sidebar component. Can we instead store the state in a hook here I wonder?

```suggestion

export function MenuContext() {
  const [openCount, setOpenCount] = React.useState(0);
  return <MenuContext.Provider value={{ openCount, setOpenCount }}>{children}</MenuContext.Provider>
}

const useMenuContext = () => React.useContext(SidebarContext);
```

You can now use the hook in both the sidebar and menu components:

```
const { openCount, setOpenCount } = useMenuContext();
```

I think this will work?

tommoor (changes requested):

app/components/ContextMenu/index.tsx
```diff
@@ -50,20 +53,38 @@ export default function ContextMenu({
...
+  const { setOpenCount } = useMenuContext();
 
   React.useEffect(() => {
     if (rest.visible && !previousVisible) {
       if (onOpen) {
         onOpen();
       }
+      if (rest["aria-label"] !== t("Submenu")) {
+        setOpenCount(true);
+      }
```
iamsaumya, line 67: The useEffect doesn't run if I click on MenuItems of Submenu. So it leads to incorrect menuCount.

app/components/ContextMenu/index.tsx
```diff
@@ -50,20 +53,38 @@ export default function ContextMenu({
...
+  const { ui } = useStores();
+  const { t } = useTranslation();
+  const { setIsMenuOpen } = useMenuContext();
 
   React.useEffect(() => {
     if (rest.visible && !previousVisible) {
       if (onOpen) {
         onOpen();
       }
+      if (rest["aria-label"] !== t("Submenu")) {
```
tommoor, line 65: Oof, this is unfortunate but I understand why it's needed

tommoor (approved):
I hope you agree that this design is much nicer than the original pass

iamsaumya:
Indeed. Much better 💯

PR #3219, opened 2022-03-10
fix: Detect Pomerium proxy

tommoor:
Forward cookies with API requests if used, adds support for using Outline behind https://www.pomerium.com/


app/utils/ApiClient.ts
```diff
@@ -203,15 +212,12 @@ class ApiClient {
...
-  constructQueryString = (data: Record<string, any>) => {
+  private constructQueryString = (data: Record<string, any>) => {
     return map(
       data,
       (v, k) => `${encodeURIComponent(k)}=${encodeURIComponent(v)}`
     ).join("&");
   };
 }
 
-export default ApiClient; // In case you don't want to always initiate, just import with `import { client } ...`
```
tommoor, line 215: Not used in 4 years, I think it's safe to remove

app/utils/ApiClient.ts
```diff
@@ -19,21 +19,28 @@ import {
...
+   * If the user is behind a proxy that requires authentication, currently we
+   * automatically recognize Cloudflare Access and Pomerium.
+   */
+  isBehindAuthenticatedProxy = false;
 
   constructor(options: Options = {}) {
     this.baseUrl = options.baseUrl || "/api";
-    this.userAgent = "OutlineFrontend";
+    this.isBehindAuthenticatedProxy = !!(
+      getCookie("CF_Authorization") || getCookie("_pomerium")
```
thenanyu, line 42: !

PR #3220, opened 2022-03-10
fix: SVGs without a natural px width are invisible

### LINKED ISSUE #3211: SVG images are not scaled correctly in Firefox
- Upload a large SVG image
- It uploads correctly but is incredibly small (invisible) in the doc
- Only in Firefox. Removing the max-height css clause gets it to appear.

tommoor:
closes #3211

### DIFF
--- a/shared/editor/nodes/Image.tsx
+++ b/shared/editor/nodes/Image.tsx
@@ -295,35 +295,12 @@ export default class Image extends Node {
   };
 
   component = (props: ComponentProps) => {
-    const { theme, isSelected } = props;
-    const { alt, src, layoutClass } = props.node.attrs;
-    const className = layoutClass ? `image image-${layoutClass}` : "image";
-
     return (
-      <div contentEditable={false} className={className}>
-        <ImageWrapper
-          className={isSelected ? "ProseMirror-selectednode" : ""}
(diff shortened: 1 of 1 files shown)


shared/editor/nodes/Image.tsx
```diff
@@ -520,6 +497,53 @@ export default class Image extends Node {
...
+  const [width, setWidth] = React.useState(0);
+
+  return (
+    <div contentEditable={false} className={className}>
+      <ImageWrapper
+        className={isSelected ? "ProseMirror-selectednode" : ""}
+        onClick={props.onClick}
+        style={{ width }}
+      >
+        <Button onClick={props.onDownload}>
```
tommoor, line 519: Move `onClick` to the button, no idea why it was on the icon

shared/editor/nodes/Image.tsx
```diff
@@ -576,10 +600,12 @@ const Caption = styled.p`
   }
 `;
 
-const ImageWrapper = styled.span`
+const ImageWrapper = styled.div`
   line-height: 0;
-  display: inline-block;
```
tommoor, line 581: Not using `inline-block` also pretty important for Firefox compat

PR #3226, opened 2022-03-10
fix: navigate to all the pages of settings through command bar

### LINKED ISSUE #3053: Navigation to different pages of settings through command bar
We should add actions in the command bar to navigate all the settings pages. All the pages have associated icons, so there is no stopping.

iamsaumya:
closes #3053 

- [x] check if the initial bundle size is increased.


app/actions/definitions/navigation.tsx
```diff
@@ -57,6 +57,7 @@ export const navigateToRecentSearchQuery = (searchQuery: SearchQuery) =>
     perform: () => history.push(searchPath(searchQuery.query)),
   });
 
+// could be a document action
```
iamsaumya, line 60: We need to assign a section to search.

Screenshot 2022-03-11 at 2 07 23 AM

iamsaumya:
RelativeCI doesn't reflect any major increase. good to go

app/hooks/useSettingsAction.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
```
tommoor, line 1:
```suggestion
import * as React from "react";
```

Prefer for consistency

app/hooks/useSettingsAction.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { createAction } from "~/actions";
+import history from "~/utils/history";
+import useAuthorizedSettingsConfig from "./useAuthorizedSettingsConfig";
+
+const useSettingsActions = () => {
+  const config = useAuthorizedSettingsConfig();
+  const actions = React.useMemo(() => {
+    return config.map((item) => {
+      const Icon = item.icon;
+      return createAction({
```
tommoor, line 11: I'd say we want these to be children of the "Settings" action, because it's a bit confusing in your screenshot. With the parent it would read "Settings -> Features" for example, and child actions are still searchable

app/routes/settings.tsx
```diff
@@ -1,42 +1,25 @@
...
-import Zapier from "~/scenes/Settings/Zapier";
 import Route from "~/components/ProfiledRoute";
-import env from "~/env";
-
-const isHosted = env.DEPLOYMENT === "hosted";
+import useAuthorizedSettingsConfig from "~/hooks/useAuthorizedSettingsConfig";
 
 export default function SettingsRoutes() {
+  const configs = useAuthorizedSettingsConfig();
+  const Routes = React.useMemo(
```
tommoor, line 8: No need for the memo here, just inline the map

app/hooks/useAuthorizedSettingsConfig.ts
```diff
@@ -0,0 +1,196 @@
...
+  | "Shares"
+  | "Import"
+  | "Export"
+  | "Slack"
+  | "Zapier";
+
+export type ConfigItem = {
+  name: string;
+  path: string;
+  icon: typeof NewDocumentIcon;
```
tommoor, line 54:
```suggestion
  icon: typeof NewDocumentIcon;
```

This type seems funky

tommoor (commented):
_almost_ – thank you

app/hooks/useSettingsAction.tsx
iamsaumya: What would be the group of the parent "Settings" action?

tommoor: I think "Navigation" would be fine

tommoor (approved):
Noice 👏

PR #3229, opened 2022-03-11
Feat: new arrow key navigation

thenanyu:
- removes boundless arrow key navigation dependency
- creates internal arrow key navigation component
- this is used in a heck of a lot of places through document list and paginated list so do some click testing!


app/components/PaginatedList.tsx
```diff
@@ -129,44 +133,47 @@ class PaginatedList extends React.Component<Props> {
...
-                // should render it, otherwise the item can go under the previous
-                // heading
-                if (!previousHeading || currentHeading !== previousHeading) {
-                  previousHeading = currentHeading;
-                  return (
-                    <React.Fragment key={item.id}>
-                      {renderHeading(currentHeading)}
-                      {children}
-                    </React.Fragment>
+            <ArrowKeyNavigation>
```
thenanyu, line 136: this change is just re-nesting in a render prop

webpack.config.js
```diff
@@ -49,7 +49,6 @@ module.exports = {
       "~": path.resolve(__dirname, 'app'),
       "@shared": path.resolve(__dirname, 'shared'),
       "@server": path.resolve(__dirname, 'server'),
-      'boundless-arrow-key-navigation': 'boundless-arrow-key-navigation/build',
```
thenanyu, line 52: I don't actually know what this does, I just removed all references to the library

tommoor: I believe it was a fix for ES6 imports

app/components/ArrowKeyNavigation.tsx
```diff
@@ -0,0 +1,51 @@
...
+import * as React from "react";
+import {
+  useCompositeState,
+  Composite,
+  CompositeStateReturn,
+} from "reakit/Composite";
+
+type Props = {
+  children: (composite: CompositeStateReturn) => React.ReactNode;
+  onEscape?: (ev: React.KeyboardEvent<HTMLDivElement>) => void;
```
tommoor, line 11:
```suggestion
  onExit?: (ev: React.KeyboardEvent<HTMLDivElement>) => void;
```

Maybe? "escape" is just mapped so thoroughly to the key in my head

app/components/ArrowKeyNavigation.tsx
```diff
@@ -0,0 +1,51 @@
...
+    },
+    [composite.currentId, composite.items, onEscape]
+  );
+
+  return (
+    <Composite
+      {...composite}
+      onKeyDown={handleKeyDown}
+      role="menu"
+      aria-label="My toolbar"
```
tommoor, line 43: Allow this to be passed in.

app/components/List/Item.tsx
```diff
@@ -63,13 +63,13 @@ const ListItem = (
...
   return (
-    <Wrapper $border={border} $small={small} {...rest}>
+    <Wrapper ref={ref} $border={border} $small={small} {...rest}>
       {content(false)}
     </Wrapper>
   );
 };
 
-const Wrapper = styled.div<{ $small?: boolean; $border?: boolean }>`
+const Wrapper = styled.a<{ $small?: boolean; $border?: boolean; to?: string }>`
```
tommoor, line 72: Should we just make this a `CompositeItem`? Would avoid needing to do it separately in every item that inherits from this.

app/components/PaginatedEventList.tsx
```diff
@@ -29,15 +29,18 @@ const PaginatedEventList = React.memo<Props>(function PaginatedEventList({
...
+          <EventListItem
+            key={item.id}
+            event={item}
+            document={document}
+            latest={index === 0}
+            composite={composite}
+            {...rest}
+          />
+        );
+      }}
```
tommoor, line 43:
```suggestion
      renderItem={(item, index, props) => {
        return (
          <EventListItem
            key={item.id}
            event={item}
            document={document}
            latest={index === 0}
            {...props}
            {...rest}
          />
        );
      }}
```

Prefer?

app/components/ArrowKeyNavigation.tsx
tommoor: Sure, not a blocker just an opinion

app/components/List/Item.tsx
thenanyu: I played with a lot of permutations of this, including what you suggested, but I went with this way in order to support inactive items (which are removed from the composite keyboard navigation but not disabled)

If we make these composite items we can give them the disabled prop but it removes all of the interactivity including mouseovers of the Time component and such. If that's a tradeoff you're happy with, we can go that way

app/components/ArrowKeyNavigation.tsx
thenanyu: yeah I prefer escape for now

app/components/PaginatedEventList.tsx
thenanyu: I can go either way — 
- `props` is nice but very generic and doesn't give you a hint as to what it's for. if you don't need keyboard nav, you can actually not include this
- `composite` breaks a bit of abstraction but I think in an instructive way
- `compositeItemProps` is the actually most self-documenting but it's... clunky

happy to defer to you on this one - what would you prefer?

tommoor: `compositeItemProps` and spread would be my preferred, but again It's not a hill I'm going to die on.

thenanyu: ok, changed to `compositeProps` and spread

PR #3254, opened 2022-03-16
chore: `documentStructure` database locking

### LINKED ISSUE #3239: `documents.move` performance
It is possible to create a document move request that sets `collection.documentStructure` many times within the one request, I'm not sure how this happens yet but it's disastrous for performance, [this is](https://app.datadoghq.com/apm/trace/6274297265365529821?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=6274297265365529821&spanViewType=metadata&timeHint=1647263292495.001) a particularly egregious example.

image

tommoor:
In the future we'll move towards wrapping all mutating API requests in transactions, for now this touches the minimum to fix the issue at hand.

closes #3239


tommoor:
@thenanyu would appreciate you checking out tomorrow and seeing if you can break it at all with rapid document moves

server/commands/documentMover.ts
```diff
@@ -38,154 +38,161 @@ async function documentMover({
...
-          );
-        };
-
-        await loopChildren(document.id);
-
-        const pin = await Pin.findOne({
+      // Efficiently find the ID's of all the documents that are children of
+      // the moved document and update in one query
+      const childDocumentIds = await document.getChildDocumentIds();
+      await Document.update(
```
tommoor, line 121: Moved this to a single `UPDATE` instead of a loop of select, update ad infinitum

server/commands/pinDestroyer.ts
```diff
@@ -24,31 +23,22 @@ export default async function pinDestroyer({
   user,
   pin,
   ip,
-  transaction: t,
+  transaction,
 }: Props): Promise<Pin> {
-  const transaction = t || (await sequelize.transaction());
```
tommoor, line 29: Had to remove the in-command transaction here so that it can be passed in without committing. As a rule going forward all commands will take a `transaction` argument.

server/models/Document.ts
```diff
@@ -232,35 +232,50 @@ class Document extends ParanoidModel {
...
-  static async updateInCollectionStructure(model: Document) {
-    if (!model.publishedAt || model.template) {
+  static async updateTitleInCollectionStructure(model: Document) {
+    // templates, drafts, and archived documents don't appear in the structure
+    // and so never need to be updated when the title changes
+    if (
+      model.archivedAt ||
+      model.template ||
+      !model.publishedAt ||
+      !model.changed("title")
```
tommoor, line 242: This is important, we're now only reading and updating the collection structure if the document title was edited.

server/routes/api/documents.ts
```diff
@@ -1041,28 +1042,11 @@ router.post("documents.update", auth(), async (ctx) => {
   document.lastModifiedById = user.id;
   const { collection } = document;
   const changed = document.changed();
-  let transaction;
 
-  try {
```
tommoor, line 1046: Removed pointless transaction

thenanyu:
The case I'm able to produce involves moving a tree of docs to `collection A`, and then moving a subnode to `collection B` while the first api request is still in motion. When this happens, a clone in the tree is made and it's actually very hard to get out of this state. Renames and moving things around subsequently can result in completely missing nodes

- in this screenshot, both `doc 1` and `doc 2` have been cloned

image

thenanyu (approved):
lgtm, race condition is eliminated

Issue #2976, opened 2022-01-22
Fade out navigation when editing and mouse hasn't moved


tommoor:
This would be a nice small detail to increase "focus mode" – between this and collapsing the sidebar it would become a very nice writing experience.

I think wait 5 seconds is probably about right, when the mouse moves the navigation should immediately fade back in.

PR #3256, opened 2022-03-16
fix: Fade out navigation when editing and mouse hasn't moved

iamsaumya:
towards #2976


app/hooks/useMouseMove.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from "react";
+
+const useMouseMove = () => {
```
tommoor, line 3:
```suggestion
const useMouseMove = (timeout: number = 5000) => {
```

Lets make the `5000` an argument

tommoor, line 3: Quick documentation comment? Trying to do this with all new hooks/methods

app/scenes/Document/components/Header.tsx
```diff
@@ -337,4 +343,10 @@ const Status = styled(Action)`
   color: ${(props) => props.theme.slate};
 `;
 
+const Wrapper = styled(Flex)<{ $hideHeader: boolean }>`
+  opacity: ${(props) => (props.$hideHeader ? 0 : 1)};
+  visibility: ${(props) => (props.$hideHeader ? "hidden" : "visible")};
```
tommoor, line 348: Intrigued as to why `visibility` was also needed? I'm sure there's a good reason

app/scenes/Document/components/Header.tsx
```diff
@@ -211,7 +217,7 @@ function DocumentHeader({
           </>
         }
         actions={
-          <>
+          <Wrapper $hideHeader={hideHeader}>
             <ObservingBanner />
```
tommoor, line 221: Lets keep `<ObservingBanner />` visible

tommoor (commented):
Nice, I think we'll also end up making this potentially a user preference but we'll see what feedback comes in

app/scenes/Document/components/Header.tsx
iamsaumya: To not trigger the click/hover events.

app/scenes/Document/components/Header.tsx
iamsaumya: Cool. Btw the navigation is always visible in collab editing mode. Do we want to fade out in it too?

tommoor: Yep, same behavior in collab is ideal

app/scenes/Document/components/Header.tsx
```diff
@@ -163,6 +170,35 @@ function DocumentHeader({
...
+        placement="bottom"
+      >
+        <Button icon={<PlusIcon />} {...props} neutral>
+          {t("New doc")}
+        </Button>
+      </Tooltip>
+    ),
+    [t]
+  );
+
```
iamsaumya, line 201: Inline usage to rerendering in collab mode.

app/editor/components/EmojiMenu.tsx
```diff
@@ -21,7 +21,7 @@ const searcher = new FuzzySearch<{
   sort: true,
 });
 
-class EmojiMenu extends React.Component<
+class EmojiMenu extends React.PureComponent<
```
iamsaumya, line 24: It takes 10ms to render even when it doesn't render any of the children's 😮
Screenshot 2022-05-01 at 4 27 32 AM

tommoor:
After playing with it, it feels quite good – a couple of tweaks:

- The fade out could be a touch slower, maybe another 500ms.
- I think we should also fade out the "References" area at the bottom of the doc that shows nested and backlinked docs
- I'm wondering if this is contentious enough to make it a user preference, thoughts?

iamsaumya:
> I'm wondering if this is contentious enough to make it a user preference, thoughts?

Sounds good. I think we will have to create a preferences page. Do you have a page in mind where we can put this?

tommoor:
I can add user preferences in the next week, just need to figure out how I want to store them 😋

iamsaumya:
Could we review and get this merge? Next week, we can always add this to the preference page - I assume we would keep it enabled by default.

tommoor (approved):

PR #3257, opened 2022-03-16
fix: prevent history from crashing due to missing EditorView

thenanyu:


app/scenes/Document/components/DataLoader.tsx
```diff
@@ -272,7 +272,11 @@ class DataLoader extends React.Component<Props> {
           revision,
           abilities,
           isEditing: this.isEditing,
-          readOnly: !this.isEditing || !abilities.update || document.isArchived,
+          readOnly:
+            !this.isEditing ||
+            !abilities.update ||
+            document.isArchived ||
+            !!revisionId,
```
thenanyu, line 279: put the editor into readonly if we're examining a revision

tommoor (approved):

Issue #2769, opened 2021-11-15
Outline is incompatible with "Block All Public Access" on the S3 bucket.


netik:
There's been a lot of back and forth in the issue reporting about how outline should handle images.  Some people are okay with private attachments and public avatars, but that's a bad security model in general. 

It's very desirable to have block all public access on your amazon account for all of S3, as well as on individual buckets for security purposes. #1337 fixes most of this problem and was merged back in Feb 2020 but no one took the team logo or avatars into account.

The current fix of adding TEAM_LOGO only fixes the image on the logged out page (which is a bit of a hack) but does not fix the logo on the team's name once logged in (why even have TEAM_LOGO if this is the case?)

Attachments work perfectly fine with S3 signing and Block All Public Access enabled, because the signing is correct and objects are private. When BPA is on, you can't have **any** public objects at all in your account, and this is good for security-focused installations

Outline should sign every image and treat avatars as private if AWS_S3_ACL is set to private (or if some other flag is set to force private avatars). Currently they are hardcoded as "public". When this happens and Block All Public Access is enabled, _uploads and downloads will silently fail_.

Further compounding this problem is the fact that some sites cannot disable BPA. It can be set on an account-wide basis by administrators and individual buckets cannot bet set public after that policy decision has been made.

**To Reproduce**
1. Enable Block Public access on a bucket.
1. Set AWS_S3_ACL to the default of private.
1. Attempt to upload attachments in a new post, it will work.
1. Try the same thing for TEAM or USER Avatars. It will fail.

**Expected behavior**
Consistent behavior across all attachments and images. Proper signing used on uploads and downloads of Avatars,

Get rid of TEAM_LOGO entirely and sign the request for that image even for logged out users.

**Outline (please complete the following information):**
- Install: self hosted
- Version: latest (b616292fce719bf9acdc2cba6cad4d3aa68db34b)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Chrome 95.0.x
 - Version: any

jongomes:
same problem

jongomes:
![image](https://user-images.githubusercontent.com/57728651/144492984-be8203dd-c0b6-4c1f-9c03-d337a734d121.png)

romeupalos:
Public access to AWS buckets are not acceptable in current industry standard for security.
Please, use signed URLs.

tommoor:
If someone wants to work on it we would accept the PR. The downside is slower loading avatar images

[...]

PR #3277, opened 2022-03-22
feat: refactor s3 configurations

frankli0324:
breaking quite a lot of stuff here

* dev compose file needs a rewrite
* previous deployments would have to be reconfigured

however it (aims to) bring:

* compatibility with more S3 providers
* sanity


phongkt-dev:
hmm, I'm using both Minio and AWS S3, and have no problem with current version of Outline, why we need to change env variable name?

frankli0324:
I find the current implementation quite confusing and seems to be suffering from burdens of the past.
look at this, just... how?

https://github.com/outline/outline/blob/7f15eb287debce76529a50625901ee1933498968/server/utils/s3.ts#L132-L145

also, I'm not expecting this PR to be merged soon. I changed the variable names because I expect to find a way to add backward compatibility, by "inferring" the added variables from the old variable values.

https://github.com/outline/outline/discussions/3159

tommoor:
This is a great initiative

server/presenters/document.ts
```diff
@@ -20,7 +20,10 @@ async function replaceImageAttachments(text: string) {
         const signedUrl = await getSignedUrl(attachment.key, 3600);
         text = text.replace(
           new RegExp(escapeRegExp(attachment.redirectUrl), "g"),
-          signedUrl
+          // keep "attachments.redirect" string for files in shared urls to be recognized correctly
+          // this behavior should be changed upon switching to `json` based document transfer
+          //     in `documents.info` endpoint
+          signedUrl + "# attachments.redirect"
```
tommoor, line 26:
```suggestion
          signedUrl + "#attachments.redirect"
```

Is there a reason for the space here?

tommoor:
> is there any other changes that must be applied?

It's honestly difficult to tell as there are so many different deployments out there for this particular part of the config. It might be best to put it out as a patch release or something before the next minor to try and gather any bugs

server/utils/s3.ts
```diff
@@ -1,110 +1,66 @@
...
-import { addHours, format } from "date-fns";
 import fetch from "fetch-with-proxy";
 import { v4 as uuidv4 } from "uuid";
 import Logger from "@server/logging/logger";
 
+// backward compatibility
 const AWS_S3_ACCELERATE_URL = process.env.AWS_S3_ACCELERATE_URL;
+const AWS_S3_UPLOAD_BUCKET_URL = process.env.AWS_S3_UPLOAD_BUCKET_URL;
+const AWS_S3_UPLOAD_BUCKET_NAME = process.env.AWS_S3_UPLOAD_BUCKET_NAME;
+const AWS_S3_FORCE_PATH_STYLE = process.env.AWS_S3_FORCE_PATH_STYLE;
```
tommoor, line 11: Thoughts on logging out a warning of deprecation and future removal if these are used?

tommoor:
Thanks for the ongoing work, I just tested locally with my existing config – which looks like:

```
AWS_REGION=us-east-2
AWS_S3_ACCELERATE_URL=https://outline-staging-attachments.s3-accelerate.amazonaws.com
AWS_S3_FORCE_PATH_STYLE=false
AWS_S3_UPLOAD_BUCKET_NAME=outline-staging-attachments
AWS_S3_UPLOAD_BUCKET_URL=https://outline-staging-attachments.s3.us-east-2.amazonaws.com
AWS_S3_UPLOAD_MAX_SIZE=26214400
```

This resulted in an upload failure, it seems like the bucket name ends up getting doubled on the request:

image

frankli0324:
hmmmm... may I just ...`.replace(`${bucket_name}.`, '');`?

tommoor:
Maybe it's the only way 😢

server/utils/s3.ts
```diff
@@ -1,110 +1,86 @@
...
-import { addHours, format } from "date-fns";
 import fetch from "fetch-with-proxy";
 import { v4 as uuidv4 } from "uuid";
 import Logger from "@server/logging/logger";
 
-const AWS_S3_ACCELERATE_URL = process.env.AWS_S3_ACCELERATE_URL;
+// backward compatibility
+function removeBucketName(url: string | undefined) {
+  if (
+    !process.env.AWS_S3_FORCE_PATH_STYLE &&
```
tommoor, line 10:
```suggestion
    process.env.AWS_S3_FORCE_PATH_STYLE !== "true" &&
```

tommoor:
Once this merges I'm going to do a big centralization and refactor of env handling, that will help a lot.

redraskal:
> is there any other changes that must be applied?

The ACL option in presigned posts should be optional.

1) Amazon says it is https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html
2) Not every s3 alternative supports it, mine does not support ACL and rejects attachment uploads

This ties in with https://github.com/outline/outline/issues/2769

NK74:
Any updates ?

almereyda:
As far as I remember, @tommoor still considered this a useful improvement over the current way of configuring S3. Is there a chance this can be reopened, and have a timely (in the sense of the bot) review?

frankli0324:
I'm very sorry but I simply don't have the effort and motivation to work on this after I've graduated and start working...
I'll still try to finish this up if no one else do

Hoshino-Yumetsuki:
It's 2025 and Outline still doesn't seem to have made any improvements to S3 API compatibility, and hopefully this pull request will be merged

PR #3282, opened 2022-03-24
refactor: ♻️ Flex type

huruji:
just using React.CSSProperties to replace types of flex component


tommoor (changes requested):
Good improvement! Please use double quotes and run `yarn lint && yarn tsc` to ensure it passes  (It won't right now)

tommoor (approved):

PR #3283, opened 2022-03-24
refactor: ♻️ del children type

huruji:
The children property does not need to appear in the custom props type


tommoor (approved):
Solid cleanup 👍  – a lot of these Props definitions came across from a fairly recent Flow conversion.

Issue #457, opened 2017-11-29
Comments


tommoor:
Tracking issue for implementing comments.

tommoor:
### Concept

After researching this week, the current plan is to implement something akin to Etherpad's [atext](http://etherpad.org/doc/v1.6.0/#index_atext) / [apool](http://etherpad.org/doc/v1.6.0/#index_apool) concepts that track additional attributes in the document on a per-character basis by mapping an array of attributes to line and character markers.

Incidentally using a format such as this would also allow us to track contributions by author at a granular level.

### Slate

This additional data would be stored on the document database model, serialized separately from the main Markdown body. This could be implemented as an additional serializer/deserializer pair that handles only attributes, mapping them to/from Slate "Marks".

jorilallo:
@tommoor all this sounds good to me. If someone updates the document via the API without the additional data, should we just clear the metadata/comments at that point? I don't think it will be common, edge case.

MarceloCajueiro:
Hey guys,

How is this going?

We started to use outline at Portabilis (our company) and we are missing this feature. Would be great to have the ability to discuss about a document in order to make it better or just to complement the author if it is something very awaited.

I don't know if we have someone with the necessary skills to help you out in the code but we can help in other ways.

almereyda:
This issue does not seem to be about conversational comments next to a document, but rather about annotations in the text itself.

kwix:
Is this feature still being considered?

tommoor:
It is, it just got delayed behind work on improved permissioning

kwstasna:
@tommoor it would be awesome for an ETA of this feature! As soon as its available our workspace will mosdef move there because you have done an excellent work here guys.

tommoor:
Work has begun, keep me honest 😄

pihentagy:
> Work has begun, keep me honest smile

Commit hash or didn't happen :)

milof:
so is this feature is about disqus type comments below a document (like in the title of this issue) or about google docs like comments in the sidebar (like in the commenting spec https://app.getoutline.com/share/79435f04-c5f3-4a74-8e6f-af5607e5d69d) ???

anyhow I'm really excited about progress on this issue 🤘👍

tommoor:
This feature is for reference based comments, there is no plan to have comments below the document that don't reference the content in some way.

milof:
> This feature is for reference based comments, there is no plan to have comments below the document that don't reference the content in some way.

[...]

PR #3286, opened 2022-03-26
chore: Editor refactor

tommoor:
This refactor is to enable us to easily have different variants of editors within Outline, the primary usecase is to allow a minimal editor within the upcoming comments feature which includes basic formatting but not things like embeds and tables.

Towards #457


server/editor/index.ts
```diff
@@ -1,75 +1,9 @@
 import { Schema } from "prosemirror-model";
 import ExtensionManager from "@shared/editor/lib/ExtensionManager";
-
-// marks
-import Bold from "@shared/editor/marks/Bold";
```
tommoor, line 5: This is nice, no separate config for server now

app/editor/index.tsx
```diff
@@ -316,138 +289,36 @@ export class Editor extends React.PureComponent<
...
 
-  createExtensions() {
-    const { dictionary } = this.props;
-
-    // adding nodes here? Update server/editor/renderToHtml.ts for serialization
-    // on the server
-    return new ExtensionManager(
-      [
-        ...[
-          new Doc(),
```
tommoor, line 327: Feels good to lose all this

app/editor/index.tsx
```diff
@@ -754,86 +618,90 @@ export class Editor extends React.PureComponent<
...
+            dir={dir}
+            rtl={isRTL}
+            grow={grow}
+            readOnly={readOnly}
+            readOnlyWriteCheckboxes={readOnlyWriteCheckboxes}
+            ref={this.element}
+          />
+          {!readOnly && this.view && (
+            <>
+              <SelectionToolbar
```
tommoor, line 641: Refactoring these is next, but they'll need converting to functional components first to allow hooks usage.

PR #3310, opened 2022-03-31
fix: Cleanup totally empty drafts on leave

### LINKED ISSUE #3179: the app should clean up documents that have no title and no body content
It's now very easy to create a document, with the addition of the [+] button in the collection sidebar item. It's really easy and natural to click that button and leave before writing anything. The intuitive thing to do when coming back is not to look up the blank document in drafts but to click the [+] button again.

The app should clean up totally blank documents once it has zero websocket connections open.

tommoor:
I looked at doing this from the side of the collaboration server but I feel like it introduces more race conditions honestly. There are two edge cases for client-side that I can see but neither feels terrible considering the current behavior:

1. two people are looking at a totally empty doc and the creator leaves causing it to be deleted, in this case it can be restored.
2. close tab with totally empty doc means it won't have a chance to cleanup so draft is left

closes #3179

### DIFF
--- a/app/scenes/Document/components/Document.tsx
+++ b/app/scenes/Document/components/Document.tsx
@@ -102,6 +102,16 @@ class DocumentScene extends React.Component<Props> {
     this.updateIsDirty();
   }
 
+  componentWillUnmount() {
+    if (
+      this.isEmpty &&
+      this.props.document.isDraft &&
+      this.props.document.title === ""
+    ) {
+      this.props.document.delete();
+    }
+  }
+
   componentDidUpdate(prevProps: Props) {
     const { auth, document, t } = this.props;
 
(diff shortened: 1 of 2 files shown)


thenanyu:
- agreed that the known edge cases are very tolerable
- I noticed a slight inconsistency in the way that trimming is done. it seems that titles are not trimmed — I think they should be, at least for trailing whitespace (I can see a weird situation where a user is using leading whitespace to create strange indenting in the sidebar)
- it's very easy to violate an invariant by typing a whitespace into the body of an empty document and clicking out quickly. The signal to delete outraces the signal to update and `document not found` is triggered. This might also be tolerable

app/scenes/Document/components/Document.tsx
```diff
@@ -138,6 +138,18 @@ class DocumentScene extends React.Component<Props> {
     }
   }
 
+  componentWillUnmount() {
+    if (
+      this.isEmpty &&
+      this.props.document.createdBy.id === this.props.auth.user?.id &&
+      this.props.document.isDraft &&
+      this.props.document.isActive &&
+      this.props.document.hasEmptyTitle
```
thenanyu, line 147: a straightforward way to solve for the two people in the doc situation is to test for if there are any views on the doc that are not the author. I think that would be a very sane reason to retain the doc

tommoor:
> I noticed a slight inconsistency in the way that trimming is done. it seems that titles are not trimmed

Will tackle separately 👍

Issue #3314, opened 2022-03-31
HEIC images do not attach to documents successfully


zacwest:
**To Reproduce**

1. In a document, type /image2. 
2. Pick a .heic image to upload.3. 
3. Note the broken embed.
4. Reload the page, note the image is entirely gone.5. 

It does appear that the upload is successful - in a self-hosted instance, it's uploaded to S3 - but upon reloading the document doesn't show the content anymore.

**Expected behavior**
HEIC-formatted images are attached successfully and visible.

**Screenshots**
<img width="763" alt="image" src="https://user-images.githubusercontent.com/74188/161102668-97f8411d-8f7b-40e2-a7ef-6a1ec40b6dd1.png">

**Outline (please complete the following information):**
- Install: [getoutline.com or self hosted] both getoutline.com & v0.62.0 self-hosted
- Version: [commit sha if self hosted] docker image, latest & version ^

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] macOS
 - Browser [e.g. chrome, safari] Safari
 - Version [e.g. 22] Version 15.4 (17613.1.17.1.6)

tommoor:
Hi @zacwest – currently no browser supports HEIC format natively, we should honestly just block it's upload right now.

PR #3324, opened 2022-04-02
fix: Add stricter validation around image file type uploads

tommoor:
It's frustrating to upload an unsupported file only to find out after the upload has complete, or even worse have it show a broken image tag in the doc.

closes #3324


tommoor:
closes #3314

PR #3337, opened 2022-04-06
chore: Refactor worker, emails and data cleanup to task system

### LINKED ISSUE #2505: Send all emails in worker process
Follow on to #2495 – currently some emails are sent from the web process, originally due to them needing database models to construct the email.

We should refactor the `Mailer` setup so that there is a queue/processor combo that has a single job of taking email content and delivering it, currently the queue takes a "template" and some parameters which the email is then constructed with. It seems like the queue should be moved until later in the process for higher reliability. We usually have the needed models in the web process anyway.

Refactors as part of this:

`/server/mailer` -> `server/emails/index.js`
`/server/emails/*` -> `server/emails/templates/*`
`/server/queues/processors/emails.js` should contain SMTP sending logic.

Depends on #2460

tommoor:
closes #2505


server/queues/tasks/EmailTask.ts
```diff
@@ -0,0 +1,13 @@
...
+import BaseTask from "./BaseTask";
+
+type Props = {
+  type: EmailTypes;
+  options: EmailSendOptions;
+};
+
+export default class EmailTask extends BaseTask<Props> {
+  public async perform(props: Props) {
+    await mailer[props.type](props.options);
```
tommoor, line 11: Next up is to refactor the mailer and move some more logic in here.

PR #3347, opened 2022-04-08
feat: adds "/" keyboard shortcut to share, focus content body on search selection

thenanyu:
- adds "/" keyboard shortcut to focus search input on share screen
- auto-focus content body on search item selection

### DIFF
--- a/app/components/SearchPopover.tsx
+++ b/app/components/SearchPopover.tsx
@@ -11,6 +11,8 @@ import InputSearch from "~/components/InputSearch";
 import Placeholder from "~/components/List/Placeholder";
 import PaginatedList from "~/components/PaginatedList";
 import Popover from "~/components/Popover";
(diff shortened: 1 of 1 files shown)


app/components/SearchPopover.tsx
```diff
@@ -126,8 +149,17 @@ function SearchPopover({ shareId }: Props) {
...
+    popover.hide();
+    if (searchInputRef.current) {
+      searchInputRef.current.value = "";
+      focusRef.current = document.getElementById(bodyContentId);
+    }
+  }, [popover.hide]);
+
   return (
     <>
+      <RegisterKeyDown trigger="/" handler={handleSearchShortcut} />
```
tommoor, line 162: Prefering `useKeyDown` btw, this component is just for use in the old class components where we can't use hooks

tommoor (approved):

tommoor (approved):

PR #3365, opened 2022-04-12
feat: make ioredis configurable via environment variables

heilerich:
As mentioned in #2608 we need support for Redis Sentinel to increase fault tolerance. This PR makes the all `ioredis` client options available to administrators by setting `REDIS_URL` to `ioredis://<base64 encoded json>` instead of the usual `redis(s)://` URL (which will still work).

I first started to implement parsing of a custom `redis+sentinel://` URI scheme, which is not officially supported by redis, but used by some other redis client projects. Then I realized that I also need access to some other TLS related options. So I came up with the current solution, which basically gives us access to all redis related options and adds minimal complexity / maintenance burden to this project (as opposed to parsing a bunch of options to just hand them down to `ioredis`).

I realize that this approach is a little unusual, but I can't really see much downsides right now. That being said, I am open to other suggestions if this approach should not be acceptable.

### DIFF
--- a/server/redis.ts
+++ b/server/redis.ts
@@ -1,6 +1,6 @@
 import Redis from "ioredis";
-import Logger from "./logging/logger";
 import { defaults } from "lodash";
+import Logger from "./logging/logger";
 
 const defaultOptions = {
   maxRetriesPerRequest: 20,
@@ -15,14 +15,14 @@ const defaultOptions = {
   tls:
     process.env.REDIS_URL && process.env.REDIS_URL.startsWith("rediss://")
       ? {
-        rejectUnauthorized: false,
-      }
+          rejectUnauthorized: false,
+        }
       : undefined,
 };
 
-function newRedisClient(url: String | undefined): Redis.Redis {
+function newRedisClient(url: string | undefined): Redis.Redis {
   if (!(url || "").startsWith("ioredis://")) {
-    return new Redis(process.env.REDIS_URL, defaultOptions)
+    return new Redis(process.env.REDIS_URL, defaultOptions);
   }
 
   const decodedString = Buffer.from(url!.slice(10), "base64").toString();
--- a/server/redis.ts
+++ b/server/redis.ts
@@ -20,25 +20,35 @@ const defaultOptions = {
       : undefined,
 };
 
-function newRedisClient(url: string | undefined): Redis.Redis {
-  if (!(url || "").startsWith("ioredis://")) {
-    return new Redis(process.env.REDIS_URL, defaultOptions);
+export default class RedisAdapter extends Redis {
+  constructor(url: string | undefined) {
+    if (!(url || "").startsWith("ioredis://")) {
+      super(process.env.REDIS_URL, defaultOptions);
+    } else {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const decodedString = Buffer.from(url!.slice(10), "base64").toString();
+      const customOptions = JSON.parse(decodedString);
+      const mergedOptions = defaults(defaultOptions, customOptions);
+
+      super(mergedOptions);
+    }
+
+    // More than the default of 10 listeners is expected for the amount of queues
+    // we're running. Increase the max here to prevent a warning in the console:
+    // https://github.com/OptimalBits/bull/issues/1192
+    this.setMaxListeners(100);
   }
 
-  const decodedString = Buffer.from(url!.slice(10), "base64").toString();
-  const customOptions = JSON.parse(decodedString);
-  const mergedOptions = defaults(defaultOptions, customOptions);
+  private static _client: RedisAdapter;
+  private static _subscriber: RedisAdapter;
 
-  return new Redis(mergedOptions);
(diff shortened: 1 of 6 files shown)


tommoor:
It's an interesting solution, I think i'd probably extend the Redis class and move this logic into the extended constructor if this was to be included.

Do you have any examples of other projects using a similar approach? I suppose the other thing to do would be to allow reading of a redis config file that's mounted into the container…

heilerich:
I will move the parsing logic from the exported function to an extended constructor.

As for the general approach, I just kind of followed my urge to have all the configuration in one place and everything else is in the environment variables 🤷🏻
I am mostly dealing with DevOps/infrastructure tasks and passing around configuration as base64 encoded JSON/YAML is very common in this area, e.g. Kubernetes Secrets/ConfigMaps, cloud-init/ignition configuration etc. So this approach just felt like a natural solution for me, I haven't thought too hard about it.

If you prefer having it in a separate file, I could just load the same ioredis parameters from a JSON file and pass something like `REDIS_URL=file:///path/to/redis-config.json`. How does that sound?

Also how would you feel about YAML in this case? I see that `js-yaml` is already in the (indirect) dependencies for this project and as someone who is writing a lot of configuration files I would much prefer YAML. I think it is just much more readable, more flexible and a lot of infrastructure tooling is built to handle configuration as YAML due to its usage in docker/kubernetes/public clouds etc.

tommoor:
> passing around configuration as base64 encoded JSON/YAML is very common in this area, e.g. Kubernetes Secrets/ConfigMaps, cloud-init/ignition configuration etc. So this approach just felt like a natural solution for me, I haven't thought too hard about it.

Feels good, I don't really want to introduce YAML config or any sort of custom file config here. Lets go with cleaning up this solution with the class extension and I think it's very mergable.

server/services/websockets.ts
```diff
@@ -49,8 +49,8 @@ export default function init(app: Koa, server: http.Server) {
 
   io.adapter(
     socketRedisAdapter({
-      pubClient: client,
-      subClient: subscriber,
+      pubClient: Redis.Client,
+      subClient: Redis.Subscriber,
```
tommoor, line 53: What made you do this stylistic change – could we not have kept the original exports or is there something I'm missing?

The capitalization makes it look like these are classes that need initializing.

server/redis.ts
```diff
@@ -18,12 +19,36 @@ const options = {
...
-export { client, subscriber };
+
+export default class RedisAdapter extends Redis {
+  constructor(url: string | undefined) {
+    if (!(url || "").startsWith("ioredis://")) {
+      super(process.env.REDIS_URL, defaultOptions);
+    } else {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const decodedString = Buffer.from(url!.slice(10), "base64").toString();
+      const customOptions = JSON.parse(decodedString);
```
tommoor, line 30: We should try/catch around this with a useful error.

server/services/websockets.ts
heilerich: No, you are absolutely right, this not a necessary change. I just think it is good practice to store global singletons of a class as static properties of said class, because is makes the relationship between them very clear everywhere you import/use them. In fact, many projects I am working on use this pattern, e.g. Apple even recommends this pattern in their developer guidelines.

You are right about the naming though. Maybe we can rename the clients to `Redis.defaultClient`/`Redis.defaultSubscriber` or `sharedClient`. Would that work for you? Otherwise I'll revert those changes.

server/redis.ts
heilerich: In my view this alert is a false positive, so there is no error to catch here. The if clause directly above this code will catch the null case (which is a perfectly fine way to initialize ioredis with the default options). It is a possible way to initialize the redis class in the current master so I wanted to preserve this behavior. Am I missing something here?

heilerich: On second though, you are probably talking about everything else that can go wrong here except the error that is explicitly ignored ^^
I will add the try/catch.

tommoor: Yea, someone will add invalid json and then complain that it breaks in a weird way :)

server/services/websockets.ts
tommoor: > Maybe we can rename the clients to Redis.defaultClient/Redis.defaultSubscriber 

This works for me

tommoor (approved):

tommoor:
Interestingly deploying this change to seems to have subtly changed the behavior of redis on the production collaboration service. It's now calling the `BRPOPLPUSH` command where it wasn't before. Nothing is _broken_, but it's concerning

image

image

heilerich:
Hm, that is interesting indeed. I have been running an earlier version of this change for a few weeks now and had no problems with it. I just took a ~5 minute sample of the redis monitor logs and couldn't find any `BRPOPLPUSH` calls here. I will try to find time to migrate to the current version and check again.

Can we obtain some monitor logs for the environment in which you are seeing the `BRPOPLPUSH` calls? It would probably be helpful to see which lists are being manipulated by the call to track down the component that is causing it.

PR #3381, opened 2022-04-15
feat: allow admins to require invites before user accounts can be created

thenanyu:
Adds a new option in security to disable "Allow authorized signups"
image

Adds a general confirmation dialog component used to confirm the decision to disable
image

When users try to join a team but are not invited, they are ultimately presented with this error state. (I think it looks a little bad stylistically, but that's what's there right now)
image


app/components/Button.tsx
```diff
@@ -111,7 +111,7 @@ const RealButton = styled.button<{
       }
 
       &:disabled {
-        background: none;
+        background: ${lighten(0.05, props.theme.danger)};
```
thenanyu, line 114: this was displaying as white! I'm a bit surprised this wasn't clearly an issue elsewhere but I guess disabled danger buttons are rare enough

app/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,58 @@
+import { observer } from "mobx-react";
```
thenanyu, line 1: A very general confirmation dialog. Good for simple use cases like interrupting a sensitive config change

app/components/Modal.tsx
```diff
@@ -82,9 +82,11 @@ const Modal: React.FC<Props> = ({
...
                         </Text>
                       )}
-                      <NudeButton onClick={onRequestClose}>
-                        <CloseIcon color="currentColor" />
-                      </NudeButton>
+                      <Text as="span" size="large">
+                        <NudeButton onClick={onRequestClose}>
+                          <CloseIcon color="currentColor" />
+                        </NudeButton>
+                      </Text>
```
thenanyu, line 89: the extra text wrapper helps vertically center the glyph visually with the first line of text

app/menus/CollectionMenu.tsx
```diff
@@ -143,7 +143,7 @@ function CollectionMenu({
   const handleDelete = React.useCallback(() => {
     dialogs.openModal({
       isCentered: true,
-      title: t("Delete collection"),
+      title: t("Are you sure you want to delete this collection?"),
```
thenanyu, line 146: changing to long-title format, sorry for the translation churn!

app/scenes/Login/Notices.tsx
```diff
@@ -71,6 +71,12 @@ export default function Notices() {
           admin.
         </NoticeAlert>
       )}
+      {notice === "invite-required" && (
```
thenanyu, line 74: the content of this notice originates with 
```
new Error("Invite required");
```
I'm not thrilled about this, as it seems a bit brittle; open to having a more rigid taxonomy for error types

server/middlewares/passport.ts
```diff
@@ -13,6 +13,10 @@ export default function createMiddleware(providerName: string) {
       },
       async (err, user, result: AccountProvisionerResult) => {
         if (err) {
+          if (err.message === "Invite required") {
+            return ctx.redirect(`/?notice=invite-required`);
+          }
```
thenanyu, line 18: Again, this seems a bit brittle -- open to suggestions

app/components/Button.tsx
tommoor: If something dangerous is disabled then it's usually hidden I suppose!

app/scenes/Login/Notices.tsx
tommoor: You should create an error in `server/errors` and then it'll be easy to use the existing logic…

app/scenes/Login/Notices.tsx
```diff
@@ -73,8 +73,10 @@ export default function Notices() {
       )}
       {notice === "invite-required" && (
         <NoticeAlert>
-          Sorry, you must first be invited to your team before you can create an
-          account – To request an invite, please contact a team admin.
+          The team you are trying to join requires an invite before you can
```
tommoor, line 76: This is good 👍

server/errors.ts
```diff
@@ -20,6 +20,14 @@ export function AuthorizationError(
   });
 }
 
+export function InviteRequiredError(
+  message = "You need an invite join this team"
```
tommoor, line 24:
```suggestion
  message = "You need an invite to join this team"
```

app/scenes/Settings/Security.tsx
```diff
@@ -38,22 +42,74 @@ function Security() {
...
-  const handleDefaultRoleChange = async (newDefaultRole: string) => {
-    const newData = { ...data, defaultUserRole: newDefaultRole };
-    setData(newData);
-    await auth.updateTeam(newData);
-    showSuccessMessage();
-  };
+  const handleDefaultRoleChange = React.useCallback(
+    async (newDefaultRole: string) => {
+      const newData = { ...data, defaultUserRole: newDefaultRole };
+      await saveData(newData);
```
tommoor, line 71:
```suggestion
      await saveData({ ...data, defaultUserRole: newDefaultRole });
```

app/scenes/Settings/Security.tsx
```diff
@@ -38,22 +42,74 @@ function Security() {
...
+        dialogs.openModal({
+          isCentered: true,
+          title: t("Are you sure you want to disable authorized signups?"),
+          content: (
+            <DisableSignupsConfirmation
+              onSubmit={async () => {
+                await saveData(newData);
+              }}
+              submitText={t("I’m sure — Disable")}
+              savingText={`${t("Disabling…")}`}
```
tommoor, line 91:
```suggestion
              savingText={`${t("Disabling")}…`}
```

We've been keeping these out of the translation string for consistency

app/scenes/Settings/Security.tsx
```diff
@@ -38,22 +42,74 @@ function Security() {
...
+              danger
+            >
+              <Trans
+                defaults="Team members will no longer be able to create accounts on their own with <em>{{authMethodSnippet}}</em> and will need to be invited first."
```
tommoor, line 95:
```suggestion
                defaults="New account creation using {{ authenticationMethods }} will be disabled. New users will need to be invited."
```

This appears in the translation interface, maybe change to `authenticationMethods` ? Not sure the bolding is needed here as it should read like part of the sentance.

app/scenes/Settings/Security.tsx
```diff
@@ -117,6 +173,28 @@ function Security() {
         />
       </SettingRow>
 
+      <SettingRow
+        label={t("Allow authorized signups")}
+        name="allowSignups"
+        description={
+          <Trans
+            defaults="Allow authorized <em>{{ authMethodSnippet }}</em> users to create new accounts without first receiving an invite"
```
tommoor, line 181:
```suggestion
            defaults="Allow authorized {{ authMethodSnippet }} users to create new accounts without first receiving an invite"
```

thenanyu: I would rather keep the <em>s around authMethodSnippets

tommoor (approved):

server/commands/userCreator.test.ts
```diff
@@ -210,4 +210,33 @@ describe("userCreator", () => {
     expect(user.email).toEqual(invite.email);
     expect(isNewUser).toEqual(true);
   });
+
+  it.only("should reject an uninvited user when invites are required", async () => {
```
tommoor, line 214:
```suggestion
  it("should reject an uninvited user when invites are required", async () => {
```

thenanyu: oops

app/actions/definitions/documents.tsx
```diff
@@ -306,17 +312,43 @@ export const createTemplate = createAction({
...
 
     stores.dialogs.openModal({
       title: t("Create template"),
+      isCentered: true,
       content: (
-        <DocumentTemplatize
-          documentId={activeDocumentId}
-          onSubmit={stores.dialogs.closeAllModals}
-        />
+        <ConfirmationDialog
```
tommoor, line 326: I wish this was still in it's own component, ngl.

tommoor (approved):

thenanyu:
A couple of final changes and then we can merge
- factored the template confirmation dialog into its own file again but using the new dialog component
- reverted the inline language strings to not affect the translation file keys and updated the values in the english file instead. I don't know if this is over-optimizing but I wasn't comfortable causing so much translation churn with such a small change

gemathus:
Hey guys. I'm curious as to why this setting is only available to cloud-hosted deployments. [This is the commit I'm referring to](https://github.com/outline/outline/commit/964b4ef97dd604862b784af789888e124f21ac8f)

I couldn't find a concrete explanation in commits or pull requests.

tommoor:
We decided to make it cloud+enterprise only.

gemathus:
Got it. Thanks!

PR #3384, opened 2022-04-15
Feat: Allowed Users to customize OIDC Team name

Akash-Ramjyothi:
## Description:

Included OIDC support, but the team name was hardly written in the code. Used environment variables to allow users to set the team name when customizing OIDC.


tommoor:
This was discussed in the initial implementation PR, an additional environment variable is not necessary as the default team name is easily changed in the existing settings screen.

I'd really recommend creating issues for discussion before PR's as described in the contributing guide, it helps to prevent unnecessary work!

PR #3386, opened 2022-04-15
Feat: Added adjust for Editable title area

Akash-Ramjyothi:
# Description:

It's difficult to click the tail of title, because the star button is so close to the title. I only star the very important documents, therefore I don't use the star button too much, but editing title is a frequently used function. To resolve this, moved the star button to the top area, near the New doc button. It's very useful to fix a button to a certain location just like Share and New doc, but now if we wanna star the document, we need to find the tail of the title, which takes some time because document lengths vary. This will also allow us to edit the title without accidentally pressing the star button.


tommoor:
You can click anywhere in the title line _except_ the star, all the way to the right. This wasn't the case until recently so perhaps this patch is from an older version of the software?

PR #3387, opened 2022-04-15
Fix: Draft document are readable from collection member

Akash-Ramjyothi:
## Description:

Currently, the draft documents are being readable by collection members through the URL. This should not happen. Made a resolution to successfully fix the same.


tommoor:
This is by design, so that you can share a draft with a team member and collaborate on it.

PR #3389, opened 2022-04-15
Feat: Added Method for Local development environment

Akash-Ramjyothi:
## Description:

The compose and docker file files in the current project cannot be implemented in the local environment, and all use docker for rapid development. Developers need to install too many dependencies, configure additional authorization services, and distract their efforts outside the outline.

Considering that the current configuration file is still used by other processes, added 2 additional files as a new solution for local development.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,11 @@
+FROM node:14-alpine
+
+WORKDIR /opt/outline
+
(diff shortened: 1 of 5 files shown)


tommoor (changes requested):
The current Dockerfile included in the project is for development, if improvements to be made we should do them there.

This PR also includes a bunch of other unrelated changes.

PR #3390, opened 2022-04-15
Feat: Attachment of Files to Documents

Akash-Ramjyothi:
## Description:

Currently uploading and maintaining files except for images on the outline that we are using. Added a feature to attach files to documents using S3.


tommoor:
I don't really follow what this is doing, we already have file attachments and this change doesn't seem to achieve anything?

tommoor:
I'm closing this out as we already support file attachments, if you have more thoughts on what this PR is offering feel free to leave them as a comment

PR #3391, opened 2022-04-15
Feat: Added Attachment of files as Block to Documents

Akash-Ramjyothi:
## Description:

Implemented a feature to add attachment of files as blocks to the relevant Documents with the help of S3.


tommoor:
Duplicate of #3390

PR #3393, opened 2022-04-15
Feat: Added Google Maps integration support

Akash-Ramjyothi:
## Description:

Added support for Google Maps integration support. Still needs to be worked on using browser address.

### DIFF
--- a/app/embeds/GoogleMaps.js
+++ b/app/embeds/GoogleMaps.js
@@ -1,0 +1,28 @@
+// @flow
+import * as React from "react";
+import Frame from "./components/Frame";
+
+const URL_REGEX = new RegExp(
+    "^https?://(?:www).google.com/maps/embed?pb=(.*)$"
+);
+
+type Props = {|
+    attrs: {|
+        href: string,
+            matches: string[],
+  |},
+|};
+
+export default class GoogleMaps extends React.Component<Props> {
+    static ENABLED = [URL_REGEX];
+
+    render() {
+        return (
+            <Frame
+                {...this.props}
+                src={this.props.attrs.href}
+                title="Google Maps Embed"
+            />
+        );
+    }
+}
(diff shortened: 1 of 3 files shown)


app/embeds/GoogleMaps.js
```diff
@@ -0,0 +1,28 @@
+// @flow
```
tommoor, line 1: We moved over to Typescript quite a while ago now, all types should be Typescript and all file extensions should be `.ts` – see all the other extensions in the codebase.

app/embeds/GoogleMaps.js
```diff
@@ -0,0 +1,28 @@
...
+import * as React from "react";
+import Frame from "./components/Frame";
+
+const URL_REGEX = new RegExp(
+    "^https?://(?:www).google.com/maps/embed?pb=(.*)$"
```
tommoor, line 6: It seems like this is quite hard to get, you'd need to go to Share -> Embed in Google Maps and then copy and paste the html and grab the link out of it – is that correct?

shared/editor/embeds/index.tsx
```diff
@@ -202,6 +203,13 @@ const embeds: EmbedDescriptor[] = [
     component: GoogleSlides,
     matcher: matcher(GoogleSlides),
   },
+  {
+    title: "Google Maps",
+    keywords: "maps",
+    icon: () => <Img src="/images/google-maps.png" />,
```
tommoor, line 209: Missing image

app/embeds/GoogleMaps.js
```diff
@@ -0,0 +1,28 @@
...
+);
+
+type Props = {|
```
tommoor, line 9: The spacing is all off in this file, `lint --fix` needs running

tommoor (changes requested):

PR #3394, opened 2022-04-15
Feat: Added Giphy embed

Akash-Ramjyothi:
## Description:

Added a feature to easily embed a GIF. Giphy helps to accomplish the same in the most efficient manner.

### DIFF
--- a/app/embeds/Giphy.js
+++ b/app/embeds/Giphy.js
@@ -1,0 +1,28 @@
+// @flow
+import * as React from "react";
+import Frame from "./components/Frame";
+
+const URL_REGEX = new RegExp("^https?://giphy.com/gifs/(.*)$");
+
+type Props = {|
+    attrs: {|
+        href: string,
+            matches: string[],
+  |},
+|};
+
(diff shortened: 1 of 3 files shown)


app/embeds/Giphy.js
```diff
@@ -0,0 +1,28 @@
+// @flow
```
tommoor, line 1: We moved over to Typescript quite a while ago now, all types should be Typescript and all file extensions should be `.ts`

shared/editor/embeds/index.tsx
```diff
@@ -202,6 +203,13 @@ const embeds: EmbedDescriptor[] = [
     component: GoogleSlides,
     matcher: matcher(GoogleSlides),
   },
+  {
+    title: "Giphy",
+    keywords: "gif video",
+    icon: () => <Img src="/images/giphy.png" />,
```
tommoor, line 209: Need the matching image for this

tommoor (changes requested):

dkkb:
Great job.

PR #3407, opened 2022-04-17
fix: Embed disabled state should persist

### LINKED ISSUE #3001: Locally persist "Disable Embeds" per document
I tick the Disable Embeds option, but after refreshing the page, the embeds return back. The embeds are very annoying.

tommoor:
Also normalizes `localStorage` access to a `Storage` class.

closes #3001

### DIFF
--- a/app/models/Document.ts
+++ b/app/models/Document.ts
@@ -1,11 +1,12 @@
 import { addDays, differenceInDays } from "date-fns";
 import { floor } from "lodash";
-import { action, computed, observable } from "mobx";
+import { action, autorun, computed, observable } from "mobx";
 import parseTitle from "@shared/utils/parseTitle";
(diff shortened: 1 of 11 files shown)


app/models/Document.ts
```diff
@@ -18,11 +19,28 @@ type SaveOptions = {
...
 
 export default class Document extends ParanoidModel {
+  constructor(fields: Record<string, any>, store: DocumentsStore) {
+    super(fields, store);
+
+    if (this.isPersistedOnce && this.isFromTemplate) {
+      this.title = "";
+    }
+
+    this.embedsDisabled = Storage.get(`embedsDisabled-${this.id}`) ?? false;
```
tommoor, line 29: I hate this, will try and move to a decorator.

PR #3434, opened 2022-04-22
chore: Refactor data import

### LINKED ISSUE #3431: Refactor importer code path
I gotta do this before asking someone to realistically build Notion or Confluence importers, it's currently a mess of spaghetti code.

tommoor:
Here's how the import process goes down:

- User uploads their file from the client -> storage, creating an `Attachment` row in the process
- Client POSTs `collections.import` with the attachment ID, this part could probably be moved but is kept as-is for now.
- A `FileOperation` is created with the correct file format and status
- A worker picks up the `fileOperations.create` event and schedules the correct task for the format
- A worker picks up the task (eg `ImportMarkdownZipTask`) and calls `perform` to run the import

To create new import capabilities we sub-class `ImportTask` and override the `parseData` method and optionally the `fetchData` method.

closes #3431


server/queues/tasks/ImportMarkdownZipTask.test.ts
```diff
@@ -0,0 +1,87 @@
+import fs from "fs";
+import path from "path";
+import { FileOperation } from "@server/models";
+import { buildFileOperation } from "@server/test/factories";
+import { flushdb } from "@server/test/support";
+import ImportMarkdownZipTask from "./ImportMarkdownZipTask";
+
+beforeEach(() => flushdb());
+
+describe("ImportMarkdownZipTask", () => {
```
tommoor, line 10: These tests are essentially copied across from what used to be in `collectionImporter.test`

server/queues/tasks/ImportMarkdownZipTask.ts
```diff
@@ -0,0 +1,215 @@
...
+  /** The file name including extension */
+  name: string;
+  /** The full path to within the zip file */
+  path: string;
+  /** The nested children */
+  children: FileTreeNode[];
+};
+
+export default class ImportMarkdownZipTask extends ImportTask {
+  public async parseData(
```
tommoor, line 23: `parseData` is now the only method that custom importers have to implement it. It accepts the file as input and `StructuredImportData` object as output which will be persisted by the parent class.

server/queues/tasks/ImportTask.ts
```diff
@@ -0,0 +1,308 @@
...
+import BaseTask, { TaskPriority } from "./BaseTask";
+
+type Props = {
+  fileOperationId: string;
+};
+
+/**
+ * Standardized format for data importing, to be used by all import tasks.
+ */
+export type StructuredImportData = {
```
tommoor, line 25: Contains all shared logic for processing imports of different formats

server/routes/api/collections.ts
```diff
@@ -134,22 +142,45 @@ router.post("collections.info", auth(), async (ctx) => {
...
+  const { attachmentId } = ctx.body;
   assertUuid(attachmentId, "attachmentId is required");
+
   const { user } = ctx.state;
   authorize(user, "importCollection", user.team);
+
   const attachment = await Attachment.findByPk(attachmentId);
   authorize(user, "read", attachment);
-  await Event.create({
-    name: "collections.import",
```
tommoor, line 145: Custom event is gone 👋

server/routes/api/collections.ts
```diff
@@ -134,22 +142,45 @@ router.post("collections.info", auth(), async (ctx) => {
...
-      type,
-    },
-    ip: ctx.request.ip,
+
+  await sequelize.transaction(async (transaction) => {
+    const fileOperation = await FileOperation.create(
+      {
+        type: FileOperationType.Import,
+        state: FileOperationState.Creating,
+        format: FileOperationFormat.MarkdownZip,
```
tommoor, line 159: This will be passed in the request from the client once we have more format support

server/utils/zip.ts
```diff
@@ -173,78 +173,3 @@ export async function archiveCollections(collections: Collection[]) {
 
   return archiveToPath(zip);
 }
-
-export async function parseOutlineExport(
```
tommoor, line 177: The equivalent is now part of `MarkdownZipImporterTask`

server/commands/collectionImporter.ts
```diff
@@ -1,206 +0,0 @@
...
-      documents[item.path] = document;
-      continue;
-    }
-
-    if (item.type === "attachment") {
-      const buffer = await item.item.async("nodebuffer");
-      const attachment = await attachmentCreator({
-        source: "import",
-        name: item.name,
-        type,
```
tommoor, line 154: Turns out there was a bug here: This was writing `outline` as the mime type for all attachments created through import. `ImportMarkdownZipTask.ts` now looks up the correct mimetype.

server/commands/collectionImporter.ts
```diff
@@ -1,206 +0,0 @@
...
-      const collectionDir = item.dir.split("/")[0];
-      const collection = collections[collectionDir];
-      invariant(collection, `Collection must exist for document ${item.dir}`);
-
-      // we have a document
-      const content = await item.item.async("string");
-      const name = path.basename(item.name);
-      const tmpDir = os.tmpdir();
-      const tmpFilePath = `${tmpDir}/upload-${uuidv4()}`;
-      await fs.promises.writeFile(tmpFilePath, content);
```
tommoor, line 107: All this pointless tmp file writing was removed 👏

server/commands/collectionImporter.ts
```diff
@@ -1,206 +0,0 @@
...
-      attachments[item.path] = attachment;
-      continue;
-    }
-
-    Logger.info("commands", `Skipped importing ${item.path}`);
-  }
-
-  // All collections, documents, and attachments have been created - time to
-  // update the documents to point to newly uploaded attachments where possible
-  for (const attachmentPath of keys(attachments)) {
```
tommoor, line 168: This second loop and re-save of documents was removed, the text replacement now happens before the initial save. Something that was enabled by a two-pass approach.

server/commands/documentImporter.ts
```diff
@@ -143,42 +150,48 @@ async function confluenceToMarkdown(file): Promise<string> {
...
-  file,
+  mimeType,
+  fileName,
+  content,
   user,
   ip,
+  transaction,
 }: {
   user: User;
-  file: File;
```
tommoor, line 151: It was necessary to change this command so that we could stop writing temporary files and just pass a `string` or a `Buffer` directly instead. As a bonus it allowed removing some `@ts-expect-error`.

server/queues/processors/ImportsProcessor.ts
```diff
@@ -1,79 +0,0 @@
-import fs from "fs";
-import os from "os";
-import File from "formidable/lib/file";
-import invariant from "invariant";
-import collectionImporter from "@server/commands/collectionImporter";
-import { Event, FileOperation, Attachment, User } from "@server/models";
-import { Event as TEvent } from "@server/types";
-import BaseProcessor from "./BaseProcessor";
-
-export default class ImportsProcessor extends BaseProcessor {
```
tommoor, line 10: This file was _weird_, it had such mixed responsibilities.

PR #3441, opened 2022-04-22
Trying to chase missing translations

Limezy:
In this PR I will try to find the few missing translations.
Now that I use Outline on a regular basis and that I have a working dev environment, I should be able to spot and correct even the most hidden ones !


Limezy:
@tommoor how can I translate "Or" here ? https://github.com/outline/outline/blob/bdcfaae025f00f4eb70533bea12c8399eeca0ed1/app/scenes/Login/index.tsx#L282

tommoor:
The most direct change would be to put the translated "Or" in a data attribute, and then access it in the css with [attr](https://developer.mozilla.org/en-US/docs/Web/CSS/attr)

Limezy:
> The most direct change would be to put the translated "Or" in a data attribute, and then access it in the css with [attr](https://developer.mozilla.org/en-US/docs/Web/CSS/attr)

I think I got it ! https://github.com/outline/outline/pull/3441/commits/0b1c310387dbdd6352d0fd6d5b7979f79983b529

Limezy:
@tommoor https://github.com/outline/outline/blob/bdcfaae025f00f4eb70533bea12c8399eeca0ed1/app/components/EventListItem.tsx#L116 
This looks completely wrong in French (and probably also in other languages). How could we have the format dependant on the language ?

tommoor:
We could have the `format` prop optionally take an object where it's keyed by locale perhaps…

```
format={{
  en_US: "MMM do, h:mm a",
  fr_FR: "'Le 'd MMMM 'à' H:mm"
}}
```

Internally within `LocaleTime` we'd just the one that's available or fall back to the default format.

Limezy:
> We could have the `format` prop optionally take an object where it's keyed by locale perhaps…
> 
> ```
> format={{
>   en_US: "MMM do, h:mm a",
>   fr_FR: "'Le 'd MMMM 'à' H:mm"
> }}
> ```
> 
> Internally within `LocaleTime` we'd just the one that's available or fall back to the default format.

Many thanks for your answer ! Do you mean that these formats would be "hardcoded" into the codebase and not through the standard crowdin / i18n pipeline ?

tommoor:
I'm not sure what using the "standard" looks like here? We want quite an explicit short format that shows the timestamp for these revisions.

Limezy:
> I'm not sure what using the "standard" looks like here? We want quite an explicit short format that shows the timestamp for these revisions.

I can try to check if date-fns library comes with a standard format. I'm also not quite sure about the tooltip, which doesn't give much additional information (only the year, currently).
image

Limezy:
Hi @tommoor I have done what you suggested above, for your review.
I have also added a long date format using the same principle.

I'm still trying to find places that are not yet translated. Please tell me if you know some !

The ShareButton seems to be quite broken for now : 
https://github.com/outline/outline/blob/eeb8008927e81330299652a4addec6be2ca4b2bb/app/scenes/Document/components/ShareButton.tsx#L38-L41
These lines seem not to be working for me, do you have an idea why ? Screenshot below
image

You can also see that the "4 days ago" was not translated

Limezy:
> You can also see that the "4 days ago" was not translated

The above commit https://github.com/outline/outline/pull/3441/commits/4c61dd55d80208b569c52a710fc1c05888355521 corrects that, but I wonder if "time ago" shouldn't be retrieved from LocaleTime instead of being recomputed inside SharePopover ?

Limezy:
https://github.com/outline/outline/blob/eeb8008927e81330299652a4addec6be2ca4b2bb/app/scenes/Settings/Members.tsx#L167-L171 
These lines are not working too

image

Limezy:
I have tried to translate the LilBadge with commit https://github.com/outline/outline/pull/3441/commits/172d24a88eebd0aef83e56bf0443906aa6fefd43
I'm not perfectly happy with the solution as the string has been exploded into three sub-strings.
In French this will look like "3 versions de retard", which is correct, but what if another language needs to change the substrings order, like "en retard de 3 versions" (which is also correct in French) ?

Limezy:
This is not working too : https://github.com/outline/outline/blob/eeb8008927e81330299652a4addec6be2ca4b2bb/app/scenes/Settings/Tokens.tsx#L48-L57
image

Limezy:
Found another place where there is no translation !

image

https://github.com/outline/outline/blob/eeb8008927e81330299652a4addec6be2ca4b2bb/app/scenes/Document/components/Contents.tsx#L69

app/components/Sidebar/components/Version.tsx
```diff
@@ -37,10 +39,10 @@ export default function Version() {
...
           <LilBadge>
             {releasesBehind === 0
-              ? "Up to date"
-              : `${releasesBehind} version${
-                  releasesBehind === 1 ? "" : "s"
-                } behind`}
+              ? t("Up to date")
+              : `${releasesBehind} ` +
+                (releasesBehind === 1 ? t("version") : t("versions")) +
+                t("behind")}
```
tommoor, line 45:
```suggestion
              : t(`{{ releasesBehind }} versions behind`, {
                  releasesBehind,
                  total: releasesBehind, 
                })}
```

More like this, will generate the plural and non-plural strings in translation files

app/components/LocaleTime.tsx
```diff
@@ -26,7 +26,7 @@ type Props = {
   addSuffix?: boolean;
   shorten?: boolean;
   relative?: boolean;
-  format?: string;
+  format?: object;
```
tommoor, line 29: This can be typed so that the object keys have to be valid locale strings. Something similar to:

```suggestion
  format?: Record<keyof locales, string>;
```

tommoor:
I think you should try and land the fixes that are here to far and we can look at the ones that aren't translating correctly separately  :)

Limezy:
> I think you should try and land the fixes that are here to far and we can look at the ones that aren't translating correctly separately :)

Many thanks for your kind review... I'm still a very early beginner in Typescript (and dev in general as it's not my background) but it's very nice to learn by contributing. Sure, I'll do as you propose, which is to fix all missing translations I can find, and fixing stuff that is translated but not working for a reason will be a second step (and probably not done by me as it's maybe above my level)

app/components/Sidebar/components/Version.tsx
Limezy: OK. But I didn't understand how you'll manage the non-plural strings with your suggestion ?

tommoor: The translation key will be generated with and without a _plural suffix automatically.

Limezy: This is thanks to the `total: releasesBehind` line ?

app/components/LocaleTime.tsx
Limezy: That one is giving me a hard time. I don't understand where I can find the locales object definition to get its keys with keyof ? I made several attemps (see last ocmmits below) without any success for now

tommoor:
Have you merged `main` recently? The update of the translation file on commit was probably added since you started this branch. Previously you had to manually run `yarn build:i18n`

Limezy:
> Have you merged `main` recently? The update of the translation file on commit was probably added since you started this branch. Previously you had to manually run `yarn build:i18n`

I have just merged main. However I already had automatic update of the translation file on commit. I understood that it wasn't updating the translation file because the chain was already there ! Looks like this was already translated, and then lost the translation for some reason, so when I put it back in it's already translated.

Limezy:
Hi @tommoor sorry I'll need your help on the `Record<keyof locales, string>` thing as I didn't manage to get it working yet (I'm not a dev so this may be a bit too high for my current TS level... ) https://github.com/outline/outline/pull/3441#discussion_r865525592
This is the last blocking step before I can mark this PR as "mergeable" !
Would be nice to have it as part of the coming v0.64 I guess
Thanks !

app/components/LocaleTime.tsx
```diff
@@ -26,7 +26,7 @@ type Props = {
   addSuffix?: boolean;
   shorten?: boolean;
   relative?: boolean;
-  format?: string;
+  format?: Record<keyof typeof locales, string>;
```
tommoor, line 29:
```suggestion
  format?: Partial<Record<keyof typeof locales, string>>;
```

shared/i18n/locales/en_US/translation.json
```diff
@@ -161,6 +161,8 @@
   "Show more": "Show more",
   "Toggle sidebar": "Toggle sidebar",
   "Delete {{ documentName }}": "Delete {{ documentName }}",
+  "Up to date": "Up to date",
+  "{{ releasesBehind }} versions behind": "{{ releasesBehind }} versions behind",
```
tommoor, line 165:
```suggestion
  "{{ releasesBehind }} versions behind": "{{ releasesBehind }} version behind",
  "{{ releasesBehind }} versions behind_plural": "{{ releasesBehind }} versions behind",
```

app/components/Sidebar/components/Version.tsx
```diff
@@ -37,10 +39,11 @@ export default function Version() {
...
           <LilBadge>
             {releasesBehind === 0
-              ? "Up to date"
-              : `${releasesBehind} version${
-                  releasesBehind === 1 ? "" : "s"
-                } behind`}
+              ? t("Up to date")
+              : t(`{{ releasesBehind }} versions behind`, {
+                  releasesBehind,
+                  total: releasesBehind,
```
tommoor, line 45:
```suggestion
                  count: releasesBehind,
```

tommoor (commented):
These suggested changes should fix all the typescript and lint issues

Limezy:
All CI tests are passing, I have built on my local machine and all changes seem to be working well.
I mark this PR as "ready for review".
Normally all places I could find without translation are now translatable
However, in the process I have also found a few places where the translation was existing but not working
Once merged I'll probably create a dedicated issue to gather them in one place for the record

tommoor (approved):

PR #3448, opened 2022-04-24
chore: refactor domain parsing to be more general

thenanyu:
This is an extensive refactor of how domains are parsed in Outline. 

- the functionality of `parseDomain` now relies on the existence of `env.URL` to do the parsing. URLs are parsed into 2 sections: domain and subdomain
- previously: 1, 2, and 3 segment domains were each explicitly handled, and everything else was out of spec. the new approach is completely coupled to outline, but handles all domain lengths
- `tld` never really mattered; it was only used in comparisons or to re-interpolate the full domain string, so we dropped it altogether
- `subdomain.localhost:3000` is now properly recognized as a custom subdomain in development


shared/utils/domains.ts
```diff
@@ -1,69 +1,53 @@
...
+  return trim(url.replace(/(https?:)?\/\//, "")).split(/[/:?]/)[0];
+}
+
 // we originally used the parse-domain npm module however this includes
 // a large list of possible TLD's which increase the size of the bundle
 // unnecessarily for our usecase of trusted input.
+
+// TODO: make this private
+// this only applies to internal domains (hosted domains)
 export function parseDomain(url?: string): Domain | null | undefined {
```
thenanyu, line 22: A lot of complexity in handing and using `parseDomain` comes from the optional argument for `url` — do you know what situations we would run into where url would come in as `undefined`?

shared/utils/domains.test.ts
```diff
@@ -2,133 +2,149 @@ import { stripSubdomain, parseDomain, isCustomSubdomain } from "./domains";
 // test suite is based on subset of parse-domain module we want to support
 // https://github.com/peerigon/parse-domain/blob/master/test/parseDomain.test.js
 describe("#parseDomain", () => {
+  beforeEach(() => {
+    process.env.URL = "https://example.com";
+  });
```
thenanyu, line 7: These are all the same tests, plus a few more that I added — the functionality of `parseDomain` now relies on the existence of `env.URL` to do the parsing, so I added setup to the beginning of each test

shared/utils/domains.ts
```diff
@@ -1,69 +1,54 @@
...
-export function parseDomain(url?: string): Domain | null | undefined {
-  if (typeof url !== "string") {
-    return null;
-  }
-  if (url === "") {
+
+// TODO: make this private
+// this only applies to internal domains (hosted domains)
+export function parseDomain(url: string): Domain | null | undefined {
+  if (!url) {
```
thenanyu, line 24: `""` is falsey, and so is `undefined`. Other than `undefined`, the input should never be a non-string datatype according to the function typing

shared/utils/domains.ts
```diff
@@ -1,69 +1,48 @@
...
-  const parts = normalizedDomain.split(".");
+  const normalBaseUrl = normalizeUrl(env.URL);
+  const normalUrl = normalizeUrl(url);
 
-  // ensure the last part only includes something that looks like a TLD
-  function cleanTLD(tld = "") {
-    return tld.split(/[/:?]/)[0];
+  const baseUrlStart = normalUrl.indexOf(`.${normalBaseUrl}`);
+  if (baseUrlStart === -1) {
+    return { subdomain: "", domain: normalUrl, custom: true };
```
thenanyu, line 29: 🍖 coupling the result of a parse with whether or not it is a custom domain. Otherwise, there was a lot of violation of [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) going on whenever domains were being handled

shared/utils/domains.ts
```diff
@@ -1,73 +1,52 @@
...
     return hostname;
   }
-  if (parsed.tld) {
-    return `${parsed.domain}.${parsed.tld}`;
-  }
   return parsed.domain;
 }
 
-export function isCustomSubdomain(hostname: string) {
+export function isHostedSubdomain(hostname: string) {
```
thenanyu, line 49: we're going to lean into "hosted" meaning hosted by Outline

shared/utils/domains.ts
```diff
@@ -1,73 +1,52 @@
 import { trim } from "lodash";
+import env from "../env";
 
 type Domain = {
-  tld: string;
-  subdomain: string;
+  subdomain?: string;
   domain: string;
+  custom?: boolean;
```
tommoor, line 7: could just always return the `custom` bool for completeness

shared/utils/domains.ts
```diff
@@ -82,6 +61,10 @@ export function isCustomSubdomain(hostname: string) {
   return true;
 }
 
+export function getCookieDomain(domain: string) {
+  return env.SUBDOMAINS_ENABLED ? stripSubdomain(domain) : domain;
```
tommoor, line 65: I think we should remove `SUBDOMAINS_ENABLED` soon, just tie it to hosted deployment

server/routes/api/auth.ts
```diff
@@ -75,13 +76,11 @@ router.post("auth.config", async (ctx) => {
 
   // If subdomain signin page then we return minimal team details to allow
   // for a custom screen showing only relevant signin options for that team.
-  if (
+  else if (
```
tommoor, line 79: 👍

server/routes/api/auth.ts
```diff
@@ -75,13 +76,11 @@ router.post("auth.config", async (ctx) => {
...
     process.env.SUBDOMAINS_ENABLED === "true" &&
-    isCustomSubdomain(ctx.request.hostname) &&
-    !isCustomDomain(ctx.request.hostname)
+    isHostedSubdomain(ctx.request.hostname)
   ) {
-    const domain = parseDomain(ctx.request.hostname);
-    const subdomain = domain ? domain.subdomain : undefined;
```
thenanyu, line 84: is the `undefined` here explicitly necessary? this seems like it's very purposeful but if `isCustomSubdomain` comes back as true, this will never be empty

shared/utils/domains.test.ts
```diff
@@ -1,149 +1,171 @@
...
+      domain: "medium.com",
     });
   });
 
   it("should not include private domains like blogspot.com by default", () => {
     expect(parseDomain("foo.blogspot.com")).toMatchObject({
-      subdomain: "foo",
-      domain: "blogspot",
-      tld: "com",
+      subdomain: "",
```
tommoor, line 72: This behavior change is pretty worrying, maybe these tests just don't make sense any more?

thenanyu: we should talk this one through live, I'll ping once I've cleaned the PR up and we can review some of the stranger corners

shared/utils/domains.ts
thenanyu: yeah they do seem to be 1:1 paired

shared/utils/domains.ts
```diff
@@ -1,73 +1,52 @@
...
-  // simplistic subdomain parse, we don't need to take into account subdomains
-  // with "." characters as these are not valid in Outline
-  if (parts.length >= 3) {
-    return {
-      subdomain: parts[0],
-      domain: parts[1],
-      tld: cleanTLD(parts.slice(2).join(".")),
-    };
-  }
+  // we consider anything in front of the base url to be the subdomain
```
thenanyu, line 32: 🍖 this is the core of the changes, we're making the calculation of subdomain and domain very specific to Outline

shared/utils/domains.ts
tommoor: I'm not aware of anywhere that matters

shared/utils/domains.ts
tommoor: I'm down with this approach, I think the only thing that's rubbing me the wrong way is retaining the generic naming of methods and the filename. It seems inevitable that these methods will be accidentally misused down the line… 

the main one is `subdomain`, maybe we make it `teamSubdomain` ? Or we need to make `parseDomain` a method that isn't exposed as you put in the TODO

tommoor: I think `parseTeamDomain`, or `TeamDomain.parse(url).subdomain` could do the trick

thenanyu: I've had similar thoughts, it's all very overloaded making this hard to name correctly.
I think the way to go is to include all the relevant properties, and try to make them all as semantically correct as possible (kind of like the built in URL class)
```
type OutlineDomain = {
  teamSubdomain: string; // only the team subdomain or empty string, never `www` or `app` 
  host: string; // including subdomain, without protocol or path... e.g. www.getoutline.com, app.getoutline.com, myteam.getoutline.com, customdomain.com
  custom: boolean;
}
```
The insight is that we done really need the domain segment just by itself. If we ever need to examine it, it's in conjunction with subdomain as well, so we're comparing the entire host string.

tommoor:
Are you waiting for me here?

app/scenes/Document/components/Document.tsx
```diff
@@ -614,7 +614,7 @@ class DocumentScene extends React.Component<Props> {
                 </Flex>
               </React.Suspense>
             </MaxWidth>
-            {isShare && !isCustomDomain() && (
+            {isShare && !parseDomain(window.location.origin).custom && (
```
thenanyu, line 617: I changed back to having a single interface for all the domain parsing stuff -- this mimics the native `URL` class. The call sites that previously referenced any of the exported functions now all just call `parseDomain` directly and use the properties of the return object

server/models/Team.ts
```diff
@@ -121,16 +121,20 @@ class Team extends ParanoidModel {
   }
 
   get url() {
+    // custom domain
```
thenanyu, line 124: I believe this comment is correct - the `domain` field on a team represents a custom domain right @tommoor ?

app/scenes/Settings/Details.tsx
```diff
@@ -141,7 +141,9 @@ function Details() {
             subdomain ? (
               <>
                 <Trans>Your knowledge base will be accessible at</Trans>{" "}
-                <strong>{subdomain}.getoutline.com</strong>
+                <strong>
+                  {subdomain}.{new URL(env.URL).host}
```
thenanyu, line 145: it's a safe assumption that `env.URL` is a valid URL

server/models/Team.ts
```diff
@@ -121,16 +121,20 @@ class Team extends ParanoidModel {
...
+    if (!this.subdomain || env.SUBDOMAINS_ENABLED !== "true") {
+      return env.URL;
     }
 
-    const url = new URL(process.env.URL || "");
-    url.host = `${this.subdomain}.${stripSubdomain(url.host)}`;
+    // TODO: HACK: new URL("") will throw an exception
+    // this is just to make the typing work but reall we should
+    // guarantee the existence of env.URL
+    const url = new URL(env.URL || "");
```
thenanyu, line 136: I assume this should be squared away as soon as the new env var stuff is merged. Leaving this big ugly comment here until that's done

shared/utils/urls.ts
```diff
@@ -1,32 +1,27 @@
...
-const env = typeof window !== "undefined" ? window.env : process.env;
-
 export function cdnPath(path: string): string {
   return `${env.CDN_URL}${path}`;
 }
 
+// TODO: HACK: if this is called server-side, it will always return false.
+// - The only call sites to this function and isExternalUrl are on the client
+//   - The reason this is in a shared util is because it's used in an editor plugin
+//     which is also in the shared code
```
thenanyu, line 11: I believe this is all true - the behavior here is very unexpected so I wanted to leave an ugly comment calling it out

shared/utils/domains.ts
```diff
@@ -1,85 +1,64 @@
...
 
-  if (
-    !parsed ||
-    !parsed.subdomain ||
-    parsed.subdomain === "app" ||
-    parsed.subdomain === "www"
-  ) {
-    return false;
+    if (!parsed.custom) {
+      return normalizeUrl(env.URL);
```
thenanyu, line 57: now that I look at it, I think this should also be `new URL(env.URL).host`

app/scenes/Settings/Details.tsx
tommoor: This is going to show a double subdomain though…? I think it was fine hardcoded as subdomains are a cloud-only feature.

server/models/Team.ts
tommoor: Correct, It's a FQDN – eg `docs.mycompany.com`

shared/utils/domains.ts
tommoor: Agreed, it should be `getoutline.com` in production

app/scenes/Settings/Details.tsx
thenanyu: It's not. It's using the env URL which will not have the team subdomain (or any subdomain presumably?)

shared/utils/domains.ts
thenanyu: What's the env URL -- is it http://getoutline.com ?

tommoor: It's `https://app.getoutline.com`

thenanyu: Ah ok - need to change some assumptions then

server/routes/auth/providers/email.test.ts
```diff
@@ -41,8 +42,8 @@ describe("email", () => {
   });
 
   it("should respond with redirect location when user is SSO enabled on another subdomain", async () => {
-    env.URL = "http://localoutline.com";
-    env.SUBDOMAINS_ENABLED = true;
+    env.URL = sharedEnv.URL = "http://localoutline.com";
+    env.SUBDOMAINS_ENABLED = sharedEnv.SUBDOMAINS_ENABLED = true;
```
thenanyu, line 46: Because the new env interface decouples `process.env` from `server/env` it also means any setting of values doesn't propagate.

Because the dependency injection for `shared/env` comes directly from `process.env` we need to set it explicitly as well.

I hate this!

tommoor: Not ideal, but I can deal.

tommoor (approved):

PR #3453, opened 2022-04-25
feat: Allow Document to be fetched without Slug

### LINKED ISSUE #3423: Remove requirement of passing slug with urlId
For all API endpoints that accept a document `id`, we also accept a `urlId`, however for the `urlId` to work it must include the slug even though this isn't used in the query. eg.

`my-doc-L64iqYevsI`, when `L64iqYevsI` is enough. We can safely loosen this restriction.

coreyja:
Fixes #3423

This PR refactors the `Document.findByPk` method to not require the
`slug` portion of the urlID.

Before this function accepted two different 'formats' for the ID.

 - The `uuid` ID of the Document
 - The full `urlID` which looked something like
   `some-document-1234567890`

However the `some-document` slug portion of this identifier wasn't
actually used when looking for a document.

We now allow searching by JUST the postfix of the `urlID`, in the above
example that is `1234567890`.
We do this via a new Regex pattern to match on that just looks for the
right looking id alone, without the prefix.

This codepath looks the same as when we find it by the full `urlID`
besides the different regex that we match on.

The issue #3423 mentions that this should apply to all the API
endpoints. I believe that this `findByPk` method is all that should be
needed for that change. But if this is incorrect, OR you would like more
test coverage on the API endpoints as a more 'end to end test' please
let me know!

### DIFF
--- a/server/models/Collection.test.ts
+++ b/server/models/Collection.test.ts
@@ -372,6 +372,13 @@ describe("#findByPk", () => {
     expect(response!.id).toBe(collection.id);
   });
 
+  test("should return collection when urlId is present, but missing slug", async () => {
+    const collection = await buildCollection();
+    const id = collection.urlId;
+    const response = await Collection.findByPk(id);
+    expect(response!.id).toBe(collection.id);
+  });
+
   test("should return undefined when incorrect uuid type", async () => {
     const collection = await buildCollection();
     const response = await Collection.findByPk(collection.id + "-incorrect");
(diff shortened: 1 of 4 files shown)


coreyja:
Changed the regex to have the slug be optional to avoid the extra regex match, thanks for that suggestion!

This had the side effect of making the change also apply to `Collection`. [The only two usages of the regex were in the `#findByPK` methods on `Document` and `Collection`
]
I believe this should be just as safe as stripping of the slug from Document, but please let me know if that isn't right

tommoor (approved):

Issue #3480, opened 2022-04-30
Cannot read properties of undefined (reading 'split')


EinarArnason:
Cannot start server on fresh install.

**To Reproduce**
1. git clone
2. yarn install
3. Fill out .env
4. yarn sequelize db:create
5. yarn start

**Expected behavior**
Server starts

**Screenshots**
```bash
/opt/outline/node_modules/toidentifier/index.js:24
    .split(' ')
     ^
Cannot read properties of undefined (reading 'split')
    at toIdentifier (/opt/outline/node_modules/toidentifier/index.js:24:6)
    at forEachCode (/opt/outline/node_modules/raw-body/node_modules/http-errors/index.js:245:16)
    at Array.forEach (<anonymous>)
    at populateConstructorExports (/opt/outline/node_modules/raw-body/node_modules/http-errors/index.js:243:9)
    at Object.<anonymous> (/opt/outline/node_modules/raw-body/node_modules/http-errors/index.js:30:1)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/opt/outline/node_modules/raw-body/index.js:16:19)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
```

**Outline (please complete the following information):**
- Install: self hosted
- Version: 0.63.0

**Desktop (please complete the following information):**
 - OS: Arch Linux

tommoor:
Cannot reproduce, most likely there is something in `.env` missing… There's nothing actually from Outline's code in that stacktrace, so that's confusing.

EinarArnason:
It's an error in the raw-body package. https://github.com/stream-utils/raw-body/issues/77
I'll open a PR there for a fix.

tommoor:
[...]

PR #3486, opened 2022-05-01
Upgrade raw-body to solve issue regarding wrong object lookup in http…

EinarArnason:
Related to https://github.com/outline/outline/issues/3480

This change was required to run from source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,6 +151,7 @@
     "query-string": "^7.0.1",
     "quoted-printable": "^1.0.1",
     "randomstring": "1.1.5",
+    "raw-body": "2.5.1",
     "raw-loader": "^0.5.1",
     "react": "^17.0.2",
     "react-avatar-editor": "^11.1.0",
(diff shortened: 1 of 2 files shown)


tommoor:
This isn't needed and just installs an additional copy of raw-body – you can see in the `yarn.lock` that now both versions are installed. The previous version specified in the lockfile is correct, so as long as you install properly with a frozen-lockfile you'll get the right version already.

EinarArnason:
> This isn't needed and just installs an additional copy of raw-body – you [can see in the yarn.lock](https://github.com/outline/outline/pull/3486/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR12343-R12353) that now both versions are installed. The previous version specified in the lockfile is correct, so as long as you install properly with a frozen-lockfile you'll get the right version already.

Ran into this issue with frozen-lockfile as well.

tommoor:
The issue you are having is weird, for sure 😞

Adding this module as a direct dependency isn't the solution though… this dependency version hasn't changed in 2+ years if the line in the lockfile is blamed.

PR #3489, opened 2022-05-03
feat: Migrate `allowedDomains` to a Team Level Settings

### LINKED ISSUE #3412: Add user domain restriction to team settings
Currently domain restriction is only possible for self hosted installations through the `ALLOWED_DOMAINS` env variable. We'd like to move this to a team level setting which will have some affects: 

- It can enable account creation via email without first receiving an invite.
- Restrict SSO access within a larger organization to a subset of the SSO group
- One less env var to configure, (we should use the env as the default during migration though)
- An extra layer of security

coreyja:
Fixes #3412

Previously the only way to restrict the domains for a Team were with the `ALLOWED_DOMAINS` enviroment variable for self hosted instances.
This PR migrates this to be a database backed setting on the Team object. This is done through the creation of a `TeamDomain` model that is associated with the `Team` and contains the domain name

This settings is updated on the Security Tab. Here domains can be added or removed from the Team.

On the server side, we take the code paths that previously were using `ALLOWED_DOMAINS` and switched them to use the Team allowed domains instead

### UI

1. [x] How should saving work? 
	- Currently that settings page saves on each change to the other inputs, since they are simple switches or selects. How its currently wired up is that when any of the switches or selects is changed the whole form saves, INCLUDING the allowed domains. The allowed domains do NOT save on each change to the text boxes, they require the new 'Save' button to be clicked. However clicking the remove button DOES remove the domain immediately
	- Possible Solution: Should I hide the 'Save' button until their are allowed domain changes to 'save'?

## TODO 

- [x] Fix issues with the form on small screen sizes
image

- [x] Add tests and fix issue where we aren't properly checking the domain for new users on existing teams.


app/scenes/Settings/Security.tsx
```diff
@@ -220,8 +247,62 @@ function Security() {
...
+                  key={index}
+                  id={`allowedDomains${index}`}
+                  value={domain}
+                  placeholder="example.com"
+                  onChange={(ev) => {
+                    const newData = { ...data };
+
+                    newData.allowedDomains[index] = ev.target.value;
+                    setData(newData);
+                  }}
```
coreyja, line 272: TODO: Break this out to a method

app/scenes/Settings/Security.tsx
```diff
@@ -220,8 +247,62 @@ function Security() {
...
+          </Button>
+        </Flex>
+      </SettingRow>
     </Scene>
   );
 }
 
+const Remove = styled("div")`
+  margin-top: 6px;
+`;
```
coreyja, line 306: This was copied from the Invite form

The `6px` seems like a magic number to just about center the icon. Didn't look pixel perfect to me but didn't futz with it too much. Let me know if you'd like me to!

server/commands/teamCreator.ts
```diff
@@ -60,19 +59,23 @@ async function teamCreator({
...
+          "authenticationProvider",
+          authenticationProvider
+        );
+        return {
+          authenticationProvider: authP,
+          team,
+          isNewTeam: false,
+        };
+      } else {
+        throw AuthenticationError(`Domain ${domain} is not on the allow list`);
```
coreyja, line 77: This was replicating the old code path as closely as possible, except I added in this slightly more descriptive error message

server/commands/teamUpdater.ts
```diff
@@ -1,9 +1,9 @@
 import { Transaction } from "sequelize";
 import { sequelize } from "@server/database/sequelize";
-import { Event, Team, User } from "@server/models";
+import { Event, Team, TeamDomain, User } from "@server/models";
 
 type TeamUpdaterProps = {
-  params: Partial<Team>;
+  params: Partial<Team> & { allowedDomains: string[] };
```
coreyja, line 6: Not sure if this wants to be cleaned up

`allowedDomains` is not a _real_ attribute of a `Team` but its what the front-end sends us. We translate these into the TeamDomain models

server/migrations/20220419052832-create-team-domains.js
```diff
@@ -0,0 +1,87 @@
...
+                    createdAt: now,
+                    updatedAt: now,
+                  },
+                  transaction,
+                }
+              );
+            }
+          }
+        }
+      }
```
coreyja, line 81: No automated tests for this (should I add some?)
But I tested this with varying different `ALLOWED_DOMAINS` env vars and got the expected results

server/models/Team.ts
```diff
@@ -144,6 +145,11 @@ class Team extends ParanoidModel {
     );
   }
 
+  getAllowedDomains = async function (): Promise<string[]> {
+    const teamDomains: TeamDomain[] = await this.getTeamDomains();
+    return (teamDomains || []).map((teamDomain) => teamDomain.name);
+  };
```
coreyja, line 151: I wrote this method so that `isDomainAllowed` would work without needing to know if `this.teamDomains` was hydrated already

But after being in the repo for a bit, maybe that isn't the best way? It seemed like it would be easy to control that all the usages DID have `TeamDomain` hydrated already, so I could lean that direction.

Do you have a preference?

server/routes/auth/providers/google.ts
```diff
@@ -69,10 +65,6 @@ if (GOOGLE_CLIENT_ID && GOOGLE_CLIENT_SECRET) {
             throw GoogleWorkspaceRequiredError();
           }
 
-          if (!isDomainAllowed(domain)) {
-            throw GoogleWorkspaceInvalidError();
-          }
```
coreyja, line 74: I think I need to replicate a bit of this functionality still

I removed this to rely on the logic in `TeamCreator` to not create a new team in the domain wasn't allowed. But I think I also need to make changes in `accountProvisioner` to make sure the domain of the user ALSO matches one of the allowed domains

shared/i18n/locales/en_US/translation.json
```diff
@@ -654,6 +654,10 @@
   "Allow authorized <em>{{ authenticationMethods }}</em> users to create new accounts without first receiving an invite": "Allow authorized <em>{{ authenticationMethods }}</em> users to create new accounts without first receiving an invite",
   "Default role": "Default role",
   "The default user role for new accounts. Changing this setting does not affect existing user accounts.": "The default user role for new accounts. Changing this setting does not affect existing user accounts.",
+  "Allowed Domains": "Allowed Domains",
+  "The domains which should be allowed to create accounts. This applies to both SSO and Email logins. Changing this setting does not affect existing user accounts.": "The domains which should be allowed to create accounts. This applies to both SSO and Email logins. Changing this setting does not affect existing user accounts.",
+  "Remove Domain": "Remove Domain",
+  "Save Domains": "Save Domains",
```
coreyja, line 660: These being auto-generated is _really_ cool! 💯

Limezy: Not yet auto-translated 🤣

app/scenes/Settings/Security.tsx
```diff
@@ -220,8 +247,62 @@ function Security() {
...
+          data.allowedDomains[data.allowedDomains.length - 1] !== "" ? (
+            <Button type="button" onClick={handleAddDomain} neutral>
+              <Trans>Add another</Trans>
+            </Button>
+          ) : (
+            <span />
+          )}
+
+          <Button type="button" onClick={handleChange}>
+            <Trans>Save Domains</Trans>
```
tommoor, line 296:
```suggestion
            <Trans>Save changes</Trans>
```

I think your idea of only showing this button when there have been edits is a good one, we can combine that with a wording change here and it should be quite clear. Once saved the button can disappear again.

app/scenes/Settings/Security.tsx
```diff
@@ -220,8 +247,62 @@ function Security() {
...
+          {!data.allowedDomains?.length ||
+          data.allowedDomains[data.allowedDomains.length - 1] !== "" ? (
+            <Button type="button" onClick={handleAddDomain} neutral>
+              <Trans>Add another</Trans>
+            </Button>
+          ) : (
+            <span />
+          )}
+
+          <Button type="button" onClick={handleChange}>
```
tommoor, line 295: Disable when saving

app/scenes/Settings/Security.tsx
tommoor: I think it's because inputs have a margin at the bottom but buttons don't so you can't flex align center. Seems okay.

server/models/Team.ts
```diff
@@ -252,8 +264,10 @@ class Team extends ParanoidModel {
   @HasMany(() => AuthenticationProvider)
   authenticationProviders: AuthenticationProvider[];
 
-  // hooks
+  @HasMany(() => TeamDomain)
+  teamDomains: TeamDomain[];
```
tommoor, line 268:
```suggestion
  allowedDomains: TeamDomain[];
```

server/models/Team.ts
tommoor: With the naming change below I think this will do the same thing:

```
await team.$get("allowedDomains");
```

server/routes/auth/providers/google.ts
tommoor: Yep, sounds right

server/commands/teamUpdater.ts
```diff
@@ -58,10 +61,47 @@ const teamUpdater = async ({ params, user, team, ip }: TeamUpdaterProps) => {
   if (inviteRequired !== undefined) {
     team.inviteRequired = inviteRequired;
   }
+  if (allowedDomains !== undefined) {
```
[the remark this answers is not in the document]
coreyja: I took a look at the `$get` and `$set` methods but wasn't exactly sure how to use those here to cleanup this code

I think most of the complexity in this code is due to the fact that the frontend sends us an array of strings, and we need to 'translate' those to an array of `TeamDomain` objects.

I could try to use `$add` and `$remove` instead of the creation on line 82 and the deletion on line 98 but didn't seem like that was any simpler when I was playing around with it. Let me know if you disagree and happy to change it!

Or let me know if there is something I am missing where I should be able to use `$set` with an array of domain strings and not `TeamDomain` objects!

server/errors.ts
```diff
@@ -28,6 +28,14 @@ export function InviteRequiredError(
   });
 }
 
+export function DomainNotAllowedError(
+  message = "The domain is not allowed for this team"
+) {
+  return httpErrors(403, message, {
+    id: "domain_not_allowed",
```
tommoor, line 35: We also need to add a quick error message that's displayed to the user in `Login/Notices.tsx` when they hit this case, it'll appear above the login buttons.

server/commands/teamUpdater.ts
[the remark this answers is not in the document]
tommoor: As it's contained within this command I don't feel too bad about it, you're right there would need to be some translation to model objects either way

server/commands/teamUpdater.ts
```diff
@@ -58,10 +61,47 @@ const teamUpdater = async ({ params, user, team, ip }: TeamUpdaterProps) => {
...
 
-  const transaction: Transaction = await sequelize.transaction();
+    // Add new domains
+    const existingDomains = team.allowedDomains.map((x) => x.name);
+    const newDomains = allowedDomains.filter(
+      (newDomain) => newDomain !== "" && !existingDomains.includes(newDomain)
+    );
+    for (const newDomain of newDomains) {
+      newAllowedDomains.push(
+        await TeamDomain.create(
```
tommoor, line 82: This could be a `await Promise.all(newDomains.map(` type of thing to parallelize the queries

server/commands/teamCreator.ts
tommoor: This should probably be the new `DomainNotAllowedError` though, ye?

tommoor:
There are a couple of references to `ALLOWED_DOMAINS` knocking about in .env.sample, app.json, etc – would be good to remove those at the same time too otherwise I'll forget 😆

tommoor:
closes https://github.com/outline/outline/issues/2678

app/scenes/Login/Notices.tsx
```diff
@@ -21,12 +21,6 @@ export default function Notices() {
           installation. Try another?
         </NoticeAlert>
       )}
-      {notice === "hd-not-allowed" && (
-        <NoticeAlert>
-          Sorry, your Google apps domain is not allowed. Please try again with
-          an allowed team domain.
-        </NoticeAlert>
-      )}
```
coreyja, line 29: Copied the error message from this older error

Also realized this error was dead code now, so cleaned it up!

coreyja:
Got the references to `ALLOWED_DOMAINS` cleaned up and the final comments addressed!

I think this should be good to go once CI passes

server/commands/teamCreator.ts
```diff
@@ -93,6 +93,7 @@ async function teamCreator({
     });
   }
 
+  // eslint-disable-next-line   @typescript-eslint/no-non-null-assertion
```
tommoor, line 96: I think maybe your editor is applying additional eslint rules?

tommoor (approved):

server/commands/teamCreator.ts
tommoor: It would be nice if we could have the `any` rule but because the codebase was relatively recently converted from Flow there's still some work to do there 😅

PR #3523, opened 2022-05-12
feat: Add mermaidjs integration

ArcticXWolf:
As with discussion #1364 this PR tracks the integration of [mermaidjs](https://github.com/mermaid-js/mermaid) into outline.

# New features

* Addition of a new "language" called "Mermaid Diagrams" for the code-block node
* Rendering the mermaid diagram SVG instead of showing the code-block
* Toggling between editing the contents of the code-block and showing the diagram

# UI Demo
capture

# Todo
- [X] Better error handling on mermaid syntax errors
- [X] Fix random generated id so that we do not leak \<div /\> elements on mermaid errors
- [X] ~~Add syntax highlighting for mermaidjs (to Prism)~~ Prism has no highlighting for mermaidjs. Another highlighter is too much work for this PR.
- [X] Verify mobile usability
- [X] Fix tests
- [X] Review on code style / best practices

# Questions to other maintainers
* Is this UI/UX suitable for outline? Do you have other/better ideas?
* Shall we provide a separate /command which creates a code-block with language mermaidjs already selected? (as a quicklink for new users to find)
* Do Prosemirror nodes have something like a unique ID? The mermaid diagram needs an ID and currently I am just creating one with Math.random, so we have no duplicate IDs. But this isn't elegant and leads to problems like leaking div-elements on syntax errors.
* As this is my first contribution to a typescript/node project, are there any best practices / code styles that I have violated? Feedback is appreciated :)
* Mermaidjs is provided with the MIT License, is that compatible to outline?


ArcticXWolf:
Okay, I have just two small things to fix, but I need help for them:

1. It seems like the `const node = view.state.doc.nodeAt(result.pos);` function returns `null` if the code block has no content. That is bad for me, as I use this in the diagram toggle to find the node and this does not work if there is no code in the codeblock, leaving the user stuck with a non-usable diagram. They can only recover by choosing a different code language and entering something. Is there something for me to use the get the node that triggered the click event?

2. This concerns the mobile UX. How does a mobile user select a code language? The action buttons are triggered by :hover events, and those dont work on touch interfaces, do they? I just noticed this during mobile UX testing and it would prevent a user from editing a diagram on mobile (as they cannot see the button).

If I can fix both and you have no feedback concerning a review, then this PR is finished.

shared/editor/nodes/CodeFence.ts
```diff
@@ -91,6 +94,9 @@ export default class CodeFence extends Node {
         language: {
           default: DEFAULT_LANGUAGE,
         },
+        diagram: {
+          default: DEFAULT_DIAGRAM_TOGGLE_STATE,
+        },
```
tommoor, line 99: This works well as a first pass but it has a couple of disadvantages which we need to overcome:

1. The toggle state is synced between editors, so if you have two people in the same doc they will see the state switching – ideally we'd keep this local so that one person can be editing the diagram and the other is just viewing it, or editing elsewhere without the content jumping because of the toggle.
2. This new `diagram` attribute will get added to every code block and cause "edits" to happen to existing documents.

We probably need to keep the toggle inside local state instead, this will mean that the class is added based on a Node decoration that is controlled by that state. I realize that this is quite involved if not familiar with Prosemirror – If  everything else is taken care of I could make this translation.

shared/editor/nodes/CodeFence.ts
```diff
@@ -134,12 +140,56 @@ export default class CodeFence extends Node {
           option.selected = node.attrs.language === value;
           select.appendChild(option);
         });
+        const codeClass = !node.attrs.diagram ? "editor-visible" : "";
+
+        const diagram = document.createElement("div");
+        if (node.attrs.language === "mermaidjs") {
```
tommoor, line 146: Shall we refactor this out into a private method for readability?

shared/editor/nodes/CodeFence.ts
```diff
@@ -134,12 +140,56 @@ export default class CodeFence extends Node {
...
+          const id = Math.round(Math.random() * 100000);
```
tommoor, line 147: This is going to generate a new `id` every time it's rendered – that seems, bad?

shared/editor/nodes/CodeFence.ts
```diff
@@ -134,12 +140,56 @@ export default class CodeFence extends Node {
...
+          const diagramId = "mermaid-diagram-" + id;
+
+          diagram.classList.add("mermaid-diagram");
+          diagram.id = diagramId;
+          if (node.attrs.diagram) {
+            diagram.classList.add("diagram-visible");
+          }
+          diagram.setAttribute("contentEditable", "false");
+          try {
+            mermaid.mermaidAPI.render(diagramId, node.textContent, function (
```
tommoor, line 157: I'm worried at the performance impact of running this constantly, it's going to get triggered a lot – even when nothing changed in the diagram source. At the least we should probably get an idea of the performance hit here, but possibly move this to a plugin too so the rendering can react to changes in the node content.

shared/editor/nodes/CodeFence.ts
```diff
@@ -134,12 +140,56 @@ export default class CodeFence extends Node {
...
+            console.log(error);
+            const errorNode = document.getElementById("d" + diagramId);
+            if (errorNode) {
+              diagram.appendChild(errorNode);
+            }
+          }
+
+          const editButton = document.createElement("button");
+          if (node.attrs.diagram) {
+            editButton.innerText = "Show code";
```
tommoor, line 172:
```suggestion
            editButton.innerText = this.options.dictionary.showCode;
```

These interface strings need adding to the dictionary object so that they can be translated.

shared/editor/nodes/CodeFence.ts
```diff
@@ -134,12 +140,56 @@ export default class CodeFence extends Node {
...
+            editButton.setAttribute("data-shown", "");
+          } else {
+            editButton.innerText = "Show diagram";
```
tommoor, line 175:
```suggestion
            editButton.innerText = this.options.dictionary.showDiagram;
```

shared/editor/nodes/CodeFence.ts
```diff
@@ -234,6 +284,51 @@ export default class CodeFence extends Node {
     }
   };
 
+  handleDiagramEdit = (event: InputEvent) => {
+    const { view } = this.editor;
+    const { tr } = view.state;
+    const element = event.target;
```
tommoor, line 290:
```suggestion
    const element = event.currentTarget;
```

I believe you want this to ensure that the event is always coming from the button

shared/editor/nodes/CodeFence.ts
```diff
@@ -234,6 +284,51 @@ export default class CodeFence extends Node {
...
+            if (errorNode) {
+              diagram.appendChild(errorNode);
+            }
+          }
+        }
+
+        const transaction = tr
+          .setSelection(Selection.near(view.state.doc.resolve(result.inside)))
+          .setNodeMarkup(result.inside, undefined, {
+            language: "mermaidjs",
```
tommoor, line 324: We need to register this, yea?

shared/editor/plugins/Prism.ts
```diff
@@ -7,6 +7,7 @@ import refractor from "refractor/core";
 
 export const LANGUAGES = {
   none: "None", // additional entry to disable highlighting
+  mermaidjs: "Mermaid Diagrams",
```
tommoor, line 10:
```suggestion
  mermaidjs: "Mermaid Diagram",
```

Prefer singular

tommoor: Note it would probably be relatively easy as a followup to create a /mermaid slash command that creates a new code block with the language pre-selected.

tommoor (changes requested):
Really great start!

> This concerns the mobile UX. How does a mobile user select a code language?

I think we can add the `::focus-within` css selector as well as hover so that the buttons always appear when cursor is focused within the code block?

image

The default rendering seems to include text in Serif? We definitely want to override this to use the same typeface as elsewhere.

ArcticXWolf:
Hey,

I dug a bit into the ProseMirror documentation and refactored everything into a ProseMirror plugin. Thus a lot of your points should be fixed, but now I have two problems that I do not know how to solve, maybe you can help me:

1. Is there any way to position the decorators better? It seems that I can only specify character-wise location. For example: I want to put the diagram-wrapper INSIDE the `.code-block`, but I can only put it outside or IN the `pre > code`. Or even worse: I want to annotate the `.code-block` with the diagramId (so the toggle button knows which diagram was triggered), but my only option was to do an inline decorator that spans the whole text. Really ugly :(

2. How can I attach data locally onto nodes/blocks? My problem is the diagramId, it needs to stay stable (so I can save the diagramVisibilty state) and thus I need a way to retrieve data concerning a specific block in the plugin. I've only found two ways:
    - either I use the pluginState, but that is global for all blocks with that plugin, so it does not help here
    - or I use a map (like currently in the commit) which maps block.positions to diagramIds, but that does not work if I add text before a diagram, then the position changes and the map is invalidated...

Sadly I was not able to find anything for that in the ProseMirror docs, but maybe thats just my inexperience with it.
But I hope this refactor is more in the direction that you thought architecture-wise.

ArcticXWolf:
Okay, I got it now in a working state. Still the two issues from above make it quite... ugly? So any help/ideas would be appreciated.
Lastly I added your idea of :focus-within, however this seems to not trigger on my Desktop-Firefox. Dont know if that is just my browser so would be awesome if someone can test this in a mobile browser :)

ArcticXWolf:
I've got an update for the `:focus-within`: apparently the focus selectors do not work properly with ContentEditable. Maybe we need some additional js?

ArcticXWolf:
Has anyone any ideas concerning my three problems (decorator positioning, local statePerNode for plugins and :focus-within)? Just so I can cleanup this PR and we can merge it. I was sadly not able to find anything elegant to solve it.

tommoor:
Hi, I've been traveling and had less time for community PR's – hope to take a detailed look in the next few days.

tommoor:
> Is there any way to position the decorators better? 

I think you have the best solution here already, decorator positioning is quite limited.

> How can I attach data locally onto nodes/blocks? My problem is the diagramId, it needs to stay stable

Well, you already have a stable diagram id through the decoration. I think the trick is to read it out of the decoration from the previous state and pass it through to the next state – no need for the separate object mapping.

A couple of other things I noticed…

- If the code block is empty and you choose to change it to a Mermaid Diagram it immediately swaps to the rendered version and shows an error state. In the case of their being no code yet it makes sense to just stay on the code view.
- We need to show a light, rounded, border around the rendered view similar to how we do for embeds otherwise it's too hard to tell where the image ends to place your cursor.
- We need to figure out the selected state for the diagram, I can probably help with that after as it might be awkward but at the moment it's really hard to select a diagram to delete it
- The text in diagrams is still Serif, which is odd.

shared/editor/plugins/Mermaid.ts
```diff
@@ -0,0 +1,175 @@
...
+    (item) => item.node.type.name === name
+  );
+
+  blocks.forEach((block) => {
+    const language = block.node.attrs.language;
+    if (!language || language !== "mermaidjs") {
+      return;
+    }
+
+    let diagramId: number = positionDiagramIdMap[block.pos];
```
tommoor, line 34: So here, I think you can use the `block.node + block.node.nodeSize` to get the position of the decoration, from which you can extract the previous diagramId?

ArcticXWolf:
> > How can I attach data locally onto nodes/blocks? My problem is the diagramId, it needs to stay stable
> 
> Well, you already have a stable diagram id through the decoration. I think the trick is to read it out of the decoration from the previous state and pass it through to the next state – no need for the separate object mapping.

How can I access the DOM from the PluginState modification / transaction handling? I dont know how to get an instance of the EditorView at that place, which would allow me to use view.domAtPos. 

> 
> A couple of other things I noticed…
> 
>     * If the code block is empty and you choose to change it to a Mermaid Diagram it immediately swaps to the rendered version and shows an error state. In the case of their being no code yet it makes sense to just stay on the code view.

Yeah, I think when I've implemented the stable diagramId, then I disable the automatic switch to the diagram view on language change. But before I cannot see which of the blocks is being the one thats new.

>     * We need to show a light, rounded, border around the rendered view similar to how we do for embeds otherwise it's too hard to tell where the image ends to place your cursor.

I've added the border and background. I chose the theme.codeBackground and theme.codeBorder, because theme.embedBorder has no background equivalent. All embeds seem to bring their own background via the iFrame, so I chose to use codeBorder and codeBackground as they should fit to each other.

>     * We need to figure out the selected state for the diagram, I can probably help with that after as it might be awkward but at the moment it's really hard to select a diagram to delete it

Yeah, if you start deleting from the back, you start deleting the (hidden) code and need a lot of keystrokes (and it breaks the diagram). But here I have no idea how that would work in Prosemirror, so your help (or at least some direction) would be appreciated after we fixed the rest.

>     * The text in diagrams is still Serif, which is odd.
I fixed that now, mermaidjs was applying their own theme on top. Now the diagram inherits the fontFamily from outside. On that note, if outline wants to have a special theme for the mermaid diagrams (I am not much of a designer), then we can modify those values anytime.

shared/editor/plugins/Mermaid.ts
ArcticXWolf: Good idea! But how can I access the DOM from here? I dont know how to get an instance of the EditorView at that place, which would allow me to use view.domAtPos to get the diagramId

tommoor: You don't need to, you can get it off the decoration directly I think?

tommoor: The old state is available, but you need to pass it into getNewState

ArcticXWolf: Ah, you are completely right. Thanks!

ArcticXWolf:
Okay, fixed also the diagramId and that we stay on code view on new diagram creation.

tommoor:
Okay, you need to run `yarn.lock` on your branch and move the PR from draft state :) 

I changed the base branch to `mermaid-integration` so that we can do some final polish such as the focus state and some translation changes before merging to main.

tommoor:
No problem, you just need to:

- Merge `main` in
- You'll find that `yarn.lock` is conflicting
- Run `yarn install` to fix the conflict
- Commit the merge.

ArcticXWolf:
Okay, done :)

Shide:
Closes https://github.com/outline/outline/discussions/2870

dkkb:
Great job.

max-mykhailenko:
@ArcticXWolf it works super slow in my cases. I can't typing normally. In Chrome it works at least in five time faster compared to Safari. Looks like problem in Style recalculating.
image

tommoor:
This is just while editing the source of the diagram? 🤔

max-mykhailenko:
> This is just while editing the source of the diagram? 🤔

Yes, just typing inside mermaid code block any letter

PR #3547, opened 2022-05-17
fix: Replaced the latin letter

rusakovdenis:


tommoor:
Hi, these files are auto-generated – fixes must be submitted through the translation portal:

https://translate.getoutline.com/

Many thanks

Issue #3325, opened 2022-04-02
Upgrade to React 18


tommoor:
React 18 is now stable – It's gotta be done.

max-mykhailenko:
Large documents lags about 500ms for each char. I hope that react 18 at least partially can improve that 
<img width="518" alt="image" src="https://user-images.githubusercontent.com/1213198/169655853-e4499b70-7cb7-4f0c-8257-94c982f7843e.png">

I can help you to fix a lot of react render problems if you can guide me about local development (currently it's more complicated than creating production env, I stuck with `[nodemon] app crashed - waiting for file changes before starting...` during make up)

tommoor:
We can actually remove that call to serialize altogether with some work, weird thing is even in a large document it only takes `0.5ms` on my machine, orders of magnitude difference. I wonder if there is something in your document content that is hitting a slow codepath within that method…

max-mykhailenko:
My example that works slow
[slow-example.md](https://github.com/outline/outline/files/8748019/slow-example.md)

tommoor:
Nice, got it down to `18ms` in this sample doc in the above PR.

max-mykhailenko:
@tommoor so cool, thank you

max-mykhailenko:
@tommoor Works better, thank you 🙌
<img width="493" alt="image" src="https://user-images.githubusercontent.com/1213198/170885287-df45e95b-8e98-48d5-bb3f-b5e0074e5005.png">

PR #3567, opened 2022-05-21
perf: Remove markdown serialize from editor render path

tommoor:
see: https://github.com/outline/outline/issues/3325#issuecomment-1133643088

Also brought some perf wins across from https://github.com/outline/outline/pull/3503 as not sure when that will land.

### DIFF
--- a/app/scenes/Document/components/Document.tsx
+++ b/app/scenes/Document/components/Document.tsx
@@ -1,4 +1,4 @@
-import { debounce, isEqual } from "lodash";
+import { debounce } from "lodash";
 import { action, observable } from "mobx";
 import { observer } from "mobx-react";
 import { AllSelection } from "prosemirror-state";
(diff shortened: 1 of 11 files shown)


app/scenes/Document/components/Document.tsx
```diff
@@ -375,13 +379,21 @@ class DocumentScene extends React.Component<Props> {
...
+    const tasks = this.editor.current?.getTasks();
+    const total = tasks?.length ?? 0;
+    const completed = tasks?.filter((t) => t.completed).length ?? 0;
+    document.updateTasks(total, completed);
+
+    // If the multiplayer editor is enabled we're done here as changes are saved
+    // through the persistence protocol. The rest of this method is legacy.
     if (auth.team?.collaborativeEditing) {
-      action(() => {
-        document.text = this.getEditorText();
```
tommoor, line 382: 🔥

PR #3590, opened 2022-05-26
feat: Validate Google, Azure, OIDC SSO access

tommoor:
Stores `expiresAt` on user authentications, validates continued access to SSO account when loading the app.

Also sets up the framework for adding the same validation to Slack in the near future but Slack is fucking awful as usual and refresh tokens are hidden behind a flag that you have to enable 🤷. Have I [mentioned](https://twitter.com/tommoor/status/1527785670852362240) that I hate the Slack API?

closes https://github.com/outline/outline/issues/2336


server/emails/mailer.tsx
```diff
@@ -44,6 +44,7 @@ export class Mailer {
             "email",
             "Couldn't generate a test account with ethereal.email at this time – emails will not be sent."
           );
+          return;
```
tommoor, line 47: Unrelated, but allows offline development

PR #3607, opened 2022-05-31
Webhooks

### LINKED ISSUE #3534: Webhooks
See [RFC](https://wiki.generaloutline.com/share/47305ccd-9f29-4edf-b48a-21c239bd625e) for more details.

coreyja:
Fixes #3534

This PR adds Webhooks to Outline!

You can create, list and delete Webhooks from an Admin only page in Settings. You can subscribe to all events, different event 'groups', or individual events.

Once created Webhooks are sent via a new `WebhookProcessor` to the custom URL. We record the attempt to delivery a webhook, along with if it succeeded or failed.

If we certain WebhookSubscription has failed on 25 of its last attempts, it's automatically disabled.

`CleanupWebhookDeliveriesTask` is added to the cron webhook, which will delete all deliveries that are older than 7 days

## Screenshots

Disabled Subscription
image

## **Not Implemented Here**

There are a couple things mentioned in the [RFC](https://wiki.generaloutline.com/share/47305ccd-9f29-4edf-b48a-21c239bd625e) which are not included here at the moment. I can add them here if we want, but thought it might be good to get this landed and add them in as follow ups

1. Webhook Secrets
2. `/webhooks.deliveries` and `/webhooks.info` endpoints
3. View of past Webhook Deliveries

## Todo

- [x] Improve UI
- [x] Add all events to the UI
- [x] Add `WebhookProcessor` to backend
- [x] Record Deliveries
- [x] Disable Webhooks that have failed too often recently
- [x] Decide on Webhook payloads. Are the presenters enough?
- [x] Add rest of Events to Processor
- [x] Add `CleanupWebhookDeliveriesTask`
- [x] Add events for new Webhook Subscriptions
- [x] Implement Webhook Secrets or drop the column for now
- [x] Allow WebhookSubscription Events to be added in the UI
- [x] Display enabled status on Webhook list view


server/policies/webhooks.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  return user.isAdmin;
+});
+
+allow(
+  User,
+  ["read", "update", "delete"],
+  WebhookSubscription,
+  (user, webhook): boolean => {
+    return !!user && !!webhook && user.id === webhook.createdById;
```
tommoor, line 17: I think this can match the logic above, any admin can manage webhooks – not just the creator.

server/policies/webhooks.ts
```diff
@@ -0,0 +1,19 @@
+import { User, Team, WebhookSubscription } from "@server/models";
```
tommoor, line 1: This file should be named `webhookSubscription.ts` to be in-keeping

tommoor (commented):
Awesome start :pray:

server/queues/processors/WebhookProcessor.ts
```diff
@@ -0,0 +1,105 @@
...
+      });
+
+      await WebhookDelivery.create({
+        webhookSubscriptionId: webhookSubscription.id,
+        statusCode: response.status,
+        requestBody: body,
+        requestHeaders: headers,
+        responseBody: response.body,
+        responseHeaders: response.headers,
+      });
```
coreyja, line 75: TODO: Pull this out into something that can be more shared between all the different webhook types

server/routes/api/webhookSubscriptions.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (events.length === 0) {
+    throw ValidationError("events are required");
+  }
+
+  console.debug("Creating webhook subscription", {
+    name,
+    events,
+    url,
+    count: events.length,
+  });
```
coreyja, line 49: TODO: Remove this eventually

coreyja:
Ok made a bit more progress here!

Now we actually send Webhooks (for some of the events currently) and record their deliveries! Still need to implement disabling after so many 'unsuccessful' tries

The UI is the same as before, so needs to be finished up there as well

server/queues/processors/WebhookProcessor.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      invariant(hydratedUser, "User not found");
+
+      const webhookUrl = webhookSubscription.url;
+      const headers = {
+        "Content-Type": "application/json",
+      };
+      const body = {
+        event: event.name,
+        user: presentUser(hydratedUser),
```
coreyja, line 87: Using the presenters here for the webhook body. Was thinking that for each Event type, I'd figure out the most relevant object to hydrate and 'present' that as the body

Let me know what you are thinking here if you had something else in mind

coreyja: Love that!

Refactored a bit and using that JSON in a presenter. I think that bit is looking pretty nice, but still need to add most of the other events into there

app/components/Input.tsx
```diff
@@ -104,31 +104,17 @@ export const LabelText = styled.div`
   display: inline-block;
 `;
 
-export type Props = Omit<React.HTMLAttributes<HTMLInputElement>, "onChange"> & {
+export type Props = React.InputHTMLAttributes<
+  HTMLInputElement | HTMLTextAreaElement
+> & {
```
coreyja, line 109: These types had to get changed a bit to play nicely with `react-hook-form`

I thin the impactful change is around the `innerRef` the rest was making typescript happy while I was here

app/components/Input.tsx
```diff
@@ -104,31 +104,17 @@ export const LabelText = styled.div`
...
-  name?: string;
-  pattern?: string;
-  minLength?: number;
-  maxLength?: number;
-  autoFocus?: boolean;
-  autoComplete?: boolean | string;
-  readOnly?: boolean;
-  required?: boolean;
-  disabled?: boolean;
-  placeholder?: string;
```
coreyja, line 126: Most of these are already defined on `React.HTMLAttributes<HTMLInputElement` so no need to 'add' them here

app/components/Input.tsx
```diff
@@ -186,20 +170,35 @@ class Input extends React.Component<Props> {
             ))}
           <Outline focused={this.focused} margin={margin}>
             {icon && <IconWrapper>{icon}</IconWrapper>}
-            <InputComponent
-              // @ts-expect-error no idea why this is not working
```
coreyja, line 190: Changed to the ternary to remove this.

Did it while debugging to make sure this type issue wasn't my issue, and thought it was worth keeping in the end. Let me know if you have other thoughts!

app/components/Input.tsx
```diff
@@ -186,20 +170,35 @@ class Input extends React.Component<Props> {
...
       </Wrapper>
     );
   }
 }
 
+export const ReactHookWrappedInput = React.forwardRef(
+  (props: Omit<Props, "innerRef">, ref: React.Ref<any>) => {
+    return <Input {...{ ...props, innerRef: ref }} />;
+  }
+);
```
coreyja, line 202: `react-hook-form` expects things that accept a `ref` prop

The `Input` accepts an `innerRef` so this simply exposes a `ref` and passes it down as the `innerRef`

app/components/withStores.tsx
```diff
@@ -6,15 +6,13 @@ import useStores from "~/hooks/useStores";
...
   ResolvedProps = JSX.LibraryManagedAttributes<
     P,
     Omit<React.ComponentProps<P>, StoreProps>
   >
->(WrappedComponent: P): React.FC<Omit<ResolvedProps, StoreProps>> {
-  const ComponentWithStore = (
-    props: Omit<React.ComponentProps<P>, StoreProps>
-  ) => {
+>(WrappedComponent: P): React.FC<ResolvedProps> {
+  const ComponentWithStore = (props: ResolvedProps) => {
```
coreyja, line 15: Could be reverted!

But happened to be poking around here and this type could be simplified a bit

app/hooks/useAuthorizedSettingsConfig.ts
```diff
@@ -158,6 +161,14 @@ const useAuthorizedSettingsConfig = () => {
         group: t("Team"),
         icon: DownloadIcon,
       },
+      Webhooks: {
+        name: t("Webhooks"),
+        path: "/settings/webhooks",
+        component: Webhooks,
+        enabled: can.createWebhookSubscription,
+        group: t("Team"),
+        icon: GlobeIcon,
```
coreyja, line 170: Picked the `GlobeIcon` but can definitely change this!

app/scenes/Settings/components/WebhookSubscriptionNew.tsx
```diff
@@ -0,0 +1,292 @@
...
+    font-size: 2em;
+    padding-bottom: 1em;
+  }
+`;
+
+const EventCheckboxText = styled.span`
+  margin-right: 0.5rem;
+`;
+
+const WEBHOOK_EVENTS = {
```
coreyja, line 46: This is just a giant structure to hold all the possible events on the Frontend

Thought it would be interesting to try and generate this from the Events on the server side but wasn't sure if mixing across the top-level dirs was a good pattern (or would even work). So I took the easier way and just manually added all the events to the frontend

This was created by looking at the `Events` type here: https://github.com/outline/outline/pull/3607/files#diff-9aa02e81e82dfd51bd95a1dd4046be4847afefab02af12edde04b215e840f2acL231

server/presenters/webhook.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export interface WebhookPresentation {
+  id: string;
+  webhookSubscriptionId: string;
+  teamId: string;
+  event: string;
+  payload: { model: unknown; id?: string };
+  createdAt: Date;
+}
```
coreyja, line 18: This is a type that represents the final shape that gets sent to each webhook url

server/queues/processors/WebhookProcessor.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      expect.anything()
+    );
+    const parsedBody = JSON.parse(
+      fetchMock.mock.calls[0]![1]!.body!.toString()
+    );
+    expect(parsedBody.webhookSubscriptionId).toBe(subscription.id);
+    expect(parsedBody.event).toBe("users.signin");
+    expect(parsedBody.teamId).toBe(subscription.teamId);
+    expect(parsedBody.payload.id).toBe(signedInUser.id);
+    expect(parsedBody.payload.model).toBeDefined();
```
coreyja, line 50: This was the best way I could figure out the grab the JSON out of the fetch mock

Don't love all the non-null assertions so if anyone knows a way to clean it up let me know!

server/migrations/20220521164111-create-webhook-subscription.js
```diff
@@ -0,0 +1,59 @@
...
+          },
+          updatedAt: {
+            allowNull: false,
+            type: Sequelize.DATE,
+          },
+        },
+        { transaction }
+      );
+    });
+  },
```
tommoor, line 55: We need to add an index for `teamId` and `enabled` here as this is the most common query pattern. See previous migrations for examples

server/routes/api/webhookSubscriptions.ts
```diff
@@ -0,0 +1,98 @@
...
+import pagination from "./middlewares/pagination";
+
+const router = new Router();
+
+router.post("webhookSubscriptions.list", auth(), pagination(), async (ctx) => {
+  const { user } = ctx.state;
+  authorize(user, "listWebhookSubscription", user.team);
+  const webhooks = await WebhookSubscription.findAll({
+    where: {
+      createdById: user.id,
```
tommoor, line 19:
```suggestion
      teamId: user.teamId,
```

Suggest we list all webhooks in the team to allow any admin to manage?

server/migrations/20220606031139-create-webhook-delivieries.js
```diff
@@ -0,0 +1,61 @@
...
+            allowNull: false,
+          },
+          updatedAt: {
+            type: Sequelize.DATE,
+            allowNull: false,
+          },
+        },
+        { transaction }
+      );
+    });
```
tommoor, line 56: Need separate indexes on `webhookSubscriptionId` and `createdAt` for this one

server/queues/processors/WebhookProcessor.ts
```diff
@@ -0,0 +1,383 @@
...
+        where: { webhookSubscriptionId: subscription.id },
+        order: [["createdAt", "DESC"]],
+        limit: 25,
+      });
+
+      const allFailed = recentDeliveries.every(
+        (delivery) => delivery.status === "failed"
+      );
+
+      if (allFailed) {
```
tommoor, line 378: This is actually immediately disabling the webhook on the first failure, need to check that `recentDeliveries` actually returns 25 results.

```suggestion
      if (recentDeliveries.length === 25 && allFailed) {
```

server/queues/processors/WebhookProcessor.ts
```diff
@@ -0,0 +1,383 @@
...
+    });
+
+    const newStatus = response.ok ? "success" : "failed";
+
+    await delivery.update({
+      status: newStatus,
+      statusCode: response.status,
+      requestBody: jsonBody,
+      requestHeaders: headers,
+      responseBody: response.body?.toString(),
```
tommoor, line 363: This is just saving "[object Object]" in the db, maybe should just change that column to be text instead of blob…

server/queues/processors/WebhookProcessor.ts
```diff
@@ -0,0 +1,383 @@
...
+
+  async perform(event: Event) {
+    const webhookSubscriptions = await WebhookSubscription.findAll({
+      where: {
+        enabled: true,
+        teamId: event.teamId,
+      },
+    });
+
+    await Promise.all(
```
tommoor, line 59: I think we need to split this processor into two. One part that finds all the applicable subscriptions as you have done here and then just schedules a `DeliverWebhookTask` (or similar) for each. There are a couple of things this achieves…

- If there is an error as part of one webhook delivery and this job retries then it won't cause a duplicate delivery for another webhook in the team that succeeded.
- We can lean on the existing queue retry mechanism to retry individual webhooks
- Spreads the load across multiple machines (there are many workers in the cloud hosted instance) and hopefully results in the http requests overall happening faster

app/scenes/Settings/components/WebhookSubscriptionNew.tsx
```diff
@@ -0,0 +1,292 @@
...
+            Object.keys(WEBHOOK_EVENTS).includes(value) &&
+            "group-label"
+        )}
+      >
+        <EventCheckboxText>{label}</EventCheckboxText>
+        <input
+          type="checkbox"
+          defaultValue={value}
+          {...register("events", {})}
+        />
```
tommoor, line 223:
```suggestion
        <input
          type="checkbox"
          defaultValue={value}
          {...register("events", {})}
        />
        <EventCheckboxText>{label}</EventCheckboxText>
```

Lets flip these around so that the checkboxes are aligned vertically

tommoor (commented):
Great progress – I think the things you haven't included in this first pass are fine. The one thing I would consider is an edit screen for the webhook as it kind of sucks if you make a mistake and need to change the url or add a single event it all needs recreating right now.

app/scenes/Settings/Webhooks.tsx
```diff
@@ -45,7 +45,7 @@ function Webhooks() {
     >
       <Heading>{t("Webhooks")}</Heading>
       <Text type="secondary">
-        <Trans defaults="WEBHOOKS are really cool" />
+        <Trans defaults="You can make an unlimited number of Webhooks. They are URLs that will be hit when events happen in Outline, and can be used to create custom integrations." />
```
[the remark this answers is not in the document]
coreyja: Ya I pulled the copy from the API Token page, agreed that unlimited webhooks wouldn't be great lol

Thanks for the changes down below 🙏

app/scenes/Settings/Webhooks.tsx
```diff
@@ -0,0 +1,69 @@
...
+function Webhooks() {
+  const team = useCurrentTeam();
+  const { t } = useTranslation();
+  const { webhookSubscriptions } = useStores();
+  const [newModalOpen, handleNewModalOpen, handleNewModalClose] = useBoolean();
+  const can = usePolicy(team.id);
+
+  return (
+    <Scene
+      title={t("Webhook Subscriptions")}
```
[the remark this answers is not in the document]
coreyja: Sounds good!

app/scenes/Settings/components/WebhookSubscriptionListItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+      content: (
+        <WebhookSubscriptionRevokeDialog
+          onSubmit={dialogs.closeAllModals}
+          webhook={webhook}
+        />
+      ),
+    });
+  }, [t, dialogs, webhook]);
+
+  const postfix = webhook.enabled ? "" : " (disabled)";
```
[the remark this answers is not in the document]
coreyja: Oh I like the badge! Using the yellow variant now but can change it if you like the gray one better

Got the `t` added too, will make sure I catch all those in this PR!

app/scenes/Settings/components/WebhookSubscriptionNew.tsx
coreyja: Ahh ya this looks much cleaner this way 👍

app/scenes/Settings/components/WebhookSubscriptionNew.tsx
```diff
@@ -0,0 +1,292 @@
...
+      <EventCheckbox label={t("All events")} value="*" />
+
+      <FieldSet
+        disabled={isAllEventSelected}
+        className={joinClasses(isAllEventSelected && "disabled")}
+      >
+        <GroupGrid className={joinClasses(isMobile && "mobile")}>
+          {Object.entries(WEBHOOK_EVENTS).map(([group, events], i) => (
+            <GroupWrapper key={i} className={joinClasses(isMobile && "mobile")}>
+              <EventCheckbox label={t(`All ${group} Events`)} value={group} />
```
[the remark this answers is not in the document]
coreyja: Thanks!

server/presenters/webhook.ts
```diff
@@ -0,0 +1,34 @@
+import { WebhookDelivery, WebhookSubscription } from "@server/models";
+import { Event } from "@server/types";
+
+interface WebhookProps {
+  event: Event;
+  delivery: WebhookDelivery;
+  subscription: WebhookSubscription;
+  payload: { model: unknown; id?: string };
```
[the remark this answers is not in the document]
coreyja: This is the presented models at this point, so I don't `BaseModel` will work here

I could work it so that we pass the model and the presenter down farther. The one place that is a bit weird is with the `user.invite` event, cause we don't have an actual User model to hydrate or present. That's one reason I did the presenting a bit higher up, in the WebhookProcessor, to accommodate not always having a real domain object

server/queues/processors/WebhookProcessor.ts
coreyja: 🤦 

Nice catch! Fixed and tested

server/queues/processors/WebhookProcessor.ts
coreyja: Ya I think changing the column type is a good idea, I'll do that!

This actually was saving the contents in the DB correctly, but it wasn't clear why! The `body` property here returns a `ReadbleString` so `toString` actually did the 'correct' thing, at least for what was already buffered.
But I also realized there was an easier way with the fetch api `await response.text()`!'

It also wasn't tested well before but is now!

server/queues/processors/WebhookProcessor.ts
coreyja: Ahh! Ya makes a ton of sense to split the deliveries into separate jobs

Will do!

server/routes/api/webhookSubscriptions.ts
coreyja: Ahh yup that makes more sense, especially since we already changed the authorize layer to let any admin through!

app/scenes/Settings/components/WebhookSubscriptionEdit.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+        const toSend = {
+          ...data,
+          events,
+        };
+
+        await webhookSubscription.save(toSend);
+
+        showToast(
+          t("Webhook subscription updated", {
```
coreyja, line 35:
```suggestion
          t("Webhook updated", {
```

coreyja:
Ok!

Feedback addressed and added an Edit form to be able to edit the existing webhooks!

I read through the PR and think everything is in a good spot, and is a good first pass at webhooks!

tommoor:
Nice nice, we getting there. I found some issues while testing, going to move this over to a branch on the main repo so that I can collaborate on the remaining things

Issue #3388, opened 2022-04-15
Unable to disable collaborative editing


daladim:
Thanks for this great piece of software!

On my self-hosted test instance, collaborative mode is enabled (by default). I'd like to disable it but the switch to control this behaviour is disabled (I cannot click it).

Reading at the code (but I'm not a web-development expert), there is
```
        <Switch
          id="collaborativeEditing"
          name="collaborativeEditing"
          checked={data.collaborativeEditing}
          disabled={data.collaborativeEditing}
          onChange={handleChange}
        />
```
which looks like this switch is disabled if (and only if) collaborative editing is enabled. Is this a bug, or is it by design that an instance that has opted in collaborative editing cannot go back?

Thanks

## Versions

I am using Outline 0.63.0-1, and the latest redis, postgres and minio dependencies, using the recommended Docker Compose setup

## Context

In case you're interested :)
I'm trying Outline (and am puzzled by the recent changes regarding collaborative editing).
I'd like to revert to non-collaborative editing, because (and I would definitely love at least #3132) I'm not adverse to collaborative editing, but I am to this "always-edit" mode, that feels like it is easy to inadvertently delete/change some content (yes, the history is kept so that's not a big deal...unless we inadvertently change something _without noticing_)

daladim:
As a workaround (or for testing reasons), is there a way (environment variable, Docker settings...) to disable collaborative editiing on 0.63.0-1? (ideally without changing anything in the Docker container)

daladim:
I could run Outline with this setting disabled by migrating from a 0.62.0 version with collaborative editing disabled...but I'm sure there should be an easier way :)

tommoor:
This is by design as disabling collaborative editing after enabling will result in data loss, non-collab editing will also be removed in a future version of the software so that there aren't two different code paths to maintain.

If you've just installed and are not worried about data loss you can either disable in the database directly or through the API as two workaround options 😄

daladim:
All right, thanks for your answer.
I may try this on this weekend then.

(and Happy Easter)

daladim:
Just to be sure, what "data loss" are we talking about? Will we lose data of documents being currenlty edited? (and as such, it is "safe" to disable this setting via the DB on an instance that is not in use) Or could we lose already edited documents already edited?

tommoor:
When collaborative editing is enabled the data is stored in a different database column. So if you:

- disable it
- edit documents
- re-enable it

[...]

PR #3618, opened 2022-06-02
add warning on collaborative editing feature

psifertex:
Per: https://github.com/outline/outline/issues/3388#issuecomment-1127248086

Hopefully this won't break the translation just combining the translated strings. Do I need to also manually add the second string to the translation API or is that auto-generated somehow?

### DIFF
--- a/app/scenes/Settings/Features.tsx
+++ b/app/scenes/Settings/Features.tsx
@@ -48,7 +48,9 @@ function Features() {
         label={t("Collaborative editing")}
         description={t(
           "When enabled multiple people can edit documents at the same time with shared presence and live cursors."
-        )}
+          ) + " " + t(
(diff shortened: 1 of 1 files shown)


psifertex:
Looks like the linter failed -- tried to match the syntax of the previous string since it seems odd to have one separated out and not the other but can amend if the preference is to strictly follow that style.

tommoor:
Hi Jordan – I think we were thinking something with a little more panache than just capitalized text, such as a confirmation dialog. 😉

psifertex:
I'd like that too, but given my skills with the codebase that was all I was capable of and perfect is the enemy of good enough. My feelings aren't hurt if this isn't sufficient, I just wanted to at least offer to make a small improvement since I complained about it in an associated issue.

tommoor:
Closed with confirmation dialog approach here: 88b3b5033

Issue #3611, opened 2022-06-01
Nested pages of publicly shared documents are not (Google) search indexed


tommoor:
Google has no problem correctly parsing and indexing the main shared link, however sub-pages should also be individually indexed and that isn't happening. For example, see the [indexed pages for wiki.generaloutline.com](https://www.google.com/search?q=site%3Awiki.generaloutline.com&sxsrf=ALiCzsZiuKkuXjlpjiAvRhIgOV4US4ikzw%3A1654069110252&ei=dheXYrjxDsSQxc8PuMqd6Ac&ved=0ahUKEwi4hIDO34v4AhVESPEDHThlB30Q4dUDCA4&uact=5&oq=site%3Awiki.generaloutline.com&gs_lcp=Cgdnd3Mtd2l6EANKBAhBGAFKBAhGGABQywJYywJg-gNoAXAAeACAAQCIAQCSAQCYAQCgAQHAAQE&sclient=gws-wiz) where the documentation lives – no sub-pages are listed.

This issue has also been raised through [support](https://secure.helpscout.net/conversation/1891576093/1425).

Testing how GoogleBot sees the pages, all of the anchor links are there correctly. ([Search Report](https://search.google.com/test/mobile-friendly/result?id=ENHi_kyh5MQbcmPEofRbPQ) -> View Tested Page -> HTML)

tommoor:
They're search indexed slowly now

PR #3646, opened 2022-06-08
fix: SSR meta data for nested shared documents

tommoor:
Related #3611

### DIFF
--- a/server/commands/documentLoader.ts
+++ b/server/commands/documentLoader.ts
@@ -1,0 +1,187 @@
+import invariant from "invariant";
+import { Op } from "sequelize";
+import {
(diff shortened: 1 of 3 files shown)


server/commands/documentLoader.ts
```diff
@@ -0,0 +1,187 @@
...
+  user?: User;
+};
+
+type Result = {
+  document: Document;
+  share?: Share;
+  collection: Collection;
+};
+
+export default async function loadDocument({
```
tommoor, line 24: Just a direct move of `loadDocument` in `/api/documents.ts` which is already tested very well in that context

PR #3647, opened 2022-06-08
fix: Enable `documents.update` with collab editing

### LINKED ISSUE #3424: Enable `documents.update` endpoint when collaborative editing is enabled
Currently this is not enabled as it was previously not possible to update the CRDT reliably from the server. With the move to a fork of `y-prosemirror` we can use `updateYFragment` to do so, it's not simple though – the full flow would look something like:

- POST request to `/api/documents.update` with Markdown
- Request is forwarded to a collaboration server
- Collaboration server loads CRDT state or uses in-memory copy if it has one
- Markdown converted to Prosemirror JSON
- ProsemirrorJSON mapped onto CRDT with `updateYFragment`
- Persist state to db and remove from memory if no-one else is connected still

tommoor:
This works with the gotcha that if someone is viewing the document when it's updated via the API their changes will overwrite the API request. Definitely not what we want in the end, but it might be a pragmatic way to get this 99% working again quickly.

closes #3424


smartygus:
@tommoor I just setup an automation to update a particular page via the API and thought I was going crazy, because it wasn't updating. But I had the page open the whole time my browser, which based on the info here was obviously overwriting the update triggered via the API.

Are there plans to make updates via the API work even if someone is viewing the document? How complex a problem would that be?

Thanks heaps

tommoor:
Yea, obviously the current situation isn't ideal – there is quite a lot of complexity in getting this to work and very few folks using the API in this way is the truth.

To support viewer commenting we'll need to solve it though, so that's most likely going to be the point at which it gets fixed.

smartygus:
Yeah fair call, this is an edge case 😃 
Thanks for the info 👍

PR #3652, opened 2022-06-10
feat: allow personal gmail accounts to be used to sign into teams with an existing invite

thenanyu:
This is only the simplest part of the spec, but it proves some important concepts
- allows personal gmail auth to accounts with an existing invite to that email
- redirects user to originating domain/subdomain on the callback from oauth


shared/utils/domains.test.ts
```diff
@@ -42,6 +42,13 @@ describe("#parseDomain", () => {
     });
   });
 
+  it("should return the same result when parsing the returned host", () => {
+    const customDomain = parseDomain("www.example.com");
+    const subDomain = parseDomain("myteam.example.com");
+    expect(parseDomain(customDomain.host)).toMatchObject(customDomain);
+    expect(parseDomain(subDomain.host)).toMatchObject(subDomain);
+  });
```
thenanyu, line 50: The redirect logic relies on this behavior. It's always been true, just specifically testing for it now

server/utils/passport.ts
```diff
@@ -6,48 +6,78 @@ import {
...
         false,
         state
       );
     }
 
-    ctx.cookies.set(this.key, "", {
+    // Oauth callbacks are hard-coded to come to the apex domain, so we
+    // redirect to the original app domain before attempting authentication.
+    // If there is an error during auth, the user will end up on the same domain
+    // that they started from.
```
thenanyu, line 54: 🍖 redirect the oauth callback to the originating host (team subdomain or custom domain)

server/utils/passport.ts
```diff
@@ -6,48 +6,78 @@ import {
...
-import { getCookieDomain } from "@shared/utils/domains";
-import { OAuthStateMismatchError } from "../errors";
+import { getCookieDomain, parseDomain } from "@shared/utils/domains";
+import { AuthRedirectError, OAuthStateMismatchError } from "../errors";
 
 export class StateStore {
   key = "state";
 
-  store = (ctx: Request, callback: StateStoreStoreCallback) => {
+  store = (req: Request, callback: StateStoreStoreCallback) => {
```
thenanyu, line 15: did some renaming of key variables to better match their semantic descriptions

server/routes/auth/providers/google.ts
```diff
@@ -57,37 +58,74 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
+              authentication: {
+                providerId: profile.id,
+                accessToken,
+                refreshToken,
+                scopes,
+              },
+            });
+          } else {
+            // No domain means it's a personal Gmail account
+            // We only allow sign-in to existing invites here
```
thenanyu, line 99: 🍖 Only the else clause is new, I changed the order of the if statement
Here we do a simple team and user lookup. In the future, we will need to make decisions about how much provisioning to allow from personal gmail accounts. This covers the simple case of an already existing email invite

server/utils/passport.ts
```diff
@@ -6,48 +6,78 @@ import {
...
+    const appDomain = parseDomain(host);
+    if (appDomain.host !== parseDomain(req.hostname).host) {
+      const reqProtocol = req.protocol;
+      const requestHost = req.get("host");
+      const requestPath = req.originalUrl;
+      const requestUrl = `${reqProtocol}://${requestHost}${requestPath}`;
+      const url = new URL(requestUrl);
+
+      url.host = appDomain.host;
+
```
thenanyu, line 64: I chose to use the existing codepath of redirecting a handled error — It seems sane but open to suggestions

app/scenes/Login/Notices.tsx
```diff
@@ -21,7 +24,7 @@ export default function Notices() {
           installation. Try another?
         </NoticeAlert>
       )}
-      {notice === "malformed_user_info" && (
+      {notice === "malformed-user-info" && (
```
thenanyu, line 27: this was almost certainly a bug before

app/scenes/Login/Notices.tsx
```diff
@@ -38,7 +41,7 @@ export default function Notices() {
           try again in a few minutes.
         </NoticeAlert>
       )}
-      {notice === "auth-error" &&
+      {(notice === "auth-error" || notice === "state-mismatch") &&
```
thenanyu, line 44: state mismatch provided no error message previously

server/routes/auth/providers/google.ts
```diff
@@ -57,37 +58,74 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
         ) => void
       ) {
         try {
+          const state = req.cookies.get("state");
+          const host = state ? parseState(state).host : req.hostname;
+          const appDomain = parseDomain(host);
```
tommoor, line 63: `appDomain` feels a bit overloaded here, `originalDomain`, `startingDomain` `originDomain` something along those lines maybe?

server/routes/auth/providers/google.ts
```diff
@@ -57,37 +58,74 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
+                accessToken,
+                refreshToken,
+                scopes,
+              },
+            });
+          } else {
+            // No domain means it's a personal Gmail account
+            // We only allow sign-in to existing invites here
+            let team;
+            if (appDomain.custom) {
```
tommoor, line 101: If self-hosted we can just use the first team here to add support for that scenario

server/utils/passport.ts
```diff
@@ -6,48 +6,78 @@ import {
...
+import { AuthRedirectError, OAuthStateMismatchError } from "../errors";
 
 export class StateStore {
   key = "state";
 
-  store = (ctx: Request, callback: StateStoreStoreCallback) => {
+  store = (req: Request, callback: StateStoreStoreCallback) => {
     // Produce a random string as state
-    const state = crypto.randomBytes(8).toString("hex");
+    console.log({ storeRequest: req, hostname: req.hostname });
```
tommoor, line 17: Hanging log

shared/utils/urlHelpers.ts
```diff
@@ -49,7 +49,7 @@ export function changelogUrl(): string {
 }
 
 export function signin(service = "slack"): string {
-  return `${env.URL}/auth/${service}`;
```
tommoor, line 52: Why changed?

server/routes/auth/providers/google.ts
```diff
@@ -57,37 +58,74 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
+                where: { subdomain: appDomain.teamSubdomain },
+              });
+            }
+
+            if (!team) {
+              throw TeamDomainRequiredError();
+            }
+
+            const user = await User.findOne({
+              where: { teamId: team.id, email: profile.email },
```
tommoor, line 114:
```suggestion
              where: { teamId: team.id, email: profile.email.toLowerCase() },
```

tommoor (changes requested):

tommoor:
The `email.callback` endpoint also takes into account the user being suspended and sending the welcome email:

https://github.com/outline/outline/blob/41e425756d43c361b6890634aa97f0da8322e8ec/server/routes/auth/providers/email.ts#L139

shared/utils/urlHelpers.ts
thenanyu: this url constructor is used in exactly 1 place — the destination for the auth provider buttons. We want it to preserve the domain/subdomain the user landed on, so this url should be relative

server/routes/auth/providers/google.ts
```diff
@@ -57,37 +58,74 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
+            });
+
+            if (!user) {
+              throw InviteRequiredError();
+            }
+
```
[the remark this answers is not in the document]
thenanyu: ah ok, I didn't realize these were explicitly done at login; are they updated anywhere else from normal activity?

thenanyu:
> The email.callback endpoint also takes into account the user being suspended and sending the welcome email:

I didn't add any logic to look for suspended users because I wanted the UX to be closer to the regular auth with google flow, which does auth, logs you in and then you hit the in-app suspension case.

server/routes/auth/providers/google.ts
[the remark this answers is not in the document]
tommoor: They are updated every 5minutes or so, but it does need to happen on login as it's the presence of this timestamp that turns an invite into a user

tommoor (approved):

PR #3659, opened 2022-06-14
feat: added mattermost oauth2 strategy

chloyka:
Hi there. Here is an implementation of Mattermost oauth strategy based on https://www.npmjs.com/package/passport-mattermost-oauth2. 
So now, for those of us, that using Mattermost - it is possible to log in via Mattermost now
Requires few new env variables:
`MATTERMOST_CLIENT_ID`, `MATTERMOST_CLIENT_SECRET`, `MATTERMOST_BACKEND_URL`
`MATTERMOST_BACKEND_URL` is required because Mattermost can be self hosted.


tommoor:
Hi @chloyka – thanks for your contribution. In fact we already support all OpenID compatible OAuth providers, of which Mattermost is one so this additional authentication method isn't needed. You can setup the OIDC provider with the values [found on this page](https://developers.mattermost.com/integrate/admin-guide/admin-oauth2/#oauth-endpoints) – something like:

```
OIDC_CLIENT_ID=
OIDC_CLIENT_SECRET=
OIDC_AUTH_URI=https://mymattermostinstall.com/oauth/authorize
OIDC_TOKEN_URI=https://mymattermostinstall.com/oauth/access_token
OIDC_USERINFO_URI=https://mymattermostinstall.com/api/v4/users/me
OIDC_DISPLAY_NAME=Mattermost
```

Tobix99:
Hey,

I was trying to authenticate via a selfhosted Mattermost instance (created OAuth2 Client, added the env's to Dockerstack), but the Authentication always fails. Are there any prerequires to set in Mattermost other than to enable OAuth2 and create a Client?

Here is a screenshot from the Frontend, after clicking on the Button.
login_fails


And here the Log from the Backend (prettified it myself):
```json
{
  "error": "Failed to obtain access token",
  "stack": "InternalOAuthError: Failed to obtain access token
    at OAuth2Strategy._createOAuthError (/opt/outline/node_modules/passport-oauth2/lib/strategy.js:423:17)
    at /opt/outline/node_modules/passport-oauth2/lib/strategy.js:177:45
    at /opt/outline/node_modules/oauth/lib/oauth2.js:191:18
    at ClientRequest.<anonymous> (/opt/outline/node_modules/oauth/lib/oauth2.js:162:5)
    at ClientRequest.emit (node:events:526:28)
    at ClientRequest.emit (node:domain:475:12)
    at TLSSocket.socketErrorListener (node:_http_client:442:9)
    at TLSSocket.emit (node:events:526:28)
    at TLSSocket.emit (node:domain:475:12)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)",
  "level": "error",
  "message": "Error during authentication"
}
```

tommoor:
Well you definitely need to add some scopes, I suggest searching the discussions for the error: "Failed to obtain access token" – it usually means that the authentication server cannot be accessed by Outline.

Tobix99:
@tommoor I tried it with scopes and without (an only ```openid email```) but it was always the same (and I looked at the Mattermost code and there they only commented, that they should implement scopes :D).


I got it working with a local, self hosted OpenID Server, but not with the Mattermost instance. But I'll check the network tomorrow, maybe I misconfigured something in Mattermost (but I saw the Token and callback traffic in the Chrome Dev Console).

Thanks for your help.

Tobix99:
Ok, so I found the Problem: The selfhosted variant of Mattermost does provide a wrongly formatted Userinfo-Uri. The userinfo is located at ```https://selfhostedmattermost.tld/api/v4/users/me``` and is a JSON provide by its own API, so it's not necessarily compatible with the OIDC Standard. I just created a simple Webserver with provided a valid JSON and this time the Login work (however Just for me as the JSON is hardcoded).

Is there any plan to reevaluate this PR and merge it? Or what should I do when the OAuth Provider is no handling request like it should?

(If the PR should be reevaluted then I would change some things)

tommoor:
You can petition Mattermost to support the OIDC standard or create a PR to add support as it's open source. In general it seems like in most setups Mattermost would be an SSO consumer like Outline, not an authoritative identity provider.

We're not currently adding any new identity providers to Outline

edsky:
mattermost doesn't work, is this merged to support mattermost?

PR #3694, opened 2022-06-28
remove `public` option from avatar & logo upload

### LINKED ISSUE #2769: Outline is incompatible with "Block All Public Access" on the S3 bucket.

cognvn:
Fix #2769.
Image upload from `Settings` area is public by default, remove this option from request body will allow the use of signed urls for avatar and logo when `AWS_S3_ACL = private`

### DIFF
--- a/app/scenes/Settings/components/ImageUpload.tsx
+++ b/app/scenes/Settings/components/ImageUpload.tsx
@@ -66,7 +66,7 @@ class ImageUpload extends React.Component<RootStore & Props> {
       });
       const attachment = await uploadFile(compressed, {
         name: this.file.name,
-        public: true,
+        // public: true,
       });
       this.props.onSuccess(attachment.url);
     } catch (err) {


tommoor:
This is not considered a bug, public permissions are needed to display these images on unauthenticated page.

nrukavkov:
Is it possible to use another public bucket?

PR #3697, opened 2022-06-29
chore: Upgrade socket.io

### LINKED ISSUE #3698: Upgrade socket.io
Socket.io is on v4, we're on v2. There are meaningful security, performance, and reliability updates between the two.

tommoor:
Upgrading from v2 -> v3.

closes #3698


server/services/websockets.ts
```diff
@@ -72,20 +75,13 @@ export default function init(
     Metrics.gaugePerInstance("websockets.count", 0);
   });
 
-  // Forbid all unauthenticated connections
```
tommoor, line 75: Removed as this was relying on internal properties of SocketIO, it's not actually required in our implementation because the user is only ever subscribed to rooms after authentication is complete. Data is not sent over any default namespaces.

server/services/websockets.ts
```diff
@@ -72,20 +75,13 @@ export default function init(
...
-
   io.adapter(
-    socketRedisAdapter({
+    createAdapter({
       pubClient: Redis.defaultClient,
       subClient: Redis.defaultSubscriber,
     })
   );
 
-  io.origins((_req, callback) => {
```
tommoor, line 85: Moved to `cors` configuration as part of migration

server/services/websockets.ts
```diff
@@ -102,18 +98,10 @@ export default function init(
       socket.client.conn.server.clientsCount
     );
 
-    socket.auth = false;
```
tommoor, line 105: Use `socket.client.user` instead

PR #3709, opened 2022-06-29
feat: Custom Length decorator for UTF-8 chars len 

### LINKED ISSUE #3705: Attempting to import a document through `ImportTask` with a title longer than 100 chars results in error
**To Reproduce**
1. Create a document in Notion with a title longer than 100 chars
2. Create export
3. Attempt to import into Outline
4. Error importing.

**Expected behavior**

- The title should be chopped to the maximum allowable characters

CuriousCorrelation:
## Description
Fixes error when importing document through `ImportTask` with a title longer than 100 chars.

Uses same strategy as in

https://github.com/outline/outline/blob/68dd76cfa37cc4559ffc323937456b8afb652889/server/commands/documentImporter.ts#L224

## Issue
Closes #3705

### DIFF
--- a/server/models/Document.ts
+++ b/server/models/Document.ts
@@ -17,7 +17,6 @@ import {
   BelongsTo,
   Column,
   Default,
-  Length,
   PrimaryKey,
   Table,
   BeforeValidate,
@@ -51,6 +50,7 @@ import User from "./User";
 import View from "./View";
 import ParanoidModel from "./base/ParanoidModel";
 import Fix from "./decorators/Fix";
+import { Length } from "./decorators/Length";
 
 export type SearchResponse = {
   results: {
--- a/server/models/decorators/Length.ts
+++ b/server/models/decorators/Length.ts
@@ -1,0 +1,19 @@
+import { size } from "lodash";
+import { addAttributeOptions } from "sequelize-typescript";
+
+/**
+ * A decorator that calculates size of the string based on lodash's size function.
+ * particularly useful for strings with unicode characters of variable lengths.
+ */
+export function Length({ msg, min, max }: { msg?: string; min: number; max: number }): Function {
+    return (target: any, propertyName: string) =>
+        addAttributeOptions(target, propertyName, {
+            validate: {
+                validLength(value: string) {
+                    if (size(value) > max || size(value) < min) {
+                        throw new Error(msg);
+                    }
+                }
+            },
+        });
+}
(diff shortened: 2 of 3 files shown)


tommoor:
That's correct, the base class is an abstract class and as such can't be instantiated. Adding a test to eg the Notion importer would work well 👍

tommoor:
Sorry, bug in GitHub keeps closing PR's when I comment 🙄

CuriousCorrelation:
Haha no worries at all.

---

`ImportNotionTask` uses `documentImporter` to set title

https://github.com/outline/outline/blob/930bfd5391c93352374b85f1409085e1b1a2e6f4/server/queues/tasks/ImportNotionTask.ts#L93

and `documentImporter`

https://github.com/outline/outline/blob/68dd76cfa37cc4559ffc323937456b8afb652889/server/commands/documentImporter.ts#L224

correctly truncates the long title. So I think any tests added to `ImportNotionTask` will pass regardless. Should I add one anyway or is there another way to isolate `ImportTask` for this case? Not 100% sure.

tommoor:
so, looking at it now I don't think this is the correct solution in that case. The argument passed to documentCreator can only come from that documentImporter and should already be truncated. 

Definitely need some tests to figure out what's going on – maybe one place is doing < max_length and another is doing <= max_length.

CuriousCorrelation:
Makes sense and

> maybe one place is doing < max_length and another is doing <= max_length.

seems the likely culprit. 
I'll do some digging and let you know what I find.

tommoor:
> I changed the title of first file in notion-markdown.zip to one with 160 chars and re-ran tests

Same here, it's a weird one – perhaps a character encoding issue or something. Will try and get hold of the original data that caused the error

tommoor:
Okay, it is indeed a UTF-8 character issue. In this case it's the "🛡️" character which lodash truncate is counting as `1` but the database validation is counting as `3`

image

For this particular usecase the lodash version is more "correct" and therefore the bug/problem is in the sequelize validation library.

tommoor:
So the solution is to change `Length` to use a custom validation that measures the string with https://lodash.com/docs/4.17.15#size

CuriousCorrelation:
That was indeed the issue! This patch should work now. 

I changed the commit message to `feat` instead of `fix` and also added a test to specifically check for UTF-8 char length. 

Let me know if there's something that needs a second look.

tommoor:
Just need to run `yarn lint --fix` to reformat that file and fix the build, looks like you did a great job here

CuriousCorrelation:
> Just need to run `yarn lint --fix` to reformat

Oops forgot about that. Latest commit should fix that.

tommoor (approved):

tommoor:
@CuriousCorrelation thanks for taking this on. If you're interested in taking on other issues (and getting paid if you like) hit me up.

CuriousCorrelation:
I would love to, it sounds like an exciting opportunity (with perfect timing)! 

I can drop an email or message via preferred comms channel, just let me know when and where.

PR #3710, opened 2022-06-30
fix: show a distinct error message when a user tries to create an account using a personal gmail

thenanyu:
As discussed in the [slack thread](https://getoutline.slack.com/archives/DKTBDD5EU/p1656075504147239)

New messaging
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ build
 node_modules/*
 .env
 .log
+.vscode/*
 npm-debug.log
 stats.json
 .DS_Store
--- a/app/scenes/Login/Notices.tsx
+++ b/app/scenes/Login/Notices.tsx
@@ -18,6 +18,13 @@ export default function Notices() {
           invite email.
         </NoticeAlert>
       )}
+      {notice === "gmail-account-creation" && (
+        <NoticeAlert>
+          Sorry, accounts cannot be created using a personal Gmail address.
+          <hr />
+          Please use a Google Workspaces account instead.
+        </NoticeAlert>
+      )}
       {notice === "maximum-teams" && (
         <NoticeAlert>
           The team you authenticated with is not authorized on this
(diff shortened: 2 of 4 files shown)


server/routes/auth/providers/google.ts
```diff
@@ -112,6 +117,17 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
             }
 
             if (!team) {
+              // No team usually means this is the apex domain
+              // Throw different errors depending on whether we think the user is
+              // trying to create a new account, or log-in to an existing one
+              const userExists = await User.count({
+                where: { email: profile.email.toLowerCase() },
```
tommoor, line 124: is this purposefully a global find without `teamId`?

app/scenes/Login/Notices.tsx
```diff
@@ -18,6 +18,13 @@ export default function Notices() {
           invite email.
         </NoticeAlert>
       )}
+      {notice === "gmail-account-creation" && (
+        <NoticeAlert>
+          Sorry, a new account cannot be created with a personal Gmail address.
```
tommoor, line 23:
```suggestion
          Sorry, a new team cannot be created with a personal Gmail address.
```

Maybe clearer, we do allow user accounts after all in the case of an invite

server/routes/auth/providers/google.ts
thenanyu: yeah, it's to determine if this is a new email altogether — the case that it's nested inside is most likely coming from the apex domain

app/scenes/Login/Notices.tsx
thenanyu: I think the current language has less of a tendency to confuse. There isn't really a point in the process where a user thinks "I'm creating a team".

They're either signing up or logging in. If they've been invited, it's more of a logging in concept than account creation.

.gitignore
```diff
@@ -3,6 +3,7 @@ build
 node_modules/*
 .env
 .log
+.vscode/*
```
thenanyu, line 6: I added this because new vscode started spitting logfiles into the project. I don't think I messed with any settings — lmk if you know something

tommoor (approved):

PR #3714, opened 2022-07-01
fix: Don't show /share/ link in popover when team sharing disabled

### LINKED ISSUE #3708: Provide a global setting for Share buttons to return normal doc URL instead of creating Share link
## Context
- Instance/account is on getoutline.com account (not self-hosted)
- _Public document sharing_ under Security settings is disabled
- We use Outline as an internal wiki only and do not wish to share docs publicly
- After publishing a document, users click _Share > Copy link_, which is the most intuitive option to obtain a link to share (e.g. in Slack) with colleagues

## Current behavior

- While the minimal view of documents provided via the _Share_ link are suitable for external sharing, for internal use, using _Share_ links results in an experience that lacks important features that are needed for internal use, and is inconsistent with the ordinary browsing experience
- When viewing a share link, users cannot navigate back to the rest of Outline via menus, they become confused
- Additional context such as the document location, authors/editors/views, are unavailable
- No menu pops out to the left, as users would normally expect
- As the Outline/wiki maintainer for my organisation, I find that the _Share_ feature is actively undermining user confidence with Outline and most do not understand why this behaviour is occurring.
- I do not want to discourage users from clicking on the most intuitive option available (the _Share_ button), because doing so would harm efforts to generally encourage contributions to our wiki, and non-technical users may have difficulty understanding that the alternative is to copy the URL from the address bar

outline_share

## Expected behavior

- There should be a toggle in the admin settings which, when enabled, changes the _Share_ button behaviour to return the normal document URL, instead of creating and returning a share link. This option should become available to select when _Public document sharing_ setting is disabled.
- This would require a small front-end change only
- This would allow organisations, who are using Outline only for internal use, to guarantee a consistent experience for their users

tommoor:
fix: Docs appear to be publicly shared when sharing previously enabled

closes #3708

### DIFF
--- a/app/scenes/Document/components/SharePopover.tsx
+++ b/app/scenes/Document/components/SharePopover.tsx
@@ -14,6 +14,7 @@ import Input from "~/components/Input";
 import Notice from "~/components/Notice";
 import Switch from "~/components/Switch";
 import Text from "~/components/Text";
+import useCurrentTeam from "~/hooks/useCurrentTeam";
 import useKeyDown from "~/hooks/useKeyDown";
 import usePolicy from "~/hooks/usePolicy";
 import useStores from "~/hooks/useStores";
@@ -36,8 +37,9 @@ function SharePopover({
   onRequestClose,
   visible,
 }: Props) {
+  const team = useCurrentTeam();
   const { t } = useTranslation();
-  const { shares, auth } = useStores();
+  const { shares } = useStores();
   const { showToast } = useToasts();
   const [isCopied, setIsCopied] = React.useState(false);
   const timeout = React.useRef<ReturnType<typeof setTimeout>>();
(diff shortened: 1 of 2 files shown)


app/scenes/Document/components/SharePopover.tsx
```diff
@@ -47,22 +49,23 @@ function SharePopover({
   const canPublish =
     can.update &&
     !document.isTemplate &&
-    auth.team?.sharing &&
+    team.sharing &&
     documentAbilities.share;
   const isPubliclyShared =
-    (share && share.published) ||
-    (sharedParent && sharedParent.published && !document.isDraft);
+    team.sharing &&
```
tommoor, line 55: I considered adding this to the presenter, but this would mean reloading all of the share records cached in the client when the team setting changes

PR #3729, opened 2022-07-03
fix: Models are not all removed from local store upon access change

tommoor:
This also fixes an issue where if a document is deleted or your access is removed while viewing the document it would remain on screen.

In the process I converted `DataLoader` from a class component with local copies of data to a functional component that relies on the stores. So this is the thing to test in all it's variances. This was required as the class was caching the model.

### DIFF
--- a/app/components/SocketProvider.tsx
+++ b/app/components/SocketProvider.tsx
@@ -6,6 +6,7 @@ import * as React from "react";
 import io from "socket.io-client";
 import RootStore from "~/stores/RootStore";
 import withStores from "~/components/withStores";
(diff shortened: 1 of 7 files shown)


thenanyu:
Everything seems to work for the primary changes — I used 2 separate user accounts to test the following scenarios
- remove permission by removing group membership to a group with access
- remove permission by removing group with access
- remove permission by removing individual access
- I validated that data loader continues to work for shared trees of public documents

thenanyu (approved):

tommoor:
This ended up being a pretty good example of how hooks can be so much simpler then class components.

PR #3731, opened 2022-07-03
Fix: clear localstore after logout

thenanyu:
- fixes a bug that left user, team and policy information in localstorage after logout.
- the problem was that localstore was being cleared but the store itself was not, so `autorun` repopulated the values when `team` or `token` was modified

### DIFF
--- a/app/stores/AuthStore.ts
+++ b/app/stores/AuthStore.ts
@@ -204,6 +204,7 @@ export default class AuthStore {
     runInAction("AuthStore#updateUser", () => {
       this.user = null;
       this.team = null;
+      this.policies = [];
       this.token = null;
     });
   };
@@ -282,7 +283,6 @@ export default class AuthStore {
       setCookie("sessions", JSON.stringify(sessions), {
         domain: getCookieDomain(window.location.hostname),
       });
-      this.team = null;
     }
 
     // clear all credentials from cache (and local storage via autorun)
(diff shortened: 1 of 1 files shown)


app/stores/AuthStore.ts
```diff
@@ -259,14 +260,6 @@ export default class AuthStore {
 
     client.post(`/auth.delete`);
 
-    // remove user and team from localStorage
-    Storage.set(AUTH_STORE, {
-      user: null,
-      team: null,
-      policies: [],
-    });
-    this.token = null;
```
thenanyu, line 268: previously, as soon as `token` was modfied, `autorun` would repopulate local storage

app/stores/AuthStore.ts
```diff
@@ -290,7 +283,12 @@ export default class AuthStore {
...
         domain: getCookieDomain(window.location.hostname),
       });
-      this.team = null;
     }
+
+    // clear all credentials from cache (and local storage via autorun)
+    this.user = null;
+    this.team = null;
+    this.policies = [];
+    this.token = null;
```
thenanyu, line 292: by setting values to null, we don't need to explicitly modify localstorage

app/stores/AuthStore.ts
```diff
@@ -204,6 +204,7 @@ export default class AuthStore {
     runInAction("AuthStore#updateUser", () => {
       this.user = null;
       this.team = null;
+      this.policies = [];
```
thenanyu, line 207: I added this line because it seemed like an oversight to not have it

tommoor (approved):

PR #3732, opened 2022-07-03
feat: merge a new authentication method onto existing user records when emails match

thenanyu:


server/commands/userCreator.ts
```diff
@@ -139,20 +154,22 @@ export default async function userCreator({
...
 
-    const inviter = await invite.$get("invitedBy");
-    if (inviter) {
-      await InviteAcceptedEmail.schedule({
-        to: inviter.email,
-        inviterId: inviter.id,
-        invitedName: invite.name,
-        teamUrl: invite.team.url,
-      });
+    if (isInvite) {
```
tommoor, line 157: It struck me now that we can move a lot of things like this to be triggered from event processors, eg could have a UserCreatedProcessor that triggers this email is the user was invited, thoughts? (not for this PR)

Downside is it makes things a little harder to know what side effects there may be but it seems much more maintainable.

thenanyu: yeah I like the concept in general when scale is required but I think we're a bit far from that point. 
worth considering when the bits of side effect code start to make the main flow of logic hard to read

thenanyu:
I tried to get cute with git and this branch is borked, switching to https://github.com/outline/outline/pull/3739

PR #3739, opened 2022-07-05
Feat: add auth provider to users on sign in

thenanyu:
- when a user signs in via SSO, we look up existing user records on the same team by email and associate the SSO auth with that record if found
- this essentially treats all existing user records as if they were "invites"

### DIFF
--- a/server/commands/userCreator.test.ts
+++ b/server/commands/userCreator.test.ts
@@ -1,3 +1,4 @@
+import { v4 as uuidv4 } from "uuid";
 import { TeamDomain } from "@server/models";
 import { buildUser, buildTeam, buildInvite } from "@server/test/factories";
 import { flushdb, seed } from "@server/test/support";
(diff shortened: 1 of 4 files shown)


server/commands/userCreator.ts
```diff
@@ -89,24 +88,51 @@ export default async function userCreator({
...
   // We have an existing invite for his user, so we need to update it with our
   // new details, link up the authentication method, and count this as a new
   // user creation.
-  if (invite && !invite.authentications.length) {
+  if (existingUser) {
+    // A `user` record might exist in the form of an invite.
+    // In Outline an invite is a shell user record with no authentication method
+    // that's never been active before.
+    const isInvite =
+      !existingUser.lastActiveAt && !existingUser.authentications.length;
```
tommoor, line 111:
```suggestion
      existingUser.isInvited && !existingUser.authentications.length;
```

Hmm, wish I could remember what the `authentications.length` check is for 🤔

server/commands/userCreator.test.ts
```diff
@@ -37,6 +38,43 @@ describe("userCreator", () => {
...
+  it("should add authentication provider to existing user", async () => {
+    const team = await buildTeam({ inviteRequired: true });
+    const teamAuthProviders = await team.$get("authenticationProviders");
+    const authenticationProvider = teamAuthProviders[0];
+
+    const email = "mynam@email.com";
+    const existing = await buildUser({
+      email,
+      teamId: team.id,
+      lastActiveAt: null,
```
tommoor, line 50: By setting `lastActiveAt` to `null` here you're essentially creating an invite, so it's not testing what you're wanting. Should just `buildUser` and let it create an existing authentication too?

server/test/factories.ts
```diff
@@ -179,6 +179,7 @@ export async function buildInvite(overrides: Partial<User> = {}) {
     name: `User ${count}`,
     createdAt: new Date("2018-01-01T00:00:00.000Z"),
     invitedById: actor.id,
+    authentications: [],
```
thenanyu, line 182: it seems correct to blank out authentications to make it as close to a "real" invite as possible

server/commands/userCreator.ts
```diff
@@ -89,48 +88,62 @@ export default async function userCreator({
...
+
+      // Regardless, create a new authentication record
+      // against the existing user (user can auth with multiple SSO providers)
+      // Update user's name and avatar based on the most recently added provider
+      await existingUser.update(
         {
           name,
           avatarUrl,
+          lastActiveAt: new Date(),
+          lastActiveIp: ip,
```
thenanyu, line 139: synchronously update these values here to ensure this path is never taken again for this user

tommoor (approved):
🙏

PR #3741, opened 2022-07-06
feat: scope login attempts to specific subdomains if available - do not switch subdomains

thenanyu:
Functionality
- if a user tries to authenticate from a specific custom domain or team subdomain, the scope of the auth attempt will be limited to that team only.
- even if they have an existing account with their SSO on a different team, we will not consider accounts outside of the team referenced by the url
- if a user authenticates from the apex domain, or if subdomains are not enabled, then it will fall back to the best-effort path, and behave as before

Basic QA test cases
1. **Attempt to auth into a subdomain with an SSO account that exists on a different subdomain. It should fail. Previously, it would have redirected to the other subdomain**
2. Attempt to auth into a subdomain with an SSO account from apex. It should find a team if it exists.
3. Attempt to auth into a subdomain with an SSO that exists on that team. It should succeed.

### DIFF
--- a/server/commands/teamCreator.ts
+++ b/server/commands/teamCreator.ts
@@ -1,6 +1,10 @@
 import { sequelize } from "@server/database/sequelize";
 import env from "@server/env";
-import { DomainNotAllowedError, MaximumTeamsError } from "@server/errors";
+import {
+  InvalidAuthenticationError,
+  DomainNotAllowedError,
+  MaximumTeamsError,
+} from "@server/errors";
 import Logger from "@server/logging/Logger";
 import { APM } from "@server/logging/tracing";
 import { Team, AuthenticationProvider, Event } from "@server/models";
@@ -55,6 +59,11 @@ async function teamCreator({
       team: authP.team,
       isNewTeam: false,
     };
+  }
+  // A team id was provided but no auth provider was found matching those credentials
+  // The user is attempting to log into a team with an incorrect SSO - fail the login
+  else if (id) {
+    throw InvalidAuthenticationError("incorrect authentication credentials");
   }
 
   // This team has never been seen before, if self hosted the logic is different
--- a/server/commands/teamCreator.ts
+++ b/server/commands/teamCreator.ts
@@ -13,6 +13,7 @@ type TeamCreatorResult = {
 };
 
 type Props = {
+  id?: string;
   name: string;
   domain?: string;
   subdomain: string;
@@ -25,6 +26,7 @@ type Props = {
 };
 
 async function teamCreator({
+  id,
   name,
   domain,
   subdomain,
@@ -33,7 +35,9 @@ async function teamCreator({
   ip,
 }: Props): Promise<TeamCreatorResult> {
   let authP = await AuthenticationProvider.findOne({
-    where: authenticationProvider,
+    where: id
+      ? { ...authenticationProvider, teamId: id }
+      : authenticationProvider,
     include: [
       {
         model: Team,
--- a/server/commands/userCreator.ts
+++ b/server/commands/userCreator.ts
@@ -40,7 +40,6 @@ export default async function userCreator({
 }: Props): Promise<UserCreatorResult> {
   const { authenticationProviderId, providerId, ...rest } = authentication;
 
-  // TODO: this needs otr be scoped to the righ tteam!!!!
   const auth = await UserAuthentication.findOne({
     where: {
       providerId,
@@ -49,6 +48,8 @@ export default async function userCreator({
       {
         model: User,
         as: "user",
+        where: { teamId },
+        required: true,
       },
     ],
   });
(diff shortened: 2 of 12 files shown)


server/routes/auth/providers/google.ts
```diff
@@ -63,27 +63,48 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
             // the query parameter.
             const avatarUrl = profile.picture.replace("=s96-c", "=s128-c");
 
+            // if a team can be inferred, we assume the user is only interested in signing into
+            // that team in particular; otherwise, we will do a best effort at finding their account
+            // or provisioning a new one (within AccountProvisioner)
             result = await accountProvisioner({
               ip: req.ip,
               team: {
+                id: team?.id,
```
thenanyu, line 107: 🥦 team id is passed into the account provisioner if available, and will be respected as the only team which the command chain will consider for login/signup

server/commands/teamCreator.ts
```diff
@@ -33,7 +35,9 @@ async function teamCreator({
   ip,
 }: Props): Promise<TeamCreatorResult> {
   let authP = await AuthenticationProvider.findOne({
-    where: authenticationProvider,
+    where: id
+      ? { ...authenticationProvider, teamId: id }
+      : authenticationProvider,
```
thenanyu, line 40: 🥦 2/ do a best effort here to find the right team. If a `teamId` is provided, we're interpreting that as the only team we're interested in.

server/commands/userCreator.ts
```diff
@@ -48,6 +48,8 @@ export default async function userCreator({
       {
         model: User,
         as: "user",
+        where: { teamId },
+        required: true,
```
thenanyu, line 52: 🥦 3/ only consider users from the target `teamId`. This opens the door to having the same external `providerId` on multiple users on multipole teams

server/commands/teamCreator.ts
```diff
@@ -33,7 +35,9 @@ async function teamCreator({
   ip,
 }: Props): Promise<TeamCreatorResult> {
   let authP = await AuthenticationProvider.findOne({
-    where: authenticationProvider,
+    where: id
+      ? { ...authenticationProvider, teamId: id }
```
tommoor, line 39: Note there is currently no index on `teamId` – need to add this now we're querying it

server/routes/auth/providers/oidc.ts
```diff
@@ -85,10 +90,13 @@ if (env.OIDC_CLIENT_ID && env.OIDC_CLIENT_SECRET) {
             throw OIDCMalformedUserInfoError();
           }
 
-          const subdomain = domain.split(".")[0];
+          // remove the TLD and form a subdomain from the remaining
+          const subdomain = domain.split(".").slice(0, -1).join("-");
```
tommoor, line 94: Wanna move this to a utils method too? Seems like it changed in a couple of spots

server/routes/auth/providers/oidc.ts
```diff
@@ -1,5 +1,5 @@
 import passport from "@outlinewiki/koa-passport";
-import { Request } from "koa";
+import type { Request } from "express";
```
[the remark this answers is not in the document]
tommoor: Yea, ideally we need to overwrite the type at the global level – it's definitely _not_ an Express request, express doesn't even exist in the lockfile – just it's types.

Then the question is why isn't `koa.Request` including the cookies property if it's actually there

tommoor (commented):
This is great

server/commands/teamCreator.ts
thenanyu: I don't think performance is any worse than before — the addition of teamId should be applied after the query planner makes use of existing indices

tommoor: I suppose it still has the providerId, ok 👍

server/errors.ts
```diff
@@ -1,16 +1,25 @@
 import httpErrors from "http-errors";
-import env from "./env";
 
 export function AuthenticationError(
-  message = "Invalid authentication",
-  redirectUrl = env.URL
+  message = "Authentication required",
+  redirectUrl = "/"
```
thenanyu, line 5: changed these to be relative to preserve subdomain and custom domains

server/routes/auth/providers/azure.ts
```diff
@@ -36,7 +40,7 @@ if (env.AZURE_CLIENT_ID && env.AZURE_CLIENT_SECRET) {
       scope: scopes,
     },
     async function (
-      req: Request,
+      req: Context,
```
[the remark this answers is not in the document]
tommoor: Ultimately I guess we should rename `req` to `ctx` to clarify then

server/routes/auth/providers/azure.ts
```diff
@@ -36,7 +40,7 @@ if (env.AZURE_CLIENT_ID && env.AZURE_CLIENT_SECRET) {
       scope: scopes,
     },
     async function (
-      req: Request,
+      ctx: Context,
```
thenanyu, line 43: changed referenced from "request" to "context"

tommoor (approved):

server/commands/teamCreator.ts
smotraghi-shopify: Sounds good - will do

PR #3744, opened 2022-07-07
feat: Add pending migrations check during startup

### LINKED ISSUE #3704: Prevent server startup if migrations are outstanding
### Requirements

Upon startup read the migrations table and see if the latest applied migration is < available migrations. Sequelize/uzmug may provide tooling / API's for this, I haven't yet researched.

If there are migrations that need to be applied then exit with an error code and explanatory message. This prevents weird errors due to missing database columns that later get reported as "bugs".

CuriousCorrelation:
# Description

PR prevents server startup if migrations are outstanding.

## Example

``` shell
[api] info: [email] SMTP_USERNAME not provided, generating test account...
[api] Running Outline in development mode. To run Outline in production [omit]
[api] You have pending migrations
[api] 1. 20220702132722-add-webhooks-deleted-at.js
[api] Please run `yarn sequelize db:migrate` to run all pending migrations
[api] node build/server/index.js --services=collaboration,websockets, [omit]
```

### Applying fix suggested

``` shell
❯ yarn sequelize db:migrate

Sequelize CLI [Node: 16.15.1, CLI: 6.4.1, ORM: 6.20.1]

Loaded configuration file "server/config/database.json".
Using environment "development".
== 20220702132722-add-webhooks-deleted-at: migrating =======
== 20220702132722-add-webhooks-deleted-at: migrated (0.012s)

Done in 0.49s
```

### Successful rerun

``` shell
[api] info: [email] SMTP_USERNAME not provided, generating test account...
[api] info: [lifecycle] Starting collaboration service
[api] info: [lifecycle] Starting websockets service
[api] info: [lifecycle] Starting admin service
[api] info: [lifecycle] Starting web service
[api] info: [lifecycle] Starting worker service
```

### Considerations

`checkPendingMigrations` uses synchronous API unlike its neighbors in `utils/startup.ts`.

This isn't exactly a programmatic approach but since we are using

https://github.com/outline/outline/blob/c6fdffba774086a4a10d9483c4e79abf81b17ed9/Makefile#L4

to run migrations, I think using a programmatic way to detect pending migrations
might diverge similar interfaces, potentially increasing friction. This also avoids
adding additional dependencies.

We could consider side-grading both to programmatic approach in the future.

## Issue

Closes #3704

### DIFF
--- a/server/utils/startup.ts
+++ b/server/utils/startup.ts
@@ -1,8 +1,34 @@
+import { execSync } from "child_process";
 import chalk from "chalk";
 import env from "@server/env";
 import Logger from "@server/logging/Logger";
 import AuthenticationProvider from "@server/models/AuthenticationProvider";
 import Team from "@server/models/Team";
+
+export function checkPendingMigrations() {
+  const commandResult = execSync("yarn sequelize db:migrate:status");
+  const commandResultArray = Buffer.from(commandResult)
+    .toString("utf-8")
+    .split("\n");
+
+  const pendingMigrations = commandResultArray.filter((line) =>
+    line.startsWith("down")
+  );
+
+  if (pendingMigrations.length) {
+    Logger.warn("You have pending migrations");
+    Logger.warn(
+      pendingMigrations
+        .map((line, i) => `${i + 1}. ${line.replace("down ", "")}`)
+        .join("\n")
+    );
+    Logger.warn(
+      "Please run `yarn sequelize db:migrate` to run all pending migrations"
+    );
+
+    process.exit(1);
+  }
+}
 
 export async function checkMigrations() {
   if (env.DEPLOYMENT === "hosted") {
--- a/server/utils/startup.ts
+++ b/server/utils/startup.ts
@@ -23,7 +23,7 @@ export function checkPendingMigrations() {
         .join("\n")
     );
     Logger.warn(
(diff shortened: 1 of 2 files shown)


server/utils/startup.ts
```diff
@@ -1,9 +1,35 @@
...
+  if (pendingMigrations.length) {
+    Logger.warn("You have pending migrations");
+    Logger.warn(
+      pendingMigrations
+        .map((line, i) => `${i + 1}. ${line.replace("down ", "")}`)
+        .join("\n")
+    );
+    Logger.warn(
+      "Please run `yarn sequelize db:migrate` to run all pending migrations"
+    );
```
tommoor, line 27:
```suggestion
    Logger.warn(
      "Please run `yarn db:migrate` or `yarn db:migrate --env production-ssl-disabled` to run all pending migrations"
    );
```

tommoor: Would be great if we could detect which one of these to print, but it's probably getting too clever

tommoor (commented):
Looks good other than this message tweak 🙏

server/utils/startup.ts
CuriousCorrelation: Could be possible. I'll let you know if I find a simple(r) solution. For the time being we could put this in a separate `feat` issue.

tommoor (approved):

PR #3760, opened 2022-07-09
feat: Redirect on unpublished share access

### LINKED ISSUE #3745: Visiting an unpublished share link should show login screen
image

### Problem
Currently we just show "Not found" page which is a bit of a rude dead end.

### Solution
It would be better to show a login screen and then once the user has logged in/created an account we can forward them back to the original document they wanted to visit. We already do this for non-share links.

CuriousCorrelation:
# Description

Unauthorized / Not logged in users should be redirected to login page. Login page should prompt user to login to colab on this document. 

# Issue 

Closes #3745

### DIFF
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -2,7 +2,7 @@ import { Location } from "history";
 import { observer } from "mobx-react";
 import * as React from "react";
 import { Helmet } from "react-helmet";
-import { RouteComponentProps, useLocation } from "react-router-dom";
+import { Redirect, RouteComponentProps, useLocation } from "react-router-dom";
 import { useTheme } from "styled-components";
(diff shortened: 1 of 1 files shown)


CuriousCorrelation:
@tommoor Right now this does close the issue but I would like to also implement that text prompt [you mentioned](https://github.com/outline/outline/issues/3745#issuecomment-1179499609) (if that's fine, although we could also create a separate issue). 

If so any recommendations on how to pass param to said text element in `Login`? 

My first thought would be to add a "?team=example" param to unpublished share links and parse those via `useQuery`.

tommoor (approved):

tommoor:
Because `Login` loads the config information that's needed for the text prompt the only way to make that work from the outside is to have a prop that takes a functional child and gets passed the config. I can do this after as it'll only take me a few minutes.

PR #3774, opened 2022-07-12
fix: `svg+xml` image type ext not assigned properly

### LINKED ISSUE #3768: Downloading an .svg downloads it as .svg+xml
**To Reproduce**
1. Create page
2. Upload .svg image
3. Download the .svg image
4. Downloaded file ends with .svg+xml

**Expected behavior**
Downloaded file should end with .svg

**Screenshots**
Bildschirmfoto 2022-07-11 um 11 39 29

- Install: getoutline.com
- Version: N/A

 - OS: mac os
 - Browse: chrome
 - Version 103.0.5060.114

CuriousCorrelation:
# Description

`svg+xml` image type wasn't being set properly because

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/shared/editor/nodes/Image.tsx#L122

which uses 

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/shared/utils/files.ts#L26

particularly 

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/shared/utils/files.ts#L36

would apply like 

``` js
"image/svg+xml".split('/')[1]

"svg+xml"
```

PR does this

``` js
"image/svg+xml".split('/')[1].split('+')[0]
"svg"
"image/jpeg".split('/')[1].split('+')[0]
"jpeg"
```

Download link is set here

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/shared/editor/nodes/Image.tsx#L128

# Issue

Closes #3768

### DIFF
--- a/shared/editor/nodes/Image.tsx
+++ b/shared/editor/nodes/Image.tsx
@@ -119,7 +119,7 @@ const downloadImageNode = async (node: ProsemirrorNode) => {
   const image = await fetch(node.attrs.src);
   const imageBlob = await image.blob();
   const imageURL = URL.createObjectURL(imageBlob);
-  const extension = imageBlob.type.split("/")[1];
+  const extension = imageBlob.type.split("/")[1].split("+")[0];
   const potentialName = node.attrs.alt || "image";
 
   // create a temporary link node and click it with our image data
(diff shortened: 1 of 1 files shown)


tommoor:
This is the right idea but let’s just change the first split to a regex that matches both characters?

CuriousCorrelation:
Does this look good? - https://regexr.com/6pj77
"Starting from but excluding" gets a bit complex with negative look-ahead but this should work.

tommoor:
Seems overly complex, why not use `\/|\+`

tommoor (approved):

Issue #3742, opened 2022-07-06
unable to create a link on inline code text


domanchi:
Outline refuses to allow me to create a hyperlink on inline text (denoted by single back-ticks in markdown, `like this`). However, if the highlighted text includes any other non-whitespace letter that isn't part of the inline text, it works.

**To Reproduce**

1. Create a new document.
2. Type in the following example:

    ```
    inline `text`.
    ```
3. Highlight the `text` snippet, and notice the toolbar pop up. Try clicking the "Create Link" button, and observe nothing happens.
4. Highlight the `text.` snippet (notice the trailing period). Try creating a link, and see that it succeeds.
5. Highlight the ` text` snippet (notice the prefixed whitespace). Try creating a link, and see observe nothing happens as well.
6. Highlight the entire example, and create a link. This works.

For all above steps, instead of clicking the button, you can also press the shortcut (cmd+k on a mac), and it will also not do anything.

**Expected behavior**

Inline code should be able to be hyperlinked (preferably, like the version SHA below).

**Outline (please complete the following information):**
- Install: Self-hosted
- Version: v0.64.3 ([`8a29a3523aecc778c31f27a28cf47cb1cda200ec`](https://github.com/outline/outline/tree/8a29a3523aecc778c31f27a28cf47cb1cda200ec))

**Desktop (please complete the following information):**
 - OS: Windows / MacOSX
 - Browser: Chrome
 - Version: 102 / 103

tommoor:
Hi, this is by design – although the _option_ to add formatting to inline code shouldn't be shown ideally, so there is definitely a bug there.

PR #3775, opened 2022-07-12
fix: ToolbarMenu popup on inline code selection

CuriousCorrelation:
# Description

This 

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/app/editor/components/SelectionToolbar.tsx#L184

combined with

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/app/editor/components/SelectionToolbar.tsx#L188-L190

takes care of code blocks like

```


```

PR adds check for `inline` code

| Before  | After |
| ------------- | ------------- |
|   Screenshot 2022-07-12 at 22-03-42 test - outline-dev-test                |          Screenshot 2022-07-12 at 22-17-04 test - outline-dev-test       |

# Issue

Mentioned in  #3742

### DIFF
--- a/app/editor/components/SelectionToolbar.tsx
+++ b/app/editor/components/SelectionToolbar.tsx
@@ -1,4 +1,5 @@
 import { some } from "lodash";
+import CodeIcon from "outline-icons/lib/components/CodeIcon";
 import { NodeSelection, TextSelection } from "prosemirror-state";
 import { CellSelection } from "prosemirror-tables";
 import { EditorView } from "prosemirror-view";
@@ -9,7 +10,6 @@ import filterExcessSeparators from "@shared/editor/lib/filterExcessSeparators";
 import getColumnIndex from "@shared/editor/queries/getColumnIndex";
 import getMarkRange from "@shared/editor/queries/getMarkRange";
 import getRowIndex from "@shared/editor/queries/getRowIndex";
-import isInCode from "@shared/editor/queries/isInCode";
 import isMarkActive from "@shared/editor/queries/isMarkActive";
 import isNodeActive from "@shared/editor/queries/isNodeActive";
 import { MenuItem } from "@shared/editor/types";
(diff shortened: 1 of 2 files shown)


CuriousCorrelation:
Works perfectly!

I see it also checks for `code_block`, so we could potentially cover for both `codeblock_inline` and

https://github.com/outline/outline/blob/e92d68a0a3328227daabea0d2db2db8f3efbb6d7/app/editor/components/SelectionToolbar.tsx#L184

with

``` jsx
  187 │ │ // toolbar is disabled in code blocks, no bold / italic etc
= 188 │ │ if (isInCode(state)) {
  189 │ │ │ return null;
  190 │ │ }
  191
```

tommoor:
Yes, this is a better approach – there is an existing system for this though that doesn't require the special case – you can set the `visible` prop on the menu item appropriately.

https://github.com/outline/outline/blob/main/app/editor/menus/formatting.ts#L39

So all of these would have `visible=false` when inside code.

CuriousCorrelation:
Should be good to go! Let me know if there's anything else that could use a second look

app/editor/components/SelectionToolbar.tsx
```diff
@@ -229,6 +232,12 @@ export default class SelectionToolbar extends React.Component<Props> {
       return null;
     }
 
+    // Only "code_inline" option should be visible if
+    // it's an inline code selection.
+    if (isCodeInlineSelection) {
+      items.map((item) => (item.visible = item.name === "code_inline"));
```
tommoor, line 238: Sorry, maybe I could have been clearer – `getFormattingMenuItems` should return the correct menu items given the state, we don't want to have additional processing here as well. The `visible` prop should be modified within that method

CuriousCorrelation: You definitely did mentioned what you were looking for

> **all of these** [items] would have **visible=false when inside code**

so it's totally my bad. Sorry.

Also

>  actually a query for this, `isInCode(state)`

and 

https://github.com/outline/outline/blob/4c05fe422c73ebd18526c37c74fce8bbdf726f1c/app/editor/menus/formatting.ts#L29-L30

should have been good hints

Third times the charm!

tommoor (approved):

PR #3780, opened 2022-07-13
chore: Remove unmaintained `socket.io-auth`

tommoor:
Related #3698

### DIFF
--- a/server/services/websockets.ts
+++ b/server/services/websockets.ts
@@ -2,17 +2,24 @@ import http, { IncomingMessage } from "http";
 import { Duplex } from "stream";
 import invariant from "invariant";
 import Koa from "koa";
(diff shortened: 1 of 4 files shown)


server/services/websockets.ts
```diff
@@ -85,195 +95,246 @@ export default function init(
...
+      }
+    });
+
+    socket.on("disconnect", async () => {
       Metrics.increment("websockets.disconnected");
       Metrics.gaugePerInstance(
         "websockets.count",
         socket.client.conn.server.clientsCount
       );
+      await Redis.defaultClient.hdel(socket.id, "userId");
```
tommoor, line 133: fix: This wasn't cleaned up before

PR #3785, opened 2022-07-14
Feature: Sync groups from OIDC provider

Frando:
This PR implements an optional feature to sync groups from an OIDC provider. I'm using this with Keycloak, which supports a "groups" claim that contains an array of group names.

When setting the `OIDC_GROUPS_CLAIM` environment variable, then on each login, this claim will be interpreted as a list of group names. Missing groups will be created. The user's group membership will be updated (discarding old memberships).

This means that groups can be fully maintained upstream in an SSO/OpenID Connect provider, and changes will be reflected into Outline on each authentication request.

See [this discussion](https://github.com/outline/outline/discussions/3779) for details on how we want to use this.

This might not be finished and I only tested this with Keycloak, but wanted to share early and maybe get feedback.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -96,6 +96,11 @@ OIDC_USERINFO_URI=
 # Supports any valid JSON path with the JWT payload
 OIDC_USERNAME_CLAIM=preferred_username
 
+# Specify a claim that contains an array of group names.
+# If set, these groups will be created an the user's group
+# membership will be updated on each login.
+OIDC_GROUPS_CLAIM=
+
 # Display name for OIDC authentication
 OIDC_DISPLAY_NAME=OpenID
 
--- a/server/commands/accountProvisioner.ts
+++ b/server/commands/accountProvisioner.ts
@@ -28,7 +28,7 @@ type Props = {
     subdomain: string;
     avatarUrl?: string | null;
   };
-  groups: string[] | undefined;
+  groups?: string[];
   authenticationProvider: {
     name: string;
     providerId: string;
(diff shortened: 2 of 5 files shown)


almereyda:
@Frando I still would like to test you branch, but would not want to downgrade where I do so. Could you rebase on `main` again, please?

@tommoor Your approval is required for the PR workflows here.

Also I am interested, would you welcome such a PR in general, and/or/if under which circumstances?

tommoor:
Group mapping is certainly a desirable feature, but I think the flexible implementation would not look much like this - we have to think in terms of solutions that work across authentication methods.

One successful way I've seen of doing this is to define rules which allow mapping of auth properties to groups, this then allows fun possibilities such as mapping email address domains or other attributes to specific groups too. 

Of course that would need more designing and planning.

AlessandroSpallina:
Waiting for this feature!

T2brozz:
Hey, are there any updates on this @tommoor ?

Frando:
Note to readers: I chose a different path for now and wrote a small webhook server with Deno. It subscribes to `users.signin` events from Outline, and syncs up their groups from Keycloak via the Outline and Keycloak REST APIs. It works well.

I just open-sourced it here: https://gist.github.com/Frando/aa561ca7e6c72ab64b5d17df911c0b1f
Feel invited to try it out and report feedback here or on the gist.

mikew:
@Frando Thanks for that! I used it as the inspiration to build similar for Authelia: https://github.com/mikew/homelab/tree/public/services/notes-outline/data/outline-authelia-users-yml-bridge

Altavion:
@mikew Right there with you! Inspired by Frando’s approach, I whipped up one for Authentik:
https://gist.github.com/Altavion/45565e0635d20fa3860f3bbe9969b4bc

And there’s even more:
https://github.com/burritosoftware/Outline-Authentik-Connector

Feels like it’s time for Outline to embrace some community-driven PRs. What do you say?

PR #3786, opened 2022-07-14
fix: Cursor disappearing behind emoji

### LINKED ISSUE #3603: [editor] can't get cursor after an emoji on firefox
I get a bug with Firefox when I add an emoji, the cursor does not appear or get positioned after it

**To Reproduce**
1. create or edit a document
2. create a line, insert some test then insert an emoji (I use `:cat` for instance) 
3. your cursor reappear after the emoji and even if you click after it on the line to position it, it never appears. 

If I put an emoji betweem 2 existing words, I don't have this issue.
I've tested with Epiphany which is webkit based and it works fine.

**Expected behavior**
the cursor to appear

**Screenshots**
If applicable, add screenshots or videos to help explain your problem.

- Install: self hosted
- Version: `0.64.2`

 - OS: Linux
 - Browser: Firefox
 - Version: 100

CuriousCorrelation:
# Description

This took _surprisingly long_ to debug, had to go down a very deep and convoluted rabbit hole but finally figured it out!

---

Here's a clip describing the issue (Two parts, bug then solution)

https://user-images.githubusercontent.com/58817502/179033019-18e27163-47a7-4238-967d-3a3da6f312a9.mp4

---

Finding this disappeared cursor becomes a guessing game if there are more than a few emojis in the document. Only way to fix it was clicking around a lot until you see it pop up somewhere.

---

Here are some reference that should explain why that happens.

ProseMirror issue 
- 1241
- 1069
- 472

[Source](https://github.com/ProseMirror/prosemirror-view/blob/089aa9bd8f261d8181782bc5f8c46c5dc5e11a5f/src/viewdesc.js#L1271)
[Solution hint](https://github.com/Patternslib/pat-tiptap/pull/8/files)
[About setting default](https://discuss.prosemirror.net/t/what-is-the-best-way-just-add-some-css-class-attribute-to-specific-node/2148/4)

# Issue
Closes #3603

Issue mentions Firefox but I was able to replicate it on Edge and Chromium.

### DIFF
--- a/app/editor/components/Styles.ts
+++ b/app/editor/components/Styles.ts
@@ -325,6 +325,20 @@ const EditorStyles = styled.div<{
     }
   }
 
+  /* Prosmirror adds a 'br' tag with class 'ProseMirror-trailingBreak'
+   * to inlined content if they are a leaf node.
+   * It makes sure empty paragraphs take up space and
+   * solves some edge case cursor positioning issues
+   * at the end of a block that in a break or non-editable node.
+   * In this case that works adversarially because of
+   * the way span elements take up space.
+   * This fixes this problem by essentially
+   * 'emulating' a 'span' with a 'div' element.
+   */
+  .emoji {
+    display: inline-block;
+  }
+
   .with-emoji {
     margin-${(props) => (props.rtl ? "right" : "left")}: -1em;
   }
--- a/app/editor/components/Styles.ts
+++ b/app/editor/components/Styles.ts
@@ -325,20 +325,6 @@ const EditorStyles = styled.div<{
     }
   }
 
-  /* Prosmirror adds a 'br' tag with class 'ProseMirror-trailingBreak'
-   * to inlined content if they are a leaf node.
-   * It makes sure empty paragraphs take up space and
-   * solves some edge case cursor positioning issues
-   * at the end of a block, especially in a break or non-editable node.
-   * In this case that works adversarially because of
-   * the way span elements take up space.
-   * This fixes the problem by essentially
-   * 'emulating' a 'span' with a 'div' element.
-   */
-  .emoji {
-    display: inline-block;
-  }
-
   .with-emoji {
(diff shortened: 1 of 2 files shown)


CuriousCorrelation:
That's the first thing I tried but somehow it didn't work (why I went on search for setting defaults - link in PR desc).

But I went to double check just to make sure and... it worked!  :monocle_face:

Thanks for the advice! 

Note to self - always triple check

CuriousCorrelation:
I also added a comment describing the issue to the style so nobody may mistakenly change it in the future.

tommoor:
Unfortunately changing this to a block element breaks copy/paste behavior 😞 which is a worse situation than the disappearing cursor.

CuriousCorrelation:
Let me know your thoughts on this latest patch. Everything should works as expected, plus no need to add styling!

tommoor:
Works well, although I've never been able to reproduce the original issue so I'm not sure how confident I am that this changes anything about the cursor behavior?

CuriousCorrelation:
Admittedly it is an edge case.

The easiest way I was able to reproduce this was (on [getoutline.com](https://www.getoutline.com)) by starting out with a blank page, start typing an emoji, say `:smi...`, pressing return to autocomplete.

Notice that the cursor disappears. This doesn't happen when you type regular text.

On this branch the cursor appears right after the emoji.

Best way to ensure that it does indeed work is by using inspect element and checking out styles applied to 
`<br class="ProseMirror-trailingBreak">`. It should be right below `span` tag that contains the emoji.

CuriousCorrelation:
Also please feel free to close this PR if you judge so. 

As I mentioned above, this is an edge case, mostly avoided by typing something after the emoji. Disappearing cursor is an issue if emoji is at the very end or very start of a line/paragraph.

We can always revisit the branch if we need to :+1:

tommoor (approved):
This is a good fix

PR #3791, opened 2022-07-15
feat: Add button to clear code blocks

### LINKED ISSUE #3471: Multiline code formatting is difficult to remove or to add normal formatting in the middle of it
Once you have a multi-line (or full one-line) code section defined it is difficult to format it differently or even to remove it.

**To Reproduce**
1. Create a multi-line code section (I do that with ```) 
```
Line 1
Line 2
```
2. Try to insert a line with normal formatting between Line 1 and Line 2
* I've tried to create several empty lines to create a break => did not work
* I've tried to insert ``` to break manually the code formatting => did not work
* I've tried to remove entirely the code formatting => did not find how
3. Eventually I deleted all the text and created again the text section **without** code formatting and finally added code formatting at the very end

**Expected behavior**
An easy to remove code formatting on a text selection

**Screenshots**
n/a

- Install: self-hosted
- Version: 0.63.0

 - OS: Linux
 - Browser: Firefox
 - Version: 99

CuriousCorrelation:
# Description

Issue #3471 describes the problem adequately. 

The gist is, unlike `inline_code` which you can clear by selecting the range and clicking `Code` option from the pop-up menu, code blocks like these

``` rust
#[derive(Debug, Clone, PartialEq)]
pub enum Role {
    Admin,
    User,
    Commander,
}
```
are quite difficult to edit once made.

One option would be to copy the content onto the clipboard via `Copy` button on the top left corner and paste the text content in a new paragraph (doesn't preserve formatting), delete the previous code block and create a new one with the contents on the clipboard. 

PR adds a `Clear` button that will change code block to paragraph (both block type nodes), thus preserving formatting and allowing users to quickly make edits.
Screenshot

# Issue
Closes #3471

### DIFF
--- a/app/hooks/useDictionary.ts
+++ b/app/hooks/useDictionary.ts
@@ -15,6 +15,7 @@ export default function useDictionary() {
       alignRight: t("Align right"),
       bulletList: t("Bulleted list"),
       checkboxList: t("Todo list"),
+      clear: t("Clear"),
       codeBlock: t("Code block"),
       codeCopied: t("Copied to clipboard"),
       codeInline: t("Code"),
--- a/shared/editor/nodes/CodeFence.ts
+++ b/shared/editor/nodes/CodeFence.ts
@@ -114,10 +114,15 @@ export default class CodeFence extends Node {
         },
       ],
       toDOM: (node) => {
-        const button = document.createElement("button");
-        button.innerText = this.options.dictionary.copy;
-        button.type = "button";
-        button.addEventListener("click", this.handleCopyToClipboard);
+        const copyButton = document.createElement("button");
+        copyButton.innerText = this.options.dictionary.copy;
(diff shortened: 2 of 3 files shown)


CuriousCorrelation:
I think putting button at the bottom right corner might be slightly better. Could prevent users from accidentally clicking clear. Although that should be easy enough to recover from.

Screenshot

tommoor:
I'm not really a fan of either of these options unfortunately, it is likely a _very_ rare circumstance that you would want to turn an existing code block into a paragraph.

If we did want this functionality then it should probably be implemented in a manner that can be extended for other block node types and not be specific to code, hope that makes sense. I'm going to close this out for that reason

CuriousCorrelation:
Thanks for the review. That makes sense, I can see it's limited extensibility. I'll look into a more general approach. 👍

PR #3798, opened 2022-07-16
feat(editor): Ability to select line in codeblocks

### LINKED ISSUE #3304: Triple click in code block should highlight line
- 1 click place cursor
- 2 click highlight word
- 3 click highlight line
- 4 click highlight entire block

CuriousCorrelation:
# Description

PR adds ability to select single line by triple clicking in code blocks. Previously a single click will place cursor, two will select a word and three will select entire block. 4th click will then deselect everything and place cursor.

Now single click will display cursor inside the block, double will highlight a word.

4 consecutive clicks will select the entire block.

## Note

These clicks should be roughly even, 3 quick ones and 1 slow will select a line and deselect it at final click.

# Issue 
Closes  #3304

### DIFF
--- a/shared/editor/nodes/CodeFence.ts
+++ b/shared/editor/nodes/CodeFence.ts
@@ -12,6 +12,8 @@ import {
   Selection,
   TextSelection,
   Transaction,
+  Plugin,
+  PluginKey,
 } from "prosemirror-state";
 import refractor from "refractor/core";
 import bash from "refractor/lang/bash";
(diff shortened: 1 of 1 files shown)


shared/editor/nodes/CodeFence.ts
```diff
@@ -283,7 +285,23 @@ export default class CodeFence extends Node {
...
-    return [Prism({ name: this.name }), Mermaid({ name: this.name })];
+    return [
+      Prism({ name: this.name }),
+      Mermaid({ name: this.name }),
+      new Plugin({
+        key: new PluginKey("triple-click"),
+        props: {
+          handleDOMEvents: {
+            mousedown(view, event) {
+              const {
```
tommoor, line 296: This is great, works really well – these plugin dom event handlers are actually document-wide in scope though so it's probably best to check that the current selection is in a code block here.

CuriousCorrelation: That's awesome! 

I was looking for a way to make sure selection is in code block, then I remembered you [mentioned that before](https://github.com/outline/outline/pull/3775#issuecomment-1182151456), and voila 

https://github.com/outline/outline/blob/9afda0a97125fa29aaa1301bc54363797a1b2a49/shared/editor/nodes/CodeFence.ts#L184 

there it was!

tommoor (approved):
Perf

PR #3813, opened 2022-07-19
feat: allow external SSO methods to log into teams as long as emails match

thenanyu:
removes special case email matching logic on google auth method, and incorporates it into the natural codepaths of team creator and user creator


server/routes/auth/providers/google.ts
```diff
@@ -63,87 +62,67 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
-              },
-              authentication: {
-                providerId: profile.id,
-                accessToken,
-                refreshToken,
-                expiresIn: params.expires_in,
-                scopes,
-              },
-            });
-          } else {
```
thenanyu, line 113: 🥦  We're removing the special personal gmail logic (except for error cases) and baking it into the functionality of the command functions.

server/routes/auth/providers/google.ts
```diff
@@ -63,87 +62,67 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
-              }
-
-              throw TeamDomainRequiredError();
-            }
+          console.log("Google sign-in", { domain, profile });
 
-            const user = await User.findOne({
-              where: { teamId: team.id, email: profile.email.toLowerCase() },
+          // No profile domain means personal gmail account
+          // No team implies the request came from the apex domain
```
thenanyu, line 72: We still need to handle the situations where we strictly forbid personal gmail auth (no team specified in the URL)

server/routes/auth/providers/google.ts
```diff
@@ -63,87 +62,67 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
...
-                where: { email: profile.email.toLowerCase() },
-              });
-
-              if (!userExists) {
-                throw GmailAccountCreationError();
-              }
-
-              throw TeamDomainRequiredError();
-            }
+          console.log("Google sign-in", { domain, profile });
```
thenanyu, line 69: hanging... will remove

server/commands/userCreator.ts
```diff
@@ -1,12 +1,18 @@
...
+  User,
+  UserAuthentication,
+  AuthenticationProvider,
+} from "@server/models";
 
 type UserCreatorResult = {
   user: User;
   isNewUser: boolean;
-  authentication: UserAuthentication;
+  authentication: UserAuthentication | null;
```
thenanyu, line 15: authentication can now return null, for cases where there is a login through simple email matching. this has no practical impact, as this part of the result was never used

server/commands/teamCreator.ts
```diff
@@ -59,11 +56,29 @@ async function teamCreator({
...
+    authP = await AuthenticationProvider.findOne({
+      where: { name: authenticationProvider.name, teamId },
+      include: [
+        {
+          model: Team,
+          as: "team",
+          required: true,
+        },
+      ],
+    });
```
thenanyu, line 73: we still need to return the auth provider associated with the team that matched the SSO provider the user was using.

thenanyu:
@tommoor I've tested with all of the cases and it seems to be working. I have some more tests to write but this should be good to review at this point

server/commands/teamCreator.ts
thenanyu: `teamCreator` is doing a lot right now, and it doesn't feel like it's named properly at this point — but I think this is coherent to the original design of this command and general pattern.

a cleaner design might be to have separate commands handle the existing team and new team cases. we can do that refactor down the line if this gets any more complicated

server/commands/accountProvisioner.ts
```diff
@@ -21,18 +21,18 @@ type Props = {
     username?: string;
   };
   team: {
-    id?: string;
+    teamId?: string;
```
tommoor, line 24: Not really sure why this change – can we just add the JSDoc comment to the property if you're finding it ambigous.

server/commands/userCreator.test.ts
```diff
@@ -277,13 +282,46 @@ describe("userCreator", () => {
...
-    expect(authentication.scopes[0]).toEqual("read");
+    expect(authentication).toBeDefined();
+    expect(authentication?.accessToken).toEqual("123");
+    expect(authentication?.scopes.length).toEqual(1);
+    expect(authentication?.scopes[0]).toEqual("read");
     expect(user.email).toEqual(invite.email);
     expect(isNewUser).toEqual(true);
   });
 
+  it.only("should create a user from an invited user who is external to the team", async () => {
```
tommoor, line 293:
```suggestion
  it("should create a user from an invited user who is external to the team", async () => {
```

server/commands/teamCreator.ts
```diff
@@ -14,23 +11,24 @@ type TeamCreatorResult = {
...
-  id?: string;
+  teamId?: string;
   name: string;
   domain?: string;
   subdomain: string;
   avatarUrl?: string | null;
   authenticationProvider: {
     name: string;
-    providerId: string;
+    providerId: string; // external id
```
tommoor, line 25:
```suggestion
    providerId: string;
```

server/commands/teamCreator.ts
```diff
@@ -59,11 +57,30 @@ async function teamCreator({
...
-  else if (id) {
-    throw InvalidAuthenticationError("incorrect authentication credentials");
+  } else if (teamId) {
+    // A team id was provided but no auth provider was found matching those credentials
+    // The user is attempting to log into a team with an external or personal SSO
+
+    // Get the AuthenticationProvider for the team that matches the SSO name (google, slack, etc)
+    authP = await AuthenticationProvider.findOne({
+      where: { name: authenticationProvider.name, teamId },
+      include: [
```
tommoor, line 67:
```suggestion
      rejectOnEmpty: true,
      include: [
```

server/commands/teamCreator.ts
```diff
@@ -59,11 +57,30 @@ async function teamCreator({
...
+        {
+          model: Team,
+          as: "team",
+          required: true,
+        },
+      ],
+    });
+
+    invariant(authP, "authentication provider type must exist on team");
```
tommoor, line 76:
```suggestion
```

Moving to the `rejectOnEmpty` style, it satisfies the type requirements too

server/commands/accountProvisioner.ts
```diff
@@ -21,18 +21,18 @@ type Props = {
...
     name: string;
     domain?: string;
     subdomain: string;
     avatarUrl?: string | null;
   };
   authenticationProvider: {
     name: string;
-    providerId: string;
+    providerId: string; // external id of the auth provider
```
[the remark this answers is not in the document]
thenanyu: is there an extension I need to install? It doesn't seem to show up on hover in vscode for me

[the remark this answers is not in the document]
tommoor: Could be because these props are defined in a number of places instead of one central spot?

server/commands/accountProvisioner.ts
thenanyu: (I added the parameter in the PR right before this one, so it's more my changing my mind on naming conventions). We specify `authenticationProviderId` no matter how it's nested, so this is more consistent and unambiguous

server/commands/accountProvisioner.ts
```diff
@@ -55,15 +57,45 @@ async function accountProvisioner({
...
-    result = await teamCreator({
+    result = await teamProvisioner({
       ...teamParams,
       authenticationProvider: authenticationProviderParams,
       ip,
     });
   } catch (err) {
-    throw InvalidAuthenticationError(err.message);
+    // The account could not be provisioned for the provided teamId
+    // check to see if we can try authentication using email matching only
```
thenanyu, line 70: 🥦  test for ability to log in via external SSO here

server/commands/accountProvisioner.ts
```diff
@@ -55,15 +57,45 @@ async function accountProvisioner({
...
+          {
+            model: Team,
+            as: "team",
+            required: true,
+          },
+        ],
+      });
+
+      if (authenticationProvider) {
+        emailMatchOnly = true;
```
thenanyu, line 87: this is passed into `userCreator` later on in the flow

tommoor:
I know we're both rightfully very cautious with this PR but the fact that all of the existing authentication tests pass without trouble and this simply adds extra cases makes it feel _a lot_ better 😄

tommoor (approved):

tommoor:
✅ Tests
✅ Tested locally the cases I can think of

Merging now for a quiet weekend rollout

PR #3800, opened 2022-07-17
feat: Port `scenes/Search` to functional component style

CuriousCorrelation:
# Description

PR ports `scene/Search/Search.tsx` to functional component style at `scene/Search/index.tsx`

# Issue
Closes  #3781 and  #3457

PR #3828, opened 2022-07-21
feat: Port `scenes/Search` to functional component style

tommoor:
Reverts outline/outline#3827
Retry https://github.com/outline/outline/pull/3800

There is a bug in here causing a request loop, needs to be fixed before merge.


CuriousCorrelation:
@tommoor I unfortunately wasn't able to reproduce the issue although I think 

https://github.com/outline/outline/blob/c3e207d644950fa56b774b17a0a0c649b25e004a/app/scenes/Search/index.tsx#L249-L251

was the culprit here.

Previously

``` js
 const handleEscape = React.useCallback(() => { 
   searchInputRef?.current?.focus(); 
 }, [searchInputRef]); 
```

Seems that `eslint` message 

> React Hook useCallback has a missing dependency: 'searchInputRef'. Either include it or remove the dependency array.

must have been a false positive because `useRef` is a mutable "container".

Found  a couple of stackoverflow answers concurring 

https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom/60476525#60476525

https://stackoverflow.com/questions/63346113/should-refs-be-in-listed-as-dependencies-for-useeffect-and-such/63353582#63353582

---

I also redid some other dependencies like 

https://github.com/outline/outline/blob/c3e207d644950fa56b774b17a0a0c649b25e004a/app/scenes/Search/index.tsx#L160

because 

https://github.com/outline/outline/blob/c3e207d644950fa56b774b17a0a0c649b25e004a/app/scenes/Search/index.tsx#L139

with its dependencies 

https://github.com/outline/outline/blob/c3e207d644950fa56b774b17a0a0c649b25e004a/app/scenes/Search/index.tsx#L146

should be enough to trigger `useEffect` because of the changes in function reference(s) _even if_ both should technically be in the same render slots with their `search` dependency. 

Similarly 

https://github.com/outline/outline/blob/c3e207d644950fa56b774b17a0a0c649b25e004a/app/scenes/Search/index.tsx#L156

with `term` as dependency.
---

There shouldn't be any other effects that might cause infinite re-renders besides that.

tommoor:
Looking at this in more detail there are a few problems causing the loop, the `offset` is never increased when there are more than one page of results which is one main reason – but also the changing `params` object makes it worse. You can reproduce the loop by temporarily setting the pagination limit to something low.

I think we need to place this existing PR to one side and start again re-build this view to use the existing `PaginatedList` component, it should vastly simplify things as it bakes on all the well-tested logic around pagination, retries, error handling, empty states.

Issue #3702, opened 2022-06-29
View of webhook deliveries


tommoor:
### Requirements

We are already recording webhook deliveries in the database, we need to create an endpoint or two to expose these via the API and enable debugging of failing deliveries.

Open to UX suggestions but I'd probably change the webhook list items to have a menu instead of buttons and then add a "Recent deliveries" item or similar in the menu to open a view similar to below…

### Inspiration

It's nice that both GitHub and Stripe let you Redeliver/Resend an individual delivery on-demand for testing. Stripe also lets you filter by success / failure deliveries. We have some existing patterns to achieve similar, eg the tabs on the Home / Collection screens that load a list endpoint based on different filters.

#### GitHub

<img width="793" alt="image" src="https://user-images.githubusercontent.com/380914/176379947-164c225d-cfde-4384-ad42-3dc769d7d4cc.png">

#### Stripe

<img width="1203" alt="image" src="https://user-images.githubusercontent.com/380914/176380427-7eed3352-ec05-44dc-b94a-582be2b151da.png">

Related https://github.com/outline/outline/pull/3691

coreyja:
Getting started on this, this evening!

PR #3833, opened 2022-07-22
Webhook Deliveries View

coreyja:
A view for Webhook Deliveries

Right now its just a table of all the deliveries, with a status icon
From here I want to add a way to view the request and response bodies and probably the headers too

Demo:


coreyja:
Resolves #3702

app/models/WebhookDelivery.ts
```diff
@@ -0,0 +1,15 @@
+import { observable } from "mobx";
+import BaseModel from "./BaseModel";
+import Field from "./decorators/Field";
+
+class WebhookDelivery extends BaseModel {
+  @Field
```
tommoor, line 6: Field decorator only required for mutable fields so can lose these

app/scenes/Settings/components/WebhookSubscriptionListItem.tsx
```diff
@@ -1,42 +1,17 @@
...
   const { t } = useTranslation();
-  const { dialogs } = useStores();
-  const [
-    editModalOpen,
-    handleEditModalOpen,
-    handleEditModalClose,
-  ] = useBoolean();
-
-  const showDeletionConfirmation = React.useCallback(() => {
-    dialogs.openModal({
```
tommoor, line 29: I'd use this style of triggering dialogs in the webhook menu too, it avoids mounting them into the React tree at all on render until triggered. I'm trying to move everything over to it.

server/routes/api/webhookDeliveries.ts
```diff
@@ -0,0 +1,35 @@
...
+import auth from "@server/middlewares/authentication";
+import { WebhookDelivery } from "@server/models";
+import { authorize } from "@server/policies";
+import { presentWebhookDelivery } from "@server/presenters";
+import pagination from "./middlewares/pagination";
+
+const router = new Router();
+
+router.post(
+  "webhookDeliverys.list",
```
tommoor, line 11:
```suggestion
  "webhookDeliveries.list",
```

server/routes/api/webhookDeliveries.ts
```diff
@@ -0,0 +1,35 @@
...
+  auth({ admin: true }),
+  pagination(),
+  async (ctx) => {
+    const { user } = ctx.state;
+    const { webhookSubscriptionId } = ctx.request.body;
+    authorize(user, "listWebhookDeliveries", user.team);
```
tommoor, line 17: Need to change this so that it queries the webhookSubscription and then have the policy run against that, currently you could pass any team's `webhookSubscriptionId` and exfiltrate data.

tommoor (commented):
Great to see progress 👏

PR #3834, opened 2022-07-22
feat: Document subscriptions

CuriousCorrelation:
# Description

First in the future family of PRs adds a subscription model with aim to de-couple notifications from events like `documents.update` or `collections.create` and moves the logic to a subscriptions system which adds another layer more fine-grained control for users.

## In this PR

1. A first look at `Subscription` model.
2. A `Subscribe`/`Unsubscribe` button in document menu (top-right corner)
3. `isSubscribed` check for document stores.
4. `Subscriptions` store.
5. Ability to subscribe/unsubscribe to `documents.update` event.
6. Tons of tests! :test_tube: :test_tube: :test_tube:


app/actions/definitions/documents.tsx
```diff
@@ -108,6 +110,63 @@ export const unstarDocument = createAction({
   },
 });
 
+export const subscribeDocument = createAction({
+  name: ({ t }) => t("Subscribe"),
+  section: DocumentSection,
+  // TODO: Use other icon.
+  icon: <BookmarkedIcon />,
```
CuriousCorrelation, line 117: Using `BookmarkIcon` and `BookmarkedIcon` for now.

app/stores/DocumentsStore.ts
```diff
@@ -753,6 +753,23 @@ export default class DocumentsStore extends BaseStore<Document> {
...
   };
 
+  subscribe = async (document: Document) => {
+    await this.rootStore.subscriptions.create({
+      documentId: document.id,
+      // REVIEW: Seems like a good location
+      // to describe event for now.
+      event: "documents.update",
+    });
+  };
```
CuriousCorrelation, line 763: Might need to move `documents.update` somewhere else in the future.

app/stores/DocumentsStore.ts
```diff
@@ -753,6 +753,23 @@ export default class DocumentsStore extends BaseStore<Document> {
...
+
+  unsubscribe = async (document: Document) => {
+    const subscription = this.rootStore.subscriptions.orderedData.find(
+      (subscription) => subscription.documentId === document.id
+    );
+
+    await subscription?.delete();
+  };
```
CuriousCorrelation, line 771: Should this also check for `subscription.userId === currentUser.id`? Probably being paranoid.

app/stores/SubscriptionsStore.ts
```diff
@@ -0,0 +1,18 @@
+import Subscription from "~/models/Subscription";
+import BaseStore, { RPCAction } from "./BaseStore";
+import RootStore from "./RootStore";
+
+export default class SubscriptionsStore extends BaseStore<Subscription> {
+  actions = [
+    RPCAction.Info,
+    // list is disabled.
+    // RPCAction.List
+    RPCAction.Create,
```
CuriousCorrelation, line 10: We could remove `subscriptions.list` route entirely before finalizing. Right now I find it really useful for testing.

server/commands/subscriptionUpdater.ts
```diff
@@ -0,0 +1,62 @@
...
+  // `userId` + `subscription` should be enough infomation.
+  subscription,
+  event,
+  enabled,
+  ip,
+  transaction,
+}: Props): Promise<Subscription> {
+  // Subscription shouldn't be allowed to move.
+  // REVIEW: Is this alright?
+  assert(event === subscription.event);
```
CuriousCorrelation, line 37: Subscriptions being _moved_ to other objects doesn't sound right. I'll look for a better way in the future if this isn't correct.

server/policies/document.ts
```diff
@@ -177,6 +177,45 @@ allow(User, ["pin", "unpin"], Document, (user, document) => {
...
+    "Collection is missing, did you forget to include in the query scope?"
+  );
+
+  if (cannot(user, "update", document.collection)) {
+    return false;
+  }
+
+  // REVIEW: subscribe outside of the team?
+  // return user.teamId === document.teamId;
+  return true;
```
CuriousCorrelation, line 216: Maybe on a public document? Not 100% sure.

server/policies/subscription.ts
```diff
@@ -0,0 +1,109 @@
...
+allow(User, "createSubscription", Document, (user, document) => {
+  // Sanity check.
+  if (!document) {
+    return false;
+  }
+
+  // REVIEW: Are viewers allowed to subscribe?
+  if (user.isViewer) {
+    return false;
+  }
```
CuriousCorrelation, line 14: Similar to comment above

https://github.com/outline/outline/pull/3834/files#r927311996

server/policies/subscription.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  // REVIEW: Show admin be allowed to prescribe
+  // a document to user?
+  if (user.isAdmin) {
+    return true;
+  }
```
CuriousCorrelation, line 20: Not sure if we should allow admins to _**prescribe**_ a subscription.

app/actions/definitions/documents.tsx
tommoor, line 117: I have some icons in the pipeline for you to use, these are good placeholders 👍

app/actions/definitions/documents.tsx
```diff
@@ -108,6 +110,63 @@ export const unstarDocument = createAction({
...
+  keywords: "subscribe",
```
tommoor, line 118:
```suggestion
```

Don't need keywords if they are just the same as the name, only if we have synonyms. In this case I can't really think of any…

app/actions/definitions/documents.tsx
```diff
@@ -108,6 +110,63 @@ export const unstarDocument = createAction({
...
+    document?.subscribe();
+  },
+});
+
+export const unsubscribeDocument = createAction({
+  name: ({ t }) => t("Unsubscribe"),
+  section: DocumentSection,
+  // TODO: Use other icon.
+  icon: <BookmarkIcon />,
+  keywords: "unsubscribe",
```
tommoor, line 147:
```suggestion
```

app/menus/DocumentMenu.tsx
```diff
@@ -327,6 +347,22 @@ function DocumentMenu({
               visible: !document.isStarred && !!can.star,
               icon: <StarredIcon />,
             },
+            {
```
tommoor, line 350: I know this is a bit confusing at the moment, this part of the codebase is kind of in permanent flux with the switch over to action definitions. Here you can use `actionToMenuItem` (see L367) to convert the action into a menu item without having to redefine the label/icon etc.

app/menus/DocumentMenu.tsx
```diff
@@ -327,6 +347,22 @@ function DocumentMenu({
...
+              icon: <BookmarkIcon />,
+            },
+            {
+              type: "button",
+              title: t("Subscribe"),
+              onClick: handleSubscribe,
+              visible: !document.isSubscribed && !!can.subscribe,
+              // TODO: Use better icon
+              icon: <BookmarkedIcon />,
+            },
```
tommoor, line 365:
```suggestion
            actionToMenuItem(subscribeDocument, context),
            actionToMenuItem(unsubscribeDocument, context),
```

app/models/Document.ts
```diff
@@ -155,6 +155,13 @@ export default class Document extends ParanoidModel {
     );
   }
 
+  @computed
```
tommoor, line 158: 👌

app/models/Subscription.ts
```diff
@@ -0,0 +1,33 @@
...
+  @observable
+  id: string;
+
+  userId: string;
+
+  documentId: string;
+
+  event: string;
+
+  enabled: boolean;
```
tommoor, line 16: All of these properties above should have the `@Field` and `@observable` decorators, likely not causing any problems rn but good for consistency – never know how the model will be used in the future.

app/stores/DocumentsStore.ts
tommoor: Not paranoid, good idea. Seems v possible we'll have other users subscriptions in this store later

app/menus/DocumentMenu.tsx
```diff
@@ -172,6 +174,24 @@ function DocumentMenu({
...
+
+  const handleUnsubscribe = React.useCallback(
+    (ev: React.SyntheticEvent) => {
+      ev.preventDefault();
+      ev.stopPropagation();
+      document.unsubscribe();
+    },
+    [document]
+  );
+
```
tommoor, line 194:
```suggestion
```

server/commands/subscriptionCreator.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    expect(subscription.documentId).toEqual(document.id);
+    expect(subscription.userId).toEqual(user.id);
+    expect(event?.name).toEqual("subscriptions.create");
+    expect(event?.modelId).toEqual(subscription.id);
+    expect(event?.actorId).toEqual(subscription.userId);
+    expect(event?.userId).toEqual(subscription.userId);
+    expect(event?.documentId).toEqual(subscription.documentId);
+  });
+
+  it("should not create anther subscription if one already exists", async () => {
```
tommoor, line 43:
```suggestion
  it("should not create another subscription if one already exists", async () => {
```

server/commands/subscriptionCreator.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        transaction,
+      })
+    );
+
+    expect(subscription1.userId).toEqual(subscription2.userId);
+    expect(subscription1.documentId).toEqual(subscription2.documentId);
+    expect(subscription1.enabled).toEqual(subscription2.enabled);
+  });
+
+  it("should not record event if the subscription already exists", async () => {
```
tommoor, line 73: I'd just combine this with the test above, add the events expects there too

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,70 @@
+import { Transaction } from "sequelize";
+import { Subscription, Event, User } from "@server/models";
+
+type Props = {
+  /** The user creating the subscription */
+  user: User;
+  /** The document to subscription */
```
tommoor, line 7:
```suggestion
  /** The document to subscribe to */
```

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,70 @@
...
+  documentId?: string;
+  /** Event to subscribe */
+  event: string;
+  /** Status of a subscription */
+  enabled: boolean;
```
tommoor, line 12:
```suggestion
```

Lets set this default on the model

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,70 @@
...
+  enabled = true,
+  ip,
+  transaction,
+}: Props): Promise<Subscription> {
+  // If a subscription already exists, fetch it,
+  // otherwise create a subscription and return it.
+  // `created` is a boolean value
+  // that indicates  whether
+  // that subscription was created
+  // or already existed.
```
tommoor, line 39:
```suggestion
```

Not a super valuable comment `findOrCreate` should be self explanatory

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,70 @@
...
+  const [subscription, created] = await Subscription.findOrCreate({
+    where: {
+      userId: user.id,
+      documentId,
+      event,
+      enabled,
```
tommoor, line 45:
```suggestion
```

Just repeating, lets remove this and move the default to the model definition. We wouldn't want this in the `where` clause here anyway as otherwise it would not find an existing disabled subscription

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+  });
+
+  // Don't emit an event if a new subscription
+  // wasn't created.
+  //
+  // `findOrCreate` is primarily used as an
+  // infallible function for valid
+  // arguments.
```
tommoor, line 54:
```suggestion
```

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,70 @@
...
+    where: {
+      userId: user.id,
+      documentId,
+      event,
+      enabled,
+    },
+  });
+
+  // Don't emit an event if a new subscription
+  // wasn't created.
```
tommoor, line 50:
```suggestion
  // Don't emit an event if a new subscription wasn't created.
```

Can you make sure that comments go to the 80 character line wrap, thanks 🙏

server/commands/subscriptionCreator.test.ts
tommoor: Might want to add a test for if the existing subscription is disabled, as it would be failing currently 😉

server/commands/subscriptionDestroyer.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  const subscribedEvent = "documents.update";
+
+  it("should destroy existing subscription", async () => {
+    const user = await buildUser();
+    const document = await buildDocument({
+      userId: user.id,
+      teamId: user.teamId,
+    });
+
+    const subscription = await sequelize.transaction(
```
tommoor, line 21: Use `buildSubscription` here?

server/commands/subscriptionDestroyer.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * This command destroys a document subscription.
+ *
+ * This just removes the subscription.
+ * Does not touch the target document.
+ *
+ * @param Props The subscription to destroy
+ * @returns void
+ */
+export default async function subscriptionDestroyer({
```
tommoor, line 23: I'm not sure if there is a usecase for destroying the subscription completely, if you did so then the next edit of the document would just result in it becoming recreated. We likely only need to mark the subscription as disabled.

server/queues/tasks/DeliverWebhookTask.ts
```diff
@@ -180,6 +183,11 @@ export default class DeliverWebhookTask extends BaseTask<Props> {
       case "shares.revoke":
         await this.handleShareEvent(subscription, event);
         return;
+      case "subscriptions.create":
```
tommoor, line 186: I appreciate the thoroughness, I don't think we need these as webhook events though – to me a document subscription is basically a personal setting akin to starring a document. There isn't value in broadcasting this, in that vein we might not even need to record events for it at all.

server/types.ts
```diff
@@ -257,6 +257,19 @@ export type ShareEvent = BaseEvent & {
...
+    | "subscriptions.create"
+    | "subscriptions.update"
+    | "subscriptions.delete";
+  modelId: string;
+  actorId: string;
+  userId: string;
+  documentId: string;
+  enabled: boolean;
+  ip: string;
+};
```
tommoor, line 271:
```suggestion
export type SubscriptionEvent = BaseEvent & {
  name:
    | "subscriptions.create"
    | "subscriptions.update"
    | "subscriptions.delete";
  modelId: string;
  userId: string;
  documentId: string;
  data: { enabled: boolean };
};
```

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,241 @@
...
+      data: subscriptions.map(presentSubscription),
+    };
+  }
+);
+
+router.post(
+  "subscriptions.info",
+  auth({
+    optional: true,
+  }),
```
tommoor, line 71:
```suggestion
  auth(),
```

Definitely not optional

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,241 @@
...
+ * 5. CRUD operation by actor on subject.
+ * 6. Ordering and pagination.
+ * 7. Setting response body.
+ */
+
+const router = new Router();
+
+const subscribableEventsDocument = ["documents.update"];
+
+router.post(
```
tommoor, line 26: Not really reviewing this file yet, it seems like it's a bit too wip

app/menus/DocumentMenu.tsx
CuriousCorrelation: Excellent! Thanks for the suggestion

server/commands/subscriptionCreator.ts
CuriousCorrelation: Will do. My auto-formatter doesn't seem to like eslint/vscode at all so it tends to format things all over the place. I'll do a full sweep after finalizing changes.

server/commands/subscriptionDestroyer.ts
```diff
@@ -0,0 +1,52 @@
...
+  const changed = subscription.changed();
+
+  // Don't emit an event if a subscription
+  // wasn't updated.
+  if (changed) {
+    await Event.create(
+      {
+        // REVIEW: Should this emit "subscriptions.delete"?
+        name: "subscriptions.update",
+        modelId: subscription.id,
```
CuriousCorrelation, line 40: Should this event name be `subscriptions.delete`? `subscriptions.update` is the _"real"_ event here, but the idea is to delete.

server/commands/subscriptionCreator.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      })
+    );
+
+    expect(subscription1.userId).toEqual(subscription2.userId);
+    expect(subscription1.documentId).toEqual(subscription2.documentId);
+    expect(subscription1.enabled).toEqual(subscription2.enabled);
+  });
+
+  it("should not enable subscription by overriding one that already exists in disabled state", async () => {
+    const user = await buildUser();
```
CuriousCorrelation, line 72: Not sure if `subscriptions.create` should enable a disabled subscription. On one hand it makes a sort of sense, on the other we might encourage use of incorrect API semantics, which might cause a ton of headache in the future if we were to add other subscribeable events.

tommoor: @thenanyu not sure if you have opinions here?

The other way we could model this is and I'm kind of leaning towards now seeing how this is progressing is to just do standard create/delete and use a paranoid model so that before auto-subscribing we check if there is a previously deleted version of the subscription in the database. We can remove the `enabled` column then too…

tommoor: The other main reason to keep it is to make sure that when you edit a document again if you have previously disabled the subscription it shouldn't get automatically recreated as if you've never edited the document before.

thenanyu: oh interesting! still think paranoid is the way to go. in this case deleted subs are acting as a tombstone, which is another valid usecase for paranoid

app/actions/definitions/documents.tsx
tommoor: Upgrade `outline-icons` to `v1.44.0`, it includes `SubscribeIcon` and `UnsubscribeIcon`

CuriousCorrelation:
@tommoor Port to Paranoid model is done! There's still some work to be done in `routes/api` but everything should work as expected.

I've changed `subscriptions.update` route to behave _as is expected_ with these change. Let me know what we should do with it. Soft block from API or completely remove or something else.

app/actions/definitions/documents.tsx
CuriousCorrelation: They look great! Checkout latest commit if you want to test them out. Dropdown menu should work.

server/scripts/20220722000000-backfill-subscriptions.ts
```diff
@@ -0,0 +1,65 @@
+import "./bootstrap";
+import { Subscription, Document, Event } from "@server/models";
+
+const limit = 100;
+let page = parseInt(process.argv[2], 10);
+page = Number.isNaN(page) ? 0 : page;
+
+export default async function main(exit = false) {
```
CuriousCorrelation, line 8: I wasn't sure if there was a recommended way to go about back-filling subscriptions, but that's when I saw https://github.com/outline/outline/blob/4f537c757829782f68c58e5c4abe384b84908ace/server/scripts/20210716000000-backfill-revisions.ts#L8
This module tries to replicate :point_up: as closely as possible.

server/scripts/20220722000000-backfill-subscriptions.test.ts
```diff
@@ -0,0 +1,151 @@
+import { Event } from "@server/models";
+import { buildDocument, buildUser } from "@server/test/factories";
+import { flushdb } from "@server/test/support";
+import script from "./20220722000000-backfill-subscriptions";
+
+beforeEach(() => flushdb());
+
+describe("#work", () => {
+  it("should create subscriptions and events for document creator and collaborators", async () => {
```
CuriousCorrelation, line 9: This test is a bit convoluted, but these cyclic dependencies help solve a couple of edge cases.

app/actions/definitions/documents.tsx
```diff
@@ -108,6 +110,65 @@ export const unstarDocument = createAction({
   },
 });
 
+export const subscribeDocument = createAction({
+  name: ({ t }) => t("Subscribe"),
+  section: DocumentSection,
+  // TODO: Use other icon.
```
tommoor, line 116:
```suggestion
```

app/actions/definitions/documents.tsx
```diff
@@ -108,6 +110,65 @@ export const unstarDocument = createAction({
...
+    );
+  },
+  perform: ({ activeDocumentId, stores, currentUserId }) => {
+    if (!activeDocumentId) {
+      return;
+    }
+
+    if (!currentUserId) {
+      return;
+    }
```
tommoor, line 164:
```suggestion
    if (!activeDocumentId || !currentUserId) {
      return;
    }
```

app/models/Subscription.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  next(): Subscription | undefined {
+    const index = this.store.orderedData.indexOf(this);
+    return this.store.orderedData[index + 1];
+  }
+
+  previous(): Subscription | undefined {
+    const index = this.store.orderedData.indexOf(this);
+    return this.store.orderedData[index + 1];
+  }
```
tommoor, line 30:
```suggestion
```

Doesn't appear to be used?

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,59 @@
...
+
+type Props = {
+  /** The user creating the subscription */
+  user: User;
+  /** The document to subscribe to */
+  documentId?: string;
+  /** Event to subscribe */
+  event: string;
+  /**  Fetch soft deleted rows? */
+  paranoid?: boolean;
```
tommoor, line 12:
```suggestion
```

Passing this is exposing too much logic outside the command, it's not needed

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,59 @@
...
+  ip,
+  transaction,
+}: Props): Promise<Subscription> {
+  const [subscription, created] = await Subscription.findOrCreate({
+    where: {
+      userId: user.id,
+      documentId,
+      event,
+    },
+    paranoid,
```
tommoor, line 39:
```suggestion
    paranoid: false,
```

This find should always find previously deleted subscriptions

server/commands/subscriptionCreator.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  // Don't emit an event if a new subscription
```
tommoor, line 42: There should be some logic here that checks if the subscription is deleted and undeletes it before returning.

server/migrations/20220719121200-create-subscriptions.js
```diff
@@ -0,0 +1,71 @@
...
+            type: Sequelize.DATE,
+            allowNull: true,
+          },
+        },
+        { transaction }
+      );
+
+      await queryInterface.addIndex(
+        "subscriptions",
+        ["userId", "documentId", "event", "deletedAt"],
```
tommoor, line 52:
```suggestion
        ["userId", "documentId", "event"],
```

server/migrations/20220719121200-create-subscriptions.js
```diff
@@ -0,0 +1,71 @@
...
+      );
+    });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeIndex("subscriptions", [
+      "userId",
+      "documentId",
+      "event",
+      "deletedAt",
```
tommoor, line 67:
```suggestion
```

server/policies/document.ts
```diff
@@ -177,6 +177,45 @@ allow(User, ["pin", "unpin"], Document, (user, document) => {
...
+    return false;
+  }
+
+  // Sanity check.
+  invariant(
+    document.collection,
+    "Collection is missing, did you forget to include in the query scope?"
+  );
+
+  if (cannot(user, "update", document.collection)) {
```
tommoor, line 210:
```suggestion
  if (cannot(user, "read", document.collection)) {
```

Don't need edit permissions to subscribe, only read permissions

server/policies/subscription.ts
```diff
@@ -0,0 +1,106 @@
...
+allow(User, "createSubscription", Document, (user, document) => {
+  // Sanity check.
+  if (!document) {
+    return false;
+  }
+
+  // REVIEW: Are viewers allowed to subscribe?
+  if (user.isViewer) {
+    return false;
+  }
```
tommoor, line 14:
```suggestion
```

Viewers can subscribe 👍

server/policies/document.ts
tommoor, line 216:
```suggestion
  return user.teamId === document.teamId;
```

Always got to check we are operating in the same team

server/policies/subscription.ts
```diff
@@ -0,0 +1,106 @@
+import { Subscription, User, Document } from "@server/models";
+import { allow, _cannot as cannot } from "./cancan";
+
+// Is `user` allowed to subscribe to `document`?
+allow(User, "createSubscription", Document, (user, document) => {
```
tommoor, line 5: This policy seems to be duplicative of the one in policies/documents?

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+import { presentSubscription } from "@server/presenters";
+import { SubscriptionEvent } from "@server/types";
+import { assertIn, assertPresent, assertUuid } from "@server/validation";
+import pagination from "./middlewares/pagination";
+
+const router = new Router();
+
+router.post(
+  "subscriptions.list",
+
```
tommoor, line 17: Please remove the extra spacing around the middlewares

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+  auth(),
+
+  pagination(),
+
+  async (ctx) => {
+    const { user } = ctx.state;
+
+    const { documentId, event } = ctx.body;
+
+    assertPresent(documentId, "documentId is required");
```
tommoor, line 27:
```suggestion
    assertUuid(documentId, "documentId is required");
```
Need to assert the UUID format or junk strings can get passed to db queries

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    // Make sure `documentId` is accompanied
+    // with valid subsribable events.
+    assertIn(
+      event,
+      ["documents.update"],
+      `${event} is not a valid subscription event for documents`
```
tommoor, line 34:
```suggestion
      `Not a valid subscription event for documents`
```

Best not to echo untrusted user input back

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  async (ctx) => {
+    const { user } = ctx.state;
+
+    const { documentId, event } = ctx.body;
+
+    assertPresent(documentId, "documentId is required");
+
+    // Make sure `documentId` is accompanied
+    // with valid subsribable events.
```
tommoor, line 30:
```suggestion
```

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+  "subscriptions.info",
+
+  auth(),
+
+  pagination(),
+
+  async (ctx) => {
+    const { user } = ctx.state;
+
+    const subscriptions = await Subscription.findAll({
```
tommoor, line 71: `info` endpoints always return the information for a single record – as such they are not paginated. In this case it should work similar to the subscriptions.list endpoint where a documentId and event are acceptable parameters and it turns the subscription record belonging to the current user (if any).

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+  "subscriptions.create",
+
+  auth(),
+
+  async (ctx) => {
+    const { user } = ctx.state;
+
+    const { documentId, event } = ctx.body;
+
+    assertPresent(documentId, "documentId is required");
```
tommoor, line 97:
```suggestion
    assertUuid(documentId, "documentId is required");
```

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    // Make sure `documentId` is accompanied
+    // with valid subsribable events.
```
tommoor, line 100:
```suggestion
```

These comments aren't particularly useful as what's happening is clear from the code itself

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+      `${event} is not a valid subscription event for documents`
+    );
+
+    const subscription = await sequelize.transaction(async (transaction) => {
+      const document = await Document.findByPk(documentId, { transaction });
+
+      authorize(user, "createSubscription", document);
+
+      return subscriptionCreator({
+        user: user,
```
tommoor, line 113:
```suggestion
        user,
```

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+      });
+    });
+
+    ctx.body = {
+      data: presentSubscription(subscription),
+    };
+  }
+);
+
+router.post(
```
tommoor, line 127: I'm not sure this endpoint needs to exist currently… aren't create/delete enough?

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+    assertUuid(id, "id is required");
+
+    const { user } = ctx.state;
+
+    const subscription = await Subscription.findByPk(id);
+
+    assertPresent(
+      subscription,
+      "subscription is required, make sure subscription id is correct"
+    );
```
tommoor, line 189:
```suggestion
    const subscription = await Subscription.findByPk(id, {
      rejectOnEmpty: true
    });
```

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    invariant(
+      subscription?.documentId,
+      "Subscription must have an associated document"
+    );
```
tommoor, line 194:
```suggestion
```

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    authorize(user, "delete", subscription);
+
+    assertIn(
+      subscription.event,
+      ["documents.update"],
+      `${subscription.event} is not a valid subscription event`
+    );
```
tommoor, line 202:
```suggestion
```

Why would we assert data already stored in the database?

server/routes/api/subscriptions.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    await subscription.destroy();
```
tommoor, line 204: `subscriptionDestroyer` should be used here…

server/scripts/20220722000000-backfill-subscriptions.ts
```diff
@@ -0,0 +1,65 @@
...
+  const work = async (page: number) => {
+    console.log(`Backfill subscription… page ${page}`);
+
+    // Retrieve all documents within set limit.
+    const documents = await Document.findAll({
+      limit,
+      offset: page * limit,
+      order: [["createdAt", "DESC"]],
```
tommoor, line 16:
```suggestion
      order: [["createdAt", "ASC"]],
```

Better to go from old -> new otherwise documents that are created while the script is running won't get processed

server/scripts/20220722000000-backfill-subscriptions.ts
```diff
@@ -0,0 +1,65 @@
...
+            await Event.findOrCreate({
+              where: {
+                name: "subscriptions.create",
+                userId: collaborator,
+                modelId: collaboratorSubscription.id,
+                documentId: collaboratorSubscription.documentId,
+                actorId: collaboratorSubscription.userId,
+              },
+            });
+          }
```
tommoor, line 45:
```suggestion
```

Lets skip the events for backfill

server/scripts/20220722000000-backfill-subscriptions.ts
```diff
@@ -0,0 +1,65 @@
...
+let page = parseInt(process.argv[2], 10);
+page = Number.isNaN(page) ? 0 : page;
+
+export default async function main(exit = false) {
+  const work = async (page: number) => {
+    console.log(`Backfill subscription… page ${page}`);
+
+    // Retrieve all documents within set limit.
+    const documents = await Document.findAll({
+      limit,
```
tommoor, line 14: Could reduce db load by only selecting the attributes we care about here, 

```suggestion
      attributes: ["collaboratorIds", "id"],
      limit,
```

server/scripts/20220722000000-backfill-subscriptions.ts
```diff
@@ -0,0 +1,65 @@
...
+          const [
+            collaboratorSubscription,
+            created,
+          ] = await Subscription.findOrCreate({
+            where: {
+              userId: collaborator,
+              documentId: document.id,
+              event: "documents.update",
+            },
+          });
```
tommoor, line 33:
```suggestion
        await Promise.all(document.collaboratorIds, (collaboratorId) => 
          Subscription.findOrCreate({
            where: {
              userId: collaboratorId,
              documentId: document.id,
              event: "documents.update",
            },
          })
        })
```

Parallelize creation will speed this script up a lot

server/policies/subscription.ts
tommoor: Maybe in the future, lets keep it locked down until the UI exists (it may never exist)

server/commands/subscriptionDestroyer.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param Props The subscription to destroy
+ * @returns void
+ */
+export default async function subscriptionDestroyer({
+  user,
+  subscription,
+  ip,
+  transaction,
+}: Props): Promise<Subscription> {
+  assert(subscription.userId === user.id);
```
tommoor, line 30: Generally authorization is not done inside commands, once we get here all parameters are validated and trusted.

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -182,3 +225,85 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+  }
+
+  if (!subject.user.email) {
+    return false;
+  }
+
+  return true;
+};
+
+const createSubscriptions = async (
```
tommoor, line 278: Lets make this a private method on the class rather than a function floating in the file

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -182,3 +225,85 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+  document: Document,
+  event: DocumentEvent
+): Promise<void> => {
+  const collaboratorIds = document.collaboratorIds;
+
+  for (const collaboratorId of collaboratorIds) {
+    // Use single transaction for both finds.
```
tommoor, line 285:
```suggestion
```

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -182,3 +225,85 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+  return true;
+};
+
+const createSubscriptions = async (
+  document: Document,
+  event: DocumentEvent
+): Promise<void> => {
+  const collaboratorIds = document.collaboratorIds;
+
+  for (const collaboratorId of collaboratorIds) {
```
tommoor, line 284: Can parallelize this with `Promise.all` and speed up the processor significantly

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -182,3 +225,85 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+type Notifiable = {
+  user: User;
+  userId: string;
+  document: Document;
+  documentId: string;
+  event: DocumentActionEvent | RevisionEvent;
+  eventName: string;
+};
+
+const shouldNotify = async (subject: Notifiable): Promise<boolean> => {
```
tommoor, line 238: Lets make this a private method on the class rather than a function floating in the file

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -182,3 +225,85 @@ export default class NotificationsProcessor extends BaseProcessor {
     }
   }
 }
+
+type Notifiable = {
```
tommoor, line 229: Lots of duplicate data in this type, could just be `user`, `document` and `event`?

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -71,66 +98,82 @@ export default class NotificationsProcessor extends BaseProcessor {
...
-      const view = await View.findOne({
-        where: {
-          userId: setting.userId,
-          documentId: event.documentId,
-          updatedAt: {
-            [Op.gt]: document.updatedAt,
-          },
+    // Document update notifications
+    // Should be similar to notifications.
+    const subscriptions = await Subscription.findAll({
```
tommoor, line 139: We still need to consider the `NotificationSetting` here as document update emails may be disabled entirely, in which case all of the users subscriptions are essentially disabled

CuriousCorrelation: Fixing this should be as simple as calling email notifier inside this `if` block https://github.com/outline/outline/blob/11b823e24266201b2c61a227d3d38c7dd4d2295e/server/queues/processors/NotificationsProcessor.ts#L291 although I am quite sure I removed it intentionally because 

https://github.com/outline/outline/blob/30190866f8c2c6189105d34cb6b508f4917b4564/server/queues/processors/NotificationsProcessor.ts#L31 (on **main**) handles both `published` and `updated` events

https://github.com/outline/outline/blob/30190866f8c2c6189105d34cb6b508f4917b4564/server/queues/processors/NotificationsProcessor.ts#L74-L75

and sends out emails

https://github.com/outline/outline/blob/30190866f8c2c6189105d34cb6b508f4917b4564/server/queues/processors/NotificationsProcessor.ts#L125

which is still the case https://github.com/outline/outline/blob/11b823e24266201b2c61a227d3d38c7dd4d2295e/server/queues/processors/NotificationsProcessor.ts#L167 (on **this branch**) although now we also include subscriptions 

https://github.com/outline/outline/blob/11b823e24266201b2c61a227d3d38c7dd4d2295e/server/queues/processors/NotificationsProcessor.ts#L139

I think this https://github.com/outline/outline/blob/11b823e24266201b2c61a227d3d38c7dd4d2295e/server/queues/processors/NotificationsProcessor.test.ts#L211 and https://github.com/outline/outline/blob/11b823e24266201b2c61a227d3d38c7dd4d2295e/server/queues/processors/NotificationsProcessor.test.ts#L255 test specifically https://github.com/outline/outline/blob/11b823e24266201b2c61a227d3d38c7dd4d2295e/server/queues/processors/NotificationsProcessor.test.ts#L309-L311 should prove the functionality.

Let me know if this sounds alright.

tommoor: I don't really follow, but we can move the check into the beforeSend, like here:

https://github.com/outline/outline/blob/863f22750ffb86fec628901a8380d285e5c12ab3/server/emails/templates/InviteAcceptedEmail.tsx#L27

CuriousCorrelation:
Rational behind https://github.com/outline/outline/pull/3834/commits/dc4b6b5785303ef94ef3925a1cea092d96627e11


`info` always returns information for a single record.

That makes updating locals/stores in `DataLoader` quite tricky because we aren't working with subscription `id`, rather a combination of `document.id`, _document event_ and `user.id`.

Using `subscriptions.fetch` will require, `id`.

See https://github.com/outline/outline/blob/086c3ec2d83d6603f8834b0a2ef5229f9638bef4/app/stores/BaseStore.ts#L194

Two ways to solve this -

1. create a new function for `SubscriptionsStore`, say `fetchSubscriptions` which would combine functionalities of `fetch` and `fetchPage` from `BaseStore` https://github.com/outline/outline/blob/abfc28f83e99f08a9bc7f733f69a8c6f75a4d3d4/app/scenes/Document/components/DataLoader.tsx#L90
2. or; because this seems like an ideal situation, use `subscriptions.list` endpoint with `fetchPage` and constrain API to only return current user subscription. A filter before call to `presentor` https://github.com/outline/outline/blob/dc4b6b5785303ef94ef3925a1cea092d96627e11/app/scenes/Document/components/DataLoader.tsx#L93

---

I have made sure all affected files are accounted for so reverting this will be just `git revert dc4b6b5`

server/queues/processors/NotificationsProcessor.ts
CuriousCorrelation: [abfc28f](https://github.com/outline/outline/pull/3834/commits/abfc28f83e99f08a9bc7f733f69a8c6f75a4d3d4) commit should better illustrate it. 

I fear we might flood user inboxes because both `revisions.create` and `documents.update` event will send emails based on same source event.

Assumption based on 
``` js
async documentUpdated(event: DocumentEvent | RevisionEvent)
```
and
https://github.com/outline/outline/blob/086c3ec2d83d6603f8834b0a2ef5229f9638bef4/server/queues/processors/NotificationsProcessor.ts#L62-L65

server/commands/subscriptionCreator.ts
```diff
@@ -36,8 +36,23 @@ export default async function subscriptionCreator({
     paranoid: false,
   });
 
-  // Don't emit an event if a new subscription
-  // wasn't created.
+  // Fetched an already deleted subscription.
+  if (subscription.deletedAt) {
+    subscription.update({ deletedAt: null });
```
tommoor, line 41:
```suggestion
    subscription.update({ deletedAt: null }, { transaction });
```

Should be updated in the same transaction

app/scenes/Document/components/DataLoader.tsx
```diff
@@ -86,6 +86,22 @@ function DataLoader({ match, children }: Props) {
     fetchRevision();
   }, [revisions, revisionId]);
 
+  React.useEffect(() => {
+    async function fetchSubscription() {
+      if (document && document.id) {
```
tommoor, line 91:
```suggestion
      if (document?.id) {
```

server/commands/subscriptionCreator.ts
```diff
@@ -36,8 +36,23 @@ export default async function subscriptionCreator({
...
+
+    await Event.create(
+      {
+        name: "subscriptions.create",
+        modelId: subscription.id,
+        actorId: user.id,
+        userId: user.id,
```
tommoor, line 48:
```suggestion
```

`userId` is for events that affect a user.

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -230,18 +230,28 @@ export default class NotificationsProcessor extends BaseProcessor {
         const user = await User.findByPk(collaboratorId, { transaction });
 
         if (user) {
-          // `user` has to have `createSubscription` permission on `document`.
-          if (can(user, "createSubscription", document)) {
+          // `user` has to have `subscribe` permission on `document`.
+          if (can(user, "subscribe", document)) {
             // `subscriptionCreator` uses `findOrCreate`.
             // A duplicate won't be created if a subscription
             // exists already.
```
tommoor, line 237:
```suggestion
```

No longer relevant

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -230,18 +230,28 @@ export default class NotificationsProcessor extends BaseProcessor {
...
-            await subscriptionCreator({
-              user: user,
-              documentId: document.id,
-              event: "documents.update",
+            const exists = await Subscription.findOne({
+              where: { userId: user.id, documentId: document.id },
```
tommoor, line 239:
```suggestion
              where: { userId: user.id, documentId: document.id, event: "documents.update" },
```

`event` should be in this query, no?

server/routes/api/subscriptions.ts
```diff
@@ -27,13 +26,17 @@ router.post("subscriptions.list", auth(), pagination(), async (ctx) => {
...
   const subscriptions = await sequelize.transaction(async (transaction) => {
-    const document = await Document.findByPk(documentId, { transaction });
-
-    authorize(user, "listSubscription", document);
+    const document = await Document.findOne({
+      where: { id: documentId },
+      attributes: ["id"],
+      rejectOnEmpty: true,
+      transaction,
+    });
```
tommoor, line 34:
```suggestion
    const document = await Document.findByPk(documentId, { userId: user.id, transaction });
    authorize(user, "read", document);
```

server/routes/api/subscriptions.ts
```diff
@@ -45,7 +48,9 @@ router.post("subscriptions.list", auth(), pagination(), async (ctx) => {
 
   ctx.body = {
     pagination: ctx.state.pagination,
-    data: subscriptions.map(presentSubscription),
+    data: subscriptions
+      .filter((subscription) => can(user, "read", subscription))
+      .map(presentSubscription),
```
tommoor, line 53:
```suggestion
    data: subscriptions.map(presentSubscription),
```

server/routes/api/subscriptions.ts
```diff
@@ -27,13 +26,17 @@ router.post("subscriptions.list", auth(), pagination(), async (ctx) => {
   );
 
   const subscriptions = await sequelize.transaction(async (transaction) => {
```
tommoor, line 28: There's not really any value to adding transactions to read operations if nothing is getting updated, feel free to remove

server/routes/api/subscriptions.ts
```diff
@@ -62,24 +67,25 @@ router.post("subscriptions.info", auth(), async (ctx) => {
     "Not a valid subscription event for documents"
   );
 
-  const subscriptions = await sequelize.transaction(async (transaction) => {
```
tommoor: Same comment RE transaction

server/routes/api/subscriptions.ts
```diff
@@ -62,24 +67,25 @@ router.post("subscriptions.info", auth(), async (ctx) => {
...
+  const subscription = await sequelize.transaction(async (transaction) => {
     const document = await Document.findByPk(documentId, { transaction });
```
tommoor, line 71:
```suggestion
    const document = await Document.findByPk(documentId, { userId: user.id, transaction });
```

I'm surprised this works without `userId` being passed in here – it probably wouldn't on a private collection

server/queues/processors/NotificationsProcessor.ts
CuriousCorrelation: Indeed should

server/routes/api/subscriptions.ts
CuriousCorrelation: Ah I assumed passing in `userId` would return documents created by/touched by user, but looking into `Document` model, it seems `findByPk` has custom implementation that checks for permission rather than narrowing down search.

tommoor: It doesn't _check_ the permission – it makes sure that the related data is returned in the query so that `authorize` can work correctly against the user in all scenarios.

CuriousCorrelation: Should the same apply here as well?
https://github.com/outline/outline/blob/ff4bb1ecbd8b075d7bf82bcd3f2b6997a0c51413/server/routes/api/subscriptions.ts#L106

server/commands/subscriptionCreator.ts
CuriousCorrelation: Gotcha! This applies everywhere subscription might emit an event, would that be correct? 

I am assuming `actorId` should be the _indicator_ of user's identity.

tommoor: Actually, I'm wrong sorry leave `userId` here as it's on the Subscription model. apologies

CuriousCorrelation: Understood, no worries :+1:

server/routes/api/subscriptions.ts
CuriousCorrelation: > It doesn't _check_ the permission – it makes sure that the related data is returned in the query so that `authorize` call can work correctly against the user in all scenarios.

Ah that makes sense

Now I understand what you meant by this
> surprised this works without `userId`

Thanks for the explanation

tommoor (approved):

tommoor:
Just realized, need to run the backfill script from migration for community edition

tommoor:
On second thoughts, it can be manually ran.

PR #3841, opened 2022-07-22
chore: Remove updatedAt column from events table

### LINKED ISSUE #3826: Drop `updatedAt` column on events
This table is considered to be immutable and massive, we might as well stop storing this column and save some space.

apoorv-mishra:
Fixes #3826

### DIFF
--- a/server/migrations/20220722184916-remove-event-updatedat.js
+++ b/server/migrations/20220722184916-remove-event-updatedat.js
@@ -1,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.removeColumn("events", "updatedAt");
+  },
+  async down (queryInterface, Sequelize) {
+    await queryInterface.addColumn("events", "updatedAt", {
+      type: Sequelize.DATE,
+      allowNull: false,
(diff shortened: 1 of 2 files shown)


server/migrations/20220722184916-remove-event-updatedat.js
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.removeColumn("events", "updatedAt");
+  },
+  async down (queryInterface, Sequelize) {
+    await queryInterface.addColumn("events", "updatedAt", {
+      type: Sequelize.DATE,
+      allowNull: false,
+      defaultValue: new Date(),
```
tommoor, line 11:
```suggestion
      defaultValue: Sequelize.fn('NOW'),
```

I believe this would be more semantically correct, of course there are no plans to reverse this migration

tommoor (commented):
Will test for unforseen side effects, but looks good 🙏

server/migrations/20220722184916-remove-event-updatedat.js
apoorv-mishra: Fixed! Thanks for the tip :pray:

PR #3855, opened 2022-07-24
feat(app): Add `Empty trash` button to `Trash` scene

### LINKED ISSUE #3848: Empty trash functionality
**Expected behavior**
On the trash page…

- Add an "Empty trash" button in the top right
- It should only be visible to admins
- It should have a confirmation modal "Are you sure you want to permanently delete all documents in the trash? This action cannot be undone."

CuriousCorrelation:
# Description 

Adds an `Empty trash` button to the `Trash` scene. This button will only be visible to team admins.

| Admin View  | User View |
| ------------- | ------------- |
| Trash icon  | Trash icon user  |

# Issue

Closes #3848


app/menus/EmptyTrashMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <Button
+        icon={<TrashIcon />}
+        danger={true}
+        onClick={() => setShowModal(true)}
+      >
+        {t("Empty trash")}
+      </Button>
+      {showModal && (
+        <Modal
+          title={t("Empty trash")}
```
CuriousCorrelation, line 61: Should this be _more_ descriptive?

CuriousCorrelation:
95a734b11e7c74df402d9698a6a34aeb699a83d6 introduces model message for situations where user clicks `Empty trash` when trash is empty already, at the cost of _more than a little_ complexity. Not sure if it is wroth it :thinking:

app/menus/EmptyTrashMenu.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [showModal, setShowModal] = React.useState(false);
+  const [isDeleting, setIsDeleting] = React.useState(false);
+  const { showToast } = useToasts();
+  const history = useHistory();
+
+  const [trashed, setTrashed] = React.useState<Document[]>([]);
+
+  React.useEffect(() => {
+    async function getTrashed() {
+      const trashedDocs = await documents.fetchDeleted();
```
tommoor, line 26: I think in hindsight the ticket should have been more detailed, while this will work on smaller amounts of docs it has a couple of problems:

- Many serial round-trips to the server
- The documents will gradually disappear from view as the iteration happen
- If there are more than a single page of deleted documents then it will not actually empty the trash

This can all be resolved by changing the implementation so there is an API endpoint to delete all of the documents in one go

app/menus/EmptyTrashMenu.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    }}
+                  />
+                </Text>
+                <Button type="submit" danger>
+                  {isDeleting ? `${t("Deleting")}…` : t("I’m sure – Delete")}
+                </Button>
+              </form>
+            </Flex>
+          ) : (
+            <Flex column>
```
tommoor, line 94: I don't see why this case is needed? Just don't show the button

app/menus/EmptyTrashMenu.tsx
```diff
@@ -0,0 +1,111 @@
...
+      }
+    },
+    [documents, history, showToast, t, trashed]
+  );
+
+  return (
+    <>
+      <Button
+        icon={<TrashIcon />}
+        danger={true}
```
tommoor, line 65: Neutral button without an icon is okay here, doesn't need attention drawn

app/menus/EmptyTrashMenu.tsx
CuriousCorrelation: Very valid points. 

It looks like most of the infrastructure requirements are already in place for this feat. `documentPermanentDeleter` also takes in `Document[]`.

So in brief - 
1. add a new route to `routes/api/documents`, 
2. make sure `user` is `admin`, 
3. find all trashed documents (with `paranoid: false`)
4. call `documentPermanentDeleter`
5. emit event(s) (single `documents.trash_empty` or multiple `documents.permanent_delete`?)

> API endpoint to delete all of the documents

What should we call the endpoint? `documents.empty_trash` maybe?

tommoor: Yes, use `findAllInBatches` and call `documentPermanentDeleter` as many times as needed – always need to protect against very large queries, someone will put 30,000 issues in trash.

> What should we call the endpoint? 

`documents.empty_trash` works for me

shared/validations.ts
```diff
@@ -30,6 +30,9 @@ export const CollectionValidation = {
 export const DocumentValidation = {
   /** The maximum length of the document title */
   maxTitleLength: 100,
+
+  /** The maximum number of documents allowed to be permanently deleted in one batch */
+  emptyTrash: 100,
```
CuriousCorrelation, line 35: Thought 100 would be a reasonable limit.

server/routes/api/documents.test.ts
```diff
@@ -2492,3 +2492,110 @@ describe("#documents.unpublish", () => {
...
+    expect(events.length).toEqual(1);
+    expect(events[0].name).toEqual("documents.permanent_delete");
+    expect(events[0].documentId).toEqual(document.id);
+    expect(events[0].collectionId).toEqual(document.collectionId);
+    expect(events[0].teamId).toEqual(document.teamId);
+    expect(events[0].actorId).toEqual(admin.id);
+    expect(events[0].data.title).toEqual(document.title);
+  });
+
+  it("should permenantly delete trashed drafts", async () => {
```
CuriousCorrelation, line 2558: This was added during testing, we could remove it now if needed. 

Context - I couldn't figure out why drafts weren't getting deleted. After a long goose chase I saw that `Document.findAll` was scoped to `publishedAt: [Op.ne]: null`. It should have been my the first thought! :sweat:

server/routes/api/documents.ts
```diff
@@ -1063,6 +1064,64 @@ router.post("documents.delete", auth({ member: true }), async (ctx) => {
...
+      where: {
+        teamId: user.teamId,
+        deletedAt: {
+          [Op.ne]: null,
+        },
+      },
+      paranoid: false,
+      limit: DocumentValidation.emptyTrash,
+    },
+    async (documents) => {
```
CuriousCorrelation, line 1083: I think there's a opportunity to speed this up with `Promise.all` after a filter on `authorize` and `documentPermanentDeleter(documents)`

app/menus/EmptyTrashMenu.tsx
CuriousCorrelation: Let me know if I am on the right track with these latest commits!

server/routes/api/documents.test.ts
```diff
@@ -2492,3 +2492,84 @@ describe("#documents.unpublish", () => {
...
+    const res = await server.post("/api/documents.empty_trash");
+    expect(res.status).toEqual(401);
+  });
+
+  it("should permenantly delete trashed documents", async () => {
+    const { admin, document } = await seed();
+    // To simulate deleting draft
+    document.publishedAt = null;
+    await document.save();
+    await document.delete(admin.id);
```
CuriousCorrelation, line 2517: Context - I couldn't figure out why drafts weren't getting deleted. Turns out it was `Document.findAll` being scoped to `publishedAt: [Op.ne]: null`. Since we already require `admin` privileges, I kept `findAll` _unscoped_ for `empty_trash` endpoint .

app/menus/EmptyTrashMenu.tsx
CuriousCorrelation: Sometimes setting PR to draft suppresses notification so @tommoor just in case. Could use your opinion on the data transfer from [here](https://github.com/outline/outline/blob/75d1329c2e548a897c130476a65050f9742e5b83/server/routes/api/documents.ts#L1121) to [here](https://github.com/outline/outline/blob/75d1329c2e548a897c130476a65050f9742e5b83/app/stores/DocumentsStore.ts#L691). I was pondering splitting it into two endpoints - one would fetch info about imminently deleted docs and other would perform the action.

PR #3857, opened 2022-07-24
feat: Put request rate limit at application server

### LINKED ISSUE #3845: Add rate limiting middleware across all API endpoints
Can start with a high rate limit across all endpoints and throttle down individually as required.

Will need some research into if any existing middleware packages are capable enough.

apoorv-mishra:
This PR adds rate limiter implementation at application level. It provides,

1. A default blanket rate limiter, blocking at IP address level.
2. Ability to add custom, route specific rate limiters.

Closes #3845


server/middlewares/rateLimiter.ts
```diff
@@ -0,0 +1,42 @@
...
+    const limiter = new RateLimiterRedis(opts);
+
+    try {
+      await limiter.consume(ctx.ip);
+    } catch (rateLimiterRes) {
+      ctx.set("Retry-After", `${rateLimiterRes.msBeforeNext / 1000}`);
+      ctx.set("X-RateLimit-Limit", `${opts.points}`);
+      ctx.set("X-RateLimit-Remaining", `${rateLimiterRes.remainingPoints}`);
+      ctx.set(
+        "X-RateLimit-Reset",
```
tommoor, line 33:
```suggestion
      ctx.set("RateLimit-Limit", `${opts.points}`);
      ctx.set("RateLimit-Remaining", `${rateLimiterRes.remainingPoints}`);
      ctx.set(
        "RateLimit-Reset",
```

`X` prefix no longer recommended on headers

tommoor (commented):
This is a good foundation, thank you. What I'm trying to figure out is how we can structure this in such a way that custom limits can be define on the route and not here in a central file – I think it'll be a much better DX and more maintainable that way.

It could be that there is another middleware that "registers" the route -> rate limit configuration in a central mapping upon execution and this is then used by the middleware you have here. 

Does that make sense to you?

.env.sample
```diff
@@ -168,3 +168,6 @@ SMTP_SECURE=true
 # The default interface language. See translate.getoutline.com for a list of
 # available language codes and their rough percentage translated.
 DEFAULT_LANGUAGE=en_US
+
+# Optionally enable rate limiter at application web server
+ENABLE_RATE_LIMITER=false
```
tommoor, line 173: Good default for now, although I'll probably make it default to enabled in the future

apoorv-mishra:
> This is a good foundation, thank you. What I'm trying to figure out is how we can structure this in such a way that custom limits can be define on the route and not here in a central file – I think it'll be a much better DX and more maintainable that way.
> 
> It could be that there is another middleware that "registers" the route -> rate limit configuration in a central mapping upon execution and this is then used by the middleware you have here.
> 
> Does that make sense to you?

Perfectly! I had been thinking along similar lines and this totally resonates. I'll try to wire it up today.

server/middlewares/rateLimiter.ts
apoorv-mishra: Hmm...Didn't know that. Will patch it up.

tommoor: https://datatracker.ietf.org/doc/html/rfc6648

apoorv-mishra: Fixed!

apoorv-mishra:
> This is a good foundation, thank you. What I'm trying to figure out is how we can structure this in such a way that custom limits can be define on the route and not here in a central file – I think it'll be a much better DX and more maintainable that way.
> 
> It could be that there is another middleware that "registers" the route -> rate limit configuration in a central mapping upon execution and this is then used by the middleware you have here.
> 
> Does that make sense to you?

Modified the implementation as per :point_up: 

So far I've tested it by nitpicking a request and replaying it manually. It seems to work fine i.e, the route level rate limiter kicks in when registered and the default(fallback) rate limiter continues to work in absence of any route specific rate limiter.

.env.sample
```diff
@@ -168,3 +168,10 @@ SMTP_SECURE=true
...
+
+# Configure default throttling paramaters for rate limiter
+MAX_ALLOWED_REQUESTS=
+DURATION_WINDOW_IN_SECONDS=
```
tommoor, line 177: Remember these are global, `DURATION_WINDOW_IN_SECONDS` is very ambiguous. We should namespace these.

`RATE_LIMITER_ENABLED`
`RATE_LIMITER_DURATION_WINDOW`
`RATE_LIMITER_REQUESTS`

server/index.ts
```diff
@@ -94,6 +96,18 @@ async function start(id: number, disconnect: () => void) {
   onerror(app);
   app.on("error", requestErrorHandler);
 
+  // set a global rate limiter map which maps route path
```
tommoor, line 99: I think functionally this is all good but it has all ended up quite scattered around the codebase. It would be nice to keep everything rate-limiter related centralized in a single module.

There isn't a great existing pattern for this, for now perhaps just a file `server/RateLimiter.ts` that can contain the mapping, default limiter, and shutdown handler rather than being defined on context. Something like:

```
class RateLimiter {
  // mapping internal to class
  
  static getRateLimiter(path: string) {
    // returns default or custom limiter
  }
  
  static setRateLimiter(path, config) {
    // sets mapping
  }
  
  static shutdownHandler() {
   // logic from shutdown hook
  }
}
```

server/index.ts
```diff
@@ -122,6 +136,35 @@ async function start(id: number, disconnect: () => void) {
...
+    // configurations would immediately come into effect
+    // on server restart. As an example,
+    // the `duration` property doesn't change
+    // for already created keys, but affects all new keys.
+    //
+    // Refer https://github.com/animir/node-rate-limiter-flexible/wiki/Options#clearexpiredbytimeout
+    Logger.info("lifecycle", `Deleting all rate limiter specific keys...`);
+
+    const scanStream = Redis.defaultClient.scanStream({
+      match: "rlflx:*",
```
tommoor, line 150: `rlflx` should be a constant as it's loosely coupled here that all the keys begin with this, why "rlflx" btw?

server/index.ts
```diff
@@ -122,6 +136,35 @@ async function start(id: number, disconnect: () => void) {
...
+      if (!isEmpty(keys)) {
+        Redis.defaultClient.unlink(keys, (err) => {
+          if (err) {
+            Logger.error("lifecycle", err);
+          }
+        });
+      }
+    });
+
+    scanStream.on("end", () => {
```
tommoor, line 164: Does server shutdown need to wait on this completing…

server/routes/api/middlewares/registerRateLimiter.ts
```diff
@@ -0,0 +1,18 @@
+import { Context, Next } from "koa";
+import { RateLimiterRedis } from "rate-limiter-flexible";
+import { RateLimiterConfig } from "@server/types";
+
+export default function registerRateLimiter(
```
tommoor, line 5: Could put this in the file with other middleware to be honest as they are highly coupled

server/index.ts
apoorv-mishra: Sure! Will iterate to organize it.

server/index.ts
apoorv-mishra: No particular reason. I borrowed it from [here](https://github.com/animir/node-rate-limiter-flexible/wiki/Options#keyprefix)(looks like it's a shorter form of `node-rate-limiter-flexible` :smile: ). We should change it since it doesn't convey much about itself. Should we make it `rate_limiter`?

server/index.ts
apoorv-mishra: Not necessarily. As long as the existing keys are flushed, we should be good. We can remove this.

server/index.ts
tommoor: Maybe just `rl:` is fine, every byte counts in keeping redis memory usage low

server/routes/api/middlewares/registerRateLimiter.ts
apoorv-mishra: Sure! Will follow it up as part of the refactor.

.env.sample
apoorv-mishra: Makes sense. Will push this.

apoorv-mishra: Fixed!

server/index.ts
apoorv-mishra: Fixed! Thanks for raising this :raised_hands:

server/index.ts
apoorv-mishra: Fixed!

server/routes/api/middlewares/registerRateLimiter.ts
apoorv-mishra: Fixed!

server/middlewares/rateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+  return async function registerRateLimiterMiddleware(
+    ctx: Context,
+    next: Next
+  ) {
+    if (!env.RATE_LIMITER_ENABLED) {
+      return next();
+    }
+
+    if (!RateLimiter.hasRateLimiter(ctx.path)) {
+      // kill the default rate limiter for this path first
```
tommoor, line 46: I'm sure it is, but it's not clear to me why this is required – I'm _guessing_ you were finding it hard to debug turning route-specific limits on/off?

apoorv-mishra: My bad! Should've explained the _why_.

> I'm guessing you were finding it hard to debug turning route-specific limits on/off?

In a way, yes. The new route-specific configuration doesn't apply unless the old keys in Redis are flushed. One way to solve this is to flush the old keys. Another solution is to specify a different `keyPrefix` from that of the default rate limiter. In the latter case, however, the old keys are not flushed forcibly, albeit upon `ttl` expiry.

server/RateLimiter.ts
```diff
@@ -0,0 +1,73 @@
...
+  constructor() {
+    throw Error(`Cannot instantiate class!`);
+  }
+
+  static readonly rateLimiterMap = new Map<string, RateLimiterRedis>();
+  static readonly defaultRateLimiter = new RateLimiterRedis({
+    storeClient: Redis.defaultClient,
+    points: env.RATE_LIMITER_REQUESTS,
+    duration: env.RATE_LIMITER_DURATION_WINDOW,
+    keyPrefix: env.RATE_LIMITER_REDIS_KEY_PREFIX,
```
tommoor, line 18: No need for this to be in `env`, it doesn't make sense to configure it separately. Just a private const on RateLimiter would be fine

server/RateLimiter.ts
```diff
@@ -0,0 +1,73 @@
...
+      if (!isEmpty(keys)) {
+        Redis.defaultClient.unlink(keys, (err) => {
+          if (err) {
+            Logger.error("lifecycle", err);
+          }
+        });
+      }
+    });
+
+    scanStream.on("end", () => {
```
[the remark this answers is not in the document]
tommoor: I suppose I was half-heartedly asking if the server should be awaiting this operation before shutting down, otherwise the process might be killed before the scanStream even returns data.

If that's the case then I'm not sure we have a pattern for that right now, It's not super important though to be honest, I think there is a world where this is only actually used in development…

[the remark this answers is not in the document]
apoorv-mishra: > otherwise the process might be killed before the scanStream even returns data.

True! I've already observed it a couple of times in dev. We have the option to either do _fire and forget_ or not do it at all. There are apparent individual trade offs in each of the two approaches. In the former, we tend to trade off _performance_ (consider having large number of keys) for _determinism_(new rate limiter configurations kick in immediately upon server restart). And, in the latter, it's vice versa.

That being said, I tend to favor the latter one since _determinism_ in the former is still not a guarantee because of what you pointed out. If you agree with me, I can get rid of it. We can put it back later once we have a pattern in place to handle shutdown gracefully.

server/middlewares/rateLimiter.ts
apoorv-mishra: >True! I've already observed it a couple of times in dev. We have the option to either do fire and forget or not do it at all. There are apparent individual trade offs in each of the two approaches. In the former, we tend to trade off performance (consider having large number of keys) for determinism(new rate limiter configurations kick in immediately upon server restart). And, in the latter, it's vice versa.

>That being said, I tend to favor the latter one since determinism in the former is still not a guarantee because of what you pointed out. If you agree with me, I can get rid of it. We can put it back later once we have a pattern in place to handle shutdown gracefully.

Quoting from [here](https://github.com/outline/outline/pull/3857#discussion_r938453895). If we're fine sacrificing determinism here as well, I can get rid of this too. Let me know your thoughts.

server/middlewares/rateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+export function rateLimiter() {
+  return async function rateLimiterMiddleware(ctx: Context, next: Next) {
+    if (!env.RATE_LIMITER_ENABLED) {
+      return next();
+    }
+
+    const limiter = RateLimiter.getRateLimiter(ctx.path);
+
+    try {
+      await limiter.consume(`${ctx.path}:${ctx.ip}`);
```
apoorv-mishra, line 19: Wondering if using this as a key for the default rate limiter is a good idea :thinking: 

This will create keys corresponding to the default rate limiter in the format `rl:<path>:<ip>`. Although, having `path` makes sense for route specific rate limiters, it seems unnecessary for the default one. For the default one, `rl:<ip>` should suffice. Besides, we would reclaim some space too. Let me know your thoughts @tommoor .

server/RateLimiter.ts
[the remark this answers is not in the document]
apoorv-mishra: Removed this.

server/RateLimiter.ts
tommoor: Lets set the rate limiter window to 60s too, much easier to reason about

server/middlewares/rateLimiter.ts
apoorv-mishra: Modified as per [`5d11311`](https://github.com/outline/outline/pull/3857/commits/5d1131126e789ccf422f6e862bb45fa6ee09f5cc).

tommoor:
I suspect this is quite mergable now as the rate limiter needs to be explicitly enabled. Next step will be to start putting some tighter limits on specific paths.

server/env.ts
```diff
@@ -495,6 +495,39 @@ export class Environment {
...
+  /**
+   * Set max allowed requests in a given duration for
+   * default rate limiter to trigger throttling.
+   */
+  @IsOptional()
+  @IsNumber()
+  @CannotUseWithout("RATE_LIMITER_ENABLED")
+  public RATE_LIMITER_REQUESTS = this.toOptionalNumber(
+    process.env.RATE_LIMITER_REQUESTS ?? "5000"
+  );
```
tommoor, line 517:
```suggestion
  public RATE_LIMITER_REQUESTS = this.toOptionalNumber(
    process.env.RATE_LIMITER_REQUESTS
  ) ?? 5000;
```

Maybe change these two numbers so they are not typed as optional, like: https://github.com/outline/outline/blob/9d792ef1cfb5082860de7eee78d14e820c131278/server/env.ts#L241

server/env.ts
```diff
@@ -495,6 +495,39 @@ export class Environment {
...
+   * Set fixed duration window(in secs) for
+   * default rate limiter, elapsing which the request
+   * quota is reset(the bucket is refilled with tokens).
+   */
+  @IsOptional()
+  @IsNumber()
+  @CannotUseWithout("RATE_LIMITER_ENABLED")
+  public RATE_LIMITER_DURATION_WINDOW = this.toOptionalNumber(
+    process.env.RATE_LIMITER_DURATION_WINDOW ?? "60"
+  );
```
tommoor, line 529:
```suggestion
  public RATE_LIMITER_DURATION_WINDOW = this.toOptionalNumber(
    process.env.RATE_LIMITER_DURATION_WINDOW
  ) ?? 60;
```

server/env.ts
apoorv-mishra: Fixed!

server/env.ts
apoorv-mishra: Fixed!

apoorv-mishra:
> Next step will be to start putting some tighter limits on specific paths.

Should we do it as part of this PR itself?

tommoor:
> Should we do it as part of this PR itself?

Oh no, not required 👍

tommoor (approved):

PR #3872, opened 2022-07-26
feat: mount app on env.URL.pathname

frankli0324:
https://github.com/outline/outline/discussions/3871

interesting fact:

https://github.com/outline/outline/blob/086c3ec2d83d6603f8834b0a2ef5229f9638bef4/server/presenters/env.ts#L12-L14

where:

https://github.com/outline/outline/blob/086c3ec2d83d6603f8834b0a2ef5229f9638bef4/app/scenes/Document/components/MultiplayerEditor.tsx#L69-L74

### DIFF
--- a/app/components/PageTitle.tsx
+++ b/app/components/PageTitle.tsx
@@ -1,7 +1,7 @@
 import { observer } from "mobx-react";
 import * as React from "react";
 import { Helmet } from "react-helmet";
-import { cdnPath } from "@shared/utils/urls";
+import { publicPath } from "@shared/utils/urls";
 import useStores from "~/hooks/useStores";
 
 type Props = {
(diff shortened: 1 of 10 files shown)


frankli0324:
I'm not sure what to do with `WebpackPwaManifest` in webpack.config.js

tommoor:
> I'm not sure what to do with WebpackPwaManifest

I'm not sure there is anything you _can_ do as webpack is build-time and the `URL` configuration is runtime, it may be that the PWA manifest just isn't going to work in this configuration

tommoor:
One way to support both might be to create a `relativePath` helper that would parse `env.URL` at startup and prefix the pathname, eg.

```
env.URL = "https://mycompany.com/wiki"

relativePath("/static/worker.js") -> "/wiki/static/worker.js"

```

frankli0324:
hmmmm.... what troubles me most is `index.html` gets generated with `{publicPath}/app.[hash].js,css`

frankli0324:
> The custom domain name functionality relies on many of the paths that you've changed to absolute being relative for this to work correctly.

oh....

tommoor:
Not sure if you wanted to take this one across the finish line based on the above suggestion?

frankli0324:
I wish so, but I'm quite busy in the last few weeks... maybe next weekend..?

PR #3873, opened 2022-07-26
chore: Add emailed confirmation code to account deletion

### LINKED ISSUE #3844: Add confirmation step to account deletion
**Desired behavior**

- Go to account delete
- Click "Request account deletion" or similar
- Generate a short code and email it to them (new email type)
- They must re-enter the matching code into the page which will be passed to the `users.delete` endpoint in order to trigger account deletion.

I think the code can be generated in such a way that it uses the jwtSecret as a seed – this has the advantage of being unique per user/installation, revoking the delete code if the user logs out, not requiring another database column that will be empty for all active users.

Possibly – hash the JWT, strip the first 6 characters and use those?

tommoor:
- Code is only required when installation has SMTP enabled

closes #3844


server/routes/api/users.ts
```diff
@@ -367,19 +370,41 @@ router.post("users.resendInvite", auth(), async (ctx) => {
   };
 });
 
-router.post("users.delete", auth(), async (ctx) => {
-  const { id } = ctx.body;
-  const actor = ctx.state.user;
-  let user = actor;
+router.post("users.requestDelete", auth(), async (ctx) => {
```
tommoor, line 373: Once #3845 merges we'll add a tight rate limit on this endpoint

server/routes/api/users.ts
```diff
@@ -367,19 +370,43 @@ router.post("users.resendInvite", auth(), async (ctx) => {
...
+  authorize(user, "delete", user);
+
+  if (emailEnabled) {
+    await ConfirmUserDeleteEmail.schedule({
+      to: user.email,
+      deleteConfirmationCode: user.deleteConfirmationCode,
+    });
+  }
 
-  if (id) {
```
tommoor, line 375: Removed ability to delete another user for now

PR #3876, opened 2022-07-26
fix: Improve document delete confirmation message

### LINKED ISSUE #3843: Improve delete document confirmation modal
After some customer mistakes it's become apparent that this confirmation dialog wording needs tweaking – we should include a count of how many nested documents will also be deleted in the message.

apoorv-mishra:
Modify document delete confirmation message to warn about the number of expected nested documents which will be deleted if confirmed.

Fixes #3843 

Zero nested documents
Screenshot from 2022-07-26 22-57-51

One nested document
Screenshot from 2022-07-26 22-14-47

More than one nested document
Screenshot from 2022-07-26 22-14-14


apoorv-mishra:
A doubt here: I used `yarn copy:i18n` to move translations to `./build` in order for the changes to reflect in UI. There's another command `yarn build:i18n` which overwrote the translations I created in `shared/i18n/locales/en_US/translation.json` every time I ran it(before running `yarn copy:i18n`). But now, it seems to have stopped overwriting translations, maybe because I ran `yarn copy:i18n` before it? Did anyone face such an issue of translations being overwritten by the `build` command?

shared/i18n/locales/en_US/translation.json
```diff
@@ -432,7 +432,8 @@
...
   "Are you sure you want to delete the <em>{{ documentTitle }}</em> template?": "Are you sure you want to delete the <em>{{ documentTitle }}</em> template?",
-  "Are you sure about that? Deleting the <em>{{ documentTitle }}</em> document will delete all of its history and any nested documents.": "Are you sure about that? Deleting the <em>{{ documentTitle }}</em> document will delete all of its history and any nested documents.",
+  "Are you sure about that? Deleting the <em>{{ documentTitle }}</em> document will delete all of its history and <em>{{ count }} nested documents</em>._{undefined}": "Are you sure about that? Deleting the <em>{{ documentTitle }}</em> document will delete all of its history and <em>{{ count }} nested documents</em>._{undefined}",
+  "Are you sure about that? Deleting the <em>{{ documentTitle }}</em> document will delete all of its history and <em>{{ count }} nested documents</em>._{undefined}_plural": "Are you sure about that? Deleting the <em>{{ documentTitle }}</em> document will delete all of its history and <em>{{ count }} nested documents</em>._{undefined}",
```
apoorv-mishra, line 436: Oops! These aren't the expected translations :thinking: I had added different ones. These are broken. I guess these got modified by the `precommit` hook. How can we fix these? Any pointers here? @tommoor

tommoor: Yea, the translation keys are generated automatically. The only time you need to edit this file is for the plural string – I'm not sure why `undefined` is here…

apoorv-mishra: Because of [this](https://github.com/outline/outline/pull/3876/files#diff-6fd6f22c7fed61dbbebc00cac6149e16260c7416cceae777c64b19095128a1dfR103). I was trying to set a [context](https://www.i18next.com/translation-function/context) for the zero nested docs scenario.

tommoor: It may be easier to split into two separate translations in the React tree based on the way this is behaving

apoorv-mishra: Yeah! Can't think of a better approach.

apoorv-mishra: Fixed!

tommoor:
Sorry, should have been clearer 3 separate translations is a bit much – we can have one for the case where there _are_ nested documents with the plural variant and one for when there are no nested documents missing that language altogether.

apoorv-mishra:
Fixed! It took some time figuring out that although the translations are auto generated(via `yarn build:i18n`), the values still need to be modified manually as required :smile:

tommoor:
> the values still need to be modified manually as required

Ah yes, but only in the case of plurals! In a lot of software you'd just use short keys like `delete_document_warning` and then write _all_ the translations in the i18n file, but this seemed much more maintainable by a small team 😄

tommoor (approved):

PR #3907, opened 2022-08-01
fix: Harden regex for embeds by escaping

### LINKED ISSUE #3899: Regex for some embeds are not escaped
Potentially allows domains which are not supported to be embedded, more details on report on huntr:
https://huntr.dev/bounties/b37145c7-796b-4b1d-994e-d0d007b426df/

apoorv-mishra:
Fixes #3899


tommoor (approved):
Killed it

PR #3909, opened 2022-08-02
fix: refactor auth flow to explicitly pass in a host

thenanyu:
- Fixes an issue that causes custom domain auth to not respect the team in the URL.
- Refactors the way redirects are done to keep users on the same domain/host during the auth process. 

Previously, a redirect would happen on the callback from SSO before the one-time-pad was checked. This worked fine for subdomains but custom domains don't have access to the cookie containing the pad.

This PR moves the redirect logic from the callback to the authentication error catch block.

### DIFF
--- a/app/scenes/Login/AuthenticationProvider.tsx
+++ b/app/scenes/Login/AuthenticationProvider.tsx
@@ -89,11 +89,15 @@ function AuthenticationProvider(props: Props) {
     );
   }
 
-  // If we're on a custom domain then the auth must point to the root
-  // app.getoutline.com for authentication so that the state cookie can be set
-  // and read.
-  const isCustomDomain = parseDomain(window.location.origin).custom;
-  const href = `${isCustomDomain ? env.URL : ""}${authUrl}`;
+  // If we're on a custom domain or a subdomain then the auth must point to the
+  // apex (env.URL) for authentication so that the state cookie can be set and read.
+  // We pass the host into the auth URL so that the server can redirect on error
+  // and keep the user on the same page.
+  const { custom, teamSubdomain, host } = parseDomain(window.location.origin);
+  const needsRedirect = custom || teamSubdomain;
+  const href = needsRedirect
+    ? `${env.URL}${authUrl}?host=${encodeURI(host)}`
+    : authUrl;
 
   return (
     <Wrapper>
--- a/server/middlewares/passport.ts
+++ b/server/middlewares/passport.ts
@@ -3,6 +3,7 @@ import { Context } from "koa";
 import env from "@server/env";
 import Logger from "@server/logging/Logger";
 import { signIn } from "@server/utils/authentication";
+import { parseState } from "@server/utils/passport";
 import { AccountProvisionerResult } from "../commands/accountProvisioner";
 
 export default function createMiddleware(providerName: string) {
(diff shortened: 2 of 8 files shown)


server/utils/passport.ts
```diff
@@ -81,6 +65,26 @@ export class StateStore {
   };
 }
 
+// Every authentication action is routed through the apex domain.
+// But when there is an error, we want to redirect the user on the
+// same domain or subdomain that they originated from (found in state).
```
thenanyu, line 70: When an auth method catches an error from a provisioner, it should wrap it using this function and re-throw.

server/utils/passport.ts
```diff
@@ -46,24 +47,7 @@ export class StateStore {
...
-      const reqProtocol = ctx.protocol;
-      const requestHost = ctx.get("host");
-      const requestPath = ctx.originalUrl;
-      const requestUrl = `${reqProtocol}://${requestHost}${requestPath}`;
-      const url = new URL(requestUrl);
-
-      url.host = appDomain.host;
-
-      return callback(AuthRedirectError(``, url.toString()), false, token);
-    }
```
thenanyu, line 66: This was always a pretty hacky but straightforward way to keep the user on the same subdomain, but it doesn't work at all for custom domains.

server/routes/auth/providers/google.ts
```diff
@@ -124,7 +128,7 @@ if (env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET) {
 
           return done(null, result.user, result);
         } catch (err) {
-          return done(err, null);
+          return done(PassportError(err, ctx), null);
```
thenanyu, line 131: 🥦 errors thrown out of authentication providers should be wrapped with `PassportError` to properly redirect

app/scenes/Login/AuthenticationProvider.tsx
```diff
@@ -89,11 +89,15 @@ function AuthenticationProvider(props: Props) {
...
-  const href = `${isCustomDomain ? env.URL : ""}${authUrl}`;
+  // If we're on a custom domain or a subdomain then the auth must point to the
+  // apex (env.URL) for authentication so that the state cookie can be set and read.
+  // We pass the host into the auth URL so that the server can redirect on error
+  // and keep the user on the same page.
+  const { custom, teamSubdomain, host } = parseDomain(window.location.origin);
+  const needsRedirect = custom || teamSubdomain;
+  const href = needsRedirect
+    ? `${env.URL}${authUrl}?host=${encodeURI(host)}`
+    : authUrl;
```
thenanyu, line 100: We're now passing in the origin host to inform possible redirects

server/routes/auth/providers/google.ts
thenanyu: (open to better names for this thing.)

server/middlewares/passport.ts
```diff
@@ -18,12 +19,28 @@ export default function createMiddleware(providerName: string) {
 
           if (err.id) {
             const notice = err.id.replace(/_/g, "-");
-            const hasQueryString = err.redirectUrl?.includes("?");
+            const redirectUrl = err.redirectUrl ?? "/";
+            const hasQueryString = redirectUrl?.includes("?");
+
+            // Every authentication action is routed through the apex domain.
+            // But when there is an error, we want to redirect the user on the
+            // same domain or subdomain that they originated from (found in state).
```
thenanyu, line 27: 🥦 move all of the redirection logic into the middleware so that users remain on the same domain when errors occur during auth

server/middlewares/passport.ts
```diff
@@ -18,12 +19,28 @@ export default function createMiddleware(providerName: string) {
 
           if (err.id) {
             const notice = err.id.replace(/_/g, "-");
-            const hasQueryString = err.redirectUrl?.includes("?");
+            const redirectUrl = err.redirectUrl ?? "/";
```
tommoor, line 22: I swear there are 10 places in the codebase where `Url` is used instead of `Path` and it's mighty confusing, need to do an audit 🙈

tommoor (approved):

tommoor:
Gonna roll this during low traffic hours

PR #3912, opened 2022-08-02
feat: Choose random color on collection creation

### LINKED ISSUE #3900: Choose a random color if none is chosen on collection creation
If a new collection is created without a color chosen, select one from the palette automatically. Ideally this logic would live on the server side to account for collections created via API

apoorv-mishra:
Fixes #3900 

Choose a random color from a shared color palette between backend and frontend during collection creation.

### DIFF
--- a/app/scenes/CollectionNew.tsx
+++ b/app/scenes/CollectionNew.tsx
@@ -3,6 +3,8 @@ import { observable } from "mobx";
 import { observer } from "mobx-react";
 import * as React from "react";
 import { withTranslation, Trans, WithTranslation } from "react-i18next";
+import { randomElement } from "@shared/random";
+import { colorPalette } from "@shared/utils/collections";
 import { CollectionValidation } from "@shared/validations";
 import RootStore from "~/stores/RootStore";
 import Collection from "~/models/Collection";
@@ -30,7 +32,7 @@ class CollectionNew extends React.Component<Props> {
   icon = "";
 
   @observable
-  color = "#4E5C6E";
+  color = randomElement(colorPalette);
 
   @observable
   sharing = true;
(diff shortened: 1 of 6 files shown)


shared/utils/collections.ts
```diff
@@ -28,3 +28,16 @@ export const sortNavigationNodes = (
       : document.children,
   }));
 };
+
+export const colorPalette = () => [
```
tommoor, line 32:
```suggestion
export const colorPalette = [
```

Just export an array, avoids needing the indirection on `Collection.COLOR_PALETTE` then

server/models/Collection.ts
```diff
@@ -390,6 +391,12 @@ class Collection extends ParanoidModel {
     });
   }
 
+  static readonly COLOR_PALETTE = colorPalette();
+
+  static randomColor() {
```
tommoor, line 396: Feels a bit out of place as a static method on `Collection` to be honest, I think we'd be okay just using `randomElement(colorPalette)` directly on the api default. It would still be very clear :)

app/scenes/CollectionNew.tsx
```diff
@@ -30,7 +30,7 @@ class CollectionNew extends React.Component<Props> {
   icon = "";
 
   @observable
-  color = "#4E5C6E";
+  color = Collection.randomColor();
```
tommoor, line 33: Same comment here as sever side, `randomElement(colorPalette)` would be totally fine and just avoid some undocumented indirection

tommoor (commented):
Thanks!

shared/utils/collections.ts
apoorv-mishra: Fixed!

app/scenes/CollectionNew.tsx
apoorv-mishra: Fixed!

tommoor (approved):

PR #3917, opened 2022-08-03
Fix state cookie login flow for future timezones

### LINKED ISSUE #3915: Users in future timezones cannot login via OAuth
The following code adds a `state` cookie as part of the login flow: https://github.com/outline/outline/blob/6fe4b1cba19cb884f4d2c38ba0232d9d5eb8f485/server/utils/passport.ts#L25-L29

The expiry is set as 10 minutes in the future of the sever's current time, as gathered by `new Date()`. This is sent by the server to the user as follows:
`Set-Cookie: state=outline.test|blahblah; path=/; expires=Wed, 03 Aug 2022 16:01:48 GMT; domain=outline.test`

If the user is in a timezone far enough in the future, relative to the server, then the cookie will immediately expire. This renders the user unable to login.

A better method would be to set the `max-age` property on the cookie, rather than specifying a timezone-specific date.

roddux:
Use `maxAge` parameter instead of `expires` for relative expiry times when setting the `state` cookie in oauth login flow.

This fixes #3915

### DIFF
--- a/server/utils/passport.ts
+++ b/server/utils/passport.ts
@@ -24,7 +24,7 @@ export class StateStore {
 
     ctx.cookies.set(this.key, state, {
       httpOnly: false,
-      expires: addMinutes(new Date(), 10),
+      maxAge: 1000 * 60 * 10, // 10 minutes in milliseconds
       domain: getCookieDomain(ctx.hostname),
     });
 


roddux:
NB this is not needed, as all `expires` flags work in GMT timezone. Non-issue, no need for this merge.

PR #3925, opened 2022-08-04
feat: add session switching to the root action menu

thenanyu:
Moves team switching to central action menu
image


app/menus/OrganizationMenu.tsx
```diff
@@ -32,32 +31,11 @@ const OrganizationMenu: React.FC = ({ children }) => {
...
-                perform: () => (window.location.href = session.url),
-              })),
-            }),
-          ]
-        : []),
-      logout,
-    ];
-  }, [team.id, team.url, sessions, t]);
+    return [navigateToSettings, separator(), changeTeam, logout];
+  }, [team.id, sessions]);
```
thenanyu, line 38: I'm not sure if doing this is necessary or prudent

app/actions/definitions/teams.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const sessions = loadSessionsFromCookie();
+    const otherSessions = sessions.filter(
+      (session) => session.teamId !== currentTeamId
+    );
+    return otherSessions.length > 0;
+  },
+  children: ({ currentTeamId }) => {
+    const sessions = loadSessionsFromCookie();
+    const otherSessions = sessions.filter(
+      (session) => session.teamId !== currentTeamId
```
thenanyu, line 21: previously this compared both `teamId` and `url` which I don't think was necessary. The sessions are de-duped on `teamId` already
https://github.com/outline/outline/blob/4d90092797e791dd7b02e0923d2e3cd2df051c66/server/utils/authentication.ts#L74-L83

app/actions/definitions/teams.tsx
```diff
@@ -0,0 +1,40 @@
...
+  section: "Account",
+  visible: ({ currentTeamId }) => {
+    const sessions = loadSessionsFromCookie();
+    const otherSessions = sessions.filter(
+      (session) => session.teamId !== currentTeamId
+    );
+    return otherSessions.length > 0;
+  },
+  children: ({ currentTeamId }) => {
+    const sessions = loadSessionsFromCookie();
```
tommoor, line 19: Ah this a shame, I was going to great effort with the hook before to only load and parse the cookie when needed, now it will be much more. Maybe it's negligible …

tommoor (approved):

PR #3935, opened 2022-08-07
Add optional export notifications

codemicro:
See: #3934

This adds a notification event called `emails.export_completed`. When a user is subscribed to this event, both success and failure emails can be sent when an export they requested is completed. Neither can be sent if the user is not subscribed to the event.

A toggle has been added to the notifications settings page for `emails.export_completed`. New users are automatically subscribed to this event when their accounts are created.

The secondary text on the export page in settings has been updated to reference how email notifications are optional, as has the toast shown when exporting a collection.

:)

### DIFF
--- a/server/emails/templates/ExportFailureEmail.tsx
+++ b/server/emails/templates/ExportFailureEmail.tsx
@@ -23,8 +23,8 @@ export default class ExportFailureEmail extends BaseEmail<Props> {
   protected async beforeSend({ userId, teamId }: Props) {
     const notificationSetting = await NotificationSetting.findOne({
       where: {
-        userId: userId,
-        teamId: teamId,
+        userId,
+        teamId,
         event: "emails.export_completed",
       },
     });
--- a/server/emails/templates/ExportFailureEmail.tsx
+++ b/server/emails/templates/ExportFailureEmail.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { NotificationSetting } from "@server/models";
 import BaseEmail from "./BaseEmail";
 import Body from "./components/Body";
 import Button from "./components/Button";
@@ -10,13 +11,27 @@ import Heading from "./components/Heading";
 
 type Props = {
   to: string;
+  userId: string;
   teamUrl: string;
+  teamId: string;
 };
 
 /**
  * Email sent to a user when their data export has failed for some reason.
  */
 export default class ExportFailureEmail extends BaseEmail<Props> {
+  protected async beforeSend({ userId, teamId }: Props) {
+    const notificationSetting = await NotificationSetting.findOne({
+      where: {
+        userId: userId,
+        teamId: teamId,
+        event: "emails.export_completed",
+      },
+    });
+
+    return notificationSetting !== null;
+  }
+
   protected subject() {
     return "Your requested export";
   }
--- a/server/emails/templates/ExportSuccessEmail.tsx
+++ b/server/emails/templates/ExportSuccessEmail.tsx
@@ -25,8 +25,8 @@ export default class ExportSuccessEmail extends BaseEmail<Props> {
   protected async beforeSend({ userId, teamId }: Props) {
     const notificationSetting = await NotificationSetting.findOne({
       where: {
-        userId: userId,
-        teamId: teamId,
+        userId,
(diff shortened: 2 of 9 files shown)


server/models/User.ts
```diff
@@ -536,6 +537,14 @@ class User extends ParanoidModel {
         },
         transaction: options.transaction,
       }),
+      NotificationSetting.findOrCreate({
```
tommoor, line 540: Lets only create this if the user is an admin as regular users cannot export anyway it'll reduce down on unused rows being written.

server/queues/tasks/ExportMarkdownZipTask.ts
```diff
@@ -70,20 +87,24 @@ export default class ExportMarkdownZipTask extends BaseTask<Props> {
         url,
       });
 
-      await ExportSuccessEmail.schedule({
-        to: user.email,
-        id: fileOperation.id,
-        teamUrl: team.url,
-      });
+      if (sendEmail) {
```
tommoor, line 90: Could we move this unto a `beforeSend` block on the emails themselves, see here as an example:

https://github.com/outline/outline/blob/863f22750ffb86fec628901a8380d285e5c12ab3/server/emails/templates/InviteAcceptedEmail.tsx#L27

Trying to go this way in the future as it feels more maintainable (the logic follows the email around irregardless of if it's triggered from multiple spots in the future.

tommoor (commented):
Seems like you figured this one out pretty fast! Just two things and one is more stylistic than functional.

Thanks!

server/queues/tasks/ExportMarkdownZipTask.ts
codemicro: Sounds good.

Is is okay to omit `teamId` from the query, or is this neccessary?

```ts
where: {
    userId: id,
    teamId: fileOperation.teamId, // omitting this line
    event: "emails.export_completed",
},

**Comment by codemicro** [comment]
> Seems like you figured this one out pretty fast!

It turned out to be quite a lot simpler than I thought it was from my cursory glance at the codebase.

I'll implement the changes in a few days time, since I'm going on holiday today.

File: server/queues/tasks/ExportMarkdownZipTask.ts (hunk 2 above)
Comment (tommoor) [reply server/queues/tasks/ExportMarkdownZipTask.ts:90]: Better to keep it, as the index is `team_id_user_id`, which makes me think that code above that I referenced really needs changing. You'll have to pass `teamId` to the email.schedule call but that's okay.

**Comment by codemicro** [comment]
I've tweaked those things

File: server/emails/templates/ExportFailureEmail.tsx [hunk 3]
```diff
@@ -10,13 +11,27 @@ import Heading from "./components/Heading";
...
 
 /**
  * Email sent to a user when their data export has failed for some reason.
  */
 export default class ExportFailureEmail extends BaseEmail<Props> {
+  protected async beforeSend({ userId, teamId }: Props) {
+    const notificationSetting = await NotificationSetting.findOne({
+      where: {
+        userId: userId,
+        teamId: teamId,
```
tommoor, line 27:
```suggestion
        userId,
        teamId,
```

server/emails/templates/ExportSuccessEmail.tsx
```diff
@@ -10,15 +11,29 @@ import Heading from "./components/Heading";
...
 /**
  * Email sent to a user when their data export has completed and is available
  * for download in the settings section.
  */
 export default class ExportSuccessEmail extends BaseEmail<Props> {
+  protected async beforeSend({ userId, teamId }: Props) {
+    const notificationSetting = await NotificationSetting.findOne({
+      where: {
+        userId: userId,
+        teamId: teamId,
```
tommoor, line 29:
```suggestion
        userId,
        teamId,
```

server/emails/templates/ExportFailureEmail.tsx
tommoor: I need to add a lint rule for these so it fixes automatically :)

tommoor (approved):

tommoor:
Thank you for the contribution

codemicro:
Thanks for the help! :)

PR #3964, opened 2022-08-12
feat: allow ad-hoc creation of new teams

thenanyu:
Changes "team" to "workspace" in all user-facing strings
image

image

We're using the normal modal with a single team name field.
image


thenanyu:
I did something weird with git and it it included the changes to actions I made earlier — but it only has minor changes in that area

app/actions/definitions/teams.tsx
```diff
@@ -1,40 +1,62 @@
...
+  children: ({ currentTeamId, isContextMenu }) => {
+    return [
+      ...getOtherSessions(currentTeamId).map((session) => ({
+        id: session.url,
+        name: session.name,
+        section: "Team",
+        icon: <Logo alt={session.name} src={session.logoUrl} />,
+        perform: () => (window.location.href = session.url),
+      })),
+      ...(isContextMenu ? [createTeam] : []),
```
thenanyu, line 26: is there a more elegant way to express this logic?

app/actions/definitions/teams.tsx
```diff
@@ -1,40 +1,62 @@
...
+}
+
 const Logo = styled("img")`
   border-radius: 2px;
   width: 24px;
   height: 24px;
 `;
 
-export const rootTeamActions = [changeTeam];
+export const rootTeamActions = [changeTeam, createTeam];
```
thenanyu, line 62: image
I think it's nicer to have the "new team" item be top level in the command menu, but the logic is slightly more annoying

thenanyu: it's nested in the context menu 
image

server/migrations/20220810185000-scope-provider-id-uniqueness-to-team.js
```diff
@@ -0,0 +1,21 @@
...
+
+  async down (queryInterface, Sequelize) {
+    await queryInterface.removeConstraint("authentication_providers", "authentication_providers_providerId_teamId_uk");
+    await queryInterface.addConstraint("authentication_providers", {
+      type: 'unique',
+      fields: ["providerId"],
+      name: "authentication_providers_providerId_key"
+    });
+  }
+};
```
thenanyu, line 21: 🥦 this is a continuation of the logic change we started in previous PRs, where auth providers are scoped to the team instead of globally. All of the lookups are already team-scoped, so releasing the constraint doesn't change any existing behavior

server/presenters/team.ts
```diff
@@ -16,6 +16,6 @@ export default function present(team: Team) {
     url: team.url,
     defaultUserRole: team.defaultUserRole,
     inviteRequired: team.inviteRequired,
-    allowedDomains: team.allowedDomains.map((d) => d.name),
+    allowedDomains: team.allowedDomains?.map((d) => d.name),
```
thenanyu, line 19: I am surprised this wasn't already causing issues! Domains are not queried or scoped in for most team objects

server/routes/api/team.ts
```diff
@@ -60,4 +62,56 @@ router.post("team.update", auth(), async (ctx) => {
   };
 });
 
+router.post("team.create", auth(), async (ctx) => {
```
thenanyu, line 65: 🥦 creates a new team with the same auth providers as the current team, and creates a user with the same email as the current user as the admin of the new team

server/routes/api/team.ts
```diff
@@ -60,4 +62,54 @@ router.post("team.update", auth(), async (ctx) => {
...
+    teamId: team.id,
+    name: user.name,
+    email: user.email,
+    service: null,
+    isAdmin: true,
+    avatarUrl: user.avatarUrl,
+    invitedById: user.id,
+  });
+
+  // make an event here?
```
thenanyu, line 105: @tommoor should we make an event here? it's not reeeally a user being created, but it might be good for... consistency?

shared/editor/nodes/ListItem.ts
```diff
@@ -259,7 +258,6 @@ export default class ListItem extends Node {
           !$pos.nodeAfter ||
           !["list_item", "checkbox_item"].includes($pos.nodeAfter.type.name)
         ) {
-          console.log("Node after not a list item");
```
thenanyu, line 262: these seemed to be hanging randomly

server/commands/teamCreator.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+async function provisionSubdomain(team: Team, requestedSubdomain: string) {
+  // filter subdomain to only valid characters
+  // if there are less than the minimum length, use a default subdomain
+  let subdomain = slugify(requestedSubdomain, { lower: true, strict: true });
+  subdomain =
+    subdomain.length < 3 || RESERVED_SUBDOMAINS.includes(subdomain)
+      ? "team"
+      : subdomain;
```
thenanyu, line 96: note this sanitizing logic for team subdomains

server/routes/api/authenticationProviders.test.ts
```diff
@@ -73,26 +73,25 @@ describe("#authenticationProviders.update", () => {
...
       body: {
-        id: authenticationProviders[0].id,
+        id: googleProvider.id,
         isEnabled: false,
         token: user.getJwtToken(),
       },
     });
     const body = await res.json();
     expect(res.status).toEqual(200);
-    expect(body.data.name).toBe("slack");
```
thenanyu, line 95: This test was failing and I think it was wrong.. but I don't know how it was ever passing

app/scenes/TeamNew.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const handleNameChange = (ev: React.ChangeEvent<HTMLInputElement>) => {
+    setName(ev.target.value);
+  };
+
+  return (
+    <>
+      <form onSubmit={handleSubmit}>
+        <Text type="secondary">
+          <Trans
+            defaults="Your are creating a new team using your current membership — <em>{{email}}</em>"
```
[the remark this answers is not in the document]
tommoor: This feature really makes the authentication management / display of linked auth providers more important as a follow up don't you think?

server/presenters/team.ts
tommoor: I made a similar change in another branch

server/routes/api/auth.ts
```diff
@@ -122,6 +122,8 @@ router.post("auth.info", auth(), async (ctx) => {
     },
     policies: presentPolicies(user, [team]),
   };
+
+  ctx.body = body;
```
tommoor, line 126: remnants of debugging? 🤷

server/routes/api/authenticationProviders.test.ts
```diff
@@ -73,26 +73,25 @@ describe("#authenticationProviders.update", () => {
     expect(res.status).toEqual(400);
   });
 
-  it("should allow admins to disable", async () => {
+  it.only("should allow admins to disable", async () => {
```
tommoor, line 76:
```suggestion
  it("should allow admins to disable", async () => {
```

server/routes/api/authenticationProviders.test.ts
tommoor: The `buildTeam` factory creates a slack auth by default, that's how it was passing

server/routes/api/team.ts
```diff
@@ -60,4 +62,65 @@ router.post("team.update", auth(), async (ctx) => {
   };
 });
 
+router.post("team.create", auth(), async (ctx) => {
+  const { user } = ctx.state;
+  const { name } = ctx.body;
+  assertLength(name, 2, "Name must be 2 or more characters");
```
tommoor, line 68: Move this validation to `models/Team` column validation?

server/routes/api/team.ts
```diff
@@ -60,4 +62,65 @@ router.post("team.update", auth(), async (ctx) => {
...
+      return {
+        name: provider.name,
+        providerId: provider.providerId,
+      };
+    }
+  );
+
+  invariant(
+    authenticationProviders?.length,
+    "authentication provideers must exist"
```
tommoor, line 87:
```suggestion
    "authentication providers must exist"
```

tommoor: Interesting though though, do they _need_ to in the long run? Wouldn't that just mean a team with magic link only signin?

server/routes/api/team.ts
```diff
@@ -60,4 +62,65 @@ router.post("team.update", auth(), async (ctx) => {
...
+    subdomain: name,
+    authenticationProviders,
+    ip: ctx.ip,
+  });
+
+  const newUser = await User.create({
+    teamId: team.id,
+    name: user.name,
+    email: user.email,
+    service: null,
```
tommoor, line 101:
```suggestion
```

Shouldn't need this? This column is hella deprecated

server/routes/api/team.ts
```diff
@@ -60,4 +62,65 @@ router.post("team.update", auth(), async (ctx) => {
...
+    isAdmin: true,
+    avatarUrl: user.avatarUrl,
+    invitedById: user.id,
```
tommoor, line 104:
```suggestion
```

server/routes/api/team.ts
```diff
@@ -60,4 +62,65 @@ router.post("team.update", auth(), async (ctx) => {
...
+  );
+
+  const team = await teamCreator({
+    name,
+    subdomain: name,
+    authenticationProviders,
+    ip: ctx.ip,
+  });
+
+  const newUser = await User.create({
```
tommoor, line 97: Should really be a transaction wrapping around this and the event creation so that one can't be committed without the other

server/routes/auth/index.ts
```diff
@@ -68,6 +70,27 @@ router.get("/redirect", auth(), async (ctx) => {
   );
 });
 
+router.get("/transfer", auth(), async (ctx) => {
```
tommoor, line 73: This method scares me, lets chat on alternatives. We can probably use the `transfer` JWT like we do from hopping from the apex -> subdomain.

https://github.com/outline/outline/blob/863f22750ffb86fec628901a8380d285e5c12ab3/server/utils/jwt.ts#L50

app/scenes/TeamNew.tsx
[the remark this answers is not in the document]
thenanyu: I think in reality, it's not as urgent as it seems because most teams have only 1 auth method

server/routes/api/auth.ts
thenanyu: yes lol

server/routes/api/team.ts
thenanyu: yes that's true; for the time being I am treating users without auth provider as "guests"

server/routes/api/team.ts
thenanyu: This field is used to populate the cookie that says "last time you signed in with Google/Slack/Email/Microsoft". What I should have done is propagate that cookie. Will update

server/routes/auth/index.ts
thenanyu: Interesting, from reading the code, this seemed to be the most straightforward way to initiate the transfer token stuff. Maybe we could do a one time pad on success of the team creation step?

thenanyu:
@tommoor do you prefer 

image

or

image

tommoor:
I believe number 2 is my preferred and the most in-keeping, you can still go straight to another team by typing the name directly then. The images do need to be a bit smaller though

server/routes/auth/index.ts
tommoor: Coming back to this… yep, it's still scary 😆. We should return the redirect url from the team creation api call which includes the temporary authentication token and uses the path that already exists… all you need is the new `team.url` and `user` on the server to get it:

`${team.url}/auth/redirect?token=${user.getTransferToken()}`

We can then remove this endpoint entirely If I'm understanding everything correctly

app/actions/definitions/teams.tsx
```diff
@@ -1,40 +1,64 @@
...
+import TeamNew from "~/scenes/TeamNew";
 import { createAction } from "~/actions";
 import { loadSessionsFromCookie } from "~/hooks/useSessions";
 
-export const changeTeam = createAction({
-  name: ({ t }) => t("Switch team"),
+export const switchTeamList = getSessions().map((session) => {
+  return createAction({
+    name: session.name,
+    section: "Switch Team",
```
tommoor, line 11: Needs defining in `actions/sections` and using the translation helper, you'll see from the ones there.

As discussed earlier and suggested by you last week lets switch to using "Workspaces" as the terminology as it better represents what's going on

app/actions/definitions/teams.tsx
```diff
@@ -1,40 +1,64 @@
...
-    );
-    return otherSessions.length > 0;
-  },
-  children: ({ currentTeamId }) => {
-    const sessions = loadSessionsFromCookie();
-    const otherSessions = sessions.filter(
-      (session) => session.teamId !== currentTeamId
-    );
+  keywords: "change switch workspace organization team",
+  section: "Team",
```
tommoor, line 24: Similar re section definition and translations

app/actions/definitions/teams.tsx
```diff
@@ -1,40 +1,64 @@
...
-      id: session.url,
-      name: session.name,
-      section: "Account",
-      icon: <Logo alt={session.name} src={session.logoUrl} />,
-      perform: () => (window.location.href = session.url),
-    }));
+export const createTeam = createAction({
+  name: ({ t }) => `${t("New team")}…`,
+  keywords: "create change switch workspace organization team",
+  section: "New Team",
```
tommoor, line 33: I think this is also fine under the same `Workspaces` section, it doesn't need to match the name of the action itself

server/routes/api/team.ts
tommoor: Existing endpoint is singular, but it's the only one that's like that – I guess the thought was there would only ever be one team in auth scope. This is probably the moment to make it `teams.create`, I can update and address backwards compat of the other endpoints separately.

server/routes/api/team.ts
tommoor: Still missing transaction

server/routes/api/team.ts
tommoor: Still weird

app/actions/definitions/teams.tsx
thenanyu: sorry, took care of this already but forgot to push!

server/routes/api/team.ts
thenanyu: you sure? this seems natural to me

server/commands/teamCreator.ts
```diff
@@ -0,0 +1,132 @@
...
+        ip,
+      },
+      {
+        transaction,
+      }
+    );
+
+    if (onSuccess) {
+      await onSuccess(team, transaction);
+    }
```
thenanyu, line 77: 🥦 this allows follow-up actions to be wrapped into the same transaction. Kinda weird looking to me but I think it's sane

server/routes/auth/index.ts
thenanyu: Yep, I think you're right

server/routes/api/team.ts
```diff
@@ -60,4 +62,77 @@ router.post("team.update", auth(), async (ctx) => {
...
+  });
+
+  ctx.body = {
+    success: true,
+    data: {
+      team: presentTeam(team),
+      transferUrl: `${
+        team.url
+      }/auth/redirect?token=${newUser?.getTransferToken()}`,
+    },
```
thenanyu, line 134: 🥦 now sending back a transfer token to log into the new team

app/scenes/GroupMembers/GroupMembers.tsx
```diff
@@ -59,7 +59,8 @@ function GroupMembers({ group }: Props) {
         <>
           <Text type="secondary">
             <Trans
-              defaults="Add and remove team members in the <em>{{groupName}}</em> group. Adding people to the group will give them access to any collections this group has been added to."
+              // TODO this string can use improvement
```
thenanyu, line 62: @tommoor the second sentence of this string is a tad bit awkward; suggestions?

app/scenes/GroupMembers/GroupMembers.tsx
```diff
@@ -59,7 +59,8 @@ function GroupMembers({ group }: Props) {
...
+              defaults="Add and remove members to the <em>{{groupName}}</em> group. Adding people to the group will give them access to any collections this group has been added to."
```
tommoor, line 63:
```suggestion
              defaults="Add and remove members to the <em>{{groupName}}</em> group. Members of the group will have access to any collections this group has been added to."
```

Slightly better. This makes me think, really need the reverse view of permissions so you can see which collections a group has been added to.

server/commands/teamCreator.ts
tommoor: Yea, it's a little weird – the way we typically do this now is just have the command accept require the transaction be passed in

server/routes/api/team.ts
```diff
@@ -60,4 +62,76 @@ router.post("team.update", auth(), async (ctx) => {
   };
 });
 
+router.post("teams.create", auth(), async (ctx) => {
```
tommoor, line 65: I need a follow up issue to pluralize the other team endpoints and deprecate the singular.

server/routes/api/team.ts
```diff
@@ -69,4 +72,80 @@ router.post(
...
+        },
+        ip: ctx.ip,
+      },
+      { transaction }
+    );
+
+    return team;
+  });
+
+  const newUser = await User.findOne({
```
tommoor, line 136: Do you remember why you re-query the user that was just created above?

thenanyu: pretty sure it was because of the previous way I did transaction passing

tommoor (approved):

server/migrations/20220810185000-scope-provider-id-uniqueness-to-team.js
xsawo: @tommoor this constraint prevents us from adding additional auth providers for a team. You can configure multiple auth providers such as Google, Slack, etc., but the login will fail if there is already an auth provider in the table for the team.

You receive the following error message after you authenticate with a provider
>Authentication failed – you do not have permission to access this workspace.

and the log message is not very enlightening either

`
{"error":"Validation error","level":"error","message":"Error during authentication","stack":"UnauthorizedError: Validation error\n    at InvalidAuthenticationError (/opt/outline/build/server/errors.js:44:34)\n    at accountProvisioner (/opt/outline/build/server/commands/accountProvisioner.js:59:54)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"}
`

tommoor: The constraint is for unique providerIds in a team, it does not prevent adding multiple unique auth providers – only duplicates.

xsawo: The `providerId` in the table is the configured domain e.g. `example.com` in our setup, Google + Gitlab as auth provider.

It can be verified in the auth provider implementations.
https://github.com/outline/outline/blob/main/plugins/google/server/auth/google.ts#L114
https://github.com/outline/outline/blob/main/plugins/oidc/server/auth/oidc.ts#L138

The constraint prevents the following combination from being added, which is absolutely valid and works as expected. Or is there something we are doing wrong?

```
id   | name   | providerId    | enabled | teamId   | createdAt          
--------------------------------------------------------------------------------
<id> | oidc   | example.com   | t       | <teamid> | 2022-09-15 11:25:56.539+00
<id> | google | example.com   | t       | <teamid> | 2023-07-06 14:14:09.233+00

```

xsawo: @tommoor have you seen my last comment?

tommoor: I would recommend just removing the constraint on your local instance in this case, thanks for raising.

xsawo: We have already done that, but thank you for taking a second look.

Issue #3950, opened 2022-08-10
Rate limiting audit


tommoor:
Once https://github.com/outline/outline/pull/3857 is merged we need to audit all API endpoints and establish tighter controls on potentially problematic ones.

apoorv-mishra:
Merged #3857. We can configure choke points now :crossed_swords:

tommoor:
closed by #4561

PR #3965, opened 2022-08-13
chore: Rate limiter audit

tommoor:
- Refactor
- api/users
- api/collections
- auth

Towards #3950

### DIFF
--- a/server/RateLimiter.ts
+++ b/server/RateLimiter.ts
@@ -1,7 +1,9 @@
-import { RateLimiterRedis } from "rate-limiter-flexible";
+import {
+  IRateLimiterStoreOptions,
+  RateLimiterRedis,
+} from "rate-limiter-flexible";
 import env from "@server/env";
 import Redis from "@server/redis";
-import { RateLimiterConfig } from "@server/types";
 
 export default class RateLimiter {
   constructor() {
@@ -22,7 +24,7 @@ export default class RateLimiter {
     return this.rateLimiterMap.get(path) || this.defaultRateLimiter;
   }
 
-  static setRateLimiter(path: string, config: RateLimiterConfig): void {
+  static setRateLimiter(path: string, config: IRateLimiterStoreOptions): void {
     const rateLimiter = new RateLimiterRedis(config);
     this.rateLimiterMap.set(path, rateLimiter);
   }
--- a/server/RateLimiter.ts
+++ b/server/RateLimiter.ts
@@ -33,3 +33,29 @@ export default class RateLimiter {
     return this.rateLimiterMap.has(path);
   }
 }
+
+/**
+ * Re-useable configuration for rate limiter middleware.
+ */
+export const RateLimiterStrategy = {
+  /** Allows five requests per minute, per IP address */
+  FivePerMinute: {
+    duration: 60,
+    requests: 5,
+  },
+  /** Allows ten requests per minute, per IP address */
+  TenPerMinute: {
+    duration: 60,
+    requests: 10,
+  },
+  /** Allows ten requests per hour, per IP address */
+  TenPerHour: {
+    duration: 3600,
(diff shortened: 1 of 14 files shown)


server/RateLimiter.ts
```diff
@@ -22,7 +24,7 @@ export default class RateLimiter {
     return this.rateLimiterMap.get(path) || this.defaultRateLimiter;
   }
 
-  static setRateLimiter(path: string, config: RateLimiterConfig): void {
+  static setRateLimiter(path: string, config: IRateLimiterStoreOptions): void {
```
tommoor, line 27: No point redefining an existing type

server/middlewares/rateLimiter.ts
```diff
@@ -4,9 +4,15 @@ import RateLimiter from "@server/RateLimiter";
...
 
-export function rateLimiter() {
+/**
+ * Middleware that limits the number of requests per IP address that are allowed
+ * within a window. Should only be applied once to a server – do not use on
+ * individual routes.
+ *
+ * @returns The middleware function.
+ */
+export function defaultRateLimiter() {
```
tommoor, line 15: Wanted to make this clearer/harder to reuse by accident

server/middlewares/rateLimiter.ts
```diff
@@ -35,7 +41,20 @@ export function rateLimiter() {
   };
 }
 
-export function registerRateLimiter(config: RateLimiterConfig) {
+type RateLimiterConfig = {
```
tommoor, line 44: Config applicable to individual routes is minimal and `points` is mapped to `requests`

server/middlewares/rateLimiter.ts
```diff
@@ -47,11 +66,18 @@ export function registerRateLimiter(config: RateLimiterConfig) {
...
-          keyPrefix: RateLimiter.RATE_LIMITER_REDIS_KEY_PREFIX,
-          storeClient: Redis.defaultClient,
-        })
+        defaults(
+          {
+            ...config,
+            points: config.requests,
+          },
+          {
+            duration: 60,
```
tommoor, line 75: While the default duration and points is configurable by the end user for these hardcoded limits we need it to be a fixed value, otherwise it'll need to be defined on every route.

server/middlewares/rateLimiter.ts
apoorv-mishra: Thanks :+1:  I was about to suggest this because `points` didn't make any sense in this context.

server/middlewares/rateLimiter.ts
```diff
@@ -35,7 +41,20 @@ export function rateLimiter() {
   };
 }
 
-export function registerRateLimiter(config: RateLimiterConfig) {
+type RateLimiterConfig = {
+  /** The window for which this rate limiter is considered (defaults to 60s) */
+  duration?: number;
+  /** The number of requests per IP address that are allowed within the window */
+  requests?: number;
```
apoorv-mishra, line 48: How about marking it `required`? Usage of `rateLimiter({})` should be prevented if there's no need to enforce readability.

server/middlewares/rateLimiter.ts
apoorv-mishra: Thanks :+1: I was about to suggest this because `points` didn't make any sense contextually.

server/middlewares/rateLimiter.ts
tommoor: Ah yes will do

server/middlewares/rateLimiter.ts
apoorv-mishra: Thanks +1 I was about to suggest this because points didn't make any sense contextually.

tommoor:
Merging what's here so far

PR #3980, opened 2022-08-16
feat: Support embed configuration

apoorv-mishra:
Enables admins to allow adding hosts for self hosted embeds like diagrams.net, grist etc...


server/models/Integration.ts
```diff
@@ -54,17 +38,17 @@ export type IntegrationSettings =
 }))
 @Table({ tableName: "integrations", modelName: "integration" })
 @Fix
-class Integration extends IdModel {
+class Integration<T> extends IdModel {
   @IsIn([Object.values(IntegrationType)])
-  @Column
-  type: string;
+  @Column(DataType.STRING)
+  type: T;
```
apoorv-mishra, line 44: Keeping it `T` so the type inference of settings below is dependent upon T(during usage, `T` is supposed to be passed as one of `IntegrationType.Embed`, `IntegrationType.Post` and `IntegrationType.Command`).

apoorv-mishra: Reverted due to type recurring type incompatibility issues.

server/models/Integration.ts
```diff
@@ -27,15 +55,18 @@ import Fix from "./decorators/Fix";
...
 
+  @IsIn([Object.values(IntegrationService)])
   @Column
   service: string;
 
   @Column(DataType.JSONB)
-  settings: Record<string, any>;
+  settings: IntegrationSettings;
 
+  @IsIn([["documents.update", "documents.publish"]])
```
apoorv-mishra, line 53: Why was this added @tommoor ?

tommoor: I was just adding validation for existing expectations, is it a requirement with this in place or still nullable?

apoorv-mishra: What if no events are sent? Or are `events` always required?

shared/editor/embeds/Diagrams.tsx
```diff
@@ -10,6 +11,16 @@ function Diagrams(props: Props) {
     ? `Diagrams.net (${params.get("title")})`
     : "Diagrams.net";
 
+  const integration = find(
+    props.integrations.orderedData,
```
apoorv-mishra, line 15: Not sure why `.orderedData` is needed here. Already passing normalized `integrations` [here](https://github.com/outline/outline/pull/3980/commits/c5a5c14089e476c661d454fa700e7507e6f46f7e#diff-18ef7e911d2802727b4354c37055ae38684c757f7de708d9010180f2ea98d248R179)

tommoor: Hmm, yea that doesn't seem to make any sense to me either

app/scenes/Document/components/DataLoader.tsx
```diff
@@ -73,6 +75,19 @@ function DataLoader({ match, children }: Props) {
     fetchDocument();
   }, [ui, documents, document, shareId, documentSlug]);
 
+  React.useEffect(() => {
+    async function fetchIntegrations() {
+      try {
+        integrations.fetchPage({
```
tommoor, line 81: I still think in the long run it might be better to just pull all these in on app load, otherwise there's a more likely race between this request completing and the document content showing.

apoorv-mishra: What's the right place? `app/components/Editor.tsx`? Any pointers?

tommoor: I'm not sure without checking but they don't _need_ to be required.

server/models/Integration.ts
tommoor: I'm not sure without checking but they don't _need_ to be required.

tommoor: I'm not sure without checking but they don't _need_ to be required.

app/scenes/Document/components/DataLoader.tsx
tommoor: Somewhere super early in the app-boot that doesn't exist yet 😄 – here is okay for this PR, but lets make sure it doesn't load integrations every time, it should only load if they haven't been already.

shared/editor/embeds/Diagrams.tsx
apoorv-mishra: Just realized that this is because of the name collision with `RootStore.integrations` [here](https://github.com/outline/outline/pull/3980/commits/c5a5c14089e476c661d454fa700e7507e6f46f7e#diff-4e4902c1514c28c9a67dea9579999973b380d7591a46c22f1574d7ebdde31389R78) :laughing:

shared/editor/embeds/index.tsx
```diff
@@ -53,13 +53,31 @@ export type EmbedProps = {
...
+    let regexes = Component.ENABLED;
+
+    if (embedConfig) {
+      const hostname = embedConfig?.settings.hostname;
+
+      try {
+        const url = new URL(hostname);
+        regexes = [
+          new RegExp(
+            `^${url.protocol}//${url.host.replace(
```
apoorv-mishra, line 71: **Note**: Not considering `url.pathname`.

apoorv-mishra (commented):
For now, modified the common `matcher` to _form_ the custom regex for self hosted embeds.

Maybe we can have a different pattern by supporting another attribute `urlRegex` or `host` in `EmbedDescriptor` which could then be mutated in `app/components/Editor.tsx` through `embedIntegrations` _already available_ via the component's `props`. Then, we can reference the regex(or host) by `this.urlRegex`(or `this.host`) within the common `matcher`.

shared/editor/rules/embeds.ts
```diff
@@ -18,7 +19,10 @@ function isLinkClose(token: Token) {
   return token.type === "link_close";
 }
 
-export default function linksToEmbeds(embeds: EmbedDescriptor[]) {
+export default function linksToEmbeds(
```
tommoor, line 22: You know, I can't find any usage of this method – remove?

shared/editor/embeds/index.tsx
```diff
@@ -53,13 +53,31 @@ export type EmbedProps = {
...
+              /\./g,
+              "\\."
+            )}/(?!proxy).*(title=\\\\w+)?`
```
tommoor, line 74: I like the idea of including this logic further down the stack as you have here, but I think this line is the giveaway that really it should remain in the `embeds/Diagrams.tsx` component – each embed will have it's own rules beyond the hostname that don't make sense to include here

apoorv-mishra: Yes! It was just an attempt to make it working. I just pushed the working version so I can improve it later, incorporating feedback. A lot of it is kind of _hardcoded_ right now. But thanks for this note since I had completely forgotten about it after initial revision :laughing: .Still looking out for better patterns.

shared/editor/rules/embeds.ts
apoorv-mishra: `server/editor/renderToHtml.ts` and `shared/editor/nodes/Embed.tsx`?

tommoor: Well... I guess my code search was broken.

shared/editor/embeds/index.tsx
```diff
@@ -53,13 +53,63 @@ export type EmbedProps = {
...
+
+    try {
+      const urlObj = new URL(this.host as string | URL);
+      regexes = [
+        new RegExp(
+          `^${urlObj.protocol}//${urlObj.host.replace(/\./g, "\\.")}${
+            // @ts-expect-error not aware of static
+            this.component.URL_PATH_REGEX
+          }`
+        ),
```
apoorv-mishra, line 108: Can be moved to `shared/utils`.

shared/editor/embeds/index.tsx
```diff
@@ -53,13 +53,63 @@ export type EmbedProps = {
...
+  }
+
+  matcher(url: string): boolean | [] | RegExpMatchArray {
     // @ts-expect-error not aware of static
-    const regexes = Component.ENABLED;
+    let regexes = this.component.ENABLED;
+
+    try {
+      const urlObj = new URL(this.host as string | URL);
+      regexes = [
```
apoorv-mishra, line 102: Should self hosted and SaaS, both be supported together?

apoorv-mishra (commented):
>For now, modified the common matcher to form the custom regex for self hosted embeds.

>Maybe we can have a different pattern by supporting another attribute urlRegex or host in EmbedDescriptor which could then be mutated in app/components/Editor.tsx through embedIntegrations already available via the component's props. Then, we can reference the regex(or host) by this.urlRegex(or this.host) within the common matcher.

_Take two_ :point_up: 

cc: @tommoor

shared/editor/embeds/index.tsx
apoorv-mishra: Fixed!

shared/editor/embeds/index.tsx
apoorv-mishra: Done!

app/components/Editor.tsx
```diff
@@ -280,7 +280,16 @@ function Editor(props: Props, ref: React.RefObject<SharedEditor> | null) {
...
           onShowToast={showToast}
-          embeds={embeds}
+          embeds={embeds.map((e) => {
+            const em = find(
+              props.embedIntegrations,
+              (i) => i.service === e.component.name.toLowerCase()
+            );
+            return new EmbedDescriptor({
+              ...e,
+              host: em?.settings.hostname,
```
tommoor, line 290: I'd suggest we spread all settings here, `hostname` won't be the only one needed in the long run. Also should choose – `host` or `hostname`, host is better but technically doesn't include the protocol so probably just use `url` is the closest to accurate?

app/components/Editor.tsx
```diff
@@ -280,7 +280,16 @@ function Editor(props: Props, ref: React.RefObject<SharedEditor> | null) {
           ref={mergeRefs([ref, handleRefChanged])}
           uploadFile={onUploadFile}
           onShowToast={showToast}
-          embeds={embeds}
+          embeds={embeds.map((e) => {
```
tommoor, line 283: Lets memoize

app/hooks/useAuthorizedSettingsConfig.ts
```diff
@@ -170,6 +172,14 @@ const useAuthorizedSettingsConfig = () => {
         group: t("Integrations"),
         icon: WebhooksIcon,
       },
+      Diagrams: {
```
tommoor, line 175: I was half way through deciding when handed over to you, lets just rename all these to `Drawio` instead of `Diagrams`? I

shared/utils/escape.ts
```diff
@@ -0,0 +1,4 @@
+// source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
+export const escapeRegExp = (text: string) => {
```
tommoor, line 2: This utility exists in lodash, please use that one

shared/utils/urls.ts
```diff
@@ -92,3 +93,13 @@ export function sanitizeUrl(url: string | null | undefined) {
   }
   return url;
 }
+
+export function urlRegex(url: string | null | undefined): RegExp | undefined {
+  if (!url || !isUrl(url)) {
```
tommoor, line 98: Lets use `sanitizeUrl` inside this method to make sure we're filtering out malicious urls from being embedded in an iframe

app/components/Editor.tsx
apoorv-mishra: Agree. `url` makes more sense. Will update that.

app/hooks/useAuthorizedSettingsConfig.ts
apoorv-mishra: Sure!

shared/utils/escape.ts
apoorv-mishra: Thanks for the tip :+1:

shared/utils/urls.ts
apoorv-mishra: Yes! Just realized this. Will update.

app/components/Editor.tsx
apoorv-mishra: Didn't understand! You mean https://lodash.com/docs/4.17.15#memoize?

tommoor: As this one has been converted to functional already can use `useMemo` hook – I want to make sure it isn't mapping to new objects very render

shared/utils/urls.ts
apoorv-mishra: Fixed!

shared/utils/escape.ts
apoorv-mishra: Fixed!

app/components/Editor.tsx
apoorv-mishra: Fixed!

app/components/Editor.tsx
apoorv-mishra: Fixed!

app/hooks/useAuthorizedSettingsConfig.ts
apoorv-mishra: Fixed!

shared/editor/embeds/index.tsx
apoorv-mishra: Made modifications to support both.

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+      url: integration?.settings.url,
+    },
+  });
+
+  const handleSubmit = React.useCallback(
+    async (data: FormData) => {
+      if (integration) {
+        await integration.save({
+          settings: {
+            url: data.url,
```
apoorv-mishra, line 49: There should be a stricter validation for `url` here.

apoorv-mishra: Added on backend.

app/components/Editor.tsx
```diff
@@ -54,10 +54,24 @@ export type Props = Optional<
 };
 
 function Editor(props: Props, ref: React.RefObject<SharedEditor> | null) {
-  const { id, shareId, onChange, onHeadingsChange } = props;
+  const { id, shareId, onChange, onHeadingsChange, embedIntegrations } = props;
   const { documents } = useStores();
   const { showToast } = useToasts();
   const dictionary = useDictionary();
+  const embedsWithSetings = React.useMemo(
```
tommoor, line 61: I'd consider pulling this into a local hook, `useEmbedIntegrations` or similar, we can pull the info from the store in the hook and totally avoid passing `embedIntegrations` into this component.

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+  url: string;
+};
+
+function Drawio() {
+  const { integrations } = useStores();
+  const { t } = useTranslation();
+
+  React.useEffect(() => {
+    integrations.fetchPage({
+      limit: 100,
```
tommoor, line 26:
```suggestion
      limit: 100,
      type: IntegrationType.Embed,
```

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+        });
+      } else {
+        await integrations.create({
+          type: IntegrationType.Embed,
+          service: "diagrams",
+          settings: {
+            url: data.url,
+          },
+        });
+      }
```
tommoor, line 60:
```suggestion
      await integrations.save({
        type: IntegrationType.Embed,
        service: "diagrams",
        settings: {
          url: data.url,
        },
      });
```

I know it was my code previously, but this should work. update/create switch is taken care of in BaseStore.

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+      title={t("Draw.io")}
+      icon={<BuildingBlocksIcon color="currentColor" />}
+    >
+      <Heading>{t("Draw.io")}</Heading>
+
+      <Text type="secondary">
+        <Trans>
+          If your team is self-hosting draw.io then adding your hosted
+          deployment url here will enable automatic embedding within documents.
+          If you are using Diagrams.net then no configuration is required.
```
tommoor, line 76:
```suggestion
          Add your self-hosted draw.io installation url here to enable automatic 
          embedding of diagrams within documents.
```

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </Trans>
+        <form onSubmit={formHandleSubmit(handleSubmit)}>
+          <p>
+            <Input
+              label={t("Draw.io Deployment")}
```
tommoor, line 81:
```suggestion
              label={t("Draw.io deployment")}
```

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+          },
+        });
+      }
+    },
+    [integrations, integration]
+  );
+
+  return (
+    <Scene
+      title={t("Draw.io")}
```
tommoor, line 67:
```suggestion
      title="Draw.io"
```

Product name never needs translating

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+      icon={<BuildingBlocksIcon color="currentColor" />}
+    >
+      <Heading>{t("Draw.io")}</Heading>
```
tommoor, line 70:
```suggestion
      <Heading>Draw.io</Heading>
```

Product name never needs translating

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+          If your team is self-hosting draw.io then adding your hosted
+          deployment url here will enable automatic embedding within documents.
+          If you are using Diagrams.net then no configuration is required.
+        </Trans>
+        <form onSubmit={formHandleSubmit(handleSubmit)}>
+          <p>
+            <Input
+              label={t("Draw.io Deployment")}
+              placeholder={"https://app.diagrams.net/"}
+              pattern="https://.*"
```
tommoor, line 83:
```suggestion
              pattern="https?://.*"
```

Probably a good idea to support non-secure deployments

app/scenes/Settings/Drawio.tsx
```diff
@@ -0,0 +1,98 @@
...
+      } else {
+        await integrations.create({
+          type: IntegrationType.Embed,
+          service: "diagrams",
+          settings: {
+            url: data.url,
+          },
+        });
+      }
+    },
```
tommoor, line 61: Need to show a toast message here too otherwise it's unclear that settings saved, something like:

```
    showToast(t("Settings saved"), {
      type: "success",
    });
```

server/routes/api/integrations.ts
```diff
@@ -43,20 +80,30 @@ router.post("integrations.update", auth(), async (ctx) => {
...
+    integration.set({
+      events: events.filter((event: string) =>
+        ["documents.update", "documents.publish"].includes(event)
+      ),
+    });
+  }
+
+  integration.set({
+    settings,
+  });
```
tommoor, line 97:
```suggestion
  integration.settings = settings;
```

Haven't used this `set` api before, seems unneccessary? :P

shared/editor/embeds/Diagrams.tsx
```diff
@@ -3,43 +3,36 @@ import Frame from "../components/Frame";
 import Image from "../components/Image";
 import { EmbedProps as Props } from ".";
 
-const URL_REGEX = /^https:\/\/viewer\.diagrams\.net\/(?!proxy).*(title=\\w+)?/;
+function Diagrams(props: Props) {
+  const embedUrl = props.attrs.matches[0];
+  const params = new URL(embedUrl).searchParams;
+  const title = params.get("title")
```
tommoor, line 9: As it displays below the embed can we show "Draw.io" as title when the embed `settings` is present?

tommoor (commented):
I think that's everything, should be good to merge once these small things are resolved.

app/components/Editor.tsx
apoorv-mishra: > we can pull the info from the store in the hook

_Pull from the server from within the hook_? As it's currently done in [`DataLoader.tsx`](https://github.com/outline/outline/pull/3980/files#diff-18ef7e911d2802727b4354c37055ae38684c757f7de708d9010180f2ea98d248R82)? Or keep the API call in `DataLoader.tsx` and just do `return integrations.orderedData(...)` in the hook? The latter should work too since `DataLoader.tsx` mounts before `Editor.tsx`.

tommoor: I was thinking about that, we could co-locate the data loading yes. Later we may want to support these integrations on public share links, that _may_ mean moving the data loading again. But it would be neat to include it in the hook for now

app/scenes/Settings/Drawio.tsx
apoorv-mishra: :point_down: seems better?

```suggestion
    service: "diagrams",
```

server/routes/api/integrations.ts
apoorv-mishra: Yes! Makes more sense when multiple attributes need to be set.

shared/editor/embeds/Diagrams.tsx
apoorv-mishra: Sure!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!
(Needed `id: integration?.id` as well for `update`)

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/components/Editor.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!
(Added for `type: "error"` as well)

server/routes/api/integrations.ts
apoorv-mishra: Fixed!

shared/editor/embeds/Diagrams.tsx
apoorv-mishra: Fixed!

app/hooks/useEmbedIntegrations.ts
```diff
@@ -0,0 +1,32 @@
...
+  >();
+
+  React.useEffect(() => {
+    async function fetchEmbedIntegrations() {
+      try {
+        await integrations.fetchPage({
+          limit: 100,
+          type: IntegrationType.Embed,
+        });
+        setEmbedIntegrations(integrations.orderedData);
```
tommoor, line 20: I don't think there's any need for the local state in here

app/scenes/Settings/Drawio.tsx
```diff
@@ -43,44 +47,43 @@ function Drawio() {
...
+          id: integration?.id,
           type: IntegrationType.Embed,
-          service: "diagrams",
+          service: SERVICE_NAME,
           settings: {
             url: data.url,
           },
         });
+
+        showToast("Settings saved", {
```
tommoor, line 60:
```suggestion
        showToast(t("Settings saved"), {
```

app/scenes/Settings/Drawio.tsx
```diff
@@ -43,44 +47,43 @@ function Drawio() {
...
+          type: "success",
+        });
+      } catch (err) {
+        showToast("Failed to save!", {
```
tommoor, line 64:
```suggestion
        showToast(err.message, {
```

app/scenes/Settings/Drawio.tsx
```diff
@@ -43,44 +47,43 @@ function Drawio() {
...
+          type: "error",
+        });
       }
     },
-    [integrations, integration]
+    [integrations, integration, showToast]
```
tommoor, line 69:
```suggestion
    [integrations, integration, t, showToast]
```

app/hooks/useEmbedIntegrations.ts
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

app/scenes/Settings/Drawio.tsx
apoorv-mishra: Fixed!

tommoor (approved):

helmut72:
> Enables admins to allow adding hosts for self hosted embeds like diagrams.net, **grist** etc...

Is this for draw.io only or also for grist?

tommoor:
draw.io currently, the same framework can be used for other integrations in the future though

Issue #3983, opened 2022-08-17
Do not add HTTPS to links if protocol is declared


lemker:
Current linking behavior seems to disregard protocols other than HTTP or HTTPS. For instance, an STSSYNC link will break the link when https:// is prepended to the link

**To Reproduce**
1. Create a new link using a protocol other than http/https, for example stssync://
2. Outline will add https:// to the link, breaking the protocol (https://stssync://)

**Expected behavior**
Leave stssync:// the way it is.

**Screenshots**
![image](https://user-images.githubusercontent.com/31901553/185226068-bb18f395-5e42-49fe-a8da-a1834019b56e.png)

**Outline (please complete the following information):**
- Install: getoutline.com

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro
 - Browser: Chromium Edge

PR #3995, opened 2022-08-19
Improving the `urls` to not break protocols and adding tests

iifawzi:
Hi, This's my first contribution to this great project, and hopefully not the last one. 
This PR should fix https://github.com/outline/outline/issues/3983
I've also added a bunch of tests, testing the urls utils. 

Signed-off-by: iifawzi <iifawzie@gmail.com>

### DIFF
--- a/shared/utils/urls.test.ts
+++ b/shared/utils/urls.test.ts
@@ -1,0 +1,70 @@
+import * as urlsUtils from "./urls";
+
+describe("IsUrl Method", () => {
+  describe("invalid urls", () => {
+    it("should return false", () => {
+      expect(urlsUtils.isUrl("")).toEqual(false);
+      expect(urlsUtils.isUrl("#invalidurl")).toEqual(false);
+      expect(urlsUtils.isUrl("mailto:")).toEqual(false);
+      expect(urlsUtils.isUrl("://")).toEqual(false);
+    });
+  });
+});
+
+describe("isInternalUrl Method", () => {
+  it("should return false if empty string", () => {
+    expect(urlsUtils.isInternalUrl("")).toEqual(false);
+  });
+
+  it("should return true if starting with relative path", () => {
+    expect(urlsUtils.isInternalUrl("/drafts")).toEqual(true);
+  });
+});
+
+describe("isExternalUrl Method", () => {
+  it("should return false if empty url", () => {
+    expect(urlsUtils.isExternalUrl("")).toEqual(false);
+  });
+
+  it("should return false if internal url", () => {
+    expect(urlsUtils.isExternalUrl("/drafts")).toEqual(false);
+  });
+});
+
+describe("sanitizeUrl Method", () => {
+  it("should return undefined if not url", () => {
+    expect(urlsUtils.sanitizeUrl(undefined)).toEqual(undefined);
+    expect(urlsUtils.sanitizeUrl(null)).toEqual(undefined);
+    expect(urlsUtils.sanitizeUrl("")).toEqual(undefined);
+  });
+
+  it("should append https:// to non-special urls", () => {
+    expect(urlsUtils.sanitizeUrl("www.google.com")).toEqual(
+      "https://www.google.com"
+    );
+  });
+
+  describe("special urls", () => {
+    it("should return the url as it's if starting with /", () => {
+      expect(urlsUtils.sanitizeUrl("/drafts")).toEqual("/drafts");
+    });
+    it("should return the url as it's if starting with #", () => {
+      expect(urlsUtils.sanitizeUrl("#home")).toEqual("#home");
+    });
+    it("should return the url as it's if it's mailto:", () => {
+      expect(urlsUtils.sanitizeUrl("mailto:outline@getoutline.com")).toEqual(
+        "mailto:outline@getoutline.com"
+      );
+    });
+    it("should return the url as it's if it's mailto:", () => {
+      expect(urlsUtils.sanitizeUrl("mailto:outline@getoutline.com")).toEqual(
+        "mailto:outline@getoutline.com"
+      );
+    });
+    it("should return the url as it's if it's a special domain", () => {
+      expect(urlsUtils.sanitizeUrl("mqtt://getoutline.com")).toEqual(
+        "mqtt://getoutline.com"
+      );
+    });
+  });
+});
--- a/shared/utils/urls.ts
+++ b/shared/utils/urls.ts
@@ -67,7 +67,11 @@ export function isUrl(text: string) {
  * @returns True if the url is external, false otherwise.
  */
 export function isExternalUrl(url: string) {
-  return !isInternalUrl(url);
+  if (url) {
+    return !isInternalUrl(url);
+  }
+
+  return false;
 }
 
 /**
@@ -86,7 +90,8 @@ export function sanitizeUrl(url: string | null | undefined) {
     !isUrl(url) &&
     !url.startsWith("/") &&
     !url.startsWith("#") &&
-    !url.startsWith("mailto:")
+    !url.startsWith("mailto:") &&
+    !url.includes("://")
   ) {
     return `https://${url}`;
   }


shared/utils/urls.ts
```diff
@@ -67,7 +67,11 @@ export function isUrl(text: string) {
  * @returns True if the url is external, false otherwise.
  */
 export function isExternalUrl(url: string) {
-  return !isInternalUrl(url);
+  if (url) {
+    return !isInternalUrl(url);
+  }
+
+  return false;
```
iifawzi, line 74: I've added this to ensure that the url is not empty, since `internalUrl` method is returning false when it's empty.

shared/utils/urls.ts
```diff
@@ -86,7 +90,8 @@ export function sanitizeUrl(url: string | null | undefined) {
     !isUrl(url) &&
     !url.startsWith("/") &&
     !url.startsWith("#") &&
-    !url.startsWith("mailto:")
+    !url.startsWith("mailto:") &&
+    !url.includes("://")
```
iifawzi, line 94: To avoid breaking the protocols if any is specified

iifawzi:
This's my first contribution,  I might not be aware of the folder structure decisions that was taken while ago, but I think it would make the code more organized if all the tests are under a separate directory, `test` for example. Even if not all the tests under a single main `test` directory, a sub-folder `test` in each folder would work too, like having all the urls tests in `test` folder under the `utils` directory.

shared/utils/urls.ts
tommoor: Hey @iifawzi this change basically nullifies the sanitize method and allows any protocol through. I suggest that we maintain a blocklist of protocols starting with `javascript` and `file` instead

iifawzi: Hey @tommoor, thank you for your time reviewing this, to make sure I got it right, you prefer to have a list of  `notAllowedProtocols` for example,  and it can initially include `javascript://` and `file://`? so we can avoid these protocols from being sanitized, right?

tommoor: Yes, `blockedProtocols` must be removed, in this case prepended with `https` as before is the way to make them "safe".

iifawzi: That's really cool, I got it, changes have been committed. 

I've used regex to detect the protocol, a simpler approach can be as simple as

```js
 blockedProtocols.some(protocol => url.includes(protocol))
```

let me know if it could be improved, or if i'm missing anything

shared/utils/urls.ts
```diff
@@ -86,12 +87,14 @@ export function sanitizeUrl(url: string | null | undefined) {
...
-    !isUrl(url) &&
-    !url.startsWith("/") &&
-    !url.startsWith("#") &&
-    !url.startsWith("mailto:") &&
-    !url.includes("://")
+    (!isUrl(url) &&
+      !url.startsWith("/") &&
+      !url.startsWith("#") &&
+      !url.startsWith("mailto:")) ||
+    (protocol && blockedProtocols.includes(protocol[0]))
```
tommoor, line 97: I think we can stick with using `startsWith("javascript://")` pattern here and inline the protocols we care about, it'll be more maintainable/simple that way and the regex above is too loose anyway

iifawzi: Changes have been committed @tommoor, let me know if you want me to squash the commits

shared/utils/urls.ts
```diff
@@ -83,10 +87,12 @@ export function sanitizeUrl(url: string | null | undefined) {
...
-    !isUrl(url) &&
-    !url.startsWith("/") &&
-    !url.startsWith("#") &&
-    !url.startsWith("mailto:")
+    (!isUrl(url) &&
+      !url.startsWith("/") &&
+      !url.startsWith("#") &&
+      !url.startsWith("mailto:")) ||
+    url.startsWith("javascript://") ||
+    url.startsWith("file://")
```
tommoor, line 95:
```suggestion
    url.startsWith("javascript:") ||
    url.startsWith("file:")
```

Javascript does not require the `//`

shared/utils/urls.test.ts
```diff
@@ -0,0 +1,81 @@
+import * as urlsUtils from "./urls";
+
+describe("IsUrl Method", () => {
```
tommoor, line 3: The tests are very appreciated 🙏

iifawzi: Thank you 🙏

shared/utils/urls.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  it("should return false if internal url", () => {
+    expect(urlsUtils.isExternalUrl("/drafts")).toEqual(false);
+  });
+});
+
+describe("sanitizeUrl Method", () => {
+  it("should return undefined if not url", () => {
+    expect(urlsUtils.sanitizeUrl(undefined)).toEqual(undefined);
```
apoorv-mishra, line 36: [`toBeUndefined()`](https://jestjs.io/docs/expect#tobeundefined)? It's just handy.
```suggestion
    expect(urlsUtils.sanitizeUrl(undefined)).toBeUndefined();
```

shared/utils/urls.test.ts
```diff
@@ -0,0 +1,81 @@
+import * as urlsUtils from "./urls";
+
+describe("IsUrl Method", () => {
+  describe("invalid urls", () => {
+    it("should return false", () => {
+      expect(urlsUtils.isUrl("")).toEqual(false);
```
apoorv-mishra, line 6: [`.toBe(value)`](https://jestjs.io/docs/expect#tobevalue) should just be fine for primitive values. Handy as well.

shared/utils/urls.ts
apoorv-mishra, line 74: [short-circuit eval](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation)?
```suggestion
  return !!url && !isInternalUrl(url);
```

apoorv-mishra (commented):
Just a few code style nitpicks!

shared/utils/urls.ts
iifawzi: Nice catch

shared/utils/urls.ts
```diff
@@ -83,10 +87,12 @@
...
   if (
-    !isUrl(url) &&
-    !url.startsWith("/") &&
-    !url.startsWith("#") &&
-    !url.startsWith("mailto:")
+    (!isUrl(url) &&
+      !url.startsWith("/") &&
+      !url.startsWith("#") &&
+      !url.startsWith("mailto:")) ||
+    url.startsWith("javascript:") ||
```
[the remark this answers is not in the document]
iifawzi: What do you think about those two protocols? I think they need to be sanitized. 
commit have been created to resolve this

@tommoor @apoorv-mishra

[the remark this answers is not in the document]
tommoor: CodeQL saving our asses

tommoor (approved):

vn-ncvinh:
@tommoor , I'm aware of a few possible XSS vulnerabilities here.

In the file `urls.ts` contains 2 code snippets with vulnerabilities.
```typescript
export function isUrl(text: string) {
  if (text.match(/\n/)) {
    return false;
  }

  try {
    const url = new URL(text);
    return url.hostname !== "";
  } catch (err) {
    return false;
  }
}
```

```typescript
export function sanitizeUrl(url: string | null | undefined) {
  if (!url) {
    return undefined;
  }

  if (
    (!isUrl(url) &&
      !url.startsWith("/") &&
      !url.startsWith("#") &&
      !url.startsWith("mailto:")) ||
    url.startsWith("javascript:") ||
    url.startsWith("file:") ||
    url.startsWith("vbscript:") ||
    url.startsWith("data:")
  ) {
    return `https://${url}`;
  }
  return url;
}
```
An attacker can easily bypass the filter to generate a url containing the XSS attack payload.
```url
jAvascript://%0Aalert(origin)
```
`jAvascript://`: will fool and pass 2 test functions.

`%0A%0Dalert(origin)`: Execute command alert origin.



# Solution
Use filter by the URL protocol
```typescript
var blacklist_protocol = ['javascript:', 'file:', 'vbscript:', 'data:']
export function isUrl(text: string) {
  if (text.match(/\n/)) {
    return false;
  }
  
  try {
    const url = new URL(text);
    return url.hostname !== "" && !blacklist_protocol.includes(url.protocol);
  } catch (err) {
    return false;
  }
}
```

```typescript
export function sanitizeUrl(url: string | null | undefined) {
  if (!url) {
    return undefined;
  }

  if (
    (!isUrl(url) &&
      !url.startsWith("/") &&
      !url.startsWith("#") &&
      !url.startsWith("mailto:"))
  ) {
    return `https://${url}`;
  }
  return url;
}
```

iifawzi:
Nice catch @vn-ncvinh, it's way better this way. Anyway, if we want to stick with the startsWith way, we can also lowercase the urls before checking them

vn-ncvinh:
> Nice catch @vn-ncvinh, it's way better this way. Anyway, if we want to stick with the startsWith way, we can lowercase the urls before checking them.

If using lowercase, there is still a way to bypass it:
```
javascript://%0Aalert(origin)
```
And of course, the javascript code is still executed as usual.

vn-ncvinh:
> Nice catch @vn-ncvinh, it's way better this way. Anyway, if we want to stick with the startsWith way, we can lowercase the urls before checking them.

If using lowercase, there is still a way to bypass it by using special characters.

apoorv-mishra:
Interesting. Thanks for pointing this out :pray: 

Maybe we should replace these with the existing [`validator`](https://github.com/outline/outline/blob/9aacd807e3d6a845b650a26312988883eaff62fb/server/validation.ts#L53-L62) module that we already use on server-side? But I wonder why isn't it used already :thinking: To keep the bundle size low @tommoor ?

iifawzi:
Ahh I see, thank you @vn-ncvinh for clarifying this, my bad.

tommoor:
I don't _think_ the validator module is doing anything too special, in fact it may be vulnerable to the use of special characters…

https://github.com/validatorjs/validator.js/blob/master/src/lib/isURL.js#L84

apoorv-mishra:
_Plug_: Seems like [`mermaid`](https://github.com/mermaid-js/mermaid/blob/cde3a7cf70cbf6f33406ebb4e83e58d7af9feb7d/package.json#L61) uses [this](https://github.com/braintree/sanitize-url/blob/main/src/index.ts).

vn-ncvinh:
I think the URL of javascript works quite well, we just need to add a blacklist of protocols.
For urls containing special characters, it will be automatically removed.
But maybe in the future there will be protocol spoofing vulnerability, but that's javascript problem, and we don't need to care about it right now.
image

tommoor:
Isn't this _fun_ 😆

apoorv-mishra:
I'm actually surprised why none of the third party validators leverage `new URL()` too as part of the validation flow :thinking:

vn-ncvinh:
I don't know about that either.
But maybe when they build a new validator function, they will avoid using the built-in validator functions.

vn-ncvinh:
I think should add `tel:` for calls.

```typescript
!url.startsWith("tel:")
```

vn-ncvinh:
I think should add `tel:`, `fax:` for calls and fax.

```typescript
||
!url.startsWith("tel:") ||
!url.startsWith("fax:")
```

vn-ncvinh:
I think should add `tel:`, `fax:`, `sms:`... or some other alternative link

tommoor:
> I think should add tel:, fax:, sms:... or some other alternative link

Not sure what you mean here?

vn-ncvinh:
> > I think should add tel:, fax:, sms:... or some other alternative link
> 
> Not sure what you mean here?

These will make it easier to the contact methods.
- `tel:0123456789` - Make a call to the phone number 0123456789
- `fax:0123456789` - Make a fax to 0123456789
- `sms:0123456789` - Make new SMS to phone number 0123456789

apoorv-mishra:
I'm thinking, maybe instead of having a _blacklist_, we should have `allowed_protocols`? Because, the blacklist above might still be bypassed by something like `new URL("somerandomprotocol://malicious.something")`. We're more confident about what we want to allow than what we don't.

iifawzi:
Although this will be more restrictive but it would make more sense to be implemented this way, and legitimate protocols can be added by the time, when people notice it, through the issues/PRs

vn-ncvinh:
whether to use blacklist or whitelist depends on each specific case.  We will prioritize using what is supposed to be the minority, and here is the blacklist.  Of the multitude of protocols available, only a few are considered dangerous, and it is easier to list them than to list the rest.

iifawzi:
I think the subsequences of forgetting any dangerous protocol are more dangerous than forgetting any allowed protocol, in the later case, people will notice them easily and will be added accordingly

vn-ncvinh:
So far, only 4 protocols have been deemed dangerous, which can lead to xss.

tommoor:
@iifawzi in the aim of maintainability let's keep going with the blocklist approach, I don't want to be fielding PR's for eternity for every desktop app possible.

vn-ncvinh:
@tommoor, What do you think about removing `url.hostname !== ""`. This prevents protocols without `//` such as `mailto:`, `cal:`... Remove it and the protocols will be valid without the need for additional whitelist. And of course, after deleting it, the protocols in blacklist are still block.
By the way, I have an unverified report on huntr.dev, can you help me finish it? Thanks.

```typescript
var blacklist_protocol = ['javascript:', 'file:', 'vbscript:', 'data:']
export function isUrl(text: string) {
  if (text.match(/\n/)) {
    return false;
  }

  try {
    const url = new URL(text);
    return !blacklist_protocol.includes(url.protocol);
  } catch (err) {
    return false;
  }
}

export function sanitizeUrl(url: string | null | undefined) {
  if (!url) {
    return undefined;
  }

  if (
    (!isUrl(url) &&
      !url.startsWith("/") &&
      !url.startsWith("#"))
  ) {
    return `https://${url}`;
  }
  return url;
}
```

image
image

vn-ncvinh:
Sorry everyone, it seems like nodejs version of replit is old, so `url.hostname` doesn't filter protocols other than http and https, which led to my mistakes in penetration testing. :(

shared/utils/urls.ts
```diff
@@ -54,7 +55,7 @@ export function isUrl(text: string) {
 
   try {
     const url = new URL(text);
-    return url.hostname !== "";
+    return url.hostname !== "" && !blacklist_protocols.includes(url.protocol);
```
iifawzi, line 58: Summing everything up, since `hostname` will always be an empty string, the `isUrl` function will always return true (which will lead to have the URLs sanitized) in all of the following cases: 

`data:text/html,<script>alert('hi');</script>` 
`vbscript:whatever`
`javascript:whatever`
`vbscript:whatever`

and the protocols check will never be reached
```JS
!blacklist_protocols.includes(url.protocol); 
```

Only in the case of the `file` protocols, the check will save us. But since we're not very sure how malicious URLs might be written and whether they may include hostnames to bypass this or not, I've kept the check. 

If any of you knows a malicious URL with a hostname, we would need to add them to tests in order for this to be really tested, because now, even if we removed the blacklist protocols check, the examples above would still be sanitized because they're not URL according to `isURL` due to the fact that there's no hostname.  

I've created a commit

shared/utils/urls.ts
```diff
@@ -55,7 +56,7 @@ export function isUrl(text: string) {
 
   try {
     const url = new URL(text);
-    return url.hostname !== "";
+    return url.hostname !== "" && !BLACKLIST_PROTOCOLS.includes(url.protocol);
```
tommoor, line 59:
```suggestion
    const blockedProtocols = [
      "javascript:",
      "file:",
      "vbscript:",
      "data:",
    ]

    return url.hostname !== "" && !blockedProtocols.includes(url.protocol);
```

Please just inline the protocols, no need to be in the separate file as it isn't used elsewhere. Otherwise this PR looks good  now 🙏

iifawzi: Changes have been comitted

tommoor (approved):

PR #4023, opened 2022-08-27
feat: Allow viewers to be upgraded to editors on individual collections

### LINKED ISSUE #4017: Allow viewers to become editors in individual groups/collections
Currently if your account is a viewer it can never become an editor, despite the UI showing that it is possible when adding users to collections / groups.

The permission logic should be changed so that individual membership to group/collection overrides the default permission.

tommoor:
Also, big improvements to types, move permissions to an enum.

closes #4017


app/scenes/CollectionPermissions/AddPeopleToCollection.tsx
```diff
@@ -57,7 +57,6 @@ class AddPeopleToCollection extends React.Component<Props> {
       this.props.memberships.create({
         collectionId: this.props.collection.id,
         userId: user.id,
-        permission: "read_write",
```
tommoor, line 60: Chosen on server side now based on the role of user being added

PR #4026, opened 2022-08-28
Throttle email notifications upon updating document frequently

### LINKED ISSUE #3467: Reduce email notifications when a document is modified multiple times
Someone has modified a page I modified previously, so I received a mail notification.
Another person modified the same page again few minutes later and I've got a second notification.

The second notification is useless (imho) because I did not visit or modified the page since I've received the first notification.

**To Reproduce**
1. Someone modify a page I previously modified
2. I receive a notification mail
3. Another person edits the same page
4. step 2. happen again

**Expected behavior**
Just being notified once.

**Screenshots**
null

- Install: self-hosted
- Version: 0.63.0

 - OS: N/a
 - Browser [e.g. chrome, safari]: N/a
 - Version [e.g. 22]: N/a

**Mobile (please complete the following information):**
 - Device: [e.g. iPhone6] N/a
 - OS: [e.g. iOS8.1] N/a
 - Browser [e.g. stock browser, safari] N/a

apoorv-mishra:
Closes #3467

### DIFF
--- a/server/emails/templates/BaseEmail.tsx
+++ b/server/emails/templates/BaseEmail.tsx
@@ -1,6 +1,7 @@
 import mailer from "@server/emails/mailer";
 import Logger from "@server/logging/Logger";
 import Metrics from "@server/logging/metrics";
+import Notification from "@server/models/Notification";
 import { taskQueue } from "@server/queues";
 import { TaskPriority } from "@server/queues/tasks/BaseTask";
 
@@ -80,6 +81,21 @@ export default abstract class BaseEmail<T extends EmailProps, S = any> {
       Metrics.increment("email.sent", {
         templateName,
       });
+
+      // @ts-expect-error does not exist on type T & S
+      if (data.notificationId) {
+        await Notification.update(
+          {
+            emailedAt: new Date(),
+          },
+          {
+            where: {
+              // @ts-expect-error does not exist on type T & S
+              id: data.notificationId,
+            },
+          }
+        );
+      }
     } catch (err) {
       Metrics.increment("email.sending_failed", {
         templateName,
--- a/server/emails/templates/BaseEmail.tsx
+++ b/server/emails/templates/BaseEmail.tsx
@@ -1,3 +1,4 @@
+import { Optional } from "utility-types";
 import mailer from "@server/emails/mailer";
 import Logger from "@server/logging/Logger";
 import Metrics from "@server/logging/metrics";
@@ -9,7 +10,10 @@ interface EmailProps {
   to: string;
 }
 
-export default abstract class BaseEmail<T extends EmailProps, S = any> {
+export default abstract class BaseEmail<
+  T extends EmailProps & Optional<{ notificationId: string }>,
+  S = Optional<{ notificationId: string }>
+> {
   private props: T;
 
   /**
@@ -81,26 +85,31 @@ export default abstract class BaseEmail<T extends EmailProps, S = any> {
       Metrics.increment("email.sent", {
         templateName,
       });
+    } catch (err) {
+      Metrics.increment("email.sending_failed", {
+        templateName,
+      });
+      throw err;
+    }
 
-      // @ts-expect-error does not exist on type T & S
-      if (data.notificationId) {
+    if (data.notificationId) {
+      try {
         await Notification.update(
           {
             emailedAt: new Date(),
           },
           {
             where: {
-              // @ts-expect-error does not exist on type T & S
               id: data.notificationId,
             },
           }
         );
+      } catch (err) {
+        Logger.error(
+          `Failed to update notification(id=${data.notificationId})`,
+          err
+        );
       }
-    } catch (err) {
-      Metrics.increment("email.sending_failed", {
-        templateName,
-      });
-      throw err;
     }
   }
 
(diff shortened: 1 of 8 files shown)


server/emails/templates/BaseEmail.tsx
```diff
@@ -80,6 +81,21 @@ export default abstract class BaseEmail<T extends EmailProps, S = any> {
       Metrics.increment("email.sent", {
         templateName,
       });
+
+      // @ts-expect-error does not exist on type T & S
```
tommoor, line 85: This is unfortunate, the types have been tricky here with it being static. Perhaps we can define S as an object which always has an optional `notificationId`?

server/emails/templates/BaseEmail.tsx
```diff
@@ -80,6 +81,21 @@ export default abstract class BaseEmail<T extends EmailProps, S = any> {
...
+      if (data.notificationId) {
+        await Notification.update(
```
tommoor, line 87: I'm worried that this within the try/catch could cause an email to be sent twice if it fails, better to move outside the try and fail with a Logger

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -237,6 +247,23 @@ export default class NotificationsProcessor extends BaseProcessor {
       return false;
     }
 
+    // Deliver only a single notification in a 12 hour window
+    const notification = await Notification.findOne({
+      order: [["createdAt", "DESC"]],
+      where: {
+        userId: user.id,
+        documentId: document.id,
+        emailedAt: {
```
tommoor, line 256: Need to make sure this has an index in the migration (see previous migrations for examples)

server/emails/templates/BaseEmail.tsx
apoorv-mishra: Thanks for spotting this :pray:

server/emails/templates/BaseEmail.tsx
apoorv-mishra: Added on both `T` and `S`.

apoorv-mishra: What is we add it `T` instead? `S = Optional<...>` thingy doesn't work since `S` would still be _unknown_. On the other hand, `S extends...` makes it _required_ which in result demands modifications at other places too.

apoorv-mishra: Or how about having and optional `metadata` field(`metadata?: Required<{ notificationId: string }>;`) in top level `EmailProps` instead? I guess that's better.

server/emails/templates/BaseEmail.tsx
```diff
@@ -86,6 +88,28 @@ export default abstract class BaseEmail<T extends EmailProps, S = any> {
...
+            where: {
+              id: metadata.notificationId,
+            },
+          }
+        );
+      } catch (err) {
+        Logger.error(
+          `Failed to update notification(id=${metadata.notificationId})`,
+          err
+        );
```
tommoor, line 110:
```suggestion
        Logger.error(
          `Failed to update notification`,
          err,
          metadata,
        );
```

You can pass structured data to be stored with the error log as third argument 👍

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -72,6 +74,13 @@ export default class NotificationsProcessor extends BaseProcessor {
...
 
       if (notify) {
+        const notification = await Notification.create({
+          event: event.name,
+          userId: recipient.user.id,
+          actorId: document.updatedBy.id,
+          teamId: team.id,
+          documentId: document.id,
+        });
         await DocumentNotificationEmail.schedule({
```
tommoor, line 84: What do you think about passing `metadata` as second argument instead of mixing in with email props?

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -237,6 +249,23 @@ export default class NotificationsProcessor extends BaseProcessor {
       return false;
     }
 
+    // Deliver only a single notification in a 12 hour window
+    const notification = await Notification.findOne({
```
tommoor, line 253: I'm tempted to suggest we abstract the entire logic into a configuration on the email class so you can define the suppression rules per-email. But it's probably over engineering in this moment as I'm not sure if any other types of email will need similar logic to this, document edits is kind of a special case.

server/queues/processors/NotificationsProcessor.ts
apoorv-mishra: Hmm...makes sense from the POV of _decoupling_ it from email notifications. It might be needed for other types of notifications(in-app etc...) as well. What's your reasoning?

server/queues/processors/NotificationsProcessor.ts
apoorv-mishra: Actually, it occurred to me as well. My suggestion is that we place the suppression rules as it is for now since we don't have many yet. The `shouldNotify` is good enough, since it's explicit and more importantly _keeps all suppression rules in one place_. Not in favor of detouring yet. Makes more sense once we have a clear distinction of suppression rules based on notification types.

server/queues/processors/NotificationsProcessor.ts
tommoor: Reasoning is mainly they are functionally different data use, the first being for template injection.

tommoor (approved):

PR #4030, opened 2022-08-30
Dismiss parent menu when a submenu opens(only on mobile devices)

### LINKED ISSUE #3948: submenus on mobile views don't dismiss the parent menu, leading to overlap
image

apoorv-mishra:
Fixes #3948

### DIFF
--- a/app/components/ContextMenu/index.tsx
+++ b/app/components/ContextMenu/index.tsx
@@ -146,7 +146,7 @@ const ContextMenu: React.FC<Props> = ({
           );
         }}
       </Menu>
-      {(rest.visible || rest.animating) && (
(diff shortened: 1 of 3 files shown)


app/components/ContextMenu/index.tsx
```diff
@@ -146,7 +146,7 @@ const ContextMenu: React.FC<Props> = ({
           );
         }}
       </Menu>
-      {(rest.visible || rest.animating) && (
+      {(rest.visible || rest.animating) && rest["aria-label"] !== t("Submenu") && (
```
apoorv-mishra, line 149: This is to prevent the issue of _double backdrop_, one from parent and other from sub-menu.

Issue #4001, opened 2022-08-22
Content is displayed wrongly when printing / Save as PDF


Nicicalu:
When printing a document or saving it as a PDF, some parts aren't displayed correctly. This is what I could find:

1. Block Quotes aren't formatted correctly when printing a document (the thin grey bar on the left of the quote is not visible).
2. Code is formatted correctly but there is a scrollbar if it overflows horizontally. In a PDF you can't click scrollbars which means you can't read the code that overflows. 
3. Highlighted text is not highlighted
4. The box of the task lists are not visible
5. The Info/Warning/Danger blocks have a white background instead of blue/yellow/red
6. The "Show source" button on mermaid diagrams is still visible
7. The background on codeblocks is white instead of grey

**To Reproduce**
1. Write the listed types of content in a document
8. Print
9. Content is displayed incorrectly

**Expected behavior**
The content is like in the editor (apart from the buttons).

**Screenshots**
In the editor:
![](https://i.imgur.com/1p701En.png)

When printing:
![](https://i.imgur.com/WhooskN.png)

**Outline:**
- Install: Self hosted
- Version: v.0.65.2 

**Desktop:**
 - OS: Windows 10
 - Browser: Google Chrome
 - Version: 103.0.5060.114

lemker:
Adding to this, when using Print to PDF my text is gray instead of black:

![image](https://user-images.githubusercontent.com/31901553/186482313-4ffa1d98-d068-49bc-a308-ccffb869a787.png)

Nicicalu:
@lemker @tommoor 
AYY I found something! When setting the CSS-Property `-webkit-print-color-adjust: exact;` on the broken elements, it works in chrome!
Or even better, just set it globally (including for non webkit browsers):
```
:root {
    print-color-adjust: exact;
    -webkit-print-color-adjust: exact;
}
```
This property is used to cut costs when printing by removing backgrounds and changing colors. The browser automatically decides which colors are changed and which backgrounds are removed. When setting this property you basically tell the browser to print everything as it is, without changing colors.
It is supported on all major browsers: https://caniuse.com/css-color-adjust

More info: 
- https://developer.mozilla.org/en-US/docs/Web/CSS/print-color-adjust?retiredLocale=de
- https://css-tricks.com/almanac/properties/p/print-color-adjust/

The only things that still need changes are the buttons or scrollbars that need to disappear:
[...]

PR #4043, opened 2022-08-31
Content is displayed wrongly when printing / Save as PDF

Nicicalu:
This first commit fixes problem number 6 in issue https://github.com/outline/outline/issues/4001, when printing a document.
The button "Show Source" in mermaid diagrams should be hidden, when printing a document.

The third commit fixes the problems number 1, 3, 4, 5, 7 in issue https://github.com/outline/outline/issues/4001, where backgrounds and task checkboxes are not visible when printing

### DIFF
--- a/app/editor/components/Styles.ts
+++ b/app/editor/components/Styles.ts
@@ -1287,6 +1287,10 @@ const EditorStyles = styled.div<{
     .page-break {
       opacity: 0;
     }
+    
+    button.show-source-button {
+      display: none;
+    }
 
     em,
     blockquote {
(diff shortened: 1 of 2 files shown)


app/editor/components/Styles.ts
```diff
@@ -1287,6 +1287,10 @@ const EditorStyles = styled.div<{
     .page-break {
       opacity: 0;
     }
+    
+    button.show-source-button {
```
tommoor, line 1291: Correct approach, but please include in the block of other hidden elements above starting L1275

Nicicalu: @tommoor Of course, I didn't see that. I committed the new file

tommoor (approved):

tommoor:
Thanks for the contribution 💯

PR #4067, opened 2022-09-05
Permanently redirect to `/s/...` for share links

### LINKED ISSUE #4041: Allow `/s/` to load share links and make it the default going forward
Cleaner and shorter

- New share links will be generated with `/s/` instead of `/share/`
- Visiting `/share/...` should permanently redirect to `/s/...` for backwards compatibility

apoorv-mishra:
Closes #4041

### DIFF
--- a/app/routes/index.tsx
+++ b/app/routes/index.tsx
@@ -55,10 +55,10 @@ export default function Routes() {
         <Route exact path="/create" component={Login} />
         <Route exact path="/logout" component={Logout} />
 
-        <Route exact path="/share/:shareId" component={SharedDocument} />
+        <Route exact path="/s/:shareId" component={SharedDocument} />
         <Route
           exact
-          path={`/share/:shareId/doc/${slug}`}
+          path={`/s/:shareId/doc/${slug}`}
           component={SharedDocument}
         />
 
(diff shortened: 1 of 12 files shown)


tommoor (changes requested):
Try sharing a document and it's nested documents, when you click on any document in the sidebar on the shared page it throws you to the login screen

app/routes/index.tsx
```diff
@@ -55,10 +55,28 @@ export default function Routes() {
         <Route exact path="/create" component={Login} />
         <Route exact path="/logout" component={Logout} />
 
-        <Route exact path="/share/:shareId" component={SharedDocument} />
+        <Route
```
tommoor, line 58: You can use the shorter `from` `to` format for these, like:

```typescript
<Redirect
  from="/share/:shareId"
  to="/s/:shareId"
  exact
/>
```

tommoor (commented):
Try searching the codebase for `/share/` – there are a number of spots left to change, the redirect is a good catch but does result in a flash of the UI. Some places:

https://github.com/outline/outline/blob/6fc1cbc0ce7870d44a8473c7e4c3c234496c9094/app/components/Editor.tsx#L163

https://github.com/outline/outline/blob/75a868e5e8db408053835ff01cde21520d540b55/app/components/SearchListItem.tsx#L41

https://github.com/outline/outline/blob/4468d2974025e10454522699b03ea8b99dfda74e/app/components/Sidebar/components/SharedDocumentLink.tsx#L95

These are all really similar in hindsight, probably worth a method in `routeHelpers`

PR #4069, opened 2022-09-05
feat: Show diff when navigating revision history

tommoor:
### Todo

- [x] fix dark mode styles
- [x] Hookup "Restore" button in header
- [x] Support RTL languages
- [x] Remove views.list and subscriptions.list requests on navigating history
- [x] Remove facepile on history
- [x] Preload diffs?
- [ ] Update metadata under title for revisions
- [ ] Improve display of complex nodes

closes https://github.com/outline/outline/discussions/2167


tommoor:
Moving the final todo's to follow-up tickets, this is already so much better than what's in production it's time to get it live.

almereyda:
Congratulations for the effort, which was well worth pursuing. 🎉

PR #4081, opened 2022-09-07
Support user and team preferences

### LINKED ISSUE #3990: Add preferences jsonb to `User` and `Team` models
In order to store preferences for several upcoming issues we should have a column that can do this without having to do a database migration for each new pref.

- https://github.com/outline/outline/issues/3091
- https://github.com/outline/outline/issues/3880
- https://github.com/outline/outline/issues/3132
- https://github.com/outline/outline/issues/4060

apoorv-mishra:
Closes #3990


server/presenters/user.ts
```diff
@@ -18,6 +19,7 @@ type UserPresentation = {
   isViewer: boolean;
   email?: string | null;
   language?: string;
+  preferences?: UserPreferences;
```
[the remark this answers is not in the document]
tommoor: > Do we want that?
Yes, I think so.

We will need to validate the keys on the server but otherwise it should be settable from the client side

server/models/User.ts
```diff
@@ -54,6 +54,12 @@ export enum UserRole {
   Viewer = "viewer",
 }
 
+export enum UserPreference {
+  LastVisitedUrl = "lastVisitedUrl",
```
apoorv-mishra, line 58: Maybe we should store `lastVisitedPath` instead, without the host? It would come as handy while handling redirection and avoid unnecessary clutter.

server/presenters/user.ts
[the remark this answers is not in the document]
apoorv-mishra: > We will need to validate the keys on the server but otherwise it should be settable from the client side

Sorry, this is in the presentation layer. I assumed we want to _restrict the visibility_ for user preferences similar to the way we've restricted `email` and `language`. Regarding validation and update, we're already doing that via `users.update`. The latter has nothing to do with former, right? Both are disjoint flows? Let me know if I misunderstood something :smile:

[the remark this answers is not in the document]
tommoor: You're correct, preferences should only be sent to the user they related to. Sorry for misdirection.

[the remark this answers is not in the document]
apoorv-mishra: Thanks! I myself realized only after pushing the suggested changes. Will revert now :laughing:

server/models/User.ts
```diff
@@ -290,6 +300,37 @@ class User extends ParanoidModel {
     return this.flags;
   };
 
+  /**
+   * User preferences store user's preferences, for example,
+   * last visited url by the user etc. The method
+   * sets a particular preference
```
tommoor, line 306:
```suggestion
   * Preferences set by the user that decide application behavior and ui.
```

server/models/User.ts
```diff
@@ -54,6 +54,12 @@ export enum UserRole {
   Viewer = "viewer",
 }
 
+export enum UserPreference {
+  LastVisitedPath = "lastVisitedPath",
```
tommoor, line 58:
```suggestion
  RememberLastPath = "rememberLastPath",
```

server/routes/api/users.ts
```diff
@@ -184,6 +186,16 @@ router.post("users.update", auth(), async (ctx) => {
   if (language) {
     user.language = language;
   }
+  if (preferences) {
+    assertKeysIn(preferences, UserPreference);
+    if (preferences.lastVisitedPath) {
```
tommoor, line 191:
```suggestion
    if (preferences.rememberLastPath) {
```

server/routes/api/users.ts
```diff
@@ -184,6 +186,16 @@ router.post("users.update", auth(), async (ctx) => {
...
+      assertUrlPath(preferences.lastVisitedPath);
```
tommoor, line 192: Lets just make this a boolean, we don't want to have to write the user record every time the path changes – it'll just be stored on the client.

tommoor (approved):
🫡

PR #4108, opened 2022-09-12
fix: Dialog doesn't close after deleting a document with a parent

### LINKED ISSUE #4107: Dialog doesn't close after deleting a document with a parent
If you delete a document that has a parent using the delete option in the app, you are successfully redirected to the parent page, but the dialog box stays up.  If you click delete on this dialog box, you'll get an unauthorized error (it does not delete the parent if clicking "I'm sure - Delete" at this point).

**To Reproduce**
1. Create a nested document that has a parent document
2. Select Delete... from the document menu
3. Select "I'm sure - Delete"
4. Observe stated bug

**Expected behavior**
The user should be returned to the parent with the dialog box closed.

**Screenshots**
outline1

- Install: self hosted
- Version: 0.66.0 (current latest Docker image)

vgwidt:
fixes #4107

* Removes return inside the if statement that checks if the parent is available in the store to allow the rest of the code to be executed. 
* Else clause added to above if statement to redirect to collection if parent is not available in the store.
* Original redirect to collection wrapped in else in case there is no parent.

Sorry if there's a cleaner way to do this so that code isn't duplicated.

### DIFF
--- a/app/scenes/DocumentDelete.tsx
+++ b/app/scenes/DocumentDelete.tsx
@@ -44,12 +44,16 @@ function DocumentDelete({ document, onSubmit }: Props) {
 
             if (parent) {
               history.push(documentUrl(parent));
-              return;
+            }
+            else {
+              // if it is not available in the store then redirect to the collection
+              history.push(collectionUrl(collection?.url || "/"));
             }
           }
-
-          // otherwise, redirect to the collection home
-          history.push(collectionUrl(collection?.url || "/"));
+          else {
+            // If it has no parent then redirect to the collection home
+            history.push(collectionUrl(collection?.url || "/"));
+          }
         }
 
         onSubmit();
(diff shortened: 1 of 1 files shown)


app/scenes/DocumentDelete.tsx
```diff
@@ -44,12 +44,16 @@ function DocumentDelete({ document, onSubmit }: Props) {
 
             if (parent) {
               history.push(documentUrl(parent));
-              return;
```
tommoor, line 47: You could achieve the same thing by called `onSubmit();` here before the return statement with no other changes

vgwidt: Brilliant!  Giving it a try now.

vgwidt: Tested with your suggestion and it works great.  Thanks!

tommoor (approved):

PR #4126, opened 2022-09-15
Introduce account preferences to remember user's previous location

### LINKED ISSUE #3091: Add user preference to remember previous location
We'd like to add a user preference to remember the last route you were looking at and redirect there upon reloading the app. This would take precedence over the team default collection if set.

apoorv-mishra:
Closes #3091


app/components/AuthenticatedLayout.tsx
```diff
@@ -39,6 +40,18 @@ const AuthenticatedLayout: React.FC = ({ children }) => {
   const location = useLocation();
   const can = usePolicy(ui.activeCollectionId);
   const { user, team } = auth;
+  const previousPath = usePrevious(location.pathname);
```
tommoor, line 43: I think you might have misunderstood the intent of the feature. The idea is that when navigating to `http://company.getoutline.com` it would remember which document or collection you were previously viewing and then redirect to that.

`usePrevious` is not going to work for this, it just gives you the value of the last render loop. We'll need to store the last viewed pathname as you navigate around and then use that here. Would potentially use the `usePersistedState` hook for this.

apoorv-mishra: Thanks!

_Note to self_ :point_down: 

There are two parts of the problem,
1. Persist last visited document or collection path.
2. Redirect to path in (1) when matched to `/`(make sure there's no UI flash in this step).

app/scenes/Login/index.tsx
```diff
@@ -62,6 +63,10 @@ function Login({ children }: Props) {
   const [error, setError] = React.useState(null);
   const [emailLinkSentTo, setEmailLinkSentTo] = React.useState("");
   const isCreate = location.pathname === "/create";
+  const rememberLastPath =
+    !!auth.user?.preferences && auth.user.preferences.rememberLastPath;
```
tommoor, line 67:
```suggestion
  const rememberLastPath = !!auth.user?.preferences?.rememberLastPath;
```

app/hooks/useAuthorizedSettingsConfig.ts
```diff
@@ -82,6 +84,14 @@ const useAuthorizedSettingsConfig = () => {
         group: t("Account"),
         icon: ProfileIcon,
       },
+      Preferences: {
+        name: t("Preferences"),
+        path: "/settings/preferences",
```
tommoor, line 89:
```suggestion
        path: accountPreferencesPath(),
```

app/scenes/Settings/Preferences.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Scene
+      title={t("Preferences")}
+      icon={<SettingsIcon color="currentColor" />}
+    >
+      <Heading>{t("Preferences")}</Heading>
+      <SettingRow
+        name="rememberLastPath"
+        label={t("Remember previous location")}
+        description={t(
+          "When enabled, user's previous location will be remembered and they'll be redirected to that location upon next visit."
```
tommoor, line 43:
```suggestion
          "Automatically return to the document you were last viewing when the app is re-opened"
```

app/scenes/Settings/Preferences.tsx
```diff
@@ -0,0 +1,57 @@
...
+        )}
+      >
+        <Switch
+          id="rememberLastPath"
+          name="rememberLastPath"
+          checked={!!user.preferences && user.preferences.rememberLastPath}
```
tommoor, line 49:
```suggestion
          checked={!!user.preferences?.rememberLastPath}
```

server/routes/api/users.ts
```diff
@@ -188,7 +189,7 @@ router.post("users.update", auth(), async (ctx) => {
   }
   if (preferences) {
     assertKeysIn(preferences, UserPreference);
-    if (preferences.rememberLastPath) {
+    if (has(preferences, UserPreference.RememberLastPath)) {
```
tommoor, line 192: We might want to move this `UserPreference` enum to `/shared/` it could be used in some of the frontend components/types then.

tommoor (commented):
Looks good, minor tweaks but works well 👍

tommoor (approved):
Great!

PR #4134, opened 2022-09-17
Share document link that opens full editor

dolsem:
It would be useful to be able to share links to documents that open internal team view with the full editor and the sidebar for those who are signed in and have permissions, and show shared view for everyone else. 

This PR adds an option to append `?edit=true` to the shared document URL that does exactly that.

UPD: my bad for not discussing this prior, just read the guidelines now 😅


tommoor:
Hey, it would be good to know more details about your usecase here that isn't fulfilled by just sharing the doc URL directly.

I take it you have seen the edit button in the top right on share links when the user has permissions?

dolsem:
@tommoor hey, so I want to have a task board inside Outline, which can be put inside an embed, like AirTable or NocoDB, put document links inside, and make it possible for team members to navigate between documents just like they would if they were just looking at the list of subdocuments for an Outline document. This would work fine if I just use internal document links. However, I also want to be able to share a document with a taskboard embed externally, and I want those documents link to still be accessible to people without an account. And if I just used regular shared documents links, this would mean team members would keep navigating away from the internal UI to shared view with no sidebar, which is inconvenient. So this is just to remove that extra manual step that requires them pressing the Edit button each time.

dolsem:
@tommoor so do you think this could be merged?

tommoor:
Hey, I think the redirect logic and the query param are fine. I'm mostly concerned about adding this as another front-and-center option in the share modal… that doesn't really make sense to me as it's never been requested as a requirement until now in over 2 years of public sharing. The option itself is also quite hard to understand, unlike the other toggles it doesnt affect the permissions at all, but the behavior of the link.

That leaves us with a couple of options.

1. Strip out the UI and land the changes fast so you can create the links you want manually 
2. Rethink the option so it's less in your face, perhaps under a "link/more options" disclosure, or more closely associated with the link input.
3. Doesn't land in mainline.
4. Something else?

dolsem:
ok let me see if I can make the UI look less awkward

dolsem:
@tommoor I made the option less in your face and updated the wording to hopefully make it less confusing. What do you think?

image

image

image

dolsem:
@tommoor sorry to bother you again, but do you think it's good enough now to be merged?

tommoor:
Yes, definitely in the right direction. Let me play with it a little, I think we should update the option to show "More settings" or something along those lines rather than just the arrow – I'm also intrigued as to why a scrollbar is appearing on your machine, that shouldn't be the case. Was that there before?

tommoor:
https://user-images.githubusercontent.com/380914/192932832-f9eff4cc-be2c-407b-9f1f-26e37bc1edc5.mp4

Tweaked UI and copy writing, thoughts?

tommoor (approved):

dolsem:
Sweet!

PR #4156, opened 2022-09-22
Sign webhook requests

### LINKED ISSUE #3699: Webhook signing
We'd like to add an additional and expected layer of security to webhooks by signing the request body and including the signature in a header as originally included in the [RFC](https://wiki.generaloutline.com/share/47305ccd-9f29-4edf-b48a-21c239bd625e) and is a follow on from https://github.com/outline/outline/pull/3691.

### Requirements

- Add a `secret` BYTEA column to `webhook_subscriptions`
- This column should contain an optional encrypted secret chosen by the webhook creator
- The "secret" field should appear on the webhook create/edit form and be user editable and nullable
- If the secret is present then generate a `Outline-Signature` header which is a signature of the request body, something like:

```
return require("crypto")
    .createHmac("sha256", secretKey)
    .update(JSON.stringify(req.body))
    .digest("hex");
```

apoorv-mishra:
Closes #3699


tommoor:
Test failed because CI environment was missing `SECRET_KEY`, I just added it

server/queues/tasks/DeliverWebhookTask.ts
```diff
@@ -539,10 +539,13 @@ export default class DeliverWebhookTask extends BaseTask<Props> {
...
           env.VERSION ? `/${env.VERSION.slice(0, 7)}` : ""
         }`,
+        "Outline-Signature": subscription.signature(
+          JSON.stringify(requestBody)
+        ),
       };
       response = await fetch(subscription.url, {
         method: "POST",
-        headers: requestHeaders,
+        headers: requestHeaders as any,
```
tommoor, line 548: I think we need to generate the signature and only add to `requestHeaders` if it's not undefined, should remove the requirement to cast to `any` here.

app/scenes/Settings/components/WebhookSubscriptionForm.tsx
```diff
@@ -250,6 +252,13 @@ function WebhookSubscriptionForm({ handleSubmit, webhookSubscription }: Props) {
           label={t("URL")}
           {...register("url", { required: true })}
         />
+        <ReactHookWrappedInput
+          flex
+          label={t("Secret")}
+          {...register("secret", {
+            required: false,
+          })}
+        />
```
apoorv-mishra, line 261: Should it be `type="password"` with a button to show/hide?

tommoor: I don't think it's neccessary at the moment

tommoor (approved):

PR #4175, opened 2022-09-26
Allow drafts to be created without requiring a collection

apoorv-mishra:
Addresses #2122 

Affected endpoints:

- [x] `documents.create`
- [ ] `documents.update`
- [ ] `documents.drafts`

...and possibly others?


server/presenters/slackAttachment.ts
```diff
@@ -22,10 +22,10 @@ function present(
     : document.getSummary();
 
   return {
-    color: collection.color,
+    color: collection?.color,
     title: document.title,
     title_link: `${team.url}${document.url}`,
-    footer: collection.name,
+    footer: collection?.name,
```
apoorv-mishra, line 28: Not tested the impact. Hoping that it's fine? :crossed_fingers:

server/routes/api/documents.ts
```diff
@@ -1162,7 +1162,17 @@ router.post("documents.create", auth(), async (ctx) => {
...
   } = ctx.body;
   const editorVersion = ctx.headers["x-editor-version"] as string | undefined;
-  assertUuid(collectionId, "collectionId must be an uuid");
+
+  if (parentDocumentId || template) {
+    assertPresent(
+      collectionId,
+      "collectionId is required to create a nested doc or a template"
+    );
+  }
```
apoorv-mishra, line 1171: I'm assuming this is both necessary and sufficient to _not break_ the existing nested doc or template creation flow _requiring_ collection?

tommoor: That sounds right, perhaps we can have nested docs without a collection later but that will require some thought given the ordering data is currently stored on the collection itself

server/routes/api/documents.ts
```diff
@@ -515,13 +519,15 @@ router.post("documents.restore", auth({ member: true }), async (ctx) => {
   // be caught as a 403 on the authorize call below. Otherwise we're checking here
   // that the original collection still exists and advising to pass collectionId
   // if not.
-  if (!collectionId && !collection) {
+  if (!document.isDraftWithoutCollection && !collectionId && !collection) {
```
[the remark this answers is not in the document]
tommoor: We could add "Drafts" as an option in this menu

server/test/factories.ts
```diff
@@ -336,7 +340,7 @@ export async function buildDocument(
     overrides.userId = user.id;
   }
 
-  if (!overrides.collectionId) {
+  if (!draft && !overrides.collectionId) {
```
tommoor, line 343: We can probably change this to check that `collectionId` isn't explicitly set as `null` then.

```
buildDraftDocument(overrides) {
  return buildDocument({ ...overrides, collectionId: null, publishedAt: null });
}
`

File: server/commands/documentUpdater.ts [hunk 5]
```diff
@@ -74,7 +77,9 @@ export default async function documentUpdater({
   const changed = document.changed();
 
   if (publish) {
-    document.lastModifiedById = user.id;
+    if (document.isDraftWithoutCollection) {
```
tommoor, line 80: I feel like draft is increasingly meaningless in this world and we can probably lose this method in favor of just checking if `collectionId` is not set and have the same behavior.

server/routes/api/documents.ts
```diff
@@ -515,13 +519,15 @@ router.post("documents.restore", auth({ member: true }), async (ctx) => {
...
     throw ValidationError(
       "Unable to restore to original collection, it may have been deleted"
     );
   }
 
-  authorize(user, "update", collection);
+  if (!document.isDraftWithoutCollection) {
```
tommoor, line 528:
```suggestion
  if (collection) {
```

Clearer?

apoorv-mishra: Makes sense. `collection` being `null` directly implies that the document isn't published either. `isDraftWithoutCollection` seems redundant. Thanks for the spot :pray:

server/models/Document.ts
```diff
@@ -527,8 +527,8 @@ class Document extends ParanoidModel {
     "teamId" = :teamId AND
     "collectionId" IN(:collectionIds) AND
     ${documentClause}
-    "deletedAt" IS NULL AND
-    "publishedAt" IS NOT NULL
+    ${!options.includeDrafts ? '"publishedAt" IS NOT NULL AND' : ""}
+    "deletedAt" IS NULL
```
tommoor, line 531: So I think this is a mistake, we're within the `searchForTeam`  method here which has no user context, so there are no drafts we can ever show.

server/models/Document.ts
```diff
@@ -569,7 +569,7 @@ class Document extends ParanoidModel {
     ]);
 
     // Final query to get associated document data
-    const documents = await this.findAll({
+    const documents = await this.scope(["withoutState", "withDrafts"]).findAll({
```
tommoor, line 572: 🙏  on including `withoutState` here though, should be a noticeable speed improvement

apoorv-mishra: Looks like [state is already excluded](https://github.com/outline/outline/blob/55e622e22ff741ad23cc52ece9b93cc9fc5bd15f/server/models/Document.ts#L83) via default scope. Will revert this change.

server/policies/document.ts
```diff
@@ -94,8 +86,7 @@ allow(User, "update", Document, (user, document) => {
   if (document.deletedAt) {
     return false;
   }
-
-  if (cannot(user, "update", document.collection)) {
+  if (document.collection && cannot(user, "update", document.collection)) {
```
apoorv-mishra, line 89: Note that this `document.collection` acts more like a _skip_ rather than a _guard_. I'm not sure if there could be a case where `cannot(user, "update")` is true for a given user against _all collections_? Thoughts @tommoor ?

tommoor: There are several places where we had an `invariant` that would throw in the case of collection not existing, this was mainly to protect against mistakes in development where `collection` is not queried with the document but could protect against other cases too.

I think we want to change these to check `if (document.collectionId)` and then within that we can still use invariant if `document.collection` does not exist

server/models/Document.ts
```diff
@@ -638,7 +630,15 @@ class Document extends ParanoidModel {
...
     "teamId" = :teamId AND
-    "collectionId" IN(:collectionIds) AND
+    ${
+      options.includeDrafts
+        ? collectionIds.length
+          ? '("collectionId" IN(:collectionIds) OR "collectionId" IS NULL) AND'
+          : '"collectionId" IS NULL AND'
+        : collectionIds.length
+        ? '"collectionId" IN(:collectionIds) AND'
+        : ""
```
tommoor, line 640: If I'm following correctly – If `includeDrafts: false` and you have access to no collections then the entire clause is removed here, giving you access to search everything in the team…

If that's true then we definitely need to add a test that would catch that.

server/models/Document.ts
```diff
@@ -620,14 +620,6 @@ class Document extends ParanoidModel {
       collectionIds = await user.collectionIds();
     }
 
-    // If the user has access to no collections then shortcircuit the rest of this
-    if (!collectionIds.length) {
```
tommoor, line 624: Probably retain but with the addition of `&& !options.includeDrafts`?

server/test/factories.ts
```diff
@@ -336,7 +351,7 @@ export async function buildDocument(
     overrides.userId = user.id;
   }
 
-  if (!overrides.collectionId) {
+  if (!overrides.collectionId && !isNull(overrides.collectionId)) {
```
tommoor, line 354:
```suggestion
  if (overrides.collectionId === undefined) {
```

Might be fine for these?

server/models/Document.ts
apoorv-mishra: Grave mistake! Thanks for the spot :pray: I've simplified this a bit now and also added supporting tests.

server/models/Document.ts
apoorv-mishra: Well, not needed anymore. If `!collectionIds.length && !options.includeDrafts` evaluates to `true`, then the `whereClause` further down evaluates to:
```sql
"searchVector" @@ to_tsquery('english', :query) AND
"teamId" = :teamId AND
"collectionId" IS NULL AND "createdById" = :userId AND
"archivedAt" IS NULL AND
"publishedAt" IS NOT NULL AND
"deletedAt" IS NULL
```
And we know, by design that if `"collectionId" IS NULL`, then the document must be an unpublished one.

server/test/factories.ts
apoorv-mishra: Yes, should be good. Though ignoring other potential _falsy_ values, but it's okay.

server/routes/api/documents.ts
```diff
@@ -820,38 +835,55 @@ router.post("documents.update", auth(), async (ctx) => {
...
-  const document = await sequelize.transaction(async (transaction) => {
-    const document = await Document.findByPk(id, {
-      userId: user.id,
-      includeState: true,
-      transaction,
-    });
-    authorize(user, "update", document);
-
-    collection = document.collection;
+  const document = await Document.findByPk(id, {
```
tommoor, line 855: Is it possible to keep this inside the database transaction? It should have had an `UPDATE` lock on the query honestly, as the model is being mutated, that's an oversight previously.

tommoor (approved):

PR #4184, opened 2022-09-27
feat: add nocodb integration

dolsem:
Adds a NocoDB integration to the editor. NocoDB is a self-hosted alternative to AirTable that allows connecting an actual relational database and serving as a frontend for it. The extra logic adds some niceties that makes the integration more useful, such as recognizing Outline doc URLs in a NocoDB table and properly navigating to them on click, without refreshing the page. Example use case is a task tracker with Outline and NocoDB.

The extra logic depends on two outstanding PRs in the NocoDB repo that are currently being merged: https://github.com/nocodb/nocodb/pull/3810 and https://github.com/nocodb/nocodb/pull/3483.


Limezy:
Because you assume NocoDB to be selfhosted, I guess that for safety reasons Tom will ask you to copy Draw.io integration style, where the url is set in the settings. Indeed, your regex could probably be used malicously to leverage your integration as a weak point. That's not the case for other SaaS integrations where the product URL is trusted by default.

Limezy:
https://github.com/outline/outline/pull/3980

dolsem:
@Limezy should be done now

app/scenes/Settings/NocoDb.tsx
```diff
@@ -0,0 +1,101 @@
...
+import useStores from "~/hooks/useStores";
+import useToasts from "~/hooks/useToasts";
+
+type FormData = {
+  url: string;
+};
+
+const SERVICE_NAME = "nocodb";
+
+function NocoDb() {
```
tommoor, line 22: Appreciate you adding this page, I think that in hindsight these individual pages will quickly get out of control – we also have one for TLDraw in the wings. It's probably better if we just rename the existing one for "Draw.io" to be "Self hosted" and then add the configuration for all self-hosted integrations in that one page under sub-headings, thoughts?

Limezy: Yep that would make sense !

dolsem: @Limezy would you be interested in giving me a hand with this? Still working on another PR for Outline I need asap, hard to find time for everything

Limezy: Yep will have a try

Limezy: Thanks @tommoor for having started the work ! Didn't find any time yet

Issue #4129, opened 2022-09-15
`Enter` in table cell should add a row after the current one


tommoor:
Currently it adds a row always at the end of the table

PabloYG:
Hey, I have modified the code to get this behavior. The README says I should comment before submitting a PR, so that's what I'm doing!

The change merely consists of calling `getRowIndex()` when Enter key is detected and using this in the same way we use the `addRowAfter` command. I've also modified `getRowIndex()` for it to work even when there's no Cell selected.

Hope it helps.

tommoor:
Sounds good, bug fixes are always welcome :)

PR #4186, opened 2022-09-28
Enter in table cell adds a row after the current one

PabloYG:
Solves issue #4129 

I changed getRowIndex to return the row index whenever the user writes in a table cell. It still returns undefined if not in a table. Also, on Enter key pressed, we get the index from the modified getRowIndex and use it to add new row below the current one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ fakes3/*
 *.pem
 *.key
 *.cert
+docker-compose.yml
+yarn-error.log
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,3 @@ fakes3/*
 *.pem
 *.key
 *.cert
-docker-compose.yml
-yarn-error.log
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,12 +3,12 @@ services:
   redis:
     image: redis
     ports:
-      - "0.0.0.0:6379:6379"
+      - "127.0.0.1:6379:6379"
     user: "redis:redis"
   postgres:
     image: postgres
     ports:
-      - "0.0.0.0:5432:5432"
+      - "127.0.0.1:5432:5432"
     environment:
       POSTGRES_USER: user
       POSTGRES_PASSWORD: pass
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,12 +3,12 @@ services:
   redis:
     image: redis
     ports:
-      - "127.0.0.1:6379:6379"
+      - "0.0.0.0:6379:6379"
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -12,3 +12,5 @@ fakes3/*
 *.pem
 *.key
 *.cert
+docker-compose.yml
```
tommoor, line 15: Please remove changes

docker-compose.yml
```diff
@@ -3,12 +3,12 @@ services:
   redis:
     image: redis
     ports:
-      - "127.0.0.1:6379:6379"
```
tommoor, line 6: Please remove changes

shared/editor/queries/getRowIndex.ts
```diff
@@ -2,10 +2,17 @@ import { CellSelection } from "prosemirror-tables";
 
 export default function getRowIndex(selection: CellSelection) {
   const isRowSelection = selection.isRowSelection && selection.isRowSelection();
-  if (!isRowSelection) {
-    return undefined;
+  if (isRowSelection) {
+    const path = (selection.$from as any).path;
+    return path[path.length - 8];
+  } else {
+    const fromStr = selection.$from.toString();
```
tommoor, line 9: I've honestly never used this method, not sure what the format of toString looks like but it's likely that as it's just stringifying the internal selection structure there is a better way of pulling the data out?

I'd also want to check that modifying the behavior of this method does not change how the selection toolbar works inside tables, does text selection in a table cell still show the correct toolbar?

.gitignore
PabloYG: Oops

docker-compose.yml
PabloYG: Is it intentional that the Postgress port was set to "127.0.0.1:5532:5432" instead of "127.0.0.1:5432:5432" (notice the 5 instead of 4 in the first one). Build was failing for me bc of that.

shared/editor/queries/getRowIndex.ts
PabloYG: You were right about both.

The toString wasn't ideal, I didn't quite understand how ProseMirror worked. I'm now using the same method you were using, just different index in the array (`path[path.length - 11]`).

To avoid changing the toolbar I created a separate function in the same `getRowIndex` file. Both methods get the row index so I figured it was ok to leave them in the same file. Let me know if it needs a different structure.

tommoor:
I removed the postgres port change from this branch to make it mergable, feel free to submit separately with the env.sample updated to match.

Issue #4204, opened 2022-10-01
Refactor remaining class components 


tommoor:
We would accept PR's to refactor any left over class components to functional:

https://github.com/outline/outline/search?q=%22extends+React.Component%22

One PR per component refactored is preferred!

honsq90:
I'm keen to give this a go

elysium001:
@honsq90, please lmk if I could join on the fun :)

PR #4228, opened 2022-10-04
Refactor DBDiagram class component to functional

pbkompasz:
Issue: #4204.
Refactor DBDiagram class component to functional
Added missing tests.


tommoor:
I think for these embed components as there are so many in the old style feel free to convert two or three in one PR if you like.

tommoor (approved):

PR #4248, opened 2022-10-09
add word count feature

domanchi:
### Summary

As discussed in https://github.com/outline/outline/discussions/3936, this adds an on-demand, minimal word count feature to Outline.

### Screenshots

* Location in `DocumentMenu`:

image

* Dialog without any words highlighted

image

* Dialog with words highlighted

image

### DIFF
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -18,6 +18,7 @@ import {
   CrossIcon,
   ArchiveIcon,
   ShuffleIcon,
+  NotepadIcon,
 } from "outline-icons";
 import * as React from "react";
 import { getEventFiles } from "@shared/utils/files";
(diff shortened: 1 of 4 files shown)


tommoor:
This is good, I'd probably put the menu item at the bottom end of menu alongside Print and Download. We should be aware that `document.text` does not always represent the exact content of the document – if recent edits were made then it may be lagging behind, this could be a reasonable tradeoff though.

Limezy:
How does it work with the history snapshots view ? Does it count the current number of words, or the displayed number of words from the history ?

domanchi:
> How does it work with the history snapshots view ? Does it count the current number of words, or the displayed number of words from the history ?

Great question. How do I check this?

tommoor:
I can tell you by looking at the code that it doesn't yet take this into account 😄 

You can view a document's history from the menu, cmd+k, or by clicking on the "Last edited by" timestamp to test it out

tommoor:
I think we need to figure out a way that the modal can pull the text from the active editor rather than `document.text`, this would resolve the revisions problem and make it always accurate to the current content.

tommoor:
Okay, I have a gift for you:
https://github.com/outline/outline/compare/tom/document-context

Rebase on this branch and you will have access to `DocumentContext` and `getPlainText`, enabling something like:

```typescript

const { editor } = useDocumentContext();
const text = editor.getPlainText();

// do your word/character calcs here

```

Issue #2359, opened 2021-07-22
Individual document sharing


tommoor:
We are exploring enabling individual document sharing within the team, there are a couple of use-cases for this that have come up over time:

- Sharing a draft with a colleague to get feedback or help contributing (cc @matbrgz)
- Sharing an individual document from a private collection for a one-off purpose. For example, imagine there is a "founders" collection that contains sensitive information but you want to share a single document with an EA

The key UX consideration is that this does not become a go-to method of sharing information that should be managed through collection permissions. All too often on other platforms you'll find that you were individually shared a bunch of documents that should have been shared within context, this makes them difficult to find and makes access control auditing difficult.

### Suggested limitations

Documents that have been shared with you…

- will **not** appear in the sidebar. This is key for performance and preventing the above ^.
- will **not** expose their collection or parent document info
- can appear in search results
- can have `read` or `read_write` permissions when shared

### API design

TODO

### UI design

TODO

Gwemox:
Good idea

napest:
Hi! Is there an approximate timeline for this feature?

Awoolyx:
Hey,

Individual document sharing seems very important to me personally. You might have a collection for a specific project where you have a few people working on. Here, you'd have a project lead and the project members. Things that you might want to limit is that not everyone can edit something like guidelines, which you'd want to be limited to the project lead.

Another example would be business operations, where you'd want to hide future business planning or strategy for safekeeping, while sharing something important for now.

Currently, you'd have to a separate collection for guidelines, future planning and current planning, even though it would be more reasonable to have this in the same place.

It seems a little overcomplicated needing a separate collection to manage who can edit or see what in larger teams or organizations.

praveenaj:
I'd love to have this as well!

tommoor:
This is the next “big feature” in the roadmap

DataGoblin:
Are the suggest limitations above still relevant? 

I would love to give permissions to nested documents that will show in the sidebar if the individual or group has collection/parent access. This prevents massive overlap of topics where more privilege users need separate collections for the same topic. 

For a visual example. 

**Normal user:** 
```
Collection: 
    --Document
        --Nested Document 1
```

**Special user:**
```
Collection: 
    --Document 1
        --Nested Document 1
        --Nested Document 2

    -- Document 2
```

tommoor:
> Are the suggest limitations above still relevant?

[...]

PR #4258, opened 2022-10-10
feat: add document level permissions

dolsem:
This PR adds what should be a reasonably efficient approach to document-level permissions.

Currently there are limitations:
* Users without collection-level edit permissions are not able to create nested documents, even if they have edit access to the parent document, because adding nested documents requires updating the collection (its document structure)
* Document permissions are not inhereted by child documents, to give someone access to a document and all of its children, permission should be granted individually for each document. If permissions are not granted on child documents, users with full permissions on the parent document will not be able to see child documents

Addresses #2359.


shane-at-algo:
Any update on this feature ? Would be an amazing addition to have

nullsumme:
I agree. Any update on this? It would be a shame if this PR gets closed without even discussing.

sorvis:
This would be of help for our company as well. Are there potential risks that are holding this PR back from going?

tommoor:
While some of this PR is inline with the plans for this feature, unfortunately much of the work here doesn't align with how we want to implement this permission system.

While the enthusiasm is appreciated – this is why we ask that contributors talk through their plans with the core team before starting on code rather than dump thousands of lines of code as a PR out of nowhere, it's always very unlikely to get merged in that scenario. It would take many hours just to even parse through the code here 😞 

This feature is still on the roadmap

PR #4266, opened 2022-10-11
Set subscribe/unsubscribe state correctly for documents

### LINKED ISSUE #4103: Subscribe/Unsubscribe state in document menu is incorrect before document has been viewed
The data for this is loaded when you view the doc, meaning if you go to a document menu before it's loaded it will always show "Subscribe" – even when you already are.

apoorv-mishra:
Closes #4103

### DIFF
--- a/app/menus/DocumentMenu.tsx
+++ b/app/menus/DocumentMenu.tsx
@@ -40,14 +40,13 @@ import {
   archiveDocument,
 } from "~/actions/definitions/documents";
 import useActionContext from "~/hooks/useActionContext";
-import useBoolean from "~/hooks/useBoolean";
 import useCurrentTeam from "~/hooks/useCurrentTeam";
 import useMobile from "~/hooks/useMobile";
 import usePolicy from "~/hooks/usePolicy";
+import useRequest from "~/hooks/useRequest";
 import useStores from "~/hooks/useStores";
 import useToasts from "~/hooks/useToasts";
 import { MenuItem } from "~/types";
-import Logger from "~/utils/Logger";
 import {
   documentHistoryUrl,
   documentUrl,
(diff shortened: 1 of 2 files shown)


app/menus/DocumentMenu.tsx
```diff
@@ -95,8 +97,35 @@ function DocumentMenu({
   });
   const { t } = useTranslation();
   const isMobile = useMobile();
+  const [subscriptionsLoaded, onLoad, onError] = useBoolean(false);
```
apoorv-mishra, line 100: Using local state purposely here since `subscriptions.isLoaded` is set _once and for all_ documents.

tommoor: We have too many one-offs like this, it's time to create something reusable… how about a hook:

```typescript
import { client } from "~/utils/ApiClient";

export default function useApiRequest<T>(
  path: string,
  params?: Record<string, any> | undefined
): {
  data: T | undefined;
  error: any;
  loading: boolean;
  request: () => Promise<void>;
} {
  const [data, setData] = React.useState();
  const [loading, setLoading] = React.useState(false);
  const [error, setError] = React.useState();

  const request = React.useCallback(async () => {
    setLoading(true);
    try {
      const response = await client.post(path, params);
      setData(response);
    } catch (err) {
      setError(err);
    } finally {
      setLoading(false);
    }
  }, [path, params]);

  return { data, loading, error, request };
}
```

tommoor: Designed as a generic hook, for this case actually it would be better if it took a promise instead of a `path` and `params` so that we can use the existing store methods like:

```typescript
const { data, loading, request } = useRequest(() => subscriptions.fetchPage({
  documentId: document.id,
  event: "documents.update",
}));

if (!data && !loading) {
  request();
}
```

apoorv-mishra: Yeah, tested and pushed. Seems to work fine.

tommoor (approved):

PR #4282, opened 2022-10-12
fix: Multiplayer changes potentially attributed to incorrect user

### LINKED ISSUE #4267: Document changes attributed to wrong user
Document changes on the History tab are attributed to a user that's not the author of the change.

**To Reproduce**
1. Open a document and do some changes
2. Have a collaborator be on the document with you clicking randomly but not writing anything
3. Abandon the document.
4. See that the history is incorrect and the email notifications quote the wrong person.

**Expected behavior**
That the changes that are highlighted on the History are attributed to the person making the change

**Screenshots**
imagen

- Install: getoutline.com
- Version: ?

 - OS: Windows
 - Chrome 105 and Firefox 105

tommoor:
Also performance improvements in collaboration persistence.

closes #4267

### DIFF
--- a/server/commands/documentCollaborativeUpdater.ts
+++ b/server/commands/documentCollaborativeUpdater.ts
@@ -4,6 +4,7 @@ import { Node } from "prosemirror-model";
 import * as Y from "yjs";
 import { sequelize } from "@server/database/sequelize";
 import { schema, serializer } from "@server/editor";
+import Logger from "@server/logging/Logger";
 import { Document, Event } from "@server/models";
 
 export default async function documentCollaborativeUpdater({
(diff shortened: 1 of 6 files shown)


server/commands/documentCollaborativeUpdater.ts
```diff
@@ -18,7 +19,10 @@ export default async function documentCollaborativeUpdater({
   return sequelize.transaction(async (transaction) => {
     const document = await Document.unscoped()
       .scope("withState")
-      .findByPk(documentId, {
+      .findOne({
```
tommoor, line 22: This change makes a big difference – `findByPk` overrides the scope and ignores our `"withState"` scope 😢 – this means that we've been writing the full state to disk on every persistence event. Should improve database load considerably.

PR #4285, opened 2022-10-13
Add team preference to use logo for branding

### LINKED ISSUE #4060: Add team preference to show logo on login screen
In the OSS we have a relatively unused env variable called `TEAM_LOGO` which when set will show the image on the login screen. I've never liked this solution, it feels hacky and disconnected from the main product – we can just show the same logo that shows inside the app on the login screen.

For cloud hosted login pages we show the Outline logo, which is fine but it would be much clearer to show the company logo.

### Implementation
Add a team preference, `publicLogo` or similar which enables the data to be included in the `auth.config` payload for the team and rendered on the login screen.

Depends on #3990

apoorv-mishra:
Closes #4060

### DIFF
--- a/app/scenes/Settings/Features.tsx
+++ b/app/scenes/Settings/Features.tsx
@@ -10,7 +10,6 @@ import Text from "~/components/Text";
 import useCurrentTeam from "~/hooks/useCurrentTeam";
 import useStores from "~/hooks/useStores";
 import useToasts from "~/hooks/useToasts";
-import isCloudHosted from "~/utils/isCloudHosted";
 import SettingRow from "./components/SettingRow";
 
 function Features() {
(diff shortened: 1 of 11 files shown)


app/scenes/Settings/Features.tsx
```diff
@@ -57,6 +58,22 @@ function Features() {
           />
         </SettingRow>
       )}
+      {team.avatarUrl && !isCloudHosted && (
```
tommoor, line 61:
```suggestion
      {team.avatarUrl && (
```

app/scenes/Settings/Features.tsx
```diff
@@ -57,6 +58,22 @@ function Features() {
...
+        <SettingRow
+          name="publicBranding"
+          label={t("Public branding")}
+          description={t(
+            "Whether to use team logo across entire application for branding, take Login screen as one example."
```
tommoor, line 66:
```suggestion
            "Show your team’s logo on public pages like login and shared documents."
```

Can be a follow up PR but I want to use this to also show the logo at the top of the left sidebar on shared documents such as: https://wiki.generaloutline.com/s/770a97da-13e5-401e-9f8a-37949c19f97e

tommoor (approved):

PR #4298, opened 2022-10-15
Add `SENTRY_TUNNEL` option

K3das:
This adds a `SENTRY_TUNNEL` config option to bypass adblockers when sending reports to sentry (a sentry tunnel basically acts as a proxy). This also fixes some spelling mistakes in the sample environment file

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -21,7 +21,7 @@ DATABASE_CONNECTION_POOL_MAX=
 
 # For redis you can either specify an ioredis compatible url like this
 REDIS_URL=redis://localhost:6379
-# or alternatively, if you would like to provide addtional connection options,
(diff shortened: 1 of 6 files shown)


tommoor (approved):
Hey, looks useful! Not sure what's going on with the CI pipeline here but it's not likely relevant

tommoor:
It seems like the tunnel implementation is very simple, like 10 lines of JS – perhaps the main application should just host a `/tunnel` endpoint and remove this configuration.

K3das:
Personally I have a central tunnel server on an edge CDN, but I could probably implement a local tunnel as a separate feature

tommoor:
Agreed it could be separate

PR #4336, opened 2022-10-25
chore: convert PlaceholderText component to functional component (#4204)

MihirGH:
Converts `PlaceholderText` component from class component to functional component as part of #4204.

Component specific quirks to look out for during review:
- Previous class level field - `width` is now migrated to a `widthRef`
- Previous shouldComponentUpdate is now migrated to React.memo using the custom comparison function


tommoor (approved):
🙏

PR #4337, opened 2022-10-25
chore: convert Avatar component to functional component (#4204)

MihirGH:
Converts Avatar component to function component as per #4204 

Brief summary:
1. Removed `@observable` in favour of `React.useState` - not sure if that's the correct migration strategy. 
2. Removed `@observer` in favour of wrapping `Avatar` with `observer`

### DIFF
--- a/app/components/Avatar/Avatar.tsx
+++ b/app/components/Avatar/Avatar.tsx
@@ -1,4 +1,3 @@
-import { observable } from "mobx";
 import { observer } from "mobx-react";
 import * as React from "react";
 import styled from "styled-components";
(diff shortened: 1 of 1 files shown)


tommoor:
> Removed @observable in favour of React.useState - not sure if that's the correct migration strategy.

It is!

app/components/Avatar/Avatar.tsx
```diff
@@ -72,4 +67,4 @@ const CircleImg = styled.img<{ size: number; $showBorder?: boolean }>`
   flex-shrink: 0;
 `;
 
-export default Avatar;
+export default observer(Avatar);
```
tommoor, line 70:
```suggestion
export default Avatar;
```

This shouldn't be needed anymore, only required when the component contains `mobx` state.

app/components/Avatar/Avatar.tsx
```diff
@@ -16,35 +15,31 @@ type Props = {
...
-@observer
-class Avatar extends React.Component<Props> {
-  @observable
-  error: boolean;
+function Avatar(props: Props) {
+  const { src, icon, showBorder, ...rest } = props;
 
-  static defaultProps = {
-    size: 24,
+  const [error, setError] = React.useState(false);
```
tommoor, line 21: Could potentially use `useBoolean` hook here to avoid defining `handleError` inline

MihirGH:
Resolved. Thanks for reviewing!

tommoor (approved):

PR #4338, opened 2022-10-25
Document emoji picker

### LINKED ISSUE #4274: Add emoji picker for document title
Currently we have a hidden feature where you can give a document an emoji that will appear in the favicon and elsewhere by adding it as the first character in the title followed by a space. This is too hidden – we should convert it to an actual emoji picker, probably [Emoji Mart](https://missiveapp.com/open/emoji-mart) with a subtle button next to the title on hover to choose an emoji when there isn't one.

To make the migration easier we could continue to store the emoji character in the title and strip it when displayed .e.g in the document itself.

apoorv-mishra:
image

Closes #4274


app/components/ContentEditable.tsx
```diff
@@ -129,6 +129,29 @@ const ContentEditable = React.forwardRef(
...
+            role="textbox"
+            {...rest}
+          >
+            {innerValue}
+          </Content>
+          {React.Children.toArray(children)[1]}
+        </div>
+      );
+    }
+
```
apoorv-mishra, line 154: Resembles a _point hack_ to _position_ the `EmojiPicker` in this component. Although, it works fine but I'm sure there are better ways. One way is to move the _first child_(referencing `EmojiPicker`) out and wrap it in `EditableTitle` but an initial trial there messed up its(emoji picker's) position in DOM. Inputs welcome! @tommoor

app/components/EmojiPicker.tsx
```diff
@@ -0,0 +1,95 @@
...
+      >
+        {(disclosureProps) => React.cloneElement(disclosure, disclosureProps)}
+      </PopoverDisclosure>
+      <BasePopover {...popover} {...rest}>
+        {showPopoverArray && <PopoverArrow {...popover} />}
+        {children}
+      </BasePopover>
+    </>
+  );
+};
```
apoorv-mishra, line 49: IMO the existing [`Popover`](https://github.com/outline/outline/blob/main/app/components/Popover.tsx) isn't sufficiently abstracted, revealing certain limitations. For example, it doesn't allow _injecting_ a custom `PopoverDisclosure` as opposed to this one. It'd be better if we can replace the existing one with this new abstraction. Leaving it here for now to contain the scope of current PR. It could be taken up separately. Thoughts on this @tommoor ?

tommoor: Yea, I can be into this – allows keeping the popover state out of the parent component

app/components/ContentEditable.tsx
```diff
@@ -129,6 +129,29 @@ const ContentEditable = React.forwardRef(
       []
     );
 
+    if (React.Children.count(children) > 1) {
```
tommoor, line 132: Clever, but I'm not a fan of the copy/paste here in `<Content` – probably better to just introduce an optional `beforeTitle` prop or similar.

app/components/Emoji.tsx
```diff
@@ -0,0 +1,39 @@
+import data from "@emoji-mart/data";
```
tommoor, line 1: We're going to have to see if this can replace the `gemoji` library usage too soon, these emoji libs are huge and now we have two + `emoji-regex` as well.

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -112,10 +120,26 @@ const EditableTitle = React.forwardRef(
...
-      fontSize,
-      lineHeight,
-    });
+    const handleEmojiSelect = React.useCallback(
+      async (emoji: string) => {
+        if (document.emoji !== emoji) {
+          await document.store.update({
+            id: document.id,
+            emoji,
+          });
```
tommoor, line 129:
```suggestion
          document.emoji = emoji;
          await document.save();
```

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -112,10 +120,26 @@ const EditableTitle = React.forwardRef(
...
+      },
+      [document.id, document.emoji, document.store]
+    );
+
+    const handleEmojiRemove = React.useCallback(async () => {
+      if (document.emoji) {
+        await document.store.update({
+          id: document.id,
+          emoji: null,
+        });
```
tommoor, line 140:
```suggestion
          document.emoji = null;
          await document.save();
```

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -128,14 +152,28 @@ const EditableTitle = React.forwardRef(
...
         maxLength={DocumentValidation.maxTitleLength}
         readOnly={readOnly}
         dir="auto"
         ref={ref}
       >
+        <EmojiPicker
+          disclosure={
+            <EmojiButton size={32} onClick={(ev) => ev.stopPropagation()}>
+              {document.emoji ? (
+                <Emoji size="24px" native={document.emoji} />
```
tommoor, line 167: Probably should be the same size?

apoorv-mishra: With and without `size="24px"` respectively :point_down: 

Screenshot from 2022-11-03 12-30-59
Screenshot from 2022-11-03 12-30-14

(The one without occupies `36px` width)

app/models/Document.ts
```diff
@@ -100,12 +103,6 @@ export default class Document extends ParanoidModel {
 
   revision: number;
 
-  @computed
-  get emoji() {
```
tommoor, line 104: Previously if you star a document then the emoji is shown in the sidebar next to the document instead of the star, that doesn't seem to be the case anymore…

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -128,14 +170,31 @@ const EditableTitle = React.forwardRef(
...
+        $containsEmoji={!!document.emoji}
+        $isFocused={isFocused}
         autoFocus={!document.title}
         maxLength={DocumentValidation.maxTitleLength}
         readOnly={readOnly}
         dir="auto"
         ref={ref}
       >
+        {(!isMobile || isFocused || document.emoji) && (
+          <EmojiPicker
```
tommoor, line 183: We're showing this edit UI to users that don't have edit permissions, the button to choose should not show on hover for viewers

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -195,13 +278,32 @@ const Title = styled(ContentEditable)<TitleProps>`
         opacity: 1;
       }
     }
+
+    ${AnimatedEmoji} {
```
tommoor, line 282: If you hover over the title and move the mouse slowly left there is a few pixels where the emoji button disappears and no longer counts as hovering.

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -112,10 +142,22 @@ const EditableTitle = React.forwardRef(
...
+      async (emoji: string) => {
+        if (document.emoji !== emoji) {
+          document.emoji = emoji;
+          await document.save();
+        }
+      },
+      [document]
+    );
+
+    const handleEmojiRemove = React.useCallback(async () => {
```
tommoor, line 155: Can we just reuse `handleEmojiSelect` with an `null` value?

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -50,6 +57,29 @@ const EditableTitle = React.forwardRef(
...
+    const pickerTheme = usePickerTheme();
+    const isMobile = useMobile();
+    const activeElement = useActiveElement();
+
+    const [isFocused, setFocus] = React.useState<boolean>(false);
+
+    React.useEffect(() => {
+      if (
+        activeElement &&
+        (["emoji-picker-disclosure", "emoji-picker", "document-title"].includes(
```
tommoor, line 70: The `<Title` component exposes ref, so we should be able to use that rather than a hardcoded `id` here for `document-title`?

tommoor (changes requested):
image

Do you know if it's possible to hide this section with a prop? Doesn't feel necessary to show the short-codes for this usecase and it's kind of large and distracting.

app/components/Emoji.tsx
tommoor: PR: https://github.com/outline/outline/pull/4390

app/scenes/Document/components/EditableTitle.tsx
apoorv-mishra: Yes, specifically a [`RefHandle`](https://github.com/outline/outline/blob/main/app/components/ContentEditable.tsx#L57-L79) which is a little problematic since `ref` in this case doesn't seem to expose the `HTMLElement` :end: 

Anyways, one way to tackle/improve this is to have three refs, i.e, `emojiPickerRef`, `disclosureRef` and `titleRef`, and snap these to the corresponding components. Then we could compare these against the `activeElement`. What do you think?

apoorv-mishra: TIL [`useId`](https://beta.reactjs.org/apis/react/useId). Maybe we can use this as an alternate?

tommoor: I don't think `useId` really offers anything here. 

> in this case doesn't seem to expose the HTMLElement

Oh true, you'd just want to add it on `useImperativeHandle` in `ContentEditable` as `element` I suppose, that seems like a useful edition.

apoorv-mishra: > I don't think `useId` really offers anything here.

True. I was implying using it in the components which form the _focus group_ in our case and then use `ref.current.id === activeDocument.id` for comparison but now I realize that this isn't needed and just `ref.current === activeDocument` would be sufficient.

> you'd just want to add it on useImperativeHandle in ContentEditable as element I suppose

Yes, tried attempting this only to run into issues. Tried :point_down: 

```js
element: () => {
  return contentRef.current;
}
```

only to get the entire custom object passed to `useImperativeHandle` :smile: 

Maybe I did something stupid, but you're right and it occurred to me as well that patching `useImperativeHandle` makes sense. I'll give it a shot again today :+1:

tommoor:
Just did a run through, still a couple of small refinements needed to get this over the finish line – mostly aesthetics and small bits of UX:

- [ ]  When scrolling the content is chopped off before the end of the popover
image

- [ ] This menu looks awful, the blur is much worse than in the [demo](https://missiveapp.com/open/emoji-mart) of emoji mart, where it's much more subtle.
  image

- [ ] The interactive elements in the picker don't respect the users cursor setting, can use `var(--pointer)` to get the correct hover cursor preference.

- [ ] When the popover opens the emoji placeholder disappears, it should remain while the popover is open otherwise it is unanchored – see how the outline around our "..." buttons remains when the menu is open.
image

- [ ] The styling of popovers on mobile has been broken, they used to be full-width and now float like on desktop.

- [ ] This "Remove" button looks disabled, can we use the standard `Button` component here?
image

- [ ] The emoji should be shown instead of a document icon in this list of child documents when available, I believe it worked before – should be super easy now there is a dedicated property for it. 
image

apoorv-mishra:
> This menu looks awful, the blur is much worse than in the [demo](https://missiveapp.com/open/emoji-mart) of emoji mart, where it's much more subtle.

This is due to `backdrop-filter: blur(4px)` set by the lib. From the lib's end, it appears to be tied to `--rgb-background` and on our end we are constrained to set its value to `props.theme.menuBackground` to ensure consistency with existing menus' styles :smile: 

> The interactive elements in the picker don't respect the users cursor setting, can use var(--pointer) to get the correct hover cursor preference.

Same goes for this. There is not way to pass down `cursor` value through any of the exposed `vars` of the lib.

We seem to be limited by the lack of `part` annotations required to access shadow DOM elements of the lib :disappointed: 
(Issue ref: https://github.com/missive/emoji-mart/pull/701)

tommoor:
I just realized that the `emoji` is not stored in the collection structure so they don't show there until the full document info is loaded either 😢

server/migrations/20230804034846-migrate-emoji-in-document-title.js
```diff
@@ -0,0 +1,61 @@
...
+      "shared/utils/parseTitle"
+    ));
+
+    await Document.scope(["withoutState", "withDrafts"]).findAllInBatches(
+      {
+        limit: 100,
+        offset: 0,
+        order: [["createdAt", "ASC"]],
+        paranoid: false,
+      },
```
apoorv-mishra, line 34: This is safe, models are initialized _before_ the sequelize context is passed on to `umzug` for running migrations.

server/migrations/20230804034846-migrate-emoji-in-document-title.js
```diff
@@ -0,0 +1,61 @@
...
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    if (process.env.DEPLOYMENT === "hosted") {
+      return;
+    }
+
+    if (process.env.NODE_ENV === "test") {
+      return;
+    }
```
apoorv-mishra, line 13: No point running in `test` env because _all_ migrations are run as part of the test harness on a freshly created test db, so there's _no test data_ yet available.

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -43,175 +50,209 @@ type Props = {
...
+      $containsEmoji={!!document.emoji}
+      $isFocused={isFocused}
+      autoFocus={!document.title}
+      maxLength={DocumentValidation.maxTitleLength}
+      readOnly={readOnly}
+      dir="auto"
+      ref={ref}
+    >
+      {can.update ? (
+        (!isMobile || isFocused || document.emoji) && (
```
apoorv-mishra, line 243: Cryptic conditional checks! :see_no_evil:

apoorv-mishra:
Fixes pending:

- [ ] The emoji skin tone setting is not respected yet, the emojis always use default skin tone

server/scripts/20230803172011-migrate-emoji-in-document-title.ts
```diff
@@ -0,0 +1,59 @@
...
+              transaction,
+            });
+          }
+        } catch (err) {
+          console.error(`Failed at ${document.id}:`, err);
+          continue;
+        }
+      }
+
+      return documents.length === limit ? work(page + 1) : undefined;
```
tommoor, line 45: Because `work` is returned here recursively the transaction is never closed, we need to make sure that the transaction is committed before the next iteration of `work` is called

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -43,175 +48,223 @@ type Props = {
...
+    >
+      {can.update ? (
+        <EmojiPicker
+          value={document.emoji}
+          onChange={handleEmojiChange}
+          // Restore focus on title
+          onClickOutside={handleClick}
+        />
+      ) : document.emoji ? (
+        <EmojiButton size={32}>
```
tommoor, line 259: A bit weird to use a component called `EmojiButton` for the non-editable state?

app/components/EmojiPicker.tsx
```diff
@@ -0,0 +1,229 @@
...
+import Emoji from "./Emoji";
+import NudeButton from "./NudeButton";
+
+type Props = {
+  value?: string | null;
+  onChange: (emoji: string | null) => Promise<void>;
+  onClickOutside: () => void;
+};
+
+const locales = {
```
tommoor, line 24: No need for this mapping? It'll always be the first part of the code as this is a standard, can split by `_`

app/components/EmojiPicker.tsx
```diff
@@ -0,0 +1,229 @@
...
+    }
+  }, [popover.visible, onClickOutside]);
+
+  return (
+    <>
+      <PopoverDisclosure {...popover}>
+        {(props) => (
+          <EmojiButton size={32} {...props} onClick={handleClick}>
+            {value ? (
+              <Emoji size="24px" native={value} />
```
tommoor, line 114: It would be good to have some hover state on this to hint that it's clickable when an emoji already exists, even just making it larger on hover would help.

server/migrations/20230804034846-migrate-emoji-in-document-title.js
```diff
@@ -0,0 +1,67 @@
...
+      "server/models"
+    ));
+    const { default: parseTitle } = require(path.join(
+      process.cwd(),
+      rootDir,
+      "shared/utils/parseTitle"
+    ));
+
+    await queryInterface.sequelize.transaction(async (transaction) =>
+      Document.unscoped().findAllInBatches(
```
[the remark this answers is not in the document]
apoorv-mishra: Hmm...you seem to be running the migration using cli. Tried with `make up` or `yarn dev`? That way all models are initialized before any migration runs.

app/components/EmojiPicker.tsx
apoorv-mishra: How about showing a border upon hover?

app/scenes/Document/components/EditableTitle.tsx
apoorv-mishra: Hmm, something like `<EmojiWrapper>`, perhaps?

package.json
```diff
@@ -24,6 +24,7 @@
     "db:rollback": "sequelize db:migrate:undo",
     "db:reset": "sequelize db:drop && sequelize db:create && sequelize db:migrate",
     "upgrade": "git fetch && git pull && yarn install && yarn heroku-postbuild",
+    "test:db:migrate": "NODE_ENV=test yarn sequelize db:migrate --env=test",
```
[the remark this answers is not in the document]
apoorv-mishra: Yes, but that doesn't set the `NODE_ENV`, which is needed to prevent the script from running in test env.

server/scripts/20230803172011-migrate-emoji-in-document-title.ts
apoorv-mishra: Good catch! Better to run it iteratively then I guess, page by page?

app/components/EmojiPicker.tsx
tommoor: sure

package.json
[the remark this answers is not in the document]
tommoor: Lets set it in the Makefile 👍

server/scripts/20230803172011-migrate-emoji-in-document-title.ts
tommoor: Just move the return statement outside of the sequelize transaction and it will solve it

server/migrations/20230804034846-migrate-emoji-in-document-title.js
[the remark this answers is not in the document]
tommoor: Yea, running `yarn db:migrate`, this has to work in production too so that needs to work independently 😬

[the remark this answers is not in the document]
apoorv-mishra: Urgh! That's gonna be a pain without the model being available :grimacing:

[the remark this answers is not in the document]
tommoor: It might be that we just shell out to the script from the migration? It would be a first… but agreed that we need to use the models here and would avoid having the logic twie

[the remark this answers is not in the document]
apoorv-mishra: Yeah, we can do for this case now. But going forward, I believe better to move to `umzug` completely and avoid sequelize-cli since sequelize-cli anyway uses `umzug` underneath.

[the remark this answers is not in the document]
apoorv-mishra: By the way, they seem to have rolled out cli as of v3 https://github.com/sequelize/umzug#cli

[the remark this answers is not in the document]
tommoor: Ok, happy to tackle as a followup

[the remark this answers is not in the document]
tommoor: > It might be that we just shell out to the script from the migration? It would be a first… but agreed that we need to use the models here and would avoid having the logic twice

It seems like you removed the migration entirely? I suggested to shell out, e.g use `exec` to call the script from the migration, that way we can assure it is run successfully still.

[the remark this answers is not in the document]
apoorv-mishra: Sorry, added it back :+1:

apoorv-mishra:
Fixes:

- [ ] Align emojis in sidebar to match that of production
- [ ] Display emojis in publish document dialog
- [ ] When there is no existing emoji and you click to open the picker the placeholder emoji disappears

app/scenes/Document/components/EditableTitle.tsx
```diff
@@ -253,7 +257,7 @@ const EditableTitle = React.forwardRef(function _EditableTitle(
           value={document.emoji}
           onChange={handleEmojiChange}
           // Restore focus on title
-          onClickOutside={handleClick}
+          onClickOutside={restoreFocus}
```
tommoor, line 260: It's no longer possible to click and move the caret in the title due to this change

apoorv-mishra: Nope! It's because of the `extraArea` mixin :see_no_evil:

tommoor (approved):
Finally done here, please don't merge.

patzm:
very nice!

PR #4359, opened 2022-10-29
chore: convert AddPeopleToGroup component to functional component (#4204)

MihirGH:
Converts `AddPeopleToGroup` component to functional component. 

Brief summary of changes:
1.  Used `useBoolean` hook to keep track of `inviteModalOpen` state 
2. Used `useState` hook to keep track of `query`
3. Since we don't have the liberty of class components anymore, we need to mindfully create the debounced function so I have created the `debouncedFetch` using useMemo hook and to keep the dependencies minimal (to avoid recreating the new instance of debounced function) I have started taking the query in the argument instead of using the state variable. I don't think it will create any problem as debounce is called for the latest arguments and hence it should reflect the latest `query` as well.
4. Removed the withStores and withTranslation hocs in favour of `useStores` and `useTranslation` hooks and removed the `@observer` directive

### DIFF
--- a/app/scenes/GroupMembers/AddPeopleToGroup.tsx
+++ b/app/scenes/GroupMembers/AddPeopleToGroup.tsx
@@ -1,9 +1,6 @@
 import { debounce } from "lodash";
-import { observable } from "mobx";
-import { observer } from "mobx-react";
 import * as React from "react";
-import { WithTranslation, withTranslation } from "react-i18next";
-import RootStore from "~/stores/RootStore";
+import { useTranslation } from "react-i18next";
 import Group from "~/models/Group";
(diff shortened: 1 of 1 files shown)


MihirGH:
I also wanted to write unit tests for this component as to ensure that I didn't introduce any breaking changes while refactoring but I wasn't able to find the testing-library setup in the codebase. I think we should integrate testing-library as well so contributors can add proper unit tests

tommoor:
We do have `enzyme` for component tests, it's used in `PaginatedList.test.tsx` – I probably wouldn't introduce additional dependencies for this

tommoor (approved):

PR #4360, opened 2022-10-29
chore: convert AddGroupToCollection component to functional component (#4204)

MihirGH:
Converts `AddPeopleToGroup` component to functional component. 

Brief summary of changes:

1.  Used `useBoolean` hook to keep track of `newGroupModalOpen` state 
2. Used `useState` hook to keep track of `query`
3. Created the `debouncedFetch` using `useMemo` hook and to keep the dependencies minimal took the `query` in the argument instead of reading from the state variable and adding it to useMemo deps.
4. Removed the `withStores` and `withTranslation` hocs in favour of `useStores` and `useTranslation` hooks and removed the `@observer` directive as well.

### DIFF
--- a/app/scenes/CollectionPermissions/AddGroupsToCollection.tsx
+++ b/app/scenes/CollectionPermissions/AddGroupsToCollection.tsx
@@ -1,4 +1,5 @@
 import { debounce } from "lodash";
+import { observer } from "mobx-react";
 import * as React from "react";
 import { useTranslation } from "react-i18next";
 import styled from "styled-components";
@@ -152,4 +153,4 @@ const ButtonWrap = styled.div`
   margin-left: 6px;
 `;
 
-export default AddGroupsToCollection;
+export default observer(AddGroupsToCollection);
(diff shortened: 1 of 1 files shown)


app/scenes/CollectionPermissions/AddGroupsToCollection.tsx
```diff
@@ -69,84 +75,81 @@ class AddGroupsToCollection extends React.Component<Props> {
...
+    </Flex>
+  );
 }
 
 const ButtonWrap = styled.div`
   margin-left: 6px;
 `;
 
-export default withTranslation()(withStores(AddGroupsToCollection));
+export default AddGroupsToCollection;
```
tommoor, line 155:
```suggestion
export default observer(AddGroupsToCollection);
```

Needs wrapping in observer as the component is reading MobX state – otherwise it won't react if underlying data changes.

tommoor (commented):
looks good other than observer wrapper

app/scenes/CollectionPermissions/AddGroupsToCollection.tsx
MihirGH: Interested in knowing how do you know by looking at props if something is reading from mobx state or not? 
I haven't used mobx so I don't know the difference.

tommoor (approved):

Issue #1038, opened 2019-09-12
MathJax support


alichtl:
My team creates a lot of documentation in a combination of Markdown and LaTex equations. We're very interested in moving to Outline, and support for MathJax would enable us to switch over.

tclain:
> My team creates a lot of documentation in a combination of Markdown and LaTex equations. We're very interested in moving to Outline, and support for MathJax would enable us to switch over.

hi @adamlichtl have you considered mathpix.com ? They have a product (snip) specially tailored for markdown/latex editing and teams support is landing soon (full disclosure, a friend of mine is the CEO). maybe it will be useful as it is specialized for scientific content authoring. 

Still outline is a great product

tommoor:
Totally agree that the rendering should be handled internally, would be very happy to see this functionality added to either Outline itself or the editor project, here: 
https://github.com/outline/rich-markdown-editor

hussaini2760:
Is good to enhance the community with good ideas over a project.

controversial:
Notion uses [KaTeX](https://katex.org/) which renders quickly and synchronously to HTML on the user’s own device (instead of using an external API to render images).

tommoor:
Agreed KaTeX would be a good choice

RyanGreenup:
KaTeX can get annoying because it doesn't support equation numbering, and the non-standard use of `tag` makes converting to LaTeX a headache, having both like [wikiJS](https://docs.requarks.io/releases) makes the most sense.

controversial:
I think for a use case like Outline, where the TeX content would exist in small blocks rather than composing the entire document, these trade-offs would be appropriate. Presenting users with two competing options to accomplish the same thing seems like it would create a much worse experience than would the few edge cases you mentioned.

matbrgz:
Keep it open please;

tommoor:
Note to self – this looks like a great way to implement support relatively quickly:

https://github.com/benrbray/prosemirror-math

soulteary:
We once built an academic community where we needed to show some formulas. Considering the rendering performance of long-form articles, we finally used server-side rendering.  @controversial @tommoor 

In fact, because outline is written by nodejs, this rendering function can be easily migrated into the project.
I have several articles before explaining how to implement a high-performance formula rendering service step by step.

- https://soulteary.com/2021/04/14/use-docker-and-node-to-build-a-formula-rendering-service-part-1.html
[...]

PR #4364, opened 2022-10-30
Add support for LaTeX inline and block expressions.

luisbc92:
### Summary

This PR addresses #1038 and adds support for including LaTeX in the document using @benrbray/prosemirror-math which provides the necessary extensions to ProseMirror.

Most of the magic comes from @tommoor's implementation at outline/rich-markdown-editor#434 with some additional changes to properly integrate it with the current code base.

I was able to make use of styled-components to introduce the required styling for math blocks in accordance to the current theme, however the styles and fonts used by KaTeX forced me to introduce css-loader and file-loader, in webpack/babel.

Please review and let me know if there are any additional changes required to accept the PR!

### Screenshots
- An example showing both inline and block expressions.
image
- The same while the inline editor is active.
image
- And with the block editor.
image
- And finally, dark mode.
image
- Both are insertable through the context menu.
image


tommoor:
Unfortunately this increases the application bundle size by 25%, which is unacceptable. One of the blockers in the previous PR was the inability to load the library async like we can with mermaid.js :(

luisbc92:
Yes I did notice that.

I think that's solvable since the size increase comes from the fonts loaded by KaTeX. The two approaches I see are:

- Integrate something like https://webpack.js.org/plugins/mini-css-extract-plugin/ to split the bundle and allow asynchronous loading.
- Import the font from KaTeX asynchronously through a style tag.

Would any of these options be acceptable?

tommoor:
Ah that does make life easier that the js increase is only 6%, as long as the fonts are only loaded when needed it's okay – however that is achieved.

luisbc92:
That should do it!

The styles are only loaded when a MathView gets created so its not possible for the fonts to load earlier.

app/editor/menus/block.tsx
```diff
@@ -111,6 +112,18 @@ export default function blockMenuItems(dictionary: Dictionary): MenuItem[] {
       icon: TableIcon,
       attrs: { rowsCount: 3, colsCount: 3 },
     },
+    {
+      name: "math_inline",
+      title: dictionary.mathInline,
+      icon: MathIcon,
+      keywords: "math latex",
```
tommoor, line 119:
```suggestion
      keywords: "math katex latex",
```

app/editor/menus/block.tsx
```diff
@@ -111,6 +112,18 @@ export default function blockMenuItems(dictionary: Dictionary): MenuItem[] {
...
+    },
+    {
+      name: "math_display",
+      title: dictionary.mathDisplay,
+      icon: MathIcon,
+      keywords: "math latex block",
```
tommoor, line 125:
```suggestion
      keywords: "math katex latex",
```

app/editor/menus/block.tsx
```diff
@@ -111,6 +112,18 @@ export default function blockMenuItems(dictionary: Dictionary): MenuItem[] {
...
       attrs: { rowsCount: 3, colsCount: 3 },
     },
+    {
+      name: "math_inline",
+      title: dictionary.mathInline,
+      icon: MathIcon,
+      keywords: "math latex",
+    },
+    {
+      name: "math_display",
```
tommoor, line 122:
```suggestion
      name: "math_block",
```

Ideally we'd name this `math_block` to match `code_block`

app/hooks/useDictionary.ts
```diff
@@ -69,6 +69,8 @@ export default function useDictionary() {
       strong: t("Bold"),
       subheading: t("Subheading"),
       table: t("Table"),
+      mathInline: t("Math In-Line (LaTeX)"),
```
tommoor, line 72:
```suggestion
      mathInline: t("Math Inline (LaTeX)"),
```

shared/editor/nodes/MathDisplay.ts
```diff
@@ -0,0 +1,53 @@
...
+import { NodeSpec, NodeType, Node as ProsemirrorNode } from "prosemirror-model";
+import { EditorState } from "prosemirror-state";
+import { MarkdownSerializerState } from "../lib/markdown/serializer";
+import mathRule from "../rules/math";
+import { Dispatch } from "../types";
+import Node from "./Node";
+
+export default class MathDisplay extends Node {
+  get name() {
+    return "math_display";
```
tommoor, line 16:
```suggestion
    return "math_block";
```

shared/editor/nodes/MathDisplay.ts
```diff
@@ -0,0 +1,53 @@
...
+    state.text(node.textContent, false);
+    state.ensureNewLine();
+    state.write("$$");
+    state.closeBlock(node);
+  }
+
+  parseMarkdown() {
+    return {
+      node: "math_display",
+      block: "math_display",
```
tommoor, line 49:
```suggestion
      node: "math_block",
      block: "math_block",
```

shared/editor/plugins/Math.ts
```diff
@@ -0,0 +1,76 @@
...
+        activeNodeViews: value.activeNodeViews,
+        macros: value.macros,
+        prevCursorPos: oldState.selection.from,
+      };
+    },
+  },
+  props: {
+    nodeViews: {
+      math_inline: createMathView(false),
+      math_display: createMathView(true),
```
tommoor, line 71:
```suggestion
      math_block: createMathView(true),
```

shared/editor/rules/math.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (state.src.slice(pos, max).trim().slice(-2) === "$$") {
+      lastPos = state.src.slice(0, max).lastIndexOf("$$");
+      lastLine = state.src.slice(pos, lastPos);
+      found = true;
+    }
+  }
+
+  state.line = next + 1;
+
+  const token = state.push("math_display", "math", 0);
```
tommoor, line 164:
```suggestion
  const token = state.push("math_block", "math", 0);
```

shared/editor/rules/math.ts
```diff
@@ -0,0 +1,180 @@
...
+    state.getLines(start + 1, next, state.tShift[start], true) +
+    (lastLine && lastLine.trim() ? lastLine : "");
+  token.map = [start, state.line];
+  token.markup = "$$";
+  return true;
+}
+
+export default function markdownMath(md: MarkdownIt) {
+  md.inline.ruler.after("escape", "math_inline", mathInline);
+  md.block.ruler.after("blockquote", "math_display", mathDisplay, {
```
tommoor, line 177:
```suggestion
  md.block.ruler.after("blockquote", "math_block", mathDisplay, {
```

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,112 @@ export type Props = {
...
+	min-width: 1em;
+	min-height: 1em;
+	font-size: 0.95em;
+	font-family: "Consolas", "Ubuntu Mono", monospace;
+	cursor: auto;
+}
+
+.math-node.empty-math .math-render::before {
+	content: "(empty math)";
+	color: red;
```
tommoor, line 29: Can use the red from the theme here

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,112 @@ export type Props = {
...
+}
+
+.math-node .math-render.parse-error::before {
+	content: "(math error)";
+	color: red;
```
tommoor, line 34: Can use the red from the theme here

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,112 @@ export type Props = {
...
+
+math-inline {
+  display: inline; white-space: nowrap;
+
+}
+
+math-inline .math-render { 
+	display: inline-block;
+	font-size: 0.85em;
+	cursor:pointer;
```
tommoor, line 59:
```suggestion
```

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,112 @@ export type Props = {
...
+}
+
+math-inline .math-src .ProseMirror {
+	display: inline;
+	/* Necessary to fix FireFox bug with contenteditable, https://bugzilla.mozilla.org/show_bug.cgi?id=1252108 */
+  border-radius: 4px;
+  border: 1px solid ${props.theme.codeBorder};
+  /* background: ${props.theme.codeBackground}; */
```
tommoor, line 67:
```suggestion
```

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,112 @@ export type Props = {
...
+math-display .math-src .ProseMirror {
+	width: 100%;
+	display: block;
+}
+
+math-display .math-src::after, math-display .math-src::before {
+	content: "$$";
+	text-align: left;
+	color: ${props.theme.codePunctuation};
+}
```
tommoor, line 103:
```suggestion
```

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,112 @@ export type Props = {
...
+const mathStyle = (props: Props) => `
+/* Based on https://github.com/benrbray/prosemirror-math/blob/master/style/math.css */
+
+/* == Math Nodes ======================================== */
+
+.math-node {
+	min-width: 1em;
+	min-height: 1em;
+	font-size: 0.95em;
+	font-family: "Consolas", "Ubuntu Mono", monospace;
```
tommoor, line 23:
```suggestion
	font-family: ${props.theme.fontFamilyMono};
```

shared/editor/components/Styles.ts
tommoor: I see the logic behind them but it doesn't follow how other blocks work in Outline to show this

shared/editor/components/Styles.ts
tommoor: A bit confusing, especially in read-only mode.

tommoor (commented):
Pretty great, works with collaborative editing too 😄 

- [x] Lets icrease the padding in the math editor to match that of code blocks – ideally these should be as visually similar as possible.
- [ ] I think we should probably remove inline math from the block menu as the interaction isn't great leaving the "(empty math)" node and the block menu only works at the beginning of lines anyway?
- [ ] Need to add these new `$$` shortcuts to the list in `<KeyboardShortcuts />` component

**Random thought**, we should probably update Mermaid blocks to work the same way for consistency where they show the source when focused or read-only and the generated image otherwise.

shared/editor/components/Styles.ts
luisbc92: Agreed on consistency with the styling. However, you can also trigger a block by typing $$<space>, but I think its enough to have it on the KeyboardShortcuts component.

shared/editor/components/Styles.ts
luisbc92: Got it on the next commit.

app/editor/menus/block.tsx
luisbc92: Oops, forgot to commit this one, but I'll leave it for the next commit.

luisbc92:
You are right, it doesn't make much sense to have inline math on the menu.

I also refactored MathDisplay to MathBlock for consistency, and got the shortcuts on the component.

And yes! It'd be nice for Mermaid to show up the same way, giving you a nice clean look when not editing.

luisbc92:
Is there anything I can do to help this feature get into mainline sooner?

tommoor (commented):
Sorry for delay, gotta make sure it's just right and that takes some testing – a couple of small things here.

The other thing that's probably not a blocker, but is a bit weird – when you're editing the math block/inline the rest of the editor loses focus. It's the only node that does this, have you seen why that happens?

luisbc92:
I hadn't noticed that. Let me look into it, I might have an idea on why that might happen.

shared/editor/components/Styles.ts
```diff
@@ -11,6 +11,106 @@ export type Props = {
...
+
+/* -- Inline Math --------------------------------------- */
+
+math-inline {
+  display: inline; white-space: nowrap;
+
+}
+
+math-inline .math-render { 
+	display: inline-block;
```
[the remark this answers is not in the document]
luisbc92: Sure!

luisbc92:
I haven't been able to replicate any loss of focus, this is what I've tried:

- Insert math inline by typing $E=mc^2$, then going in and out of it with arrow keys.
- Clicking on math inline, then going in and out with arrow keys.
- Insert code block from menu (using mouse) and start typing, then going in and our with arrow keys.
- Insert code block from menu (triggered by /) and start typing, then ...
- Inserting multiple inline/blocks interspersed with other blocks and moving cursor around.

Do you have a specific scenario where you found this?

tommoor:
We have a lot of styles that are only applied when the editor is in focus, for example the "H1" markers infront of headings (create a few headings in your doc to see), it's a clue that you're in editing mode.

When moving into a latex block these disappear as the entire editor seems to lose the appropriate classes.

luisbc92:
Found it.

**prosemirror-math** is creating an inner **EditorView** for editing LaTeX which steals focus from the outer ProseMirror. I think the only way to solve this is by re-implementing the **NodeView** provided by **prosemirror-math** and use a different method to toggle editing, which I could do in the future.

At that point however, it might make more sense to bring that code over and remove the dependency, but I'd rather leave that for later if the drawback is minimal.

TerenceLiu98:
@luisbc92 If I want to try the `latex-support` version, should I clone the [latex-support](https://github.com/luisbc92/outline/tree/latex-support) and build the image locally? :)

luisbc92:
@TerenceLiu98 yes, you should be able to build it and run it locally. Be sure to follow the instructions [here](https://app.getoutline.com/s/770a97da-13e5-401e-9f8a-37949c19f97e/doc/local-development-5hEhFRXow7) to setup your environment correctly.

TerenceLiu98:
> @TerenceLiu98 yes, you should be able to build it and run it locally. Be sure to follow the instructions [here](https://app.getoutline.com/s/770a97da-13e5-401e-9f8a-37949c19f97e/doc/local-development-5hEhFRXow7) to setup your environment correctly.

Thanks! I first only build the `Dockerfile` without building the `Dockerfile.base`, but then I figure it out.

tommoor:
Moving this work to a branch in the main repo for final touchup.

luisbc92:
Excellent! Thanks Tom.

jpedrego:
Dear Outline Team. Thanks for including Latex into the Outline Markdown.

However, in *my* ideal world, I would use an offline Markdown editor to prepare the documentation and then simply copy/paste on Outline to share it with colleagues for review or use when it is ready.

Such work flow can not be currently (easily)  implemented due to two reasons:
- First, Outline decided to choose the use of $$ x $$ for inline and $$$ x $$$ for blocks, breaking away from the standard $x$ for inline and $$ x $$ for blocks that everyone else uses. I am surprise by such decision, but maybe there is a hidden technical reason...

-Second: when I have tried doing copy/paste from Markdown formatted text (after changing $$ => $$$ and $ => $$) the Latex parts are not recognized and the $ symbols are simply displayed as dollar symbols.


I am aware that I am being picky but still... 

In any case, if anyone could address the failure when copy pasting Markdown with equations, it would be awesome!

Thanks in advance.

jpedrego:
I didn't see it coming!... I guess as I never thought about the issue with the dollar sign being used as actually being a Dollar sign !!!
PS: In my defense, an European writing here... ;-)

tommoor:
> However, GitHub and the others clearly did not think it was a problem...

They use plain markdown, not a rich editor – so they don't have the same problems.

> The problem is that, when copy/pasting the text to Outline, the $$ and the $$$ do not seem to be taken into account....

This sounds like it should be filed as a bug

jpedrego:
It seems like the copy/paste is working today. No idea what I did different today from the other day... In any case is working.

Issue #3000, opened 2022-01-25
Move to sending JSON rather than Markdown to client


tommoor:
### Problem

Currently when you request documents from the server we send a Markdown representation to the client which is rendered immediately. For collaborative editing we also render this while waiting for the collaborative connection to start to make the document appear faster.

This worked well up until now as every type of node could be serialized to Markdown in a reversible fashion – going forward this restriction should be removed so we can have items in the doc that cannot be represented in Markdown. For example fancy new node types, file attachments, comments etc.

### Solution

Markdown will be used as an import/export type and markdown shortcuts will continue to work in the editor as today, however it will not be used to render in the app and may not be sent from the server at all unless requested explicitly for a download.

To migrate we'll need to convert all the existing docs in the database and store the JSON representation alongside the markdown.

therajumandapati:
@tommoor Do you have any ideas in mind for coming up with a JSON structure for transforming markdown?

tommoor:
The format already exists, it's what the Markdown gets parsed into. It also exists inside the CRDT in the database which contains the collaborative state of any given document… eg.

```
import { parser } from "@server/editor"
const json = parser.parse(markdown)
```

tommoor:
Work has begun on this, now that collaborative editing is the default it makes things easier. Some features that this potentially unlocks:

- Different highlighting colors
- Resizing tables/rows/columns, more content support in table cells
- Multi-column layouts
- Arbitrary embed support, resizing embeds

almereyda:
🙏🏽

PR #4367, opened 2022-10-31
feat: add Page Link component to Editor #4366

thelicato:
I've added a new component to the editor that allows the user to create a button link to another page in the KB as detailed in this old discussion we had in January: https://github.com/outline/outline/discussions/2940


tommoor:
Hi @thelicato – thanks for your efforts on this one. There are a couple of thoughts blocking me here – the main one is that I believe we need to soon have display options for every link, not just "page links" and in that sense internal links aren't really unique enough to get their own special node type.

thelicato:
The initial idea was to allow any kind of link. After looking at GitBook I found out that it allows only to link internal pages. I think this is a winning choice because allows the user to quickly know that the linked page is **another document**.

andrzejwp:
Hello @tommoor @thelicato is there any progress on this? I'd be happy to assist as I see this addition really valuable.

thelicato:
@andrzejwp it was ready to be merged when I opened this PR but @tommoor thought that the component should be used for both **internal** and **external** links. I did not agree on this since the idea was just to provide a component to enhance the UX of the navigation between Outline documents. Here is the link to the discussion: https://github.com/outline/outline/discussions/2940

Since my last message I did not receive any reply from Tom so now the PR is stale.

tommoor:
Adding new nodes is something that has to happen with great consideration as once they exist there is no going back, they need to be supported by the software forever. That's the thinking behind it – if this is simply a display option for an existing node type perhaps that's how it should be represented in the data structure as well.

The other problem to overcome with this PR is that the markdown representation is not unique, until #3000 is complete every node needs to be uniquely represented in the Markdown output so that it can be rendered correctly, as it stands the page links here will not survive export<->import, for example.

thelicato:
@tommoor Could you tell me when do you think #3000 will be implemented? Currently we are using my forked repo internally but it is troublesome to keep the forked version updated with official Outline. A quick solution to overcome the "export<->import" problem could be to add and ID or a class to the Page Link to keep it unique:
```
[Click me](<internal link>){#page-link .page-link}
```

If you don't like it let me know if  and how I can help with #3000.

thelicato:
Hi, is there any way I can help you to get this on the official branch or do I have to throw away my hopes on this?

PR #4390, opened 2022-11-07
Move from `gemoji` -> `emoji-mart`

tommoor:
We need `emoji-mart` for https://github.com/outline/outline/pull/4338 – makes sense to use the same data bundle for in-editor picking and will overall reduce bundle size once both branches are merged.


tommoor:
Merging into emoji picker branch

Issue #4284, opened 2022-10-13
Refine server side validations


apoorv-mishra:
Currently, different types of validations for a given API are scattered within the handlers as a bunch of assertions, added on ad-hoc basis. This ends up making handlers _fat_ and cluttered. It'd be good if validations could be pulled out and be handled more _declaratively_, in one place, possibly through a middleware. It'd be worth exploring [Zod](https://github.com/colinhacks/zod) for this purpose.

tommoor:
Now we're talkin'. 

I think we should write our own middleware based on zod, as you say – but this express one might provide some inspiration:
https://github.com/andrewkeig/express-validation

PR #4397, opened 2022-11-07
Introduce zod for server-side validations

apoorv-mishra:
### Todo
- [x] Cover missing tests for existing assertions
- [x] Replace `DocumentSchema` in favor of separate `input` schemas for each of the routes
  - [x] `documents.list`
  - [x] `documents.archived`
  - [x] `documents.deleted`
  - [x] `documents.viewed`
  - [x] `documents.drafts`
  - [x] `documents.info`
  - [x] `documents.export`
  - [x] `documents.restore`
  - [x] `documents.search_titles`
  - [x] `documents.search`
  - [x] `documents.templatize`
  - [x] `documents.update`
  - [x] `documents.move`
  - [x] `documents.archive`
  - [x] `documents.delete`
  - [x] `documents.unpublish`
  - [x] `documents.import`
  - [x] `documents.create`
- [x] Type `input`
- [x] Middleware to validate `input`
- [x] Fix type for `ctx` in route handlers to accommodate `input`
- [x] Replace `ctx.request.body` refs with `ctx.input` in all of the route handlers
- [x] Collate duplicate schemas
- [x] Restructure `/api` to accommodate document schema

Towards #4284


tommoor:
> Refactor to use DocumentSchema in a top level blanket middleware for all documents APIs

I don't think we want this, the validated payload needs to be specific to each endpoint.

tommoor:
I'm not sure if this is already the case but we should make sure that the `body` is typed correctly as part of this work, this is one of the big advantages of using zod.

apoorv-mishra:
> I don't think we want this, the validated payload needs to be specific to each endpoint.

Thought of doing this earlier, but not pursuing it now. Makes sense to keep it to individual routes.

server/routes/api/documents.ts
```diff
@@ -37,150 +38,176 @@ import {
...
+  DocumentsArchiveReq,
+  DocumentsDeleteReqSchema,
+  DocumentsDeleteReq,
+  DocumentsUnpublishReqSchema,
+  DocumentsUnpublishReq,
+  DocumentsImportReqSchema,
+  DocumentsImportReq,
+  DocumentsCreateReqSchema,
+  DocumentsCreateReq,
+} from "@server/routes/api/types";
```
tommoor, line 78: I think this is a case where it's better to just do something like...

```
import * as T from "@server/routes/api/types";

T.DocumentsListSchema
```

Maybe we can lose `Req` as they all have it?

tommoor (commented):
This is great work 🙏

server/routes/api/documents.ts
apoorv-mishra: > Maybe we can lose Req as they all have it?

Didn't quite get. You mean rename, for example `DocumentsCreateReqSchema` -> `DocumentsCreateSchema` but keep `DocumentsCreateReq` unchanged?

apoorv-mishra: I have `Req` there purposely to distinguish with `Res`, in case if we have that in future, something like
```js
APIContext<DocumentsCreateReq, DocumentsCreateRes>
```

tommoor: ok

tommoor: > rename, for example `DocumentsCreateReqSchema` -> `DocumentsCreateSchema` but keep DocumentsCreateReq unchanged?

Sorry, to be clear – this sounds good.

server/routes/api/types.ts
```diff
@@ -0,0 +1,284 @@
+import { isEmpty } from "lodash";
```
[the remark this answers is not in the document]
apoorv-mishra: There are two things - types and schemas. I assume you're referring to move only `types`? Not sure if it's required to move schemas into `@shared`.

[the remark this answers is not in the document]
tommoor: I did mean both but it would just be to keep them colocated in that case, as you say I doubt schema would be needed on frontend

tommoor:
@SocketSecurity ignore koa-webpack-hot-middleware@1.0.3

wtf?

tommoor:
Okay, I'm thankful for the extensive tests in `api/documents` – makes this a little less scary to merge but in hindsight I do wish we'd started with a less important set of endpoints 😅

tommoor (approved):

server/routes/api/types.ts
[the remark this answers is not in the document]
apoorv-mishra: If we decide to restructure `/api` as follows, for example,

```
/api
  \--/attachments
        \--index.ts
        \--attachments.ts
        \--attachments.test.ts
        \--schema.ts
```
and if we also decide to share the types with frontend, it'd entail separating schemas and their types.

I say we keep them colocated under `schema.ts` and think about the frontend sharing part later. It's probably overengineering at this point. Thoughts @tommoor ?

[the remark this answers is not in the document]
tommoor: Okay, works for me 👍  - one improvement at a time

tommoor (approved):

server/routes/api/documents/documents.test.ts
```diff
@@ -47,6 +47,19 @@ describe("#documents.info", () => {
...
+      body: {
+        token: user.getJwtToken(),
+        id: document.urlId,
+      },
+    });
+    const body = await res.json();
+    expect(res.status).toEqual(200);
+    expect(body.data.id).toEqual(document.id);
+  });
+
```
apoorv-mishra, line 62: Missed this! Thanks for adding :pray:

PR #4400, opened 2022-11-08
feat: added user avatars in author search dropdown

### LINKED ISSUE #4193: Add user avatars in author search dropdown
When users have similar names the avatars are useful to quickly distinguish, we use them everywhere else so they should be here too.

adityash1:
fixes #4193


app/components/FilterOptions.tsx
```diff
@@ -37,6 +39,8 @@ const FilterOptions = ({
 
   const selectedLabel = selected ? `${selectedPrefix} ${selected.label}` : "";
 
+  const isAuthorDropdown = defaultLabel === "Any author" ? true : false;
```
apoorv-mishra, line 42: Doesn't seem to account for translations. Could you test this for language other than en?

adityash1: Oops! Missed that scenario. fixing it

app/components/FilterOptions.tsx
```diff
@@ -37,6 +40,9 @@ const FilterOptions = ({
 
   const selectedLabel = selected ? `${selectedPrefix} ${selected.label}` : "";
 
+  const { t } = useTranslation();
+  const isAuthorDropdown = defaultLabel === t("Any author") ? true : false;
```
tommoor, line 44: Aesthetically it looks fine but `FilterOptions` is meant to be a generic component, it doesn't make sense to have anything related specifically to avatars or authors here. 

Instead you should add an `image` or `icon` prop and then pass the `<Avatar` component in from `UserFilter`

tommoor (changes requested):

app/components/FilterOptions.tsx
adityash1: I've made the changes as you asked. Passed Avatar component as an image prop in from `UserFilter`. And I'm styling the component inside the `FilterOptions` function because of scope limitation. is this good approach or it can be done in a better way ?

tommoor: Yes, sorry bad timing – the entire `user` object must be passed to `<Avatar` now

adityash1: made the changes as required. passing user object. didn't change anything else...

adityash1:
@tommoor can you please review this if I need to make further changes. So that I can work on other issues...

apoorv-mishra:
Does the other issue that you want to pick depends upon this one in any way?

adityash1:
> Does the other issue that you want to pick depends upon this one in any way? If not, then you can just branch off latest `main` and start working on the new issue.

I thought it was better to solve issues one by one. But now I'll do as you said...

tommoor:
Hi @adityash1 you need to fix up the branch a little here, the diff includes a lot of changes from `main` that you haven't made so it's too hard to review.

adityash1:
> Hi @adityash1 you need to fix up the branch a little here, the diff includes a lot of changes from `main` that you haven't made so it's too hard to review.

It was some GitHub issue. I've fixed it. Let me know if I need to fix more things.

app/components/FilterOptions.tsx
```diff
@@ -57,6 +68,7 @@ const FilterOptions = ({
             selected={option.key === activeKey}
             {...menu}
           >
+            {Image && <StyledImage model={option.user} size={20} />}
```
tommoor, line 71: I don't know why you're designing the interface like this, so you can pass an `image` prop into `FilterOptions` but it must have a component with a `model` prop? It doesn't really make sense… `image` should just allow any component here.

adityash1: Right. It's awful. I'm making new changes, and now I'm passing `styledAvatar` component in from `UserFilter` to save it from unnecessary re-rendering in `filteroptions`...

adityash1: I've pushed the changes. and everything is as we discussed here https://github.com/outline/outline/pull/4400#discussion_r1017361813

PR #4409, opened 2022-11-11
fix: `DATABASE_URL` missing in env does not produce validation error

tommoor:
Missing `DATABASE_URL` in configuration was causing server to crash rather than produce a human readable error, see:
https://github.com/outline/outline/discussions/4408

This also fixes `DATABASE_CONNECTION_POOL_URL` never being evaluated as it sat behind the required `DATABASE_URL`.

### DIFF
--- a/server/env.ts
+++ b/server/env.ts
@@ -37,6 +37,7 @@ export class Environment {
    * @returns A promise that resolves when the environment is validated.
    */
   public validate() {
+    console.log("validating…");
     return this.validationPromise;
   }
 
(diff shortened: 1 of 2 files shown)


server/env.ts
```diff
@@ -51,14 +51,14 @@ export class Environment {
    * set or your users will be unable to login.
    */
   @IsByteLength(32, 64)
-  public SECRET_KEY = `${process.env.SECRET_KEY}`;
```
tommoor, line 54: This evaluated to `"undefined"`

PR #4412, opened 2022-11-11
fix: Self-hosted logic for allowed domains

### LINKED ISSUE #4275: Allowed Domains troubles in self-hosted
### Discussed in https://github.com/outline/outline/discussions/4083

<sup>Originally posted by **Dan6erbond** September  7, 2022</sup>
I have Outline hosted in a Kubernetes cluster, and configured with Authentik OIDC for SSO. So far this has worked well with @gmail.com Users, until I noticed that some colleagues who wanted to access with @gmx.ch, etc. were out of luck. Based on my research I then tried to add those domains to the Allowed Domains setting in Outline, but logging in with those users still doesn't work. Is this expected behavior as Outline's SSO might not cover custom OIDC providers or something else?</div>

tommoor:
closes #4275

### DIFF
--- a/server/commands/teamProvisioner.ts
+++ b/server/commands/teamProvisioner.ts
@@ -72,13 +72,13 @@ async function teamProvisioner({
     };
   } else if (teamId) {
     // The user is attempting to log into a team with an unfamiliar SSO provider
-    throw InvalidAuthenticationError();
-  }
+    if (env.DEPLOYMENT === "hosted") {
+      throw InvalidAuthenticationError();
+    }
 
-  // This team has never been seen before, if self hosted the logic is different
-  // to the multi-tenant version, we want to restrict to a single team that MAY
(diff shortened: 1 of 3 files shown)


server/commands/teamProvisioner.ts
```diff
@@ -72,13 +72,13 @@ async function teamProvisioner({
...
-    throw InvalidAuthenticationError();
-  }
+    if (env.DEPLOYMENT === "hosted") {
+      throw InvalidAuthenticationError();
+    }
 
-  // This team has never been seen before, if self hosted the logic is different
-  // to the multi-tenant version, we want to restrict to a single team that MAY
-  // have multiple authentication providers
-  if (env.DEPLOYMENT !== "hosted") {
```
tommoor, line 81: `teamId` always exists for self-hosted now, so this logic was never reached

PR #4442, opened 2022-11-17
feat: add `search title only` filter for search options

### LINKED ISSUE #4192: Add search option to "Only search titles"
Sometimes you know you're looking for something in the title and it's useful to quickly whittle down the search results.

adityash1:
fixes #4192

### DIFF
--- a/app/scenes/Search/Search.tsx
+++ b/app/scenes/Search/Search.tsx
@@ -21,6 +21,7 @@ import Flex from "~/components/Flex";
 import LoadingIndicator from "~/components/LoadingIndicator";
 import RegisterKeyDown from "~/components/RegisterKeyDown";
 import Scene from "~/components/Scene";
+import Switch from "~/components/Switch";
 import Text from "~/components/Text";
 import withStores from "~/components/withStores";
 import Logger from "~/utils/Logger";
@@ -52,11 +53,16 @@ class Search extends React.Component<Props> {
 
   lastParams: SearchParams;
 
+  lastTitleFilterState: boolean;
+
   @observable
   query: string = decodeURIComponentSafe(this.props.match.params.term || "");
 
   @observable
   params: URLSearchParams = new URLSearchParams(this.props.location.search);
+
+  @observable
+  searchTitleOnly = false;
 
   @observable
   offset = 0;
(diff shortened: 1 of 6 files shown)


app/scenes/Search/Search.tsx
```diff
@@ -52,12 +53,17 @@ class Search extends React.Component<Props> {
 
   lastParams: SearchParams;
 
+  lastTitleFilterState: boolean;
```
tommoor, line 56: Why store this separately? It should be in params so that the page can be reloaded and keep settings

app/scenes/Search/Search.tsx
```diff
@@ -171,6 +177,20 @@ class Search extends React.Component<Props> {
     });
   };
 
+  handleSearchTitleOnly = (ev: React.ChangeEvent<HTMLInputElement>) => {
+    this.searchTitleOnly = ev.target.checked;
+    this.fetchResults();
+    // Disable other filters if this filter is enabled
+    if (ev.target.checked) {
+      this.handleFilterChange({
```
tommoor, line 185: It's not great UX to just clear out all the users filters and not include them in the search, we'd need to pass these through and add support on the backend.

app/stores/DocumentsStore.ts
```diff
@@ -386,9 +386,32 @@ export default class DocumentsStore extends BaseStore<Document> {
       query,
     });
     invariant(res?.data, "Search response should be available");
+
     // add the documents and associated policies to the store
     res.data.forEach(this.add);
     this.addPolicies(res.policies);
+
+    // store a reference to the document model in the search cache instead
```
tommoor, line 394: I'm not super clear why this changed

app/scenes/Search/Search.tsx
adityash1: yes. i'll make changes on the backend to support other filters too.

app/stores/DocumentsStore.ts
adityash1: i'm storing reference to document model just like how `document_search` does and setting cache accordingly.  i'll remove extra props `ranking` `context` `id` which i'm returning. is this right?

app/scenes/Search/Search.tsx
adityash1: i've fixed it. thanks

adityash1:
@tommoor should we pass a parameter to `search` action in `DocumentsStore.ts`. if  `titleFilter` is enabled, we can fetch results from `/documents.search_titles` and merge the common result and return them.

Or we should completely change `/documents.search_titles`  to accept search parameters as we pass in `search` action in `DocumentsStore.ts`

adityash1:
@tommoor I've made necessary changes on the frontend. But as you said we should be able to combine multiple filters. But `document_search_titles` doesn't support additional filters like `document_search`. can you help me with the required changes on the backend...

tommoor:
Okay, will add backend changes in a separate PR.

adityash1:
#4587

PR #4449, opened 2022-11-20
feat: Allow data imports larger than the standard attachment size

tommoor:
- Allows configurable larger uploads for data imports
- Enforces data uploads are private and expire(delete) quickly to minimize abuse
- Cleanup/refactor of `attachments.create` endpoint
- Deprecates `public` option on `attachments.create` payload

### DIFF
--- a/server/presenters/attachment.ts
+++ b/server/presenters/attachment.ts
@@ -1,0 +1,12 @@
+import { Attachment } from "@server/models";
+
+export default function present(attachment: Attachment) {
+  return {
+    documentId: attachment.documentId,
+    contentType: attachment.contentType,
+    name: attachment.name,
+    id: attachment.id,
+    url: attachment.canonicalUrl,
+    size: attachment.size,
+  };
+}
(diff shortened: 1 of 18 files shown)


server/routes/api/attachments.ts
```diff
@@ -1,81 +1,87 @@
...
+  auth(),
+  transaction(),
+  async (ctx: TransactionContext) => {
+    const {
+      name,
+      documentId,
+      contentType = "application/octet-stream",
+      size,
+      // 'public' is now deprecated and can be removed on December 1 2022.
+      public: isPublicDeprecated,
```
tommoor, line 33: Retained for now, just so existing loaded JS continues to work

server/presenters/attachment.ts
```diff
@@ -0,0 +1,12 @@
+import { Attachment } from "@server/models";
+
+export default function present(attachment: Attachment) {
```
tommoor, line 3: Missing presenter

PR #4465, opened 2022-11-22
Add support for dark theme for mermaid diagrams

### LINKED ISSUE #4195: Mermaid.js diagrams don't work with the dark theme
**To Reproduce**
1. Create mermaid diagram
2. Select dark theme for outline

**Expected behavior**
Mermaid diagram changes the text color to white

**Screenshots**
image
image

- Install: self hosted
- Version: 0.66.1

 - OS: linux
 - Browser Firefox
 - Version 103

not7cd:
This will fix #4195,
I'm able to pull the currently chosen theme. But I don't know how to trigger re-render when the theme is changed through the user menu.

### DIFF
--- a/shared/editor/plugins/Mermaid.ts
+++ b/shared/editor/plugins/Mermaid.ts
@@ -3,6 +3,7 @@ import { Plugin, PluginKey, Transaction } from "prosemirror-state";
 import { findBlockNodes } from "prosemirror-utils";
 import { Decoration, DecorationSet } from "prosemirror-view";
 import { v4 as uuidv4 } from "uuid";
+import stores from "~/stores";
 
 type MermaidState = {
   decorationSet: DecorationSet;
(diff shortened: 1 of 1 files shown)


shared/editor/plugins/Mermaid.ts
```diff
@@ -66,6 +67,7 @@ function getNewState({
               htmlLabels: false,
             },
             fontFamily: "inherit",
+            theme: stores.ui.resolvedTheme === "light" ? "default" : "dark",
```
tommoor, line 70: Ah, thanks for trying. Unfortunately you cannot import things from `~/` into `shared` as that's importing client dependencies into the server. I think you can pass it to the plugin from `node.editor.props.theme`, you can listen for changes with `window.addEventListener("theme-changed"`

not7cd: I see that the same approach is used within `editor.components.ComponentView`. Also within this ComponentView I should have access to `node.editor.props.theme`?

tommoor:
Thanks for the push – took this over the finish line here https://github.com/outline/outline/pull/4564

not7cd:
Thanks! This definetly needed more understanding of the internals.

PR #4504, opened 2022-11-30
fix errors with custom OIDC provider

deolive:
There are 2 errors fixed in pull request:
1. if OIDC provider gets user_id as integer Postgres failed query because providerId must be a string
2. if there isn't refreshToken function getEncryptedColumn returns empty object and expression "!this.refreshToken" returns false

### DIFF
--- a/server/commands/userProvisioner.ts
+++ b/server/commands/userProvisioner.ts
@@ -60,7 +60,7 @@ export default async function userProvisioner({
   const auth = authentication
     ? await UserAuthentication.findOne({
         where: {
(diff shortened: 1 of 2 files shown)


deolive:
Should I make issues for this request?

tommoor:
Thanks, no issue needed – i'd say the bug with `getEncryptedColumn` needs fixing at the model level though, not via a workaround here – can merge the string fix.

deolive:
Yes. I will make a new pull request with only string fix

PR #4506, opened 2022-12-01
fix: string error with custom OIDC provider

deolive:
if OIDC provider gets user_id as integer Postgres failed query because providerId must be a string


deolive:
Why are there 2 commits 🤦‍♂️

PR #4507, opened 2022-12-01
fix: getEncryptedColumn for empty token

deolive:
if there isn't Token (refreshToken for example) function getEncryptedColumn returns empty object and break expression like "!this.refreshToken"


tommoor:
Thanks! I need to carefully check this doesn't break anything, it feels like `undefined` would be a better return value to represent the column not being set.

deolive:
> Thanks! I need to carefully check this doesn't break anything, it feels like `undefined` would be a better return value to represent the column not being set.

I save return type support (string), and this function already return an empty string

https://github.com/outline/outline/blob/a2f180ce5f249dffb3a032b35c340a28b456a250/server/models/decorators/Encrypted.ts#L26-L28

PR #4508, opened 2022-12-01
feat: added user avatars in author search dropdown

### LINKED ISSUE #4193: Add user avatars in author search dropdown

adityash1:
Fixes #4193

### DIFF
--- a/app/scenes/Search/components/UserFilter.tsx
+++ b/app/scenes/Search/components/UserFilter.tsx
@@ -30,7 +30,6 @@ function UserFilter(props: Props) {
     }));
     return [
       {
-        user: undefined,
         key: "",
         label: t("Any author"),
         icon: <Avatar />,
--- a/app/scenes/Search/components/UserFilter.tsx
+++ b/app/scenes/Search/components/UserFilter.tsx
@@ -1,4 +1,5 @@
 import { observer } from "mobx-react";
+import { UserIcon } from "outline-icons";
 import * as React from "react";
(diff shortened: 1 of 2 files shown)


app/scenes/Search/components/UserFilter.tsx
```diff
@@ -22,13 +23,17 @@ function UserFilter(props: Props) {
 
   const options = React.useMemo(() => {
     const userOptions = users.all.map((user) => ({
+      user,
       key: user.id,
       label: user.name,
+      icon: <Avatar model={user} />,
```
adityash1, line 29: @tommoor right now it's showing default ring on author. should we change it to something else?

app/scenes/Search/components/UserFilter.tsx
```diff
@@ -24,11 +25,13 @@ function UserFilter(props: Props) {
     const userOptions = users.all.map((user) => ({
       key: user.id,
       label: user.name,
+      icon: <Avatar model={user} />,
     }));
     return [
       {
         key: "",
         label: t("Any author"),
+        icon: <Avatar />,
```
adityash1, line 34: @tommoor right now it's showing a hollow ring. Should we change it to something else?

tommoor: How about `<UserIcon />` ?

tommoor:
I think you got it this time 🙂

PR #4526, opened 2022-12-03
Upgrade `socket.io`

tommoor:
Also drops support with v2 clients


yarn.lock
```diff
@@ -16159,11 +16099,6 @@ yarn-deduplicate@^3.1.0:
     commander "^6.1.0"
     semver "^7.3.2"
 
-yeast@0.1.2:
```
tommoor, line 16162: Always nice to remove a module called "yeast" 🤮

PR #4527, opened 2022-12-03
if OIDC provider gets user_id as integer Postgres failed query 

deolive:
Sorry for that, but previous pull request had wrong commit (((
 
if OIDC provider gets user_id as integer Postgres failed query because providerId must be a string


tommoor:
Thanks!

PR #4532, opened 2022-12-03
fix: Cleanup user menu

### LINKED ISSUE #4530: Users with long names is outfit dropdown list
Users with long names is outfit dropdown list

**To Reproduce**
1. Go to Settings->Members
2. Try change role user with very long name

**Screenshots**
image

- Install:  self hosted
- Version: latest

 - OS: MaxOs
 - Browser Chrome
 - Version 108

I think user name indropdown list is unnecessary

tommoor:
Move confirmation dialogs from window confirmations, last `window.confirm`'s in the codebase.

closes #4530


app/components/Modal.tsx
```diff
@@ -252,8 +252,9 @@ const Small = styled.div`
   animation: ${fadeAndScaleIn} 250ms ease;
 
   margin: auto auto;
+  width: 30vw;
   min-width: 350px;
-  max-width: 30vw;
```
tommoor, line 256: These were getting a bit wide

PR #4537, opened 2022-12-04
Enhance server side error handling

### LINKED ISSUE #4505: Refactor server-side error handling
Currently, the errors, if any, in top level routes, like `/_health` for example, aren't handled the way they are for the `/api/*` routes because of the absence of some equivalent of `errorHandling` middleware there. It'd be better to have a single error handler which can be wielded across server-side, i.e, both for top level as well as `/api/*` routes. Once implemented, it can replace the current `errorHandling` middleware for `/api/*` routes.

apoorv-mishra:
Closes #4505 

- [x] Replicate `koa-onerror` behavior with in-house `server/onerror`
- [x] Treat `text` type reqs with `json` error responses
- [x] `readErrorFile` to load HTML error template
- [x] Modify `json` func to identically behave as existing `errorHandling` middleware
- [x] Split error template for dev and prod envs
- [x] Tests


server/onerror.ts
```diff
@@ -0,0 +1,138 @@
...
+      err = newError;
+    }
+
+    const headerSent = this.headerSent || !this.writable;
+    if (headerSent) {
+      err.headerSent = true;
+    }
+
+    // delegate
+    this.app.emit("error", err, this);
```
apoorv-mishra, line 75: Do we need this for non-500 errors? Looks like the current `errorHandling` middleware for APIs only emits this for 500 errors.

tommoor: It's only need for real internal errors yea, not things like validation. It's actually only used in one spot to call the `requestErrorHandler` that sends errors to sentry so we could call that directly instead of emitting the event

server/static/devError.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
```
apoorv-mishra, line 1: This and `prodError.html` have been borrowed as it is from `koa-onerror` to tone down the overall scope of the PR. It'd be good if both match Outline's theme but I'm also in favor of doing that separately.

tommoor: Okay, reasonable. Lets call these `error.dev.html` and `error.prod.html`

server/onerror.ts
```diff
@@ -0,0 +1,193 @@
...
+          newError.status = err.status;
+        }
+        if (err.headers) {
+          newError.headers = err.headers;
+        }
+      }
+      err = newError;
+    }
+
+    // ENOENT support
```
tommoor, line 85:
```suggestion
```

Comment is obvious

server/onerror.ts
```diff
@@ -0,0 +1,193 @@
...
+  };
+
+  return app;
+}
+
+/**
+ * json error handler
+ * @param {Error} err
+ */
+
```
tommoor, line 135:
```suggestion
/**
 * Handle errors for json requests.
 *
 * @param err The error being handled.
 * @param ctx The request context.
 */
```

server/onerror.ts
```diff
@@ -0,0 +1,193 @@
...
+    error = "not_found";
+  }
+
+  if (/Authorization error/i.test(message)) {
+    ctx.status = 403;
+    error = "authorization_error";
+  }
+
+  if (ctx.status === 500) {
+    message = "Internal Server Error";
```
tommoor, line 162:
```suggestion
    message = "Internal server error";
```

server/onerror.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    const headerSent = this.headerSent || !this.writable;
+    if (headerSent) {
+      err.headerSent = true;
+    }
+
+    // nothing we can do here other
+    // than delegate to the app-level
+    // handler and log.
```
tommoor, line 106:
```suggestion
    // Nothing we can do here other than delegate to the app-level handler and log.
```

server/onerror.ts
```diff
@@ -0,0 +1,193 @@
...
+      path.join(__dirname, "static/devError.html")
+    ))
+  );
+};
+
+export default function onerror(app: Koa) {
+  app.context.onerror = function (err: any) {
+    // don't do anything if there is no error.
+    // this allows you to pass `this.onerror`
+    // to node-style callbacks.
```
tommoor, line 42:
```suggestion
    // Don't do anything if there is no error, this allows you to pass `this.onerror` to node-style callbacks.
```

server/onerror.ts
```diff
@@ -0,0 +1,193 @@
...
+    // @ts-expect-error ts-migrate(2571) FIXME: Object is of type 'unknown'.
+    delete ctx.body.data;
+  }
+}
+
+/**
+ * html error handler
+ * @param {Error} err
+ */
+
```
tommoor, line 185:
```suggestion
/**
 * Handle errors for html requests.
 *
 * @param err The error being handled.
 * @param ctx The request context.
 */

```

server/routes/errors.test.ts
```diff
@@ -0,0 +1,38 @@
...
+const server = getTestServer();
+
+describe("/test/error", () => {
+  it("should return error response as json", async () => {
+    const res = await server.post("/test/error", {
+      body: {},
+    });
+    const body = await res.json();
+    expect(res.status).toBe(500);
+    expect(body.message).toBe("Internal Server Error");
```
tommoor, line 12:
```suggestion
    expect(body.message).toBe("Internal server error");
```

server/routes/errors.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  it("should fallback to json err response for types other than html", async () => {
+    const res = await server.post("/test/error", {
+      headers: {
+        accept: "text/plain",
+      },
+      body: {},
+    });
+    const body = await res.json();
+    expect(res.status).toBe(500);
+    expect(body.message).toBe("Internal Server Error");
```
tommoor, line 36:
```suggestion
    expect(body.message).toBe("Internal server error");
```

tommoor (commented):
Thanks, this is exactly what was needed – most of the comments are just stylistic, I know a lot of this was copied across.

This enables moving the default rate limiter up to cover the entire application now 🙏

server/onerror.ts
apoorv-mishra: Existing `errorHandling` middleware [uses caps](https://github.com/outline/outline/blob/b0b7c7d647d50ac9fdcb6ea04ffdfb3341967307/server/middlewares/errorHandling.ts#L35). Although unlikely, but hope it doesn't introduce regressions :crossed_fingers:

tommoor: Yep 👍

PR #4544, opened 2022-12-06
perf: Improve collab persistence performance

tommoor:
A comment to stop the bot

### DIFF
--- a/server/commands/documentCollaborativeUpdater.ts
+++ b/server/commands/documentCollaborativeUpdater.ts
@@ -1,6 +1,7 @@
 import { yDocToProsemirrorJSON } from "@getoutline/y-prosemirror";
 import { uniq } from "lodash";
 import { Node } from "prosemirror-model";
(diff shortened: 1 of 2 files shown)


server/commands/documentCollaborativeUpdater.ts
```diff
@@ -36,48 +41,41 @@ export default async function documentCollaborativeUpdater({
     const node = Node.fromJSON(schema, yDocToProsemirrorJSON(ydoc, "default"));
     const text = serializer.serialize(node, undefined);
     const isUnchanged = document.text === text;
-    const hasMultiplayerState = !!document.state;
+    const lastModifiedById = userId ?? document.lastModifiedById;
 
-    if (isUnchanged && hasMultiplayerState) {
```
tommoor, line 41: This check is no longer needed as this code will never be called unless there is `state` in the database, it's created if missing as part of `onLoadDocument` hook

PR #4550, opened 2022-12-06
Ability to create share url slug

### LINKED ISSUE #4549: Share url slug
Specs:

* Add a field for custom path to Share, possibly a `urlId`.
* In the “More options” section of the share dialog, allow the user to choose and save a custom slug
* They can then access from [domain.com/s/my-custom-slug](http://domain.com/s/my-custom-slug)

apoorv-mishra:
Closes #4549 
- [x] Frontend
  - [x] Add input field for adding slug under "More Options" in share popover
  - [x] Modify `Input` component to support validation errors
  - [x] Validate input slug - only lowercase letters, digits and dashes allowed
  - [x] As user types slug, share url should be previewed in real-time in the same place as error
  - [x] "Copy Link" button should be enabled/disabled as per the input slug validation
  - [x] Clicking on "Copy Link" should copy the slug url variant instead of the `id` variant
  - [x] `debounce` like strategy for API call to eventually update slug on backend

- [x] Backend
  - [x] Add `urlId` field to `shares` table and consequently to `Share` model
  - [x] Validate `urlId` and allow update in route handler
  - [x] Share link with `id` should redirect to the slug variant


server/routes/api/shares.ts
```diff
@@ -162,9 +163,15 @@ router.post("shares.list", auth(), pagination(), async (ctx) => {
 });
 
 router.post("shares.update", auth(), async (ctx) => {
-  const { id, includeChildDocuments, published } = ctx.request.body;
+  const { id, includeChildDocuments, published, urlId } = ctx.request.body;
   assertUuid(id, "id is required");
 
+  if (urlId) {
+    if (!SHARE_URL_SLUG_REGEX.test(urlId)) {
```
tommoor, line 170: Move to model validation?

app/scenes/Document/components/SharePopover.tsx
```diff
@@ -233,6 +247,15 @@ function SharePopover({
               </SwitchText>
             </SwitchLabel>
           </SwitchWrapper>
+          <Separator />
+          <SwitchWrapper>
+            <Input
+              type="text"
+              label={t("Share url slug")}
```
tommoor, line 254:
```suggestion
              label={t("Custom link")}
```

Perhaps?

apoorv-mishra: Sure. It goes well with the actual link preview just below the input to resolve a minor confusion arising out of "Custom link" - the input field expects a slug, not a link as the label suggests :smile:

server/migrations/20221206163421-add-share-url-slug.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    try {
+      await queryInterface.addColumn("shares", "urlId", {
+        type: Sequelize.STRING,
+        allowNull: true,
+        unique: true,
```
apoorv-mishra, line 9: Modified from `unique(teamId, urlId)` to `unique(urlId)` because if we load `https://<team.url>/<urlId>`, the only available information to fetch required entities is `urlId`. And to be able to uniquely identify a `Share`(and eventually a `Document`), we need the unique constraint on `urlId`.

app/scenes/Document/components/SharePopover.tsx
```diff
@@ -233,6 +266,17 @@ function SharePopover({
...
           </SwitchWrapper>
+          <Separator />
+          <SwitchWrapper>
+            <Input
+              type="text"
+              label={t("Custom link")}
+              onChange={handleUrlSlugChange}
+              error={slugValidationError}
+              preview={urlSlug ? `${team.url}/${urlSlug}` : ""}
+              defaultValue={urlSlug || ""}
```
apoorv-mishra, line 277: Why do we need `urlSlug` in state if `share.urlId` is already observable?

Well, consider the scenario where someone types "foobar*", an invalid value in the `input` and then closes the share popover. Now, if the popover is opened again, the expected value in the input should be "foobar*", which is stored in `urlSlug`, not `share.urlId`.

server/migrations/20221206163421-add-share-url-slug.js
tommoor: We do have the team domain/subdomain from the url and a utility exists to convert that to a team – `getTeamFromContext` which should be enough. I _think_ this is needed because otherwise you can just type in "common" terms in the input and find shared content that way across all the teams…

tommoor:
A couple of small issues in UX testing:

- [ ] "Copy link" button should become enabled again when the custom link changes. After clicking once it gets stuck in a disabled state
- [ ] Validation errors are not displayed below the input anymore
- [ ] The yellow text is very hard to read against white, lets stick to `textSecondary` here in a first pass, doesn't hurt to create a link either – perhaps:
    image

- [ ] The url displayed below the input is missing the `/s/` in path
- [ ] Lets move the custom link above the "Automatically redirect" control, it's more important

app/components/Input.tsx
```diff
@@ -120,6 +121,8 @@ export type Props = React.InputHTMLAttributes<
   flex?: boolean;
   short?: boolean;
   margin?: string | number;
+  error?: string;
+  preview?: string;
```
tommoor, line 125: I think it's overkill to add this in here, can just add below the `Input` element

server/migrations/20221206163421-add-share-url-slug.js
apoorv-mishra: Ahh, didn't know that. Thanks :+1:

server/routes/api/shares.ts
```diff
@@ -191,6 +191,10 @@ router.post("shares.update", auth(), async (ctx) => {
     share.includeChildDocuments = includeChildDocuments;
   }
 
+  if (urlId) {
```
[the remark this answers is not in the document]
apoorv-mishra: Oops, this had to be only an `undefined` check. Thanks for the spot.

app/scenes/Document/components/SharePopover.tsx
```diff
@@ -211,6 +255,31 @@ function SharePopover({
...
+              error={slugValidationError}
+              defaultValue={urlSlug}
+            />
+            {!slugValidationError && urlSlug && (
+              <DocumentLinkPreviewWrapper>
+                <DocumentLinkPreview type="secondary" size="small">
+                  <Trans>The document will be available at</Trans>
+                  <br />
+                  <a
+                    href={escape(urlSlug ? `${team.url}/s/${urlSlug}` : "")}
```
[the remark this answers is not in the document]
apoorv-mishra: Added it after CodeQL spat out the warning.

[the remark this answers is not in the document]
tommoor: CodeQL was wrong here, lets dismiss that

server/routes/api/documents/documents.ts
```diff
@@ -394,18 +395,18 @@ router.post(
...
       shareId,
       user,
+      teamId: team?.id,
     });
     const isPublic = cannot(user, "read", document);
     const serializedDocument = await presentDocument(document, {
       isPublic,
     });
 
-    const team = await document.$get("team");
```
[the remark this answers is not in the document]
tommoor: But then the `team` is guaranteed to be the one that is associated with the share. If we just get the team from the context all that means is that it was loaded on a specific subdomain – it doesn't actually relate to the logged in user at all. Does that make sense?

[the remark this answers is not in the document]
apoorv-mishra: Makes sense, yes. The `share` load would fail if attempted against an incorrect `teamId`, obtained from an incorrect subdomain. The `urlId` update requires auth. So, the pair `(urlId, teamId)` used in query has an _implicit_ auth associated to it.

tommoor (approved):

PR #4563, opened 2022-12-10
Add additional filters to `search_titles` endpoint

tommoor:
Additional backend filters for https://github.com/outline/outline/pull/4442


adityash1:
Additional Filters are not working when I'm hitting the endpoint with optional filters parameters. These are the changes I made to the DocumentStore.tsx

[https://gist.github.com/adityash1/064e2ae48dd1c72bed8c9f974b82f773](url)

tommoor:
Hmm, not sure how that's possible. Will add some tests around the API endpoint as well

tommoor:
There was a bug with `dateFilter`, other than that it seems to work well

PR #4582, opened 2022-12-16
Ability to choose publish location for a document

apoorv-mishra:
Addresses #2122 

- [x] Gather collections data in a tree-like data structure
- [x] Flatten the tree and pass the flattened data to the virtualized list
- [x] Index data and make it searchable
- [x] Display data as per the [design](https://github.com/outline/outline/issues/2122#issuecomment-1344991998)
  - [x] Search bar
  - [x] Scroll view
    - [x] Row 
  - [x] Footer
    - [x] Text area to display selected node
    - [x] Publish button
- [x] Trigger API upon publish
- [x] Capability to create a draft(without collection) via "New Doc" button on homepage
- [x] Make rows collapsible
- [x] Make it a centered modal instead of a popover
- [x] Keyboard navigation
  - [x] Collapse and expand shortcut


app/scenes/Document/components/Header.tsx
```diff
@@ -314,19 +315,23 @@ function DocumentHeader({
...
-                <Tooltip
-                  tooltip={t("Publish")}
-                  shortcut={`${metaDisplay}+shift+p`}
-                  delay={500}
-                  placement="bottom"
-                >
-                  <Button
-                    onClick={handlePublish}
-                    disabled={publishingIsDisabled}
+                {document.collectionId ? (
```
apoorv-mishra, line 318: Replace this check with `false` to test out the flow in dev. This is so because we don't yet have the capability(via UI) to create a doc without collection.

.env.sample
```diff
@@ -177,3 +177,6 @@ RATE_LIMITER_ENABLED=true
 # Configure default throttling parameters for rate limiter
 RATE_LIMITER_REQUESTS=1000
 RATE_LIMITER_DURATION_WINDOW=60
+
+# Feature flag to enable new publish UI
+PUBLISH_UI_ENABLED=false
```
apoorv-mishra, line 182: Temporarily added. Acts as a feature flag. Can get rid off when we have some basic feature flag support.

app/scenes/Document/components/Header.tsx
apoorv-mishra: Not needed anymore. Added the capability to create a doc without collection.

app/scenes/Document/components/PublishPopover.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const listRef = React.useRef<List>(null);
+
+  React.useEffect(() => {
+    if (visible && selectedLocation) {
+      listRef.current?.scrollToItem(selectedLocation.index);
+    }
+  });
+
+  const searchIndex = React.useMemo(() => {
+    const data = flattenTree(collections.tree.root).slice(1);
```
apoorv-mishra, line 40: I assume we don't need to filter out templates here? From what I understood _Templatize_ option creates a copy and doesn't alter the existing `documentStructure`.

app/scenes/Document/components/PublishButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <StyledPopover {...popover} aria-label={t("Publish")}>
+        <PublishPopover document={document} {...popover} />
+      </StyledPopover>
+    </>
+  );
+}
+
+const StyledPopover = styled(Popover)`
+  z-index: ${depths.popover};
+  > :first-child {
```
apoorv-mishra, line 43: Using `> :first-child` seems like a hack(a bad one), `${Contents}` would be better?

app/scenes/Document/components/PublishPopover.tsx
```diff
@@ -0,0 +1,232 @@
...
+      } else {
+        results = searchIndex.haystack;
+      }
+    }
+
+    if (onlyShowCollections) {
+      results = results.filter(
+        (result: any) => result.data.type === "collection"
+      );
+    }
```
apoorv-mishra, line 70: This is unnecessary given that a template is published by default?

tommoor:
Great progress, some feedback on current state:

- [ ] If the modal will show then make the button "Publish…" instead of "Publish" in that scenario to hint to the user that there are more steps
- [ ] You should be able to arrow down from the search box to the first item
- [ ] You should be able to arrow up from the first item to the search box
- [ ] Arrow right can be mapped to expand
- [ ] We could try keeping the collapse controls when searching filtered unless that adds a ton of complexity
- [ ] Make the "Publish under" text secondary color
- [ ] I think this modal needs to be a bit wider, maybe 500px is a good place
- [ ] We need to combine the hover + active state into one so you can't end up in this situation. The way to do that is listen for pointermove and set the active index rather than using css hover styles.
       image

Hope this all makes sense! Almost there ✌️

apoorv-mishra:
> We could try keeping the collapse controls when searching filtered unless that adds a ton of complexity

Complexity aside, it does feel a bit off to show tree-like view in search results. What about retaining the existing document move like flat view on search trigger? (ctrl-p like view of VS Code)

tommoor:
Okay, lets try it 👍

tommoor:
Discussed some in chat but leaving here too for posterity:

- [ ] Can we keep the focus within the filter input when navigating with keyboard so you can continue to filter (See cmd+k for desired behavior)
- [ ] Can we use the emoji instead of document icon if there is one (See Starred items in sidebar for example)
- [ ] Text should be baselign aligned here, not centered
    image
- [ ] I'd probably make this text tertiary to increase the contrast, it's not easy to scan right now as it's too similar
- [ ] Don't show the breadcrumb when it's just the collection name, eg:
   image
- [ ] Breadcrumb text coloring incorrect in selected state
   image
- [ ] Can cmd+enter submit? It kind of does right now, but it ends up changing the selection and publishing in the wrong place. I haven't looked into why that might be happening in the code…

app/components/Modal.tsx
```diff
@@ -254,7 +254,7 @@ const Small = styled.div`
   margin: auto auto;
   width: 30vw;
   min-width: 350px;
-  max-width: 450px;
+  max-width: 500px;
```
apoorv-mishra, line 257: This might slightly affect all modals. Fine?

tommoor (approved):
As discussed we'll follow up with removing `tree` from collections store – I'd probably just turn it into a memoized hook inside the DocumentPublish component

PR #4606, opened 2022-12-23
refactor: add server side validation schema for attachments

elidrissidev:
I refactored the attachments API endpoints to use server-side validation schema following the same structure as in #4397.

- See outline/outline#4284


server/routes/api/attachments/attachments.ts
```diff
@@ -5,46 +5,32 @@ import { bytesToHumanReadable } from "@shared/utils/files";
...
-      size,
-      // 'public' is now deprecated and can be removed on December 1 2022.
-      public: isPublicDeprecated,
-      preset = isPublicDeprecated
-        ? AttachmentPreset.Avatar
-        : AttachmentPreset.DocumentAttachment,
-    } = ctx.request.body;
+  async (ctx: APIContext<T.AttachmentCreateReq>) => {
+    const { name, documentId, contentType, size, preset } = ctx.input;
     const { user, transaction } = ctx.state;
```
elidrissidev, line 28: Here we lose the type for `ctx.state.transaction` because I had to switch the ctx type to `APIContext`. Do you suggest introducing another type for that?

server/routes/api/attachments/attachments.ts
```diff
@@ -121,34 +107,38 @@ router.post(
...
+    });
+
+    ctx.body = {
+      success: true,
+    };
+  }
+);
 
 const handleAttachmentsRedirect = async (ctx: ContextWithState) => {
   const id = ctx.request.body?.id ?? ctx.request.query?.id;
```
elidrissidev, line 144: The validation middleware only handles the request body, so this is left as is.

server/routes/api/attachments/attachments.ts
```diff
@@ -64,7 +50,7 @@ router.post(
     }
 
     const modelId = uuidv4();
-    const acl = AttachmentHelper.presetToAcl(preset);
+    const acl = AttachmentHelper.presetToAcl(preset!);
```
elidrissidev, line 53: Had to add the non-null assertion for `preset` because AFAIK it's optional if `public` is defined in the request. Not sure if there's a better to handle this?

server/routes/api/attachments/attachments.ts
tommoor: The two middlewares are both newish and haven't been used together yet, need to figure out the pattern to enable typing though as these will be used together more and more.

server/routes/api/attachments/attachments.ts
tommoor: As commented `public` is deprecated and can be removed now, which simplifies things.

elidrissidev: Should I include that in this PR if it's a small change?

tommoor: Yea, might as well remove here as it'll make these types more accurate 👍

elidrissidev: Done. Hopefully I didn't miss something.

tommoor (approved):

PR #4610, opened 2022-12-25
fix: missing protocol occurs in faulty redirect

### LINKED ISSUE #4609: SSO URL in wrong format causes redirection problems
As soon as OIDC omits the HTTPS in the URL, the redirection to the SSO provider does not work properly anymore see video

The configuration then looks accordingly as follows
```

    - "OIDC_CLIENT_ID=wiki.example.de"
     - "OIDC_CLIENT_SECRET=example"
     - "OIDC_AUTH_URI=id.example.de/auth/realms/master/protocol/openid-connect/auth"
     - "OIDC_TOKEN_URI=id.example.de/auth/realms/master/protocol/openid-connect/token"
     - "OIDC_USERINFO_URI=id.example.de/auth/realms/master/protocol/openid-connect/userinfo"
     - "OIDC_USERNAME_CLAIM=email"
     - "OIDC_DISPLAY_NAME=Keycloak"
     - "OIDC_SCOPES=email"
     - "FORCE_HTTPS=false"
```

https://user-images.githubusercontent.com/54689374/209473664-0477f26c-ac54-48d2-8498-aa26503423a8.mp4

florian-sabonchi:
closes #4609

### DIFF
--- a/server/routes/auth/providers/oidc.ts
+++ b/server/routes/auth/providers/oidc.ts
@@ -22,6 +22,7 @@ import {
 
 const router = new Router();
 const providerName = "oidc";
+const httpRegex = /^https?:\/\//;
 const OIDC_AUTH_URI = env.OIDC_AUTH_URI || "";
 const OIDC_TOKEN_URI = env.OIDC_TOKEN_URI || "";
 const OIDC_USERINFO_URI = env.OIDC_USERINFO_URI || "";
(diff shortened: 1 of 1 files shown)


tommoor:
This isn't a change that is advisable, the expectation of all URL configuration is that the fully qualified domain is provided including the protocol and OIDC is not an exception. I think it would be more confusing for the software to automatically prepend a protocol in this one instance.

florian-sabonchi:
How should we handle it then if only the FQDN is given without protocol? In my opinion, the software should still work as usual even if the protocol is not specified.

tommoor:
This configuration is entirely in the control of whomever is running the software, just provide the correct value including the protocol in the environment variable? I'm not sure what I'm missing.

PR #4622, opened 2022-12-31
refactor: add server-side validation schema for events

elidrissidev:
Refactoring the events API endpoints to use server-side validation schema.

- See outline/outline#4284


server/routes/api/events/events.ts
```diff
@@ -0,0 +1,102 @@
...
+      where = {
+        ...where,
+        [Op.or]: [
+          {
+            collectionId: collectionIds,
+          },
+          {
+            collectionId: {
+              [Op.is]: null,
+            },
```
elidrissidev, line 73: typescript started complaining here about `collectionId` not being nullable, hence why I made the change in Event model.

server/models/Event.ts
```diff
@@ -88,7 +88,7 @@ class Event extends IdModel {
 
   @ForeignKey(() => Collection)
   @Column(DataType.UUID)
-  collectionId: string;
+  collectionId: string | null;
 
   @BelongsTo(() => Team, "teamId")
   team: Team;
```
elidrissidev, line 94: I wonder why these fields weren't nullable already, since I see from the database schema they are?

tommoor: 👍  Just a mistake

tommoor (approved):

server/models/Event.ts
elidrissidev: alright, I will include that later.

tommoor:
The database schema is probably a bit loose here, it's a very old model – there should always be an actor and a team at a minimum

elidrissidev:
> The database schema is probably a bit loose here, it's a very old model – there should always be an actor and a team at a minimum

👍 just noticed the conditions inside the route handler. Should be good now.

tommoor (approved):

PR #4623, opened 2022-12-31
chore: Move tracing decorators into the codebase

tommoor:
More control over tracing and fixing some naming issues by bringing `datadog-apm` decorators into the codebase instead of relying on forked third party module


tommoor:
@SocketSecurity ignore @datadog/native-appsec@2.0.0

PR #4657, opened 2023-01-07
feat: Server side translation setup

### LINKED ISSUE #4647: Add translation hooks to Slack messages
Similar to #4645 – use the same logic but for the slack hooks.

tommoor:
- Adds all the setup for translating strings on the server
- Adds translations for Slack hooks as examples

closes #4647


i18next-parser.config.js
```diff
@@ -8,7 +8,9 @@ module.exports = {
   defaultNamespace: "translation",
   // Default namespace used in your i18next config
 
-  defaultValue: "",
+  defaultValue(locale, namespace, key) {
```
tommoor, line 11: Bug since last bump of this dependency was leaving new strings with empty values

PR #4683, opened 2023-01-12
Allow listening on a custom address

steveh:
With the listen address hardcoded to `localhost`, it's not possible to run Outline behind an external load balancer.

A better way might be to have a single env var like `LISTEN=localhost:3000` and a corresponding flag like `-l`, but I didn't want to break backwards compatibility.

### DIFF
--- a/server/env.ts
+++ b/server/env.ts
@@ -155,6 +155,12 @@ export class Environment {
   public PORT = this.toOptionalNumber(process.env.PORT);
 
   /**
+   * The address that the server will listen on, defaults to localhost.
+   */
+  @IsOptional()
+  public LISTEN_ADDRESS = process.env.LISTEN_ADDRESS || "";
+
+  /**
    * Optional extra debugging. Comma separated
    */
   public DEBUG = process.env.DEBUG || "";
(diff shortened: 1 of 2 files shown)


tommoor:
> it's not possible to run Outline behind an external load balancer.

I think we can land this, but I'm not sure what you mean here – there are thousands of teams running this behind a load balancer…

steveh:
With `yarn run start`, it will listen on `localhost:3000`, meaning that the port is not accessible to the outside world. You can confirm that by running `lsof -i4 -P | grep LISTEN` and seeing it doesn't appear.

With this patch, if you run `LISTEN_ADDRESS=0.0.0.0 yarn run start`, it will appear in `lsof`.

```
node      97883 steve   32u  IPv4 0x92b44fa0ee3cbcb      0t0  TCP *:3000 (LISTEN)
```

Unless I've fundamentally misunderstood something?

tommoor:
You haven't misunderstood… although `localhost` is printed in the console the server actually binds to `0.0.0.0` by default, so this change isn't neccessary (I had to double check) 😄 

[Documented here](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback)

rhclayto:
I wonder why this was closed. Being able to specify an IP address to bind to is pretty common. Why not here?

PR #4691, opened 2023-01-13
[GH-4672]: Add support for `deep` optional parameter to use with `sort` to get all documents.

### LINKED ISSUE #4672: Cannot get all documents when using sort: "index" via the API.
I have a document list like the following one

```
├── first document
├── second document
│    ├── nested document
│    │    ├── nested document
│    │    ├── nested document
│    │    ├── nested document
│    │    └── nested document
│    └── nested document
│         ├── nested document
│         ├── nested document
│         └── nested document
└── third document
     ├── nested document
     │    ├── nested document
     │    ├── nested document 
     │    ├── nested document
     │    └── nested document
     └── nested document
          ├── nested document
          ├── nested document
          └── nested document
```

I'm trying to get a collection list using the API calling the `documents.list` using the following params

```json
{
  "offset": 0,
  "collectionId": "....",
  "limit":50,
  "direction": "DESC"
}
```
With that options everything works fine, I can get all the documents and the nested ones. 

The sorting it's not the ideal, I want the manual sorting I already have, so I am trying to use the sort by index option like this

```json
{
  "offset": 0,
  "sort": "index",
  "collectionId": "....",
  "limit":50,
  "direction": "DESC"
}
```
But when I do this I am getting only the 3 top level documents and the nested ones are missing.

Am I missing something ? 🤔

Thank you in advance :)

adityash1:
Fixes  #4672

### DIFF
--- a/server/routes/api/documents/documents.test.ts
+++ b/server/routes/api/documents/documents.test.ts
@@ -746,6 +746,31 @@ describe("#documents.list", () => {
     expect(body.data[1].id).toEqual(document.id);
   });
 
+  it("should return all sorted documents", async () => {
+    const { user, document, collection } = await seed();
+    const childDocument = await buildDocument({
+      title: "child document",
+      text: "random text",
+      userId: user.id,
+      teamId: user.teamId,
+      collectionId: collection.id,
+      parentDocumentId: document.id,
+    });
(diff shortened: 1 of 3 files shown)


tommoor:
closed for now, feel free to reopen if you continue with this work

PR #4705, opened 2023-01-14
show collection name in pin menu

### LINKED ISSUE #4704: "Pin to collection" should say which collection
This menu is not always displayed within a clear context of a collection – it would be better if it said "Pin to `<Collection name>`".

image

adityash1:
Fixes #4704

### DIFF
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -356,12 +356,12 @@ export const pinDocumentToCollection = createAction({
     const selectedDocument = stores.documents.get(activeDocumentId);
     const selectedDocumentCollectionName = selectedDocument
       ? stores.documents.getCollectionForDocument(selectedDocument)?.name
-      : "";
+      : undefined;
     return selectedDocumentCollectionName
       ? t("Pin to {{collectionName}}", {
           collectionName: selectedDocumentCollectionName,
         })
-      : "";
+      : t("Pin to collection");
   },
 
   section: DocumentSection,
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -352,7 +352,10 @@ export const duplicateDocument = createAction({
  * of the collection for all collection members to see.
  */
 export const pinDocumentToCollection = createAction({
-  name: ({ t }) => t("Pin to collection"),
+  name: ({ t, stores }) => {
+    const activeCollectionName = stores.collections.active?.name;
+    return t("Pin to ") + activeCollectionName;
+  },
   section: DocumentSection,
   icon: <PinIcon />,
   iconInContextMenu: false,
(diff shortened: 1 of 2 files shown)


app/actions/definitions/documents.tsx
```diff
@@ -352,7 +352,10 @@ export const duplicateDocument = createAction({
  * of the collection for all collection members to see.
  */
 export const pinDocumentToCollection = createAction({
-  name: ({ t }) => t("Pin to collection"),
+  name: ({ t, stores }) => {
+    const activeCollectionName = stores.collections.active?.name;
+    return t("Pin to ") + activeCollectionName;
```
tommoor, line 357: This should use a template within the translation rather than string concatenation – not all languages are int the same order.

app/actions/definitions/documents.tsx
```diff
@@ -352,7 +352,10 @@ export const duplicateDocument = createAction({
  * of the collection for all collection members to see.
  */
 export const pinDocumentToCollection = createAction({
-  name: ({ t }) => t("Pin to collection"),
+  name: ({ t, stores }) => {
+    const activeCollectionName = stores.collections.active?.name;
```
tommoor, line 356: The active collection is not the collection it'll be pinned to, it's the collection associated with the `activeDocument` – they may be separate things

tommoor (changes requested):

app/actions/definitions/documents.tsx
adityash1: Thanks for this info. I've pushed the changes...

app/actions/definitions/documents.tsx
```diff
@@ -352,7 +352,18 @@ export const duplicateDocument = createAction({
...
+  name: ({ activeDocumentId = "", t, stores }) => {
+    const selectedDocument = stores.documents.get(activeDocumentId);
+    const selectedDocumentCollectionName = selectedDocument
+      ? stores.documents.getCollectionForDocument(selectedDocument)?.name
+      : "";
+    return selectedDocumentCollectionName
+      ? t("Pin to {{collectionName}}", {
+          collectionName: selectedDocumentCollectionName,
+        })
+      : "";
```
tommoor, line 364:
```suggestion
    const selectedDocument = stores.documents.get(activeDocumentId);
    const selectedDocumentCollectionName = selectedDocument
      ? stores.documents.getCollectionForDocument(selectedDocument)?.name
      : undefined;
    return selectedDocumentCollectionName
      ? t("Pin to {{collectionName}}", {
          collectionName: selectedDocumentCollectionName,
        })
      : t("Pin to collection");
```

We can fall back to the old language otherwise it's possible to have a menu item with no text

tommoor (approved):

PR #4706, opened 2023-01-14
fix: Sidebar jumps when publish or unpublish document

### LINKED ISSUE #4702: Publishing or unpublishing a doc is janky in sidebar
When a document is published or unpublished the sidebar shows a flash where the document has disappeared, this makes the whole experience feel a lot less "solid" – we should find a way around this.

tommoor:
Updated the API to return the collection that's already in memory so that everything can be updated at once.

closes #4702

### DIFF
--- a/app/components/Sidebar/components/SidebarLink.tsx
+++ b/app/components/Sidebar/components/SidebarLink.tsx
@@ -163,13 +163,11 @@ const Actions = styled(EventBoundary)<{ showActions?: boolean }>`
   }
 `;
 
-type NavLinkProps = {
+const Link = styled(NavLink)<{
   $isActiveDrop?: boolean;
   $isDraft?: boolean;
   $disabled?: boolean;
-};
-
-const Link = styled(NavLink)<NavLinkProps>`
+}>`
   display: flex;
   position: relative;
   text-overflow: ellipsis;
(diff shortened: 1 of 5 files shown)


app/components/Sidebar/components/SidebarLink.tsx
```diff
@@ -194,8 +196,17 @@ const Link = styled(NavLink)<{
   ${(props) =>
     props.$isDraft &&
     css`
-      padding: 4px 14px;
```
tommoor, line 197: Now that the update is seamless it made it obvious that drafts are slightly differently sized on desktop

app/stores/DocumentsStore.ts
```diff
@@ -763,16 +772,16 @@ export default class DocumentsStore extends BaseStore<Document> {
...
       invariant(res?.data, "Data should be available");
-      document.updateFromJson(res.data);
+      document.updateFromJson(res.data.document);
+      const collection = this.getCollectionForDocument(document);
+      collection?.updateFromJson(res.data.collection);
       this.addPolicies(res.policies);
     });
-    const collection = this.getCollectionForDocument(document);
-    if (collection) {
-      collection.refresh();
```
tommoor, line 774: Avoids an extra API request here, so that's nice.

PR #4763, opened 2023-01-24
Move tree implementation out of collections store

### LINKED ISSUE #4759: Move the tree method implementation out of collections store
Refactor to move the tree implementation out as expressed [here](https://github.com/outline/outline/pull/4582#pullrequestreview-1265588907)

apoorv-mishra:
Fixes #4759 

- [x] Refactor `collections.tree` into a separate hook.
- [x] Identify and factor out the parts that overlap with `DocumentMove`


app/hooks/useCollectionTrees.ts
```diff
@@ -0,0 +1,76 @@
+import * as React from "react";
+import { NavigationNode, NavigationNodeType } from "@shared/types";
+import Collection from "~/models/Collection";
+import useStores from "~/hooks/useStores";
+
+export default function useCollectionsTree() {
```
tommoor, line 6: It would be good to explicitly define the return type here and add a jsdoc comment

app/hooks/useCollectionTrees.ts
```diff
@@ -0,0 +1,76 @@
...
+      children: collection.documents || [],
+      parent: null,
+    };
+
+    collectionNode = addType(collectionNode);
+    collectionNode = addDepth(collectionNode);
+    collectionNode = addCollectionId(collectionNode);
+    collectionNode = addParent(collectionNode);
+
+    return collectionNode;
```
tommoor, line 70:
```suggestion
    return addParent(addCollectionId(addDepth(addType(collectionNode))));
```

Could also use this style if you're going all functional like this

app/components/DocumentExplorer.tsx
```diff
@@ -0,0 +1,372 @@
...
+import useStores from "~/hooks/useStores";
+import { isModKey } from "~/utils/keyboard";
+import { flattenTree, ancestors, descendants } from "~/utils/tree";
+
+type Props = {
+  actionOnItem: () => void;
+  onSelectItem: (item: NavigationNode | null) => void;
+};
+
+function DocumentExplorer({ actionOnItem, onSelectItem }: Props) {
```
tommoor, line 32: Good name 👍

app/components/DocumentExplorer.tsx
```diff
@@ -0,0 +1,372 @@
...
+import InputSearch from "~/components/InputSearch";
+import Text from "~/components/Text";
+import useCollectionTrees from "~/hooks/useCollectionTrees";
+import useMobile from "~/hooks/useMobile";
+import useStores from "~/hooks/useStores";
+import { isModKey } from "~/utils/keyboard";
+import { flattenTree, ancestors, descendants } from "~/utils/tree";
+
+type Props = {
+  actionOnItem: () => void;
```
tommoor, line 28: Need comments on these as it's not immediately clear what the difference between `actionOnItem` and `onSelectItem` is.

`actionOnItem` -> `onSubmit` perhaps?

app/components/DocumentExplorerNode.tsx
```diff
@@ -7,49 +7,40 @@ import breakpoint from "styled-components-breakpoint";
...
 type Props = {
-  location: any;
   selected: boolean;
   active: boolean;
   style: React.CSSProperties;
-  isSearchResult: boolean;
   expanded: boolean;
   icon?: React.ReactNode;
+  title: string;
+  nestLevel: number;
```
tommoor, line 18: Keep this as `depth`?

app/scenes/DocumentPublish.tsx
```diff
@@ -200,225 +58,48 @@ function DocumentPublish({ document }: Props) {
...
-          <Text type="secondary">{t("No results found")}.</Text>
-        </NoResults>
-      )}
+    <FlexContainer column>
+      <DocumentExplorer actionOnItem={publish} onSelectItem={selectPath} />
       <Footer justify="space-between" align="center" gap={8}>
-        {selectedLocation ? (
-          <SelectedLocation type="secondary">
+        <StyledText type="secondary">
+          {selectedPath ? (
```
tommoor, line 66: "path" is better than "location" here, glad to see the change

tommoor (commented):
Looks good, I can't find any functional regressions

app/components/DocumentExplorerNode.tsx
apoorv-mishra: Yes! Now that that component has `Node` in its name, `depth` would be acceptable.

app/components/DocumentExplorer.tsx
apoorv-mishra: `onSubmit` is good, yes. Will add comments...

app/hooks/useCollectionTrees.ts
apoorv-mishra: Why not. Good spot :pray:

tommoor (approved):

PR #4765, opened 2023-01-24
Replace Webpack with Vite

hanspagel:
This PR replaces Webpack with [Vite](https://vitejs.dev/).

## Benefits
* Vite embraces native ECMAScript modules (and that’s the future we all strive for)
* Vite is really fast (HMR ~1,500ms vs ~50ms, build time ~45s vs. ~30s)
* Vite configuration is easier to work with (opinion)

## Workarounds
1. When using the prosemirror-math module an import from katex is broken. I worked around this with [patch-package](https://www.npmjs.com/package/patch-package) which applies Git patches to npm dependencies through a npm postinstall hook. 
I’d create a PR, but the package seems to be unmaintained and I wasn’t able to solve it through configuration. If we somehow could force Vite to use the CommonJS version of prosemirror-math it would work. 

2. The dependency `workbox-core` (used for the Service Worker) throws TypeScript exceptions and there’s no way to ignore them through the tsconfig. The only workaround I found is to enable [skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck), which then skips type checking for all .d.ts files … Which is not really what we want, but solves the issue. 

## Tasks
- [x] Test the Service Worker in production.
- [x] Execute service worker code for everyone (remove the localStorage check).

## Testing the Service Worker

Execute the following code in the browser console and refresh the page:

```js
localStorage.setItem("installServiceWorker", true)
```


app/index.tsx
```diff
@@ -1,4 +1,7 @@
+// eslint-disable-next-line import/no-unresolved
+import "vite/modulepreload-polyfill";
```
hanspagel, line 2: It’s recommended to add this when integrating with a backend: https://vitejs.dev/guide/backend-integration.html#backend-integration

app/index.tsx
```diff
@@ -1,4 +1,7 @@
...
 import "focus-visible";
+import "setimmediate";
```
hanspagel, line 4: Not sure how this worked before, but it was missing.

app/utils/polyfills.ts
```diff
@@ -27,8 +27,8 @@ export async function loadPolyfills() {
  */
 function supportsResizeObserver() {
   return (
-    "ResizeObserver" in global &&
-    "ResizeObserverEntry" in global &&
+    "ResizeObserver" in window &&
+    "ResizeObserverEntry" in window &&
```
hanspagel, line 31: Is there a benefit to use global in the frontend? I was able to shim global for npm dependencies, but not for our code.

server/routes/app.ts
```diff
@@ -37,20 +38,10 @@ const readIndexFile = async (ctx: Context): Promise<Buffer> => {
...
-        }
-
-        resolve(result);
-      }
-    );
-  });
+  // TODO: Should be part of the build
+  return await readFile(
+    path.join(__dirname, "../../../server/static/index.html")
+  );
```
hanspagel, line 44: Actually, I forgot why I added this TODO and I even forgot what this code is doing. I think we can keep it like that and just remove the TODO. Or am I missing something?

server/services/web.ts
```diff
@@ -28,6 +27,12 @@ const scriptSrc = [
   "cdn.zapier.com",
 ];
 
+// Allow to load assets from Vite
+if (!isProduction) {
+  scriptSrc.push("127.0.0.1:3001");
+  scriptSrc.push("localhost:3001");
+}
+
```
hanspagel, line 35: Vite serves assets from another URL. We need to allow it for non-production environments.

server/utils/manifest.ts
```diff
@@ -0,0 +1,24 @@
...
+
+try {
+  manifest = fs.readFileSync(file, "utf8") as string;
+} catch (err) {
+  console.warn(
+    `Can not find ${file}. Try executing "yarn vite:build" before running in production mode.`
+  );
+}
+
+export default JSON.parse(manifest) as ManifestStructure;
```
hanspagel, line 24: I didn’t saw there was code to read out the manifest.json already, but I like my implementation more - so I deleted the other one. Let me know if I should remove that and get the other one in again. 🙃

server/utils/startup.ts
```diff
@@ -50,10 +50,11 @@ export async function checkMigrations() {
     return;
   }
 
+  const isProduction = env.ENVIRONMENT === "production";
   const teams = await Team.count();
   const providers = await AuthenticationProvider.count();
 
-  if (teams && !providers) {
+  if (isProduction && teams && !providers) {
```
hanspagel, line 57: I was seeding the dev environment and that exception came up. @tommoor suggested to add this workaround for now.

vite.config.ts
```diff
@@ -0,0 +1,103 @@
...
+              },
+              useBuiltIns: "usage",
+              exclude: [
+                "transform-modules-commonjs",
+                "proposal-dynamic-import",
+              ],
+            },
+          ],
+        ],
+      },
```
hanspagel, line 37: To exclude the two plugins from `@babel/preset-env` I’ve had to copy the whole `presets` part of the babel config, but it’s using the existing babel config for the rest.

It’s not possible to just add this to the babel config, because it’s used for the backend and the frontend.

app/utils/polyfills.ts
tommoor: I'd imagine this was a copy/paste to be honest

server/routes/app.ts
```diff
@@ -78,14 +69,31 @@ export const renderApp = async (
...
+        import RefreshRuntime from 'http://localhost:3001/@react-refresh'
+        RefreshRuntime.injectIntoGlobalHook(window)
+        window.$RefreshReg$ = () => { }
+        window.$RefreshSig$ = () => (type) => type
+        window.__vite_plugin_react_preamble_installed__ = true
+      </script>
+      <script type="module" src="http://localhost:3001/@vite/client"></script>
+      <script type="module" src="http://localhost:3001/${entry}"></script>
+    `;
+
```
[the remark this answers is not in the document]
tommoor: You know it started off with just a couple of replacements and it didn't really seem worth the weight of a templating engine, especially as a package like that comes with even more dependencies etc. I don't know now though, there's enough replacements here that it's definitely crossed some sort of line.

tommoor:
@SocketSecurity ignore postinstall-postinstall@2.1.0
@SocketSecurity ignore esbuild@0.16.13

tommoor:
btw, tried running locally and getting this error:

image

tommoor:
@SocketSecurity ignore esbuild@0.16.17

vite.config.ts
```diff
@@ -70,6 +70,10 @@ export default () => {
       manifest: true,
       target: browserslistToEsbuild(),
       reportCompressedSize: false,
+      terserOptions: {
```
tommoor, line 73: @hanspagel the thing I'm contending with is in staging the `Collection` class is being changed to `R`, so the logic above where you removed the numbers isn't working, we need to prevent the classname mangling altogether.

Thought this would do it, but it seems not to make a difference

package.json
```diff
@@ -356,7 +356,7 @@
     "js-yaml": "^3.14.1",
     "jpeg-js": "0.4.4",
     "qs": "6.9.7",
-    "rollup": "^3.11.0"
+    "rollup": "^3.14.0"
```
tommoor, line 359: Thought this would fix the class name mangling… https://github.com/rollup/rollup/pull/4827

app/sw/registerSW.ts
```diff
@@ -0,0 +1,34 @@
+// To active execute the following code in the browser console and refresh:
+// localStorage.setItem("installServiceWorker", true)
+
+if ("serviceWorker" in navigator) {
```
tommoor, line 4: Any reason we wouldn't just keep this in app/index as before? Avoid the additional entry point and plumbing.

hanspagel: My understanding was:

1. The HTML is served through getoutline.com.
2. The service worker must be registered through JS (inline or registerSW.js) from the same domain.
3. And the actual service worker must be served through the same domain.

But I seem to be wrong. Currently, you’re serving assets from the CDN and it’s working. 🤦‍♂️ I’ll give it another try.

tommoor:
Btw, for your points (2) and (3) in the PR description are both resolved.

app/stores/BaseStore.ts
```diff
@@ -61,7 +57,7 @@ export default abstract class BaseStore<T extends BaseModel> {
   constructor(rootStore: RootStore, model: Class<T>) {
     this.rootStore = rootStore;
     this.model = model;
-    this.modelName = modelNameFromClassName(model.name);
+    this.modelName = lowerFirst(model.name).replace(/\d$/, "");
```
hanspagel, line 60: We’re still replacing digits here, is that still needed?

app/typings/window.d.ts
```diff
@@ -1,11 +1,7 @@
 declare global {
-  interface NodeRequire {
+  interface ImportMeta {
     /** A special feature supported by webpack's compiler that allows you to get all matching modules starting from some base directory.  */
```
hanspagel, line 3: It says "webpack", this is probably outdated. I’d update the description, but I don’t know what it’s doing. 🤡

app/stores/BaseStore.ts
tommoor: It might not be, was reproducible locally

tommoor:
A version of this is now in production, the only problem is that the paths in the `sw.js` do not contain the CDN host which is putting a lot of extra requests on the app servers. Looking into that now…

Issue #4769, opened 2023-01-25
Visual Basic syntax highlighting doesn't work


vgwidt:
**To Reproduce**
1. Create a code block
2. Select VB
3. Observe no line numbers or highlighting

**Expected behavior**
Line numbers should appear with syntax highlighting

**Outline (please complete the following information):**
- Install: self hosted
- Version: 67.2

Aliases "vb" or "vba" need to be used for Prism, rather than "visualbasic".

PR #4770, opened 2023-01-25
fix: Visual Basic syntax higlighting (#4769)

vgwidt:
Aliases "vb" or "vba" need to be used for Prism, rather than "visualbasic".

### DIFF
--- a/shared/editor/plugins/Prism.ts
+++ b/shared/editor/plugins/Prism.ts
@@ -40,7 +40,7 @@ export const LANGUAGES = {
   swift: "Swift",
   toml: "TOML",
   typescript: "TypeScript",
-  visualbasic: "Visual Basic",
+  vb: "Visual Basic",
   yaml: "YAML",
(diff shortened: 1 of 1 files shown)


wtfiscrq:
@vgwidt 👋🏽 - with #4771 this issue will be fixed. I've changed `visualbasic` to `visual-basic` (which actual module's name, required to load it). Would you be ok with that fix if it gets approved?
Thanks!

vgwidt:
> @vgwidt 👋🏽 - with #4771 this issue will be fixed. I've changed `visualbasic` to `visual-basic` (which actual module's name, required to load it). Would you be ok with that fix if it gets approved? Thanks!

I'm perfectly okay with that 👍

tommoor:
sgtm, lets focus efforts on that PR.

tommoor:
The other PR appears to have stalled, so merging this fix in the meantime.

PR #4771, opened 2023-01-25
feat: dynamically load code blocks language

### LINKED ISSUE #4769: Visual Basic syntax highlighting doesn't work

wtfiscrq:
# Description

Add support for languages in code blocks to be loaded dynamically. This opens the door to more easily support additional languages without affecting page loading that don't use those languages.

As a follow-up to this PR it'd be interesting to generate the list of languages dynamically (on build, maybe?) to keep it in sync with supported languages from Refractor/Prism.js. For now, I've added a language to the list: [Rego](https://www.openpolicyagent.org/docs/latest/policy-language/).

# Implementation

- Removed the auto-register logic in CodeFence
- Used [Webpack's Dynamic Imports feature](https://webpack.js.org/guides/code-splitting/#dynamic-imports) in the Prism.js plugin. On each block, if a language isn't already registered it is imported and registered. This effectively loads only the languages for the code blocks already present or added in a document.
- Toggling a code block selected language causes a new chunk to be fetched with that language's definition, if not previously fetched and registered (this behavior can be observed by looking at the network tab).

# Issues

This implementation, as of now, has a couple of issues:

- Adding a code block, adding some content and then toggling the selected language to one that hasn't been loaded/registered already doesn't re-render the code block (i.e. syntax highlighting isn't rendered). If we type something in the same code block afterwards the syntax  is correctly highlighted.
- Adding a code block, selecting the language we want first, and typing some code causes the cursor to move to the start of the line.

_See [this screencast](https://www.loom.com/share/444e7f546ff1441fb6a8842157f399ef) for a demo of the issues._

# Related Issues/Discussions

- #4764 
- fixes #4769
- closes #4770

### DIFF
--- a/shared/editor/plugins/Prism.ts
+++ b/shared/editor/plugins/Prism.ts
@@ -45,6 +45,9 @@ export const LANGUAGES = {
   zig: "Zig",
 };
 
+// Languages that are not supported by refractor or are specially handled
+const EXCLUDED_LANGUAGES = ["mermaidjs"];
+
 type ParsedNode = {
   text: string;
   classes: string[];
@@ -89,7 +92,8 @@ function getDecorations({
   blocks.forEach(async (block) => {
     let startPos = block.pos + 1;
     const language = block.node.attrs.language;
-    if (!language || language === "none") {
+    const isExcluded = EXCLUDED_LANGUAGES.includes(language);
+    if (!language || language === "none" || isExcluded) {
       return;
     }
 
--- a/shared/editor/plugins/Prism.ts
+++ b/shared/editor/plugins/Prism.ts
@@ -32,6 +32,7 @@ export const LANGUAGES = {
   php: "PHP",
   powershell: "Powershell",
   python: "Python",
+  rego: "Rego",
   ruby: "Ruby",
   rust: "Rust",
   scala: "Scala",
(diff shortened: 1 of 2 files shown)


shared/editor/plugins/Prism.ts
```diff
@@ -86,13 +90,25 @@ function getDecorations({
...
-    if (!language || language === "none" || !refractor.registered(language)) {
+    const isExcluded = EXCLUDED_LANGUAGES.includes(language);
+    if (!language || language === "none" || isExcluded) {
       return;
     }
 
+    if (!refractor.registered(language)) {
+      try {
+        // Dynamically load and register the language module
+        const mod = await import(`refractor/lang/${language}`);
```
tommoor, line 104: It might be better to leave this method as non `async` and actually use a standard then/catch here.

Two things left to consider if you're willing:

- This loop gets triggered a lot as the user is typing, we need to record that we're already loading the module and avoid doing so again, probably in a local map/object.
- When the module loads we can then dispatch a transaction with some metadata that tells the editor to re-render, this should solve the issue in the screencast. You can see how we listen to some metadata further down on L184 to trigger a render.

tommoor:
Oh, if you want to sign-in to CircleCI with your GitHub account that should allow the build pipeline to run

shared/editor/plugins/Prism.ts
wtfiscrq: Still getting used to the codebase and architecture, these suggestions surely help me understanding it better. I'll give it a shot and try to implement these changes. Thanks! 🙏🏽

PR #4781, opened 2023-01-27
feat: add support for Nix code highlight (#4781

zimbatm:
Nix is a build system and package manager and used in one of the top-10
most active repos on GitHub.


tommoor:
Thanks, we have an open branch right now to add dynamic loading of code languages so this addition depends on that being completed.

tommoor:
Seems dynamic language loading is still delayed, so merging for now.

PR #4796, opened 2023-01-28
chore: Use streams where possible in import/export to lower memory usage

tommoor:
This should address issues where the task processor crashes due to extremely large attachments.

### DIFF
--- a/server/queues/tasks/ExportDocumentTreeTask.ts
+++ b/server/queues/tasks/ExportDocumentTreeTask.ts
@@ -55,10 +55,10 @@ export default abstract class ExportDocumentTreeTask extends ExportTask {
     await Promise.all(
       attachments.map(async (attachment) => {
         try {
-          const img = await getFileByKey(attachment.key);
+          const stream = getFileByKey(attachment.key);
           const dir = path.dirname(pathInZip);
-          if (img) {
-            zip.file(path.join(dir, attachment.key), img as Blob, {
+          if (stream) {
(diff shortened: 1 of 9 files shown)


server/queues/tasks/ImportTask.ts
```diff
@@ -161,15 +161,27 @@ export default abstract class ImportTask extends BaseTask<Props> {
...
-   * and return the path.
+   * Fetch the remote data associated with the file operation as a Buffer.
    *
    * @param fileOperation The FileOperation to fetch data for
-   * @returns string
+   * @returns A promise that resolves to the data as a buffer.
    */
-  protected async fetchData(fileOperation: FileOperation) {
-    return fileOperation.buffer;
+  protected async fetchData(fileOperation: FileOperation): Promise<Buffer> {
```
tommoor, line 169: Unfortunately `JSZip` cannot be initialized with a stream, so import remains bottlenecked by loading the zip file.

server/queues/tasks/ExportDocumentTreeTask.ts
```diff
@@ -55,10 +55,10 @@ export default abstract class ExportDocumentTreeTask extends ExportTask {
     await Promise.all(
       attachments.map(async (attachment) => {
         try {
-          const img = await getFileByKey(attachment.key);
```
tommoor, line 58: Previously the entire attachment would end up in memory here.

Issue #4785, opened 2023-01-27
Some UX improvements to the document explorer


apoorv-mishra:
Now that we've [factored out document explorer](https://github.com/outline/outline/pull/4763) from publish modal, some possible UX improvements to it could be:

- [x] ~Smoothly~ scroll children into view as an item is expanded
- [ ] Prioritize recently viewed items in search results
- [x] ~Highlight matches in bold, in search results~

tommoor:
I'm not sold on the last one, on a small list like this lots of highlights in the results is more distracting than useful

tommoor:
<img width="307" alt="image" src="https://user-images.githubusercontent.com/380914/215338812-50381932-9a2d-4302-bb25-5c58b4970962.png">

I don't think we need to display the document name twice, can lose the last part of the breadcrumb.

PR #4812, opened 2023-01-31
Scroll children into view upon expansion

apoorv-mishra:
Towards #4785 

### Questions
- [ ] Disabled scroll for all of `onPointerMove`, `onClick` and `onDisclosureClick`. Hope that's fine?

### DIFF
--- a/app/components/DocumentExplorer.tsx
+++ b/app/components/DocumentExplorer.tsx
@@ -1,13 +1,5 @@
 import FuzzySearch from "fuzzy-search";
-import {
-  includes,
-  difference,
-  concat,
-  filter,
-  throttle,
-  map,
-  fill,
-} from "lodash";
+import { includes, difference, concat, filter, map, fill } from "lodash";
 import { observer } from "mobx-react";
 import { StarredIcon, DocumentIcon } from "outline-icons";
 import * as React from "react";
@@ -115,7 +107,7 @@ function DocumentExplorer({ onSubmit, onSelect, items }: Props) {
     (node: number) => {
       if (itemRefs[node] && itemRefs[node].current) {
         scrollIntoView(itemRefs[node].current as HTMLSpanElement, {
-          behavior: "smooth",
+          behavior: "auto",
           block: "center",
         });
       }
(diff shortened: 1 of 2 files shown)


app/components/DocumentExplorer.tsx
```diff
@@ -258,42 +292,55 @@ function DocumentExplorer({ onSubmit, onSelect, items }: Props) {
...
-        } else {
-          toggleSelect(activeNode);
-        }
-        break;
-      }
+    if (ev.key === "ArrowDown" || ev.key === "ArrowUp") {
+      ev.preventDefault();
     }
+
+    const throttledOnKeyDown = throttle(
```
apoorv-mishra, line 299: Had to throttle because the event fires at a rate faster than smooth scroll animation and as a result the active node quickly runs out of view after which further scrolling via keyboard stops.

app/components/DocumentExplorer.tsx
```diff
@@ -258,42 +292,55 @@ function DocumentExplorer({ onSubmit, onSelect, items }: Props) {
...
-      case "Enter": {
-        if (isModKey(ev)) {
-          onSubmit();
-        } else {
-          toggleSelect(activeNode);
-        }
-        break;
-      }
+    if (ev.key === "ArrowDown" || ev.key === "ArrowUp") {
+      ev.preventDefault();
```
apoorv-mishra, line 296: Forced to put this out of `throttle` since then there seemed to be no way preventing it, no way I could think of at the very least.

app/components/DocumentExplorer.tsx
tommoor: Unfortunately it feels unresponsive and laggy now, that 100ms response time makes all the difference.It would likely feel better to remove the smooth scroll and have it be instant

apoorv-mishra: Want both the smooth scroll and 100ms throttling removed?

tommoor: I think it's worth trying yes

apoorv-mishra: Disabled! Not lagging as before, better UX I'd say :wink:

tommoor (approved):

PR #4838, opened 2023-02-06
Capability to mention users in a document

### LINKED ISSUE #889: Mention/tag users in a document
It would be great to be able to tag another users within team as part of the document in a similar way to github mentions with "@" character.

apoorv-mishra:
Closes #889

__Todo__
- [x] Add inline node `mention` to store `id` and `name` attributes
- [x] Trigger menu open upon typing `@`
- [x] Add `MentionsProcessor`, similar to `BacklinksProcessor` that parses which mentions have been added on "documents.publish" and "documents.update" events.
  - [x] Add `parseMentions` util which parses `mention` nodes in doc text
  - [x] MentionsProcessor tests
    - [x] "documents.publish"
    - [x] "documents.update"
    - [x] "documents.delete"
    - [x] "users.delete"
- [x] Add notification email for mentions scheduled from `MentionsProcessor`

__Fixes__
- [x] ~Fix search~ Menu items are not updated immediately but updated after next keystroke - off by one issue.
- [x] Node insertion currently fails, needs to be fixed
- [x] Cursor disappears after node is inserted
- [x] ~The node flashes unstyled raw markdown upon initial load, `rulePlugin` needs to be added~ App crashes with `structure is null` TypeError
- [x] All mention nodes except the first one flashing upon reload(cause found)
- [x] Remove `Mention` model and `MentionsProcessor`
- [x] Include `actor` in node `attrs`


app/editor/components/CommandMenu.tsx
```diff
@@ -80,14 +83,6 @@ class CommandMenu<T extends MenuItem> extends React.Component<Props<T>, State> {
...
   }
 
-  shouldComponentUpdate(nextProps: Props<T>, nextState: State) {
-    return (
-      nextProps.search !== this.props.search ||
-      nextProps.isActive !== this.props.isActive ||
-      nextState !== this.state
-    );
-  }
-
```
apoorv-mishra, line 90: Removed `shouldComponentUpdate` to fix :point_down:
> Menu items are not updated immediately but updated after next keystroke - off by one issue.

`shouldComponentUpdate` seemed to [prevent re-rendering when necessary](https://reactjs.org/docs/react-component.html#shouldcomponentupdate).

Hoping that the removal isn't a problem :crossed_fingers:

tommoor: This is pretty dated stuff, I'm not too worried about it

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,187 @@
...
+            "data-id": dom.dataset.id,
+            label: dom.innerText,
+          }),
+        },
+      ],
+      toDOM: (node) => {
+        return [
+          "span",
+          {
+            class: `${this.name}`,
```
apoorv-mishra, line 55: @tommoor what style do we expect for this `node`? Currently it's indistinguishable from rest of the document text.

tommoor: Lets make it something like this for now – will definitely revisit before it goes live though:

image

app/editor/index.tsx
```diff
@@ -202,6 +206,8 @@ export class Editor extends React.PureComponent<
     this.events.on(EventType.blockMenuClose, this.handleCloseBlockMenu);
     this.events.on(EventType.emojiMenuOpen, this.handleOpenEmojiMenu);
     this.events.on(EventType.emojiMenuClose, this.handleCloseEmojiMenu);
+    this.events.on(EventType.mentionMenuOpen, this.handleOpenMentionMenu);
+    this.events.on(EventType.mentionMenuClose, this.handleCloseMentionMenu);
```
tommoor, line 210: Need to add it to the list to refactor this mess to something less repetitive, probably an `openMenu` enum at a minimum as it's not possible to have multiple open at once.

server/models/Mention.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+    {
+      model: Document,
+      as: "document",
+    },
+  ],
+}))
+@Table({ tableName: "mentions", modelName: "mention" })
+@Fix
+class Mention extends IdModel {
```
tommoor, line 36: I wasn't expecting an entire new model, what took you down this path… perhaps it can be avoided with some discussion. I think it's probably better to keep the source of truth inside the document content.

apoorv-mishra: Hmm...TBH, I just referenced the `Backlink` implementation. But now that you raised this point, I'll revisit to understand if this could be avoided.

app/editor/components/MentionMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+import useRequest from "~/hooks/useRequest";
+import useStores from "~/hooks/useStores";
+import CommandMenu, { Props } from "./CommandMenu";
+import MentionMenuItem from "./MentionMenuItem";
+
+interface MentionItem extends MenuItem {
+  name: string;
+  title: string;
+  attrs: {
+    "data-type": string;
```
tommoor, line 13:
```suggestion
    type: MentionType;
```

Lets make it an enum `MentionType.User` is the only option for now

app/editor/components/MentionMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+    "data-id": string;
```
tommoor, line 14:
```suggestion
    id: string;
    actorId: string;
```

app/editor/components/MentionMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (data) {
+      setItems(
+        data.map((d) => ({
+          name: "mention",
+          title: d.name,
+          appendSpace: true,
+          attrs: {
+            "data-type": "user",
+            "data-id": d.id,
+            "data-actor": auth.user?.id,
```
tommoor, line 45: These don't need "data-" in front, they represent the attributes stored in the document data and don't have to directly map to dom properties.

```suggestion
            type: MentionType.User,
            id: d.id,
            actorId: auth.user?.id,
```

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const document = await Document.unscoped().findByPk(documentId);
+    if (!document) {
+      return false;
+    }
+
+    return { document };
+  }
+
+  protected subject() {
+    return "You are mentioned";
```
tommoor, line 41: Ideally this would be `<actorName> mentioned you in "<document name>"`

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,88 @@
...
+    teamUrl,
+    document,
+    eventName = "mentioned",
+  }: Props): string {
+    return `
+You are mentioned
+
+${actorName} ${eventName} you in document "${document.title}".
+
+Open Document: ${teamUrl}${document.url}
```
tommoor, line 59: Lets add the `mentionId=` on here, even if it isn't supported today it will be

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  protected renderAsText({
+    actorName,
+    teamUrl,
+    document,
+    eventName = "mentioned",
+  }: Props): string {
+    return `
+You are mentioned
```
tommoor, line 55:
```suggestion
You were mentioned
```

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,88 @@
...
+    teamUrl,
+  }: Props) {
+    const link = `${teamUrl}${document.url}?ref=notification-email`;
+
+    return (
+      <EmailTemplate>
+        <Header />
+
+        <Body>
+          <Heading>You are mentioned</Heading>
```
tommoor, line 76:
```suggestion
          <Heading>You were mentioned</Heading>
```

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -99,12 +100,30 @@ export default class NotificationsProcessor extends BaseProcessor {
         );
       }
     }
+
+    // send notifs to mentioned users
+    const mentions = parseMentions(document);
+    for (const mention of mentions) {
+      const [recipient, actor] = await Promise.all([
+        User.findByPk(mention["data-id"]),
```
tommoor, line 108:
```suggestion
        User.findByPk(mention.id),
```

This will be like this when above suggestions are made

shared/editor/rules/mention.ts
```diff
@@ -0,0 +1,82 @@
...
+  const mType = tokens[idx].attrGet("data-type");
+  const mId = tokens[idx].attrGet("data-id");
+  const mActor = tokens[idx].attrGet("data-actor");
+  const label = tokens[idx].content;
+  const id = tokens[idx].attrGet("id");
+
+  return `<span class="mention" id="${id}" data-actor="${mActor}" data-type="${mType}" data-id="${mId}">${label}</span>`;
+}
+
+function parseMentions(state: StateCore) {
```
tommoor, line 15: I think you should revert the last changes to store the actor Id in the Markdown, it shouldn't be needed at all and complicates the format a lot

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+    state.write(`@[${label}](mention://m/${id}/a/${mActor}/${mType}/${mId})`);
+  }
+
+  parseMarkdown() {
+    return {
+      node: "mention",
+      getAttrs: (tok: Token) => ({
+        "data-type": tok.attrGet("data-type"),
+        "data-id": tok.attrGet("data-id"),
+        "data-actor": tok.attrGet("data-actor"),
```
tommoor, line 200:
```suggestion
        type: tok.attrGet("data-type"),
        id: tok.attrGet("data-id"),
```

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
+    const mType = node.attrs["data-type"];
+    const mId = node.attrs["data-id"];
+    const mActor = node.attrs["data-actor"];
+    const label = node.attrs.label;
+    const id = node.attrs.id;
+
+    console.log(mActor);
```
tommoor, line 190:
```suggestion
```

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+      ],
+      toDOM: (node) => {
+        return [
+          "span",
+          {
+            class: `${node.type.name}`,
+            id: node.attrs.id,
+            "data-type": node.attrs["data-type"],
+            "data-id": node.attrs["data-id"],
+            "data-actor": node.attrs["data-actor"],
```
tommoor, line 68:
```suggestion
            "data-type": node.attrs.type,
            "data-id": node.attrs.id,
            "data-actorId": node.attrs.actorId,
```

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+      group: "inline",
+      atom: true,
+      parseDOM: [
+        {
+          tag: `span.${this.name}`,
+          preserveWhitespace: "full",
+          getAttrs: (dom: HTMLElement) => ({
+            "data-type": dom.dataset.type,
+            "data-id": dom.dataset.id,
+            "data-actor": dom.dataset.actor,
```
tommoor, line 54:
```suggestion
            type: dom.dataset.type,
            id: dom.dataset.id,
            actorId: dom.dataset.actorId,
```

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+            class: `${node.type.name}`,
+            id: node.attrs.id,
+            "data-type": node.attrs["data-type"],
+            "data-id": node.attrs["data-id"],
+            "data-actor": node.attrs["data-actor"],
+          },
+          node.attrs.label,
+        ];
+      },
+      toPlainText: (node) => node.attrs.label,
```
tommoor, line 73:
```suggestion
      toPlainText: (node) => `@${node.attrs.label}`,
```

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+  get name() {
+    return "mention";
+  }
+
+  get schema(): NodeSpec {
+    return {
+      attrs: {
+        style: {
+          default: "",
+        },
```
tommoor, line 25:
```suggestion
```

Not sure how this snuck in

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,206 @@
...
+        "data-actor": {
+          default: undefined,
+        },
+        label: {
+          default: undefined,
+        },
+        id: {
+          default: undefined,
+        },
```
tommoor, line 40:
```suggestion
        type: {},
        label: {},
        id: {
          default: undefined,
        },
        actorId: {
          default: undefined,
        },
```

"type" and "label" are required, the others can be optional

shared/editor/components/Styles.ts
```diff
@@ -124,6 +124,17 @@ font-size: 1em;
 line-height: 1.6em;
 width: 100%;
 
+.mention {
+  background: ${props.theme.listItemHoverBackground};
```
tommoor, line 128: Lets add a new item to the theme for this, `mentionBackground` – it can be the same color as `listItemHoverBackground` for now.

server/utils/parseMentions.ts
```diff
@@ -0,0 +1,38 @@
...
+  const node = DocumentHelper.toProsemirror(document);
+  const mentions: Record<string, string>[] = [];
+  const visited: Map<string, boolean> = new Map();
+
+  function findMentions(node: Node) {
+    if (visited.get(node.attrs.id)) {
+      return;
+    }
+
+    visited.set(node.attrs.id, true);
```
tommoor, line 23: Seems fine to use a `Set` but why not just store the mentions in the values of the set rather than keeping two data structures? 😉

apoorv-mishra: The map keeps track of which nodes have already been visited during depth-first traversal, kind of like memoization to improve perf by not traversing same nodes again and again. May use `Set` to eliminate duplicates post traversal but doesn't help with efficient traversal.

tommoor: `descendants` should already be efficient, it shouldn't be visiting the same node more than once at least

shared/editor/nodes/Mention.ts
apoorv-mishra: > Lets add the mentionId= on here, even if it isn't supported today it will be

Let's include `mentionId` too in `attrs`?

server/utils/parseMentions.ts
apoorv-mishra: Not sure how that works but I've found for example the `doc` node(and others) being visited more than once in local testing. Either way, the `visited` checks are not completely accurate because it relies on `node.attrs.id` which is specific to `mention` nodes and not all. Hence, falling back to the `Set` approach :+1: . Wish there was a node explorer for prosemirror, would've been very helpful with debugging :disappointed:

shared/editor/nodes/Mention.ts
tommoor: `id` is mentionId?

apoorv-mishra: `id` = identity of mentioned entity, like `userId`
`mentionId` = identify of the `mention` node itself which goes into the `id` HTML attribute of the resulting `span`.

tommoor: Ah, this could be quite easily confusing – perhaps `id` should be renamed to `modelId` then?

shared/editor/components/Styles.ts
apoorv-mishra: Enhancement tip: Color mention nodes differently, each color representing a mentioned entity. Not a blocker, can be pursued later.

shared/editor/nodes/Mention.ts
apoorv-mishra: :100: On it.

apoorv-mishra: And use `id` in place of `mentionId`?

tommoor: right

app/editor/components/MentionMenuItem.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+type MentionMenuItemProps = Omit<CommandMenuItemProps, "shortcut" | "theme"> & {
+  label: string;
+};
+
+export default function MentionMenuItem({
+  label,
+  ...rest
+}: MentionMenuItemProps) {
+  return <CommandMenuItem {...rest} title={label} />;
```
tommoor, line 14: Can we pass the user avatar as the `icon` prop here

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    return { document };
+  }
+
+  protected subject({ actorName, document }: Props) {
+    return `${actorName} mentioned you in "${document.title}"`;
+  }
+
+  protected preview({ actorName, eventName }: Props): string {
+    return `${actorName} ${eventName} you`;
```
tommoor, line 46:
```suggestion
    return `${actorName} mentioned you`;
```

It's always the same 🤷

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,91 @@
...
+    actorName,
+    teamUrl,
+    document,
+    eventName = "mentioned",
+    mentionId,
+  }: Props): string {
+    return `
+You were mentioned
+
+${actorName} ${eventName} you in document "${document.title}".
```
tommoor, line 59:
```suggestion
${actorName} mentioned you in the document "${document.title}".
```

server/emails/templates/MentionNotificationEmail.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const link = `${teamUrl}${document.url}?ref=notification-email&mentionId=${mentionId}`;
+
+    return (
+      <EmailTemplate>
+        <Header />
+
+        <Body>
+          <Heading>You were mentioned</Heading>
+          <p>
+            {actorName} {eventName} you in document{" "}
```
tommoor, line 81:
```suggestion
            {actorName} mentioned you in the document{" "}
```

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -99,12 +100,31 @@ export default class NotificationsProcessor extends BaseProcessor {
         );
       }
     }
+
+    // send notifs to mentioned users
+    const mentions = parseMentions(document);
```
tommoor, line 105: I think we should send the mentioned emails _first_ and then send to the document notification recipients. If anyone was mentioned in the document we can suppress the other email so they don't receive both…

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -99,12 +100,31 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+    for (const mention of mentions) {
+      const [recipient, actor] = await Promise.all([
+        User.findByPk(mention.modelId),
+        User.findByPk(mention.actorId),
+      ]);
+      if (recipient && actor && recipient.id !== actor.id) {
+        await MentionNotificationEmail.schedule({
+          to: recipient.email,
+          documentId: event.documentId,
+          eventName: "mentioned",
```
tommoor, line 115:
```suggestion
```

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -163,6 +183,28 @@ export default class NotificationsProcessor extends BaseProcessor {
         );
       }
     }
+
+    // send notifs to newly mentioned users
+    const prev = await revision.previous();
+    const oldMentions = prev ? parseMentions(prev) : [];
+    const newMentions = parseMentions(document);
+    const mentions = differenceBy(newMentions, oldMentions, "modelId");
```
tommoor, line 191: While this accounts for new mentions it doesn't account for the same user being mentioned multiple times, e.g. you have a weekly meeting minutes doc and they get mentioned one week and then the next you'd expect to get another notification.

Might be that the only change needed is to differenceBy "id" instead of "modelId"?

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -163,6 +183,28 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+    for (const mention of mentions) {
+      const [recipient, actor] = await Promise.all([
+        User.findByPk(mention.modelId),
+        User.findByPk(mention.actorId),
+      ]);
+      if (recipient && actor && recipient.id !== actor.id) {
```
tommoor, line 197: Similar to the document creation case, lets do the mention loop _first_ – and then if we sent a mention email don't send a `DocumentNotificationEmail` as well.

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -163,6 +183,28 @@ export default class NotificationsProcessor extends BaseProcessor {
...
+        await MentionNotificationEmail.schedule({
+          to: recipient.email,
+          documentId: event.documentId,
+          eventName: "mentioned",
```
tommoor, line 201:
```suggestion
```

server/utils/parseMentions.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Parse a list of mentions contained in a document or revision
+ *
+ * @param document Document or Revision
+ * @returns An array of mentions in passed document or revision
+ */
+export default function parseMentions(
+  document: Document | Revision
+): Record<string, string>[] {
+  const node = DocumentHelper.toProsemirror(document);
+  const mentions: Set<Record<string, string>> = new Set();
```
tommoor, line 15: Is set achieving anything now? Could probably just be an array as we convert it to one at the end

shared/editor/components/Styles.ts
```diff
@@ -124,6 +124,17 @@ font-size: 1em;
...
 width: 100%;
 
+.mention {
+  background: ${props.theme.mentionBackground};
+  border-radius: 12px;
+  padding-bottom: 2px;
+  padding-top: 1px;
+  padding-left: 4px;
+  padding-right: 4px;
+  font-weight: bold;
```
tommoor, line 134:
```suggestion
  font-weight: 500;
```

server/queues/processors/NotificationsProcessor.ts
apoorv-mishra: And we don't want to suppress _mention_ notifications at all?

tommoor: Currently both notifications should be sent – I think we'll tackle grouping and delay of notifications separately at that level of the system once #4924 is complete.

server/queues/processors/NotificationsProcessor.ts
apoorv-mishra: Well, the `oldMentions` don't have `id` since those are obtained from the revision text and _not state_.

tommoor: Ah in that case we'll have to count the mentions per-user

tommoor: That won't tell us which is the _new_ one though I suppose, for that we'd have to look at the position in the document, if you pass the `pos` of each mention back from `parseMentions` then you can `differenceBy` that.

tommoor: That won't tell us which mention is the new one though I suppose… hmm, we could assume it's the revision creator which is fairly safe but not foolproof or store the `id` in the markdown 😬 (I actually thought we still were and only actorId was dropped)

apoorv-mishra: Let's store the `id` in markdown :smile:

server/utils/parseMentions.test.ts
```diff
@@ -15,14 +15,16 @@ it("should return an array of mentions", async () => {
   const document = await buildDocument({
     text: `# Header
 
-@[Alan Kay](mention://user/34095ac1-c808-45c0-8c6e-6c554497de64) :wink:
+@[Alan Kay](mention://m/2767ba0e-ac5c-4533-b9cf-4f5fc456600e/user/34095ac1-c808-45c0-8c6e-6c554497de64) :wink:
```
tommoor, line 18: What is the `/m/` signifying? Possible to lose it?

apoorv-mishra: Just `mention`. Yes we can lose it.

tommoor: We already have `mention://`, so yes feels duplicative 😄 

Sorry to be so picky, we only get one chance to get this right and then it's kind of stuck in the data forever.

apoorv-mishra: Yes. Totally fine :+1:

apoorv-mishra: Also, not a blocker but it'd be good to have support utility that enables testing the `document.state` scenario here. Currently, `parseMentions` is tested only against `document.text` and the two scenarios differ in the way that one stores `actorId` but other does not.

tommoor: Mmm yes, we can probably make a test in the factory so that every test document gets the correct `state`

shared/editor/nodes/Mention.ts
```diff
@@ -0,0 +1,192 @@
...
+import { EditorState, TextSelection, Plugin } from "prosemirror-state";
+import { MarkdownSerializerState } from "../lib/markdown/serializer";
+import { run } from "../plugins/BlockMenuTrigger";
+import isInCode from "../queries/isInCode";
+import mentionRule from "../rules/mention";
+import { Dispatch, EventType } from "../types";
+import Node from "./Node";
+
+const OPEN_REGEX = /(?:^|\s)@([a-zA-Z+-]+)?$/;
+const CLOSE_REGEX = /(?:^|\s)@(([0-9a-zA-Z_+-]*\s+)|(\s+[0-9a-zA-Z_+-]+)|[^0-9a-zA-Z_+-]+)$/;
```
apoorv-mishra, line 13: These need to be changed to support `i18n` :see_no_evil:

apoorv-mishra: I hope it's safe to use https://github.com/tc39/proposal-regexp-unicode-property-escapes :crossed_fingers:

tommoor: I wonder if we can just use `{Letter}` here, or more specifically `\p{Letter}\d` any letter in any language and digits.

apoorv-mishra: Are digits needed for name?

apoorv-mishra: By the way, it's recommended to include _combining marks_ as well.

>“character” really means “Unicode code point”. \p{L} matches a single code point in the category “letter”. If your input string is à encoded as U+0061 U+0300, it matches a without the accent. If the input is à encoded as U+00E0, it matches à with the accent. The reason is that both the code points U+0061 (a) and U+00E0 (à) are in the category “letter”, while U+0300 is in the category “mark”.

:point_up: from https://www.regular-expressions.info/unicode.html#prop

 So `\p{L}{M}` should be good.

tommoor: I probably would just because folks signing in with different OIDC providers often end up with a username-like string in the name field. It doesn't hurt

tommoor (approved):

PR #4849, opened 2023-02-09
fix japanese phrases

doracpphp:
Japanese phrasing has been corrected to be more natural.


tommoor:
Hi, thanks but translations must be updated through CrowdIn or they will be overridden – please check out the documentation here:

https://github.com/outline/outline/blob/main/docs/TRANSLATION.md#translating-strings

Issue #4854, opened 2023-02-10
Wrong download link for uploaded pdf


setavenger:
The download link to an uploaded pdf is incorrect. It points to an aws location and not to the one set in the env file.  I was able to spot the file in the digital ocean spaces I am using. File upload worked fine, it's just the download link.

**To Reproduce**
1. Upload a pdf file via drag and drop to a doc
2. click on download
3. file link will be set to "<bucket_name>.s3.<region>.amazonaws.com/uploads

**Expected behavior**
The link should point to the correct S3 storage location as defined in the env file.

**Screenshots**
If applicable, add screenshots or videos to help explain your problem.

**Outline (please complete the following information):**
- Install: self hosted
- Version: 9cb85ddfc3ea

**Desktop (please complete the following information):**
n.a.

**Mobile (please complete the following information):**
n.a.

tommoor:
What is the value of `AWS_S3_ACL` in your configuration?

tommoor:
Closing as dupe of #4853

setavenger:
AWS_S3_ACL is not set. It was not filled in the AWS example docs and I didn't know what it was for.

PR #4856, opened 2023-02-10
3810 error the server does not support ssl connections

PhoeKhwar12:
#4854


tommoor:
Why have you made the commits as me? That's super weird.

Issue #4863, opened 2023-02-12
Cannot use any Postgres schema other than public


kerwitz:
The Outline server side code currently offers no way to manipulate the schema used by Sequelize. This limits us to using dedicated databases whereas we could be using a single database with multiple schemas if desired. Because no configuration is made towards Sequelize regarding the schema, Sequelize falls back to `public` and prefixes all table names this way.

This prevents the `db:migrate` command from running through, as it exits early with this message:
```bash
$ sequelize db:migrate --env=production-ssl-disabled

Sequelize CLI [Node: 16.14.2, CLI: 6.4.1, ORM: 6.20.1]

Loaded configuration file "server/config/database.json".
Using environment "production-ssl-disabled".
== 20160619080644-initial: migrating =======
== 20160619080644-initial: migrated (0.216s)

== 20160622043741-add-parent-document: migrating =======

ERROR: relation "public.documents" does not exist

error Command failed with exit code 1.
```

This error message indicates Sequelize is trying to find the "documents" table inside of the default "public" schema.

**To Reproduce**
With Postgres it is possible to change the default search path of any role / user to a different schema if desired, i.e. I could create a schema "outline_db" and create a user "outline" that defaults to this schema instead of the regular "public" one:
```sql
create role outline with login encrypted password '1234';
# .. grant connect, schema usage and privileges to the new user here ..
alter role outline set search_path to outline_db;
```

Now I can configure outline to connect to my database using the new "outline" user (docker.env):
```
DATABASE_URL=postgres://outline:1234@my.db.com/db
```

**Expected behavior**
The Outline installation should correctly work inside of the search path defined for the user. The db:migrate commands should successfully build the database structure into the "outline_db" schema so that I can later run my Outline instance against it. Alternatively to "forcing" Outline into the desired schema via the search path, Outline should expose additional configuration to make using another schema possible.

Sequelize allows for the schema to be defined in it's options like this:
```ts
new Sequelize(url, { schema: env.DB_SCHEMA, })
```

As shown it would be really helpful if Outline were to expose this setting via a dedicated environment variable, e.g. `DB_SCHEMA`. If this variable has been set its value should be passed to Sequelize accordingly, if not Sequelize should fall back to its native defaults.

**Screenshots**
N/A

**Outline (please complete the following information):**
- Install: self hosted
- Version: 0.67.2

**Desktop (please complete the following information):**
Not applicable.

**Mobile (please complete the following information):**
Not applicable.

[...]

PR #4865, opened 2023-02-12
Expose Sequelize schema configuration through env variable

kerwitz:
This fixes https://github.com/outline/outline/issues/4863 by exposing the schema configuration value of Sequelize through a new DATABASE_SCHEMA environment variable. This way hosted installations of Outline can be directed to use custom schemas other than the default "public" schema if desired.

This is a follow up of #4864 as the earlier pull request included unintentional changes to code style.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -18,7 +18,7 @@ DATABASE_CONNECTION_POOL_MIN=
 DATABASE_CONNECTION_POOL_MAX=
 # Uncomment this to disable SSL for connecting to Postgres
 # PGSSLMODE=disable
-# Specify a specific schema to use if deviating from defaults.
+# Optionally specify a database schema if different from "public". 
 # DATABASE_SCHEMA=
 
 # For redis you can either specify an ioredis compatible url like this
--- a/.env.sample
+++ b/.env.sample
@@ -18,6 +18,8 @@ DATABASE_CONNECTION_POOL_MIN=
 DATABASE_CONNECTION_POOL_MAX=
 # Uncomment this to disable SSL for connecting to Postgres
 # PGSSLMODE=disable
+# Specify a specific schema to use if deviating from defaults.
+# DATABASE_SCHEMA=
 
 # For redis you can either specify an ioredis compatible url like this
 REDIS_URL=redis://localhost:6379
--- a/server/database/sequelize.ts
+++ b/server/database/sequelize.ts
@@ -3,6 +3,7 @@ import env from "@server/env";
 import Logger from "../logging/Logger";
 import * as models from "../models";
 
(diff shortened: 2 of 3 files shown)


.env.sample
```diff
@@ -18,6 +18,8 @@ DATABASE_CONNECTION_POOL_MIN=
 DATABASE_CONNECTION_POOL_MAX=
 # Uncomment this to disable SSL for connecting to Postgres
 # PGSSLMODE=disable
+# Specify a specific schema to use if deviating from defaults.
```
tommoor, line 21:
```suggestion
# Optionally specify a database schema if different from "public". 
```

tommoor (commented):
Looks like there's a small linter issue, otherwise I don't see any problem with merging this optional config 👍

kerwitz:
Happy to fix, thanks for the heads up!

kerwitz:
Looking at it, this does not bring the desired functionality just yet - so this PR shouldn't be merged just jet. While the desired schema name will be used in code, it is not respected currently in the CLI commands of Sequelize because they rely on the ./server/config/database.json file. I'll have a look.

tommoor:
Closing for now as it's been quiet for a few months

33Fraise33:
Hello, @tommoor 

This is still an issue when you create an outline database and an outline user/role in postgres giving it all permissions to the outline database. It will give permission errors as it will try to create items in the public schema. 

postgres permissions: (notice outline_pg_user having CTc permissions = ALL on outline db)
```
    Name     |    Owner    | Encoding | Locale Provider |  Collate   |   Ctype    | Locale | ICU Rules |        Access privileges
-------------+-------------+----------+-----------------+------------+------------+--------+-----------+---------------------------------
 cm_pg_admin | cm_pg_admin | UTF8     | libc            | en_US.utf8 | en_US.utf8 |        |           |
 outline     | cm_pg_admin | UTF8     | libc            | en_US.utf8 | en_US.utf8 |        |           | =Tc/cm_pg_admin                +
             |             |          |                 |            |            |        |           | cm_pg_admin=CTc/cm_pg_admin    +
             |             |          |                 |            |            |        |           | outline_pg_user=CTc/cm_pg_admin
```

But still when running the initialize command: 
`docker compose run --rm outline yarn db:migrate --env=production-ssl-disabled`
you get the following error: 
```
ERROR:  permission denied for schema public at character 28
2025-06-11T08:14:58.546845883Z 2025-06-11 08:14:58.546 UTC [134] STATEMENT:  CREATE TABLE IF NOT EXISTS "SequelizeMeta" ("name" VARCHAR(255) NOT NULL UNIQUE , PRIMARY KEY ("name"));
```

Issue #4862, opened 2023-02-12
Plugin architecture refactor


tommoor:
- [x] Framework
- [x] Slack
- [x] Webhooks
- [x] Google auth
- [x] Azure auth
- [x] Email auth
- [x] OIDC auth

---

- [x] Check all works in Vite refactor branch
- [x] Remove old auth provider loading
- [x] Move auth provider logo loading

PR #4866, opened 2023-02-12
Move bulk of webhook logic to plugin

tommoor:
related #4862

Everything moved except models and policies, which haven't been hooked up to plugin system yet.


tommoor:
@SocketSecurity ignore koa-webpack-hot-middleware@1.0.3

PR #4868, opened 2023-02-13
feat: add support ZhiXi MindMap

fangzhengjin:
editor support linked ZhiXi MindMap.
like this:
image


tommoor:
Thank you for your effort, we won't be adding this service at this time. Soon there will be arbitraty iframe embeds which will allow for this type of thing.

PR #4876, opened 2023-02-14
refactor: add server side validation schema for authProviders

elidrissidev:
Refactoring the Authentication Providers API endpoints to use server-side validation schema.

- See outline/outline#4284


server/routes/api/authenticationProviders/authenticationProviders.ts
```diff
@@ -1,26 +1,27 @@
...
 
 const router = new Router();
 
 router.post(
   "authenticationProviders.info",
   auth({ admin: true }),
-  async (ctx: APIContext) => {
+  validate(T.AuthenticationProvidersInfoSchema),
+  async (ctx: APIContext<T.AuthenticationProvidersInfoReq>) => {
     const { id } = ctx.request.body;
```
tommoor, line 22: In order to get the benefit of the typed response we need to change the handlers to use `ctx.input.body` instead of `ctx.request.body`, the former is typed correctly after the validate middleware.

I just realized this also applies to https://github.com/outline/outline/pull/4859where I missed it!

PR #4881, opened 2023-02-15
refactor: add server side validation schema for groups

elidrissidev:
Refactoring the Groups API endpoints to use server-side validation schema.

- See outline/outline#4284


tommoor (approved):
🙏 Took a little longer as the diff was harder to parse.

PR #4912, opened 2023-02-22
fix: Bundle size calculation includes development React

tommoor:
Testing this as a solution in CI.


tommoor:
image

Well, would you look at that.

PR #4932, opened 2023-02-26
Webhook support for comments

### LINKED ISSUE #4931: Webhook support for comments

tommoor:
closes #4931


stereo9:
@tommoor I just tested the webhooks for comments on v0.69.0 , by attaching it to a test webhook via webhook.site
it seems like comments don't trigger the request to be sent.
when assigning to all events though, the others seem to work without any problems

tommoor:
Testing here it works as expected, which event specifically did you enable the hook for?

stereo9:
thank you for the quick reply, which actually helped me to narrow down the problem.
i just clicked "all comment events"
image
but this didnt do the job. when selecting them individually it works like a charm.

tommoor:
I think this bug is wider than comments, will create a new issue.

Issue #4952, opened 2023-02-28
Google https://datastudio.google.com links changed to https://lookerstudio.google.com


phaethon:
Google has rebranded Data Studio to Looker Studio and as part of that URL has changed. If I try to use feature to embed Datastudio into Outline, I get `Sorry, that link won't work for this embed type` When I manually change beginning of URL from https://lookerstudio.google.com/... to https://datastudio.google.com/... embedding works.

Limezy:
You should be able to propose a very simple corrective PR by correcting this regex : 
https://github.com/outline/outline/blob/bb21fa725cc58bc7d397981483926df5837ed9f3/shared/editor/embeds/GoogleDataStudio.tsx#L28

Limezy:
OK let me try to do that one

PR #4967, opened 2023-03-02
Fix 4952

Limezy:
Fixes : #4952


Limezy:
image

shared/editor/embeds/GoogleLookerStudio.tsx
```diff
@@ -3,30 +3,30 @@ import Frame from "../components/Frame";
...
       border
     />
   );
 }
 
-GoogleDataStudio.ENABLED = [
+GoogleLookerStudio.ENABLED = [
   new RegExp(
-    "^https?://datastudio\\.google\\.com/(embed|u/0)/reporting/(.*)/page/(.*)(/edit)?$"
+    "^https?://lookerstudio\\.google\\.com/(embed|u/0)/reporting/(.*)/page/(.*)(/edit)?$"
```
tommoor, line 28: Can we retain support for the old embeds?

Limezy: I have tried something below but am not sure about how they are kept in the database.
Maybe it's more complex than a simple regex OR... And requires a kind of migration ?

tommoor: This is fine 👍

tommoor (approved):

shared/editor/embeds/GoogleLookerStudio.tsx
Limezy: Great !

PR #4968, opened 2023-03-02
Tldraw

Limezy:
Adding Castopod as a self-hosted option (castopod.org)
Adding back Tldraw since the beta allows iframes again

Capture d’écran 2023-03-03 à 02 38 06


shared/editor/embeds/Castopod.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from "react";
+import Frame from "../components/Frame";
+import Image from "../components/Image";
+import { EmbedProps as Props } from ".";
+
+function Castopod(props: Props) {
```
tommoor, line 6: I don't think this makes sense as a first-class integration for a team knowledgebase to be honest, we have to maintain all of these so it doesn't come without cost.

Soon we'll support arbitrary iframes which will cover these kinds of cases.

Limezy: I understand your point. In that view, legitimacy of other integrations could also be challenged. For example, Spotify ?
Are you OK to integrate back Tldraw, or shall I just close this PR ?

tommoor: Ok with tldraw!

shared/editor/embeds/Tldraw.tsx
```diff
@@ -7,13 +7,13 @@ function Tldraw(props: Props) {
     <Frame
       {...props}
       src={props.attrs.href}
-      title="Tldraw.com Embed"
+      title="Tldraw.com (beta) Embed"
```
apoorv-mishra, line 10: This "beta" here isn't necessary?

shared/editor/embeds/index.tsx
```diff
@@ -349,10 +349,9 @@ const embeds: EmbedDescriptor[] = [
     component: Spotify,
   }),
   new EmbedDescriptor({
-    title: "Tldraw",
+    title: "Tldraw (beta)",
```
apoorv-mishra, line 352: Get rid of "beta" here as well?

shared/editor/embeds/index.tsx
```diff
@@ -349,10 +349,9 @@ const embeds: EmbedDescriptor[] = [
...
     keywords: "draw schematics diagrams",
-    visible: false,
-    icon: <Img src="/images/tldraw.png" alt="Tldraw" />,
+    icon: <Img src="/images/tldraw.png" alt="Tldraw (beta)" />,
```
apoorv-mishra, line 354: Remove "beta"?

shared/editor/embeds/Tldraw.tsx
```diff
@@ -7,13 +7,13 @@ function Tldraw(props: Props) {
...
       referrerPolicy="origin"
       border
     />
   );
 }
 
-Tldraw.ENABLED = [new RegExp("https?://www\\.tldraw\\.com/r/(.*)$")];
+Tldraw.ENABLED = [new RegExp("https?://beta.tldraw.com/r/(.*)")];
```
apoorv-mishra, line 17: Oops! Should escape the `.`(dot) in regex here using `\\` as before otherwise it allows arbitrary domains like `https://betatldraw.com` which might be a trap. Refer https://github.com/outline/outline/issues/3899 for details.

Also don't forget to update tests :)

Limezy: Oh yes sorry ! I will update accordingly

shared/editor/embeds/Tldraw.tsx
Limezy: I have put bêta here and everywhere else to make sure people aren't confused. Indeed, at this moment of time, tldraw.com is still up and running. So if people paste a multiplayer link from tldraw.com, it won't work and they won't understand why. Only a multiplayer link from beta.tldraw.com will work.

shared/editor/embeds/Tldraw.tsx
Limezy: Done and done

Limezy:
Waiting for your advise on whether keeping the (beta) mention or not.
Rationale for this explained above 👆

apoorv-mishra (approved):

matbgn:
Beta is no more needed since a week or more, am I right?

https://tldraw.substack.com/p/tldraws-upcoming-re-launch?utm_source=substack&utm_medium=email

Limezy:
Yes you are right and I planned to open again a PR to correct that. The old.tldraw.com url should still be included, however, to keep support for old stuff.

PR #5034, opened 2023-03-14
Filter groups given a member

### LINKED ISSUE #5033: Filter out groups based on member
Given a `userId`, the `groups.list` endpoint should only return groups to which the given user had been added.

apoorv-mishra:
Closes #5033

### DIFF
--- a/server/routes/api/groups/groups.ts
+++ b/server/routes/api/groups/groups.ts
@@ -23,10 +23,10 @@ router.post(
   pagination(),
   validate(T.GroupsListSchema),
   async (ctx: APIContext<T.GroupsListReq>) => {
-    const { direction, sort } = ctx.input.body;
+    const { direction, sort, userId } = ctx.input.body;
     const { user } = ctx.state.auth;
 
-    const groups = await Group.findAll({
+    const groups = await Group.filterByMember(userId).findAll({
       where: {
         teamId: user.teamId,
       },
--- a/server/routes/api/groups/groups.ts
+++ b/server/routes/api/groups/groups.ts
@@ -1,5 +1,4 @@
 import Router from "koa-router";
-import { includes } from "lodash";
 import { Op } from "sequelize";
 import { MAX_AVATAR_DISPLAY } from "@shared/constants";
 import auth from "@server/middlewares/authentication";
@@ -25,10 +24,9 @@ router.post(
   validate(T.GroupsListSchema),
   async (ctx: APIContext<T.GroupsListReq>) => {
     const { direction, sort } = ctx.input.body;
-    const { userId } = ctx.input.query;
     const { user } = ctx.state.auth;
 
-    let groups = await Group.findAll({
+    const groups = await Group.findAll({
       where: {
         teamId: user.teamId,
       },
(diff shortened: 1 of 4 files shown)


server/routes/api/groups/groups.ts
```diff
@@ -35,6 +37,15 @@ router.post(
       limit: ctx.state.pagination.limit,
     });
 
+    if (userId) {
+      groups = groups.filter((group) => {
```
tommoor, line 41: This filter needs to be part of the query to function properly

tommoor (changes requested):

server/routes/api/groups/groups.ts
apoorv-mishra: Well, I tried that initially using `include` and `where` combo but it's a bit tricky since `groupMemberships` are not retained then. For example, attempted with `scope` :point_down: 
```
include: [{
  association: "groupMemberships",
  required: false,
  where: {
    userId: query.userId
  }
}]
```
but it filters out `groupMemberships` other than that of `query.userId`

Next, attempted using [top-level clauses](https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#complex-where-clauses-at-the-top-level) which resulted in same issue.

I assume we need to return all `groupMemberships` for the `group` irrespective of what `query.userId` is?

server/routes/api/groups/schema.ts
```diff
@@ -22,6 +22,9 @@ export const GroupsListSchema = z.object({
       })
       .default("updatedAt"),
   }),
+  query: z.object({
+    userId: z.string().uuid().optional(),
```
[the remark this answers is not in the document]
apoorv-mishra: Keep in both `query` and `body`?

server/routes/api/groups/groups.ts
tommoor: I haven't tried it, but perhaps we can do the include twice with different names…

```
include: [{
  association: "groupMemberships",
  required: true, 
  where: { userId: query.userId }
  as: "tmp",
}]
```

server/routes/api/groups/schema.ts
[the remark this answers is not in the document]
tommoor: Just body is fine

server/routes/api/groups/groups.ts
apoorv-mishra:
```
      {
        association: "groupMemberships",
        required: true,
        where: {
          userId: query.userId,
        },
        as: "tmp",
      },
```

throws :point_down: 

```
{
	"ok": false,
	"error": "internal_server_error",
	"status": 500,
	"message": "table name \"groupMemberships\" specified more than once"
}
```
and,

```
      {
        model: GroupUser,
        required: true,
        where: {
          userId: memberId,
        },
        as: "tmp",
      },
```

throws :point_down: 

```
{
	"ok": false,
	"error": "internal_server_error",
	"status": 500,
	"message": "group_user is associated to group using an alias. You've included an alias (tmp), but it does not match the alias(es) defined in your association (groupMemberships)."
}
```

What works is adding an extra `@HasMany` decorator with an alias(say, `members`) and then, modifying the second `include` as :point_down: 

```
      {
        association: "members",
        required: true,
        where: {
          userId: memberId,
        },
      },
```

tommoor (approved):
Thanking you

PR #5043, opened 2023-03-15
Upgrade MermaidJS (#5043

tommoor:
Adds support for [Mindmaps](https://mermaid.js.org/syntax/mindmap.html)


tommoor:
Unfortunately the same problem as before Mermaid just doesn't seem to care about their bundle size, upgrading this dependency makes the entire application 30% bigger and adds 20% to the initial js size sigh.

PR #5047, opened 2023-03-16
API to fetch users who have read/write permission on a document collection

### LINKED ISSUE #5032: @mentions shows users without access to the collection
**To Reproduce**
1. Make a collection "No access"
2. Go to a document within the collection
3. Type "@"
4. Users will appear that do not have access to the document / collection

**Expected behavior**
You should only be able to mention users with access

apoorv-mishra:
Closes #5032 

Currently, we don't have document level permissions for users so we base it on collection level permissions.

- [x] Add `documents.users` endpoint
- [x] Tests for `documents.users`
- [x] Replace `users.fetchPage` call in `MentionsMenu` with `documents.users` call

### DIFF
--- a/app/editor/components/MentionMenu.tsx
+++ b/app/editor/components/MentionMenu.tsx
@@ -35,7 +35,7 @@ type Props = Omit<
   "renderMenuItem" | "items" | "onLinkToolbarOpen" | "embeds" | "onClearSearch"
 >;
 
-function MentionMenu({ search, ...rest }: Props) {
+function MentionMenu({ search, isActive, ...rest }: Props) {
   const [items, setItems] = React.useState<MentionItem[]>([]);
   const { t } = useTranslation();
   const { users, auth } = useStores();
@@ -53,8 +53,10 @@ function MentionMenu({ search, ...rest }: Props) {
   );
 
   React.useEffect(() => {
-    request();
-  }, [request]);
+    if (isActive) {
+      request();
+    }
+  }, [request, isActive]);
 
   React.useEffect(() => {
     if (data) {
@@ -92,6 +94,7 @@ function MentionMenu({ search, ...rest }: Props) {
   return (
     <SuggestionsMenu
       {...rest}
+      isActive={isActive}
       filterable={false}
       onClearSearch={clearSearch}
       search={search}
--- a/app/editor/components/MentionMenu.tsx
+++ b/app/editor/components/MentionMenu.tsx
@@ -38,14 +38,17 @@ type Props = Omit<
 function MentionMenu({ search, ...rest }: Props) {
   const [items, setItems] = React.useState<MentionItem[]>([]);
   const { t } = useTranslation();
-  const { documents, auth } = useStores();
+  const { users, auth } = useStores();
   const location = useLocation();
   const documentId = parseDocumentSlug(location.pathname);
   const { view } = useEditor();
   const { data, request } = useRequest(
     React.useCallback(
-      () => documents.fetchUsers({ id: documentId, query: search }),
-      [documents, search]
+      () =>
+        documentId
+          ? users.fetchDocumentUsers({ id: documentId, query: search })
+          : Promise.resolve([]),
+      [users, documentId, search]
     )
   );
 
--- a/app/editor/components/MentionMenu.tsx
+++ b/app/editor/components/MentionMenu.tsx
@@ -1,9 +1,11 @@
 import { observer } from "mobx-react";
 import * as React from "react";
 import { useTranslation } from "react-i18next";
+import { useLocation } from "react-router-dom";
 import { v4 } from "uuid";
 import { MenuItem } from "@shared/editor/types";
 import { MentionType } from "@shared/types";
+import parseDocumentSlug from "@shared/utils/parseDocumentSlug";
 import User from "~/models/User";
 import Avatar from "~/components/Avatar";
 import Flex from "~/components/Flex";
@@ -36,12 +38,14 @@ type Props = Omit<
 function MentionMenu({ search, ...rest }: Props) {
   const [items, setItems] = React.useState<MentionItem[]>([]);
   const { t } = useTranslation();
-  const { users, auth } = useStores();
+  const { documents, auth } = useStores();
+  const location = useLocation();
+  const documentId = parseDocumentSlug(location.pathname);
   const { view } = useEditor();
   const { data, request } = useRequest(
     React.useCallback(
-      () => users.fetchPage({ query: search, filter: "active" }),
-      [users, search]
+      () => documents.fetchUsers({ id: documentId, query: search }),
+      [documents, search]
     )
   );
 
--- a/app/stores/DocumentsStore.ts
+++ b/app/stores/DocumentsStore.ts
@@ -11,6 +11,7 @@ import BaseStore from "~/stores/BaseStore";
 import RootStore from "~/stores/RootStore";
 import Document from "~/models/Document";
 import Team from "~/models/Team";
+import User from "~/models/User";
 import env from "~/env";
 import { FetchOptions, PaginationParams, SearchResult } from "~/types";
 import { client } from "~/utils/ApiClient";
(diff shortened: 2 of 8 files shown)


server/routes/api/documents/documents.ts
```diff
@@ -435,6 +436,49 @@ router.post(
...
+    const { user } = ctx.state.auth;
+    const document = await Document.findByPk(id);
+    authorize(user, "update", document);
+
+    let users: User[] = [];
+
+    if (document.collectionId) {
+      const collection = await Collection.findByPk(document.collectionId);
+      authorize(user, "update", collection);
+      const memberIds = await Collection.membershipUserIds(collection.id);
```
apoorv-mishra, line 454: This doesn't include the `createdBy` user of the document. Should we include it separately?

tommoor: I don't think so as they don't necessarily retain access to the document if it's been moved between collections or their permissions have changed

apoorv-mishra:
@tommoor you seem to have later modified `users.fetchPage` call [here](https://github.com/outline/outline/blob/21a44428f49d6d1d70655ed6d240964594e5d75c/app/editor/components/MentionMenu.tsx#L43). I assume it was to narrow down search space by filtering out inactive users, thereby making it efficient? If so, I don't think it's needed anymore for this new endpoint since we've already narrowed down users to `collection.membershipUserIds` here, which practically would be a much smaller number compared to _all users_.

tommoor:
It wasn’t for efficiency, it was because suspended users were showing up in the mention options

apoorv-mishra:
I see. A little eyeballing at `users.list` showed me `[Op.is]: null` for `suspendedAt` so I assumed it was the default filter anyway. Nonetheless, I believe we want to filter out suspended users for `documents.users` as well, right?

tommoor:
> I believe we want to filter out suspended users for documents.users as well, right?

Yep yep!

server/routes/api/documents/documents.ts
```diff
@@ -435,6 +436,52 @@ router.post(
   }
 );
 
+router.post(
+  "documents.users",
```
apoorv-mishra, line 440: Should we paginate this yet? :thinking:

tommoor: Yes, unfortunately we have 1000+ member collections to think of

shared/utils/parseDocumentSlug.ts
apoorv-mishra, line 1: Returning `""` instead of `undefined` because it helps dealing with nasty type errors. I don't think `undefined` here is particularly used for any specific purpose, as far as `falsy` test is concerned, `""` equally works.

shared/utils/parseDocumentSlug.ts
```diff
@@ -13,11 +13,11 @@ export default function parseDocumentSlug(url: string) {
     try {
       parsed = new URL(url).pathname;
     } catch (err) {
-      return;
+      return "";
```
apoorv-mishra, line 16: Returning `""` instead of `undefined` because it helps in dealing with nasty type errors. I don't think `undefined` here is particularly used for any specific purpose and as far as _falsy_ tests are concerned, `""` equally works.

app/editor/components/MentionMenu.tsx
```diff
@@ -36,12 +38,14 @@ type Props = Omit<
...
+  const { documents, auth } = useStores();
+  const location = useLocation();
+  const documentId = parseDocumentSlug(location.pathname);
   const { view } = useEditor();
   const { data, request } = useRequest(
     React.useCallback(
-      () => users.fetchPage({ query: search, filter: "active" }),
-      [users, search]
+      () => documents.fetchUsers({ id: documentId, query: search }),
+      [documents, search]
```
tommoor, line 48: `documentId` should be in dependencies too, no?

app/stores/DocumentsStore.ts
```diff
@@ -281,6 +282,24 @@ export default class DocumentsStore extends BaseStore<Document> {
     });
   };
 
+  @action
+  fetchUsers = async (
```
tommoor, line 286: I don't think there is a _right_ answer of where to put this, but I would sway to towards the users store as the entity being requested and stored is a `User`.

shared/utils/parseDocumentSlug.ts
tommoor: I think it's useful in that it stops you in making a useless request to `document.users` with an empty id. Instead we can guard against that and make no request at all of this returns `undefined`?

server/routes/api/documents/documents.ts
```diff
@@ -435,6 +436,59 @@ router.post(
...
+  auth(),
+  pagination(),
+  validate(T.DocumentsUsersSchema),
+  async (ctx: APIContext<T.DocumentsUsersReq>) => {
+    const { id, query } = ctx.input.body;
+    const actor = ctx.state.auth.user;
+    const { offset, limit } = ctx.state.pagination;
+    const document = await Document.findByPk(id);
+    authorize(actor, "update", document);
```
tommoor, line 449:
```suggestion
    authorize(actor, "read", document);
```

read is enough otherwise viewers won't be able to mention other users in comments

server/routes/api/documents/documents.ts
tommoor: Totally forgot we had this method

app/editor/components/MentionMenu.tsx
apoorv-mishra: Yes!

shared/utils/parseDocumentSlug.ts
apoorv-mishra: Better :+1:

server/routes/api/documents/documents.ts
apoorv-mishra: Okay...How do I test this? Sorry, but it's not clear how to add comments. Ideally it should be an item of the toolbar that pops up upon text selection but I don't see it there. Or is it hidden behind some feature flag for now?

tommoor: Currently viewers cannot add comments on a text selection – they can add comments to the document though through the sidebar, and can reply to existing comments

server/routes/api/documents/documents.ts
```diff
@@ -435,6 +436,59 @@ router.post(
...
+    const document = await Document.findByPk(id);
+    authorize(actor, "read", document);
+
+    let users: User[] = [];
+    let total = 0;
+
+    if (document.collectionId) {
+      const collection = await Collection.findByPk(document.collectionId);
+      authorize(actor, "update", collection);
+      const memberIds = await Collection.membershipUserIds(collection.id);
```
tommoor, line 457:
```suggestion
      const [collection, memberIds] = Promise.all([
        Collection.findByPk(document.collectionId),
        Collection.membershipUserIds(document.collectionId)
      ]);
      authorize(actor, "update", collection);
```

I think it would be worth running these in parallel

server/routes/api/documents/documents.ts
```diff
@@ -435,6 +436,59 @@ router.post(
...
+        User.findAll({ where, offset, limit }),
+        User.count({ where }),
+      ]);
+    }
+
+    ctx.body = {
+      pagination: { ...ctx.state.pagination, total },
+      data: users.map((user) =>
+        presentUser(user, { includeDetails: can(actor, "readDetails", user) })
+      ),
```
tommoor, line 486:
```suggestion
      data: users.map((user) => presentUser(user)),
```

No need for any extra details here

tommoor (approved):

app/stores/UsersStore.ts
```diff
@@ -159,6 +159,25 @@ export default class UsersStore extends BaseStore<User> {
     return res.data;
   };
 
+  @action
+  fetchDocumentUsers = async (params: {
```
tommoor, line 163: All these `fetchBlah` calls are super similar, it would be good to refactor and cleanup how this is done in stores if you care to in a separate PR.

apoorv-mishra: Similar in manner to `fetchDrafts`, `fetchArchived` etc. of `DocumentsStore`?

Issue #5042, opened 2023-03-15
Duplicate Docker images to GitHub Registry


tommoor:
With the announcement that Docker will no longer host images for free as of next month the plan is to move/duplicate these to GitHub's new [registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry).

Incidentally this will offer an opportunity to update the image names, perhaps:

- ghcr.io/outline/community 
- ghcr.io/outline/enterprise

tommoor:
For now I've setup a redirect rule so we're in control of the backing registry and moved the documentation to point to the new url: `docker.getoutline.com/outlinewiki/outline`

PR #5053, opened 2023-03-17
Add pushing of images to GCR

tommoor:
There's currently a duplicate build step here, it's unclear if it's possible to separate `buildx` and the `push`, discussions seem to point to no.

Note: The images still rely on the base image existing on DockerHub, which will have to change next.

related #5042


tommoor:
For now we've decided to proxy the DockerHub registry.

PR #5082, opened 2023-03-22
Deleting a collection should detach associated drafts from it

### LINKED ISSUE #4857: Can't delete/unpublish documents from admin API key
When attempting to delete/unpublish a document via the API, I receive the following error:
`{"ok":false,"error":"authorization_error","status":403,"message":"Authorization error"}`
...however read operations work just fine.

**To Reproduce**
1. Create API token with admin account
2. 
```bash
curl --fail-with-body -s -H 'Accept: application/json' -H 'Content-Type: application/json' -H "Authorization: Bearer ${TOKEN}" -X POST "${ENDPOINT}/api/documents.archive" -d "{ \"id\": \"$DOCUMENT_ID)\"}"
```

**Expected behavior**
A HTTP 200 response.

- Install: Docker, self-hosted
- Version: 0.67.2

If it helps, here's the result of `/api/auth.info`:
```json
{
  "data": {
    "user": {
      "id": "[REDACTED]",
      "name": "James Kerber",
      "avatarUrl": "[REDACTED]",
      "color": "#9E5CF7",
      "isAdmin": true,
      "isSuspended": false,
      "isViewer": false,
      "createdAt": "2022-08-09T14:02:29.993Z",
      "updatedAt": "2023-02-10T14:39:14.230Z",
      "lastActiveAt": "2023-02-10T14:39:14.227Z",
      "email": "[REDACTED]",
      "language": "en_US",
      "preferences": null
    },
    "team": {
      "id": "[REDACTED]",
      "name": "Wiki",
      "avatarUrl": "[REDACTED]",
      "sharing": true,
      "memberCollectionCreate": true,
      "collaborativeEditing": true,
      "defaultCollectionId": null,
      "documentEmbeds": true,
      "guestSignin": false,
      "subdomain": "[REDACTED]",
      "domain": null,
      "url": "[REDACTED]",
      "defaultUserRole": "member",
      "inviteRequired": false,
      "allowedDomains": [],
      "preferences": null
    },
    "availableTeams": [
      {
        "id": "[REDACTED]",
        "name": "Wiki",
        "avatarUrl": "[REDACTED]",
        "url": "[REDACTED]",
        "isSignedIn": true
      }
    ]
  },
  "policies": [
    {
      "id": "[REDACTED]",
      "abilities": {
        "createApiKey": true,
        "createAttachment": true,
        "createAuthenticationProvider": true,
        "createCollection": true,
        "importCollection": true,
        "createDocument": true,
        "createFileOperation": true,
        "createImport": true,
        "createExport": true,
        "createIntegration": true,
        "createNotificationSetting": true,
        "inviteUser": true,
        "read": true,
        "share": true,
        "createTeam": false,
        "update": true,
        "manage": true,
        "createGroup": true,
        "listWebhookSubscription": true,
        "createWebhookSubscription": true
      }
    }
  ],
  "status": 200,
  "ok": true
}
```

apoorv-mishra:
Fixes #4857

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -1,5 +1,5 @@
 import invariant from "invariant";
-import { find } from "lodash";
+import { filter, find } from "lodash";
 import { action, observable } from "mobx";
 import { observer } from "mobx-react";
 import * as React from "react";
@@ -299,7 +299,11 @@ class WebsocketProvider extends React.Component<Props> {
         const collectionId = event.modelId;
         const deletedAt = new Date().toISOString();
 
-        const deletedDocuments = documents.inCollection(collectionId);
+        // drafts eventually are to be detached from collection rather than deleted, hence excluded here
+        const deletedDocuments = filter(
+          documents.inCollection(collectionId),
+          (d) => !!d.publishedAt
+        );
         deletedDocuments.forEach((doc) => {
           doc.deletedAt = deletedAt;
           policies.remove(doc.id);
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -1,5 +1,5 @@
 import invariant from "invariant";
-import { filter, find } from "lodash";
+import { find } from "lodash";
 import { action, observable } from "mobx";
 import { observer } from "mobx-react";
 import * as React from "react";
@@ -175,7 +175,7 @@ class WebsocketProvider extends React.Component<Props> {
                 id: document.collectionId,
               });
 
-              if (!existing) {
+              if (!existing && document.collectionId) {
                 event.collectionIds.push({
                   id: document.collectionId,
                 });
@@ -298,14 +298,14 @@ class WebsocketProvider extends React.Component<Props> {
       action((event: WebsocketEntityDeletedEvent) => {
         const collectionId = event.modelId;
         const deletedAt = new Date().toISOString();
-
-        // drafts eventually are to be detached from collection rather than deleted, hence excluded here
-        const deletedDocuments = filter(
-          documents.inCollection(collectionId),
-          (d) => !!d.publishedAt
-        );
+        const deletedDocuments = documents.inCollection(collectionId);
         deletedDocuments.forEach((doc) => {
-          doc.deletedAt = deletedAt;
+          if (!doc.publishedAt) {
+            // draft is to be detached from collection, not deleted
+            doc.collectionId = null;
+          } else {
+            doc.deletedAt = deletedAt;
+          }
           policies.remove(doc.id);
         });
         documents.removeCollectionDocuments(collectionId);
--- a/app/models/Document.ts
+++ b/app/models/Document.ts
@@ -261,7 +261,7 @@ export default class Document extends ParanoidModel {
   };
 
   @action
-  pin = (collectionId?: string) =>
+  pin = (collectionId?: string | null) =>
     this.store.rootStore.pins.create({
       documentId: this.id,
       ...(collectionId ? { collectionId } : {}),
--- a/app/models/Document.ts
+++ b/app/models/Document.ts
@@ -46,7 +46,7 @@ export default class Document extends ParanoidModel {
 
   @Field
   @observable
-  collectionId: string;
+  collectionId: string | null | undefined;
 
   @Field
   @observable
--- a/app/models/Document.ts
+++ b/app/models/Document.ts
@@ -46,7 +46,7 @@ export default class Document extends ParanoidModel {
 
   @Field
   @observable
-  collectionId: string | null | undefined;
+  collectionId?: string | null;
 
   @Field
   @observable
--- a/app/stores/BaseStore.ts
+++ b/app/stores/BaseStore.ts
@@ -112,7 +112,10 @@ export default abstract class BaseStore<T extends BaseModel> {
     return this.update(params, rest);
   }
 
-  get(id: string): T | undefined {
+  get(id?: string | null): T | undefined {
+    if (!id) {
+      return;
+    }
     return this.data.get(id);
   }
 
--- a/app/stores/BaseStore.ts
+++ b/app/stores/BaseStore.ts
@@ -112,10 +112,7 @@ export default abstract class BaseStore<T extends BaseModel> {
     return this.update(params, rest);
   }
 
-  get(id?: string | null): T | undefined {
-    if (!id) {
-      return;
-    }
+  get(id: string): T | undefined {
     return this.data.get(id);
   }
 
--- a/app/stores/DocumentsStore.ts
+++ b/app/stores/DocumentsStore.ts
@@ -645,7 +645,11 @@ export default class DocumentsStore extends BaseStore<Document> {
 
   @action
   removeCollectionDocuments(collectionId: string) {
-    const documents = this.inCollection(collectionId);
+    // drafts are to be detached from collection rather than deleted, hence excluded here
+    const documents = filter(
+      this.inCollection(collectionId),
+      (d) => !!d.publishedAt
+    );
     const documentIds = documents.map((doc) => doc.id);
     documentIds.forEach((id) => this.remove(id));
   }
--- a/app/stores/DocumentsStore.ts
+++ b/app/stores/DocumentsStore.ts
@@ -799,6 +799,6 @@ export default class DocumentsStore extends BaseStore<Document> {
     find(this.orderedData, (doc) => url.endsWith(doc.urlId));
 
   getCollectionForDocument(document: Document) {
-    return this.rootStore.collections.data.get(document.collectionId);
+    return this.rootStore.collections.get(document.collectionId);
   }
 }
--- a/app/stores/DocumentsStore.ts
+++ b/app/stores/DocumentsStore.ts
@@ -799,6 +799,8 @@ export default class DocumentsStore extends BaseStore<Document> {
     find(this.orderedData, (doc) => url.endsWith(doc.urlId));
 
   getCollectionForDocument(document: Document) {
-    return this.rootStore.collections.get(document.collectionId);
+    return document.collectionId
+      ? this.rootStore.collections.get(document.collectionId)
+      : undefined;
   }
 }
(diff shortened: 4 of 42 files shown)


tommoor:
I'm going to argue that although this might be "technically" correct it will actually result in a lot of confused a frustrated users. Drafts today are disconnected enough from the collection they are in that you don't really expect them to be affected by actions like this, also other users may have items in their drafts that will be wiped away and you would not know that.

Now that we have the concept of unpublished documents without a collection we should revert the drafts to this state. Ideally this would be done in an async Processor –You'd have a processor that listens for the `collections.delete` event and moves any associated drafts.

apoorv-mishra:
Makes sense. Adding to it but unrelated to the issue, I think we can introduce one more menu item for document menu, something like "Move to drafts"?

tommoor:
I wouldn't add a new menu item for it, but "Drafts" should be a location you can move to inside the existing "Move" dialog

app/components/WebsocketProvider.tsx
```diff
@@ -299,7 +299,11 @@ class WebsocketProvider extends React.Component<Props> {
         const collectionId = event.modelId;
         const deletedAt = new Date().toISOString();
 
-        const deletedDocuments = documents.inCollection(collectionId);
+        // drafts eventually are to be detached from collection rather than deleted, hence excluded here
+        const deletedDocuments = filter(
```
tommoor, line 303: Instead of filtering perhaps we should be clearing the `collectionId` in the `deletedDocuments.forEach` below? Won't drafts be left in the collection right now on clients?

apoorv-mishra: Technically, yes but it doesn't show up visually. Regardless, you're correct, it's better to clear `collectionId` for drafts here.

app/stores/DocumentsStore.ts
```diff
@@ -660,7 +660,11 @@ export default class DocumentsStore extends BaseStore<Document> {
 
   @action
   removeCollectionDocuments(collectionId: string) {
-    const documents = this.inCollection(collectionId);
+    // drafts are to be detached from collection rather than deleted, hence excluded here
+    const documents = filter(
+      this.inCollection(collectionId),
+      (d) => !!d.publishedAt
+    );
```
apoorv-mishra, line 667: This retains drafts in two other cases, `entities` and `collections.remove_user` of `WebsocketProvider`. Hope that's fine :crossed_fingers:

server/routes/api/documents/documents.test.ts
```diff
@@ -2955,6 +2955,35 @@ describe("#documents.delete", () => {
     expect(deletedDoc?.deletedAt).toBeTruthy();
   });
 
+  it("should delete a draft under deleted collection", async () => {
```
tommoor, line 2958:
```suggestion
  it("should not delete draft under deleted collection", async () => {
```

server/queues/tasks/DetachDraftsFromCollectionTask.ts
```diff
@@ -0,0 +1,32 @@
...
+  async perform(task: Task) {
+    const collection = await Collection.findByPk(task.collectionId, {
+      paranoid: false,
+    });
+
+    if (!collection || !collection.deletedAt) {
+      return;
+    }
+
+    await Document.update(
```
tommoor, line 19: I think we should query draft documents in the collection and loop over them calling `documentUpdater` here or the correct events / websocket messages won't be triggered.

app/stores/BaseStore.ts
```diff
@@ -112,7 +112,10 @@ export default abstract class BaseStore<T extends BaseModel> {
     return this.update(params, rest);
   }
 
-  get(id: string): T | undefined {
+  get(id?: string | null): T | undefined {
```
tommoor, line 115: Not loving this change…

server/routes/api/documents/documents.test.ts
apoorv-mishra: This is for when attempt is made to delete the draft document, not collection.

app/stores/BaseStore.ts
apoorv-mishra: Agree, but the alternative is making changes at ten different places checking for `id` before calling this `get` :grimacing:

server/queues/tasks/DetachDraftsFromCollectionTask.ts
apoorv-mishra: Sadly, neither `documentUpdater` nor `documentMover` possess the capability to _update `collectionId` to `null`_ so far :man_facepalming:

apoorv-mishra: Appropriate to bake this into `documentMover` since we're _moving out of collection_?

tommoor: Oh yea, that sounds good? We should generally have the ability to move back to drafts

server/models/Document.ts
```diff
@@ -603,17 +616,22 @@ class Document extends ParanoidModel {
     this.createdById = userId;
     this.lastModifiedById = userId;
     this.publishedAt = null;
+    // Q: should it be detached from collection?
+    this.collectionId = null;
```
apoorv-mishra, line 620: I hope this is fine?(TODO: add tests if approved)

server/models/Document.ts
```diff
@@ -555,15 +561,19 @@ class Document extends ParanoidModel {
     return this.save(options);
   };
 
-  publish = async (userId: string, { transaction }: SaveOptions<Document>) => {
+  publish = async (
+    userId: string,
+    collectionId: string,
```
apoorv-mishra, line 566: Hope this makes more sense semantically from a functional perspective; a document is _supposed to be published in a collection_.

app/stores/UiStore.ts
```diff
@@ -37,7 +37,7 @@ class UiStore {
   activeDocumentId: string | undefined;
 
   @observable
-  activeCollectionId: string | undefined;
+  activeCollectionId: string | null | undefined;
```
tommoor, line 40:
```suggestion
  activeCollectionId?: string | null;
```

app/models/Document.ts
```diff
@@ -46,7 +46,7 @@ export default class Document extends ParanoidModel {
 
   @Field
   @observable
-  collectionId: string;
+  collectionId: string | null | undefined;
```
tommoor, line 49:
```suggestion
  collectionId?: string | null;
```

server/commands/documentMover.ts
```diff
@@ -5,12 +5,19 @@ import { User, Document, Collection, Pin, Event } from "@server/models";
...
+  /** Document which is being moved */
   document: Document;
-  collectionId: string;
+  /** Destination collection to which the document is moved */
+  collectionId: string | null;
+  /** ID of parent under which the document is moved */
   parentDocumentId?: string | null;
+  /** Position of moved document within document structure */
   index?: number;
+  /** The IP address of the user creating the document */
```
tommoor, line 18:
```suggestion
  /** The IP address of the user moving the document */
```

server/models/helpers/NotificationHelper.ts
```diff
@@ -164,12 +164,13 @@ export default class NotificationHelper {
...
       // Check the recipient has access to the collection this document is in. Just
-      // because they are subscribed doesn't meant they "still have access to read
+      // because they are subscribed doesn't mean they still have access to read
       // the document.
       if (
         recipient.email &&
         !recipient.isSuspended &&
-        collectionIds.includes(document.collectionId)
+        (!document.collectionId ||
+          collectionIds.includes(document.collectionId))
```
tommoor, line 173:
```suggestion
        document.collectionId && collectionIds.includes(document.collectionId))
```

server/models/Document.ts
tommoor: Kind of, but I think you've now introduced a bug where you can publish into a different collection that the `document.collectionId` contains?

You need to check that `document.collectionId` is set inside the publish method or set it to the collectionId passed as an argument

server/models/Document.ts
```diff
@@ -399,7 +405,7 @@ class Document extends ParanoidModel {
 
   @ForeignKey(() => Collection)
   @Column(DataType.UUID)
-  collectionId: string;
+  collectionId: string | null | undefined;
```
tommoor, line 408:
```suggestion
  collectionId?: string | null;
```

server/queues/tasks/DetachDraftsFromCollectionTask.ts
```diff
@@ -0,0 +1,48 @@
...
+        paranoid: false,
+      }),
+      User.findByPk(task.actorId),
+    ]);
+
+    if (!actor || !collection || !collection.deletedAt) {
+      return;
+    }
+
+    const documents = await Document.findAll({
```
tommoor, line 26:
```suggestion
    const documents = await Document.scope("withDrafts").findAll({
```

I believe you need to use the scope here or it will never find anything, drafts are filtered out of find operations by default.

server/models/Document.ts
tommoor: Is it neccessary?

tommoor (changes requested):

server/models/helpers/NotificationHelper.ts
apoorv-mishra: Should we limit it to the creator then in case of draft without collection?

server/models/Document.ts
apoorv-mishra: No, not really in the current scope.

server/queues/tasks/DetachDraftsFromCollectionTask.ts
apoorv-mishra: Gets overridden? [Here](https://github.com/outline/outline/blob/main/server/routes/api/documents/documents.ts#L356-L358) is another occurrence.

apoorv-mishra: Added scope nonetheless. No harm I suppose.

server/models/helpers/NotificationHelper.ts
tommoor: For now it's not possible to get a notification in a draft without collection, but in the future this will need updating when we have individually shared documents _soon_

tommoor (approved):

PR #5102, opened 2023-03-27
fix: Call the OIDC logout endpoint when logging out.

kvicrey-ls:
Draft for now.

This fix calls the OIDS logout endpoint to log the user out of the auth provider at the same time of the outline session.

Seeking for feedback. What I see missing:

- [ ] `${authentications[0].oidcTokenId` => not ideal. Can there be a case where a user has multiple authentication methods?
- [ ] Couple of tests

### DIFF
--- a/server/migrations/20230325030839-add-token-id-to-user-authentications.js
+++ b/server/migrations/20230325030839-add-token-id-to-user-authentications.js
@@ -1,0 +1,14 @@
+"use strict";
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn("user_authentications", "oidcTokenId", {
+      type: Sequelize.BLOB,
+      allowNull: true,
+    });
+  },
+
+  async down(queryInterface) {
+    await queryInterface.removeColumn("user_authentications", "oidcTokenId");
+  },
+};
--- a/server/models/UserAuthentication.ts
+++ b/server/models/UserAuthentication.ts
@@ -50,6 +50,9 @@ class UserAuthentication extends IdModel {
   @Column
   providerId: string;
 
+  @Column
+  oidcTokenId: string;
+
   @Column(DataType.DATE)
   expiresAt: Date;
 
--- a/server/routes/api/auth.ts
+++ b/server/routes/api/auth.ts
@@ -168,8 +168,13 @@ router.post("auth.delete", auth(), transaction(), async (ctx: APIContext) => {
     }
   );
 
+  const authentications = await user.$get("authentications");
+
   ctx.body = {
     success: true,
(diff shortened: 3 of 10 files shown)


server/routes/api/auth.ts
```diff
@@ -168,8 +168,13 @@ router.post("auth.delete", auth(), transaction(), async (ctx: APIContext) => {
     }
   );
 
+  const authentications = await user.$get("authentications");
+
   ctx.body = {
     success: true,
+    oidcEndSessionUri: authentications[0]?.oidcTokenId
```
tommoor, line 175: I'd probably make this more generic, eg `postLogoutUrl`

server/routes/api/auth.ts
```diff
@@ -168,8 +168,13 @@ router.post("auth.delete", auth(), transaction(), async (ctx: APIContext) => {
     }
   );
 
+  const authentications = await user.$get("authentications");
```
tommoor, line 171: Yea, you're right – it would be better to find any OIDC authentication specifically rather than just use the first.

server/models/UserAuthentication.ts
```diff
@@ -50,6 +50,9 @@ class UserAuthentication extends IdModel {
   @Column
   providerId: string;
 
+  @Column
+  oidcTokenId: string;
```
tommoor, line 54: I believe this column would need to be encrypted in the database? like refresh and access token

server/migrations/20230325030839-add-token-id-to-user-authentications.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn("user_authentications", "oidcTokenId", {
```
tommoor, line 5: It's a shame to have another column just for this, but hard to see a way to avoid it.

Anomen:
Will be working on improving this PR this week 👍 
Thanks for the comments!

tommoor:
Thanks for the help. If you're interested in improving OIDC further in the future we should really change to using something like OIDC_DISCOVERY_URL instead of providing all the endpoints separately.

Anomen:
Hey @tommoor, I seem to be getting these errors in local:

image

I thought it was because of my changes but I don't think so, I seem to be getting the same error from master. Do you get the same thing?

server/routes/api/auth.ts
Anomen: Can there be several OIDC authentication for one user? or it's not a use case that can happen?

tommoor: Unlikely but technically possible if we can account for it

tommoor:
> Do you get the same thing?

unrelated frontend warnings 👍

PR #5130, opened 2023-03-30
Additional policy regarding comments on a document

### LINKED ISSUE #5065: Additional policy for commenting on a document
We'd want to address the scenario where some viewers are _not allowed_ to comment. Currently, it's not possible since [all viewers can somehow comment](https://github.com/outline/outline/blob/68640860fb9cf584f155348ae06e4536226bef16/server/routes/api/comments/comments.ts#L32). Or maybe disallow commenting altogether on a document or collection.

apoorv-mishra:
Closes #5065 

Although for now it's just an alias of the `read` permission but can be later modified to support tighter constraints enforced by admins specifically for _commenting_ - allow/disallow a `group` of users, allow/disallow commenting altogether on a document etc.

### DIFF
--- a/server/policies/document.ts
+++ b/server/policies/document.ts
@@ -10,20 +10,7 @@ allow(User, "createDocument", Team, (user, team) => {
   return true;
 });
 
-allow(User, "read", Document, (user, document) => {
-  if (!document) {
-    return false;
-  }
-
-  // existence of collection option is not required here to account for share tokens
-  if (document.collection && cannot(user, "read", document.collection)) {
-    return false;
-  }
-
-  return user.teamId === document.teamId;
-});
-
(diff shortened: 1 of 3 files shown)


server/policies/document.ts
```diff
@@ -23,6 +23,19 @@ allow(User, "read", Document, (user, document) => {
   return user.teamId === document.teamId;
 });
 
+allow(User, "comment", Document, (user, document) => {
```
tommoor, line 26: Lets just add it to "read" while it's the same…

```
allow(User, ["read", "comment"], Document, (user, document) => {
```

tommoor (approved):
It would be nice to have a follow up PR to use this policy to hide the commenting functionality in the frontend

PR #5137, opened 2023-04-01
Add iframe support

Anomen:
Add the command `/iframe` in the editor. In this example, I add an iframe in the page (which is the homepage of outline itself. Sorry if it's confusing :-D ).

Kapture 2023-03-31 at 21 51 46

Related discussion: https://github.com/outline/outline/discussions/2286


tommoor:
Hi @Anomen – while it appears this simple at first this solution doesn't work for a couple of reasons:

- The regex matches every url which in-turn breaks all of the other embed integrations.
- It also means pasting any url into the editor turns it into an iframe

For this case, we probably need to use metadata hidden in the link title to trigger the embed rather than the url regex to allow this to be backwards compatible

EarMaster:
Is the only thing stopping this the regex? Because that seems an easy fix for a great feature. Just remove the match all URL regex…

ThinkSalat:
Looks great. Would be nice to be able to expand/collapse or resize. This could be utilized for transclusion

tommoor:
> Is the only thing stopping this the regex? Because that seems an easy fix for a great feature. Just remove the match all URL regex…

That's not how this works, the regex is used to decide which links should be displayed as different embed types. Ultimately this feature is probably blocked by #3000

EarMaster:
If I understand this correctly switching to the JSON format would also allow the implementation of adjustable iFrames (change size, maybe other things?), right?

DVAlexHiggs:
+1 on this PR, this would massively change how our knowledgebase integrates into other material in our organisation. Major convenience

tommoor:
This will be added once support is added through closing #3000 – as it stands this PR is not mergable

airbenich:
Whats the current status of IFrame embedding? As @tommoor said it will be added once #3000 is closed. It was closed in end of may. Is now a development of the iframe embed possible?

ThinkSalat:
Looks like #7319 is going to be merged soon! @Blendman974, will this support embedding notes/bullets themselves as well? Thinking of how logseq handles transclusion/embedding

PR #5141, opened 2023-04-01
Autorun migrations in community edition

### LINKED ISSUE #4680: Run migrations in community edition automatically
Will land this in v0.68.0, the expectation will be that any backups are done before updating the software. Required migrations will run automatically.

apoorv-mishra:
Closes #4680

### DIFF
--- a/server/utils/startup.ts
+++ b/server/utils/startup.ts
@@ -27,11 +27,24 @@ export function checkPendingMigrations() {
           .map((line, i) => `${i + 1}. ${line.replace("down ", "")}`)
           .join("\n")
       );
-      Logger.warn(
-        "Please run `yarn db:migrate` or `yarn db:migrate --env production-ssl-disabled` to run all pending migrations"
-      );
+      if (env.isCloudHosted()) {
+        Logger.warn(
+          "Please run `yarn db:migrate` or `yarn db:migrate --env production-ssl-disabled` to run all pending migrations"
+        );
 
-      process.exit(1);
+        process.exit(1);
+      } else {
+        // run migrations automatically for community edition
+        Logger.info("database", "Running migrations...");
+        const cmdResult = execSync(
+          `yarn db:migrate${
+            env.PGSSLMODE === "disable" ? " --env=production-ssl-disabled" : ""
+          }`
+        );
+        const cmdOutput = Buffer.from(cmdResult).toString("utf-8");
+        Logger.info("database", cmdOutput);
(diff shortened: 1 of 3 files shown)


server/utils/startup.ts
```diff
@@ -20,32 +21,74 @@ export function checkPendingMigrations() {
...
+function logMigrations(migrations: string[]) {
+  Logger.warn("You have pending migrations");
+  Logger.warn(
+    migrations
+      .map((line, i) => `${i + 1}. ${line.replace("down ", "")}`)
+      .join("\n")
+  );
+  Logger.warn(
+    "Please run `yarn db:migrate` or `yarn db:migrate --env production-ssl-disabled` to run all pending migrations"
+  );
```
tommoor, line 75:
```suggestion
```

It's ok, I don't need this and this is only a safety measure for cloud now.

server/utils/startup.ts
```diff
@@ -20,32 +21,74 @@ export function checkPendingMigrations() {
...
+    const cmdResult = execSync(
+      `yarn db:migrate${
+        env.PGSSLMODE === "disable" ? " --env=production-ssl-disabled" : ""
+      }`
+    );
+    const cmdOutput = Buffer.from(cmdResult).toString("utf-8");
+    Logger.info("database", cmdOutput);
+    Logger.info("database", "Done.");
   } catch (err) {
+    Logger.warn(chalk.red("Failed to run migrations!"));
```
tommoor, line 52:
```suggestion
```

This is resulting in a double error, seems fine without?
image

tommoor (approved):

tommoor:
Oh, I meant to add we should remove the `db:migrate` call from within the `Makefile` as it'll be run twice now.

server/utils/startup.ts
apoorv-mishra: I believe removing `db:migrate` from `Makefile` resolves this. No further action needed?

tommoor: I don't think it changes it to be honest, I'm easy either way here

apoorv-mishra: The first error is logged from within the `sequelize` cli :see_no_evil: 

Catching errors centrally within `checkMigrations` now.

tommoor (approved):

patzm:
awesome feature!!! Thanks @apoorv-mishra, this has always been a pain for me. In particular in a `docker-compose` stack, in which the _just run a command_ thinking isn't really native.

PR #5143, opened 2023-04-02
Toggle visibility of comment UI based on policy

### LINKED ISSUE #5142: Toggle comment functionality based on policy
[Follow up](https://github.com/outline/outline/pull/5130#pullrequestreview-1367872813) of #5130

apoorv-mishra:
Closes #5142

### DIFF
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -719,6 +719,7 @@ export const openDocumentComments = createAction({
     return (
       !!activeDocumentId &&
       can.read &&
+      can.comment &&
       !can.restore &&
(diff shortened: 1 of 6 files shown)


tommoor:
We want to retain the ability to view comments with this policy, just not create comments. So the only UI that should be hidden is:

- The reply input on comment threads
- The comment input at the bottom of the sidebar
- The button in the floating editor toolbar

tommoor (changes requested):
Requesting changes properly so you can request re-review.

tommoor:
There's no changes in this PR now 😆

apoorv-mishra:
Paused! Working on a different issue for now.

PR #5151, opened 2023-04-04
Notifications refactor (#5151

### LINKED ISSUE #4924: Refactor emails to be sent based on notification model
We're currently scheduling emails and creating notification models side by side, this doesn't make sense – the email should be scheduled from a processor that reacts to the notifications.create event.

tommoor:
Refactoring the notifications pipeline to allow for in-app notifications. This is a fair amount of changes, mainly moving code around – now the logic is:

- `NotificationsProcessor` listens for events that should have notifications and enqueues tasks to calculate them
- `XNotificationsTask`'s calculate which users need a notification for each specific event based on team, subscriptions, mentioned etc and create `Notification` rows
- `EmailsProcessor` listens for `notifications.create` event from new notification rows and enqueues one more task to email a version of the notification where applicable. (In the future this will be more advanced, right now we just email most things)

---

Notably, calculating the HTML content for emails has moved to the email presenter classes – previously it was in each `NotificationsTask`, now these just create database rows and nothing else. 

closes #4924


server/queues/processors/EmailsProcessor.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    const notificationId = notification.id;
+
+    switch (notification.event) {
+      case NotificationEventType.UpdateDocument:
+      case NotificationEventType.PublishDocument: {
+        await new DocumentPublishedOrUpdatedEmail(
+          {
+            to: notification.user.email,
```
tommoor, line 31: Going to let this sit for a day or so, but I think one last refactor is needed here so that the Email presenters just take the `notification` record and don't need it split out like this…

PR #5167, opened 2023-04-08
Revert "Upgrade MermaidJS (#5043" (#5167

tommoor:
Reverts outline/outline#5043 – caused a Vite reload loop in local dev, unclear at this time why.


Awoolyx:
@tommoor, is it possible that the Vite reload loop is caused by the move from CJS to ESM that is breaking something? CJS was dropped entirely (highlighted in the v10.0.0 release notes). There are a few other breaking changes in v10, not sure if you are aware. Check the v10.0.0 release notes.

I can investigate if you'd like, as the current version 9.3.0 is missing e.g. timeline functionality.

tommoor:
> is it possible that the Vite reload loop is caused by the move from CJS to ESM

It's possible, I probably won't have a chance to investigate for a while

PR #5193, opened 2023-04-12
OICD Profile name is not mandatory anymore

jrevillard:
Related to https://github.com/outline/outline/discussions/4766


tommoor:
Thanks, kept the check but moved it in 821c9368f

PR #5208, opened 2023-04-16
Keep nested docs in shared sidebar collapsed by default

### LINKED ISSUE #5123: Collapse nested docs in shared sidebar by default
This isn't a big enough feature request to warrant opening an issue, so I'll put it here and see what people think.

One thing I've noticed about moving my blog to outline is the [sidebar gets a bit unwieldy](https://blog.gurucomputing.com.au/s/blog). It would be nice if the default (or configurable) behavour allowed for the sidebar to be collapsed by default, stopping an information overload by having it all expanded.

_Originally posted by @routerino in https://github.com/outline/outline/discussions/5076_

apoorv-mishra:
Closes #5123

### DIFF
--- a/app/components/Sidebar/components/SharedDocumentLink.tsx
+++ b/app/components/Sidebar/components/SharedDocumentLink.tsx
@@ -1,3 +1,4 @@
+import { includes } from "lodash";
 import { observer } from "mobx-react";
 import * as React from "react";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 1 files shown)


tommoor:
I think it probably goes too far, if you look at the existing guide as an example:

image

This is a lot to display at once, but this is also not very nice:

image

Perhaps the right thing is to open the first level and close everything else?

image

almereyda:
Showing only the first-level of nesting looks also much cleaner to me.

Does the UI remember which lower levels were expanded, in case of a page reload or browsing to another page and back here?

apoorv-mishra:
> Does the UI remember which lower levels were expanded, in case of a page reload or browsing to another page and back here?

No, it doesn't _remember_ the expanded levels across multiple sessions unless you _click_ a particular document in sidebar.

app/components/Sidebar/components/SharedDocumentLink.tsx
```diff
@@ -55,12 +66,6 @@ function DocumentLink(
     }
   }, [showChildren]);
 
-  React.useEffect(() => {
```
tommoor, line 58: Still need this otherwise we lose the automatic opening behavior as you click around?

tommoor: Ping on this question

apoorv-mishra: Sorry, don't get this? What is "automatic opening"? Expansion of document node in sidebar?

tommoor: Yes, the expansion as a document becomes active

tommoor (approved):

Issue #5166, opened 2023-04-08
In-app notifications


tommoor:
With the introduction of comments and the completion of https://github.com/outline/outline/pull/5151 we're now well placed to add a notifications panel in-app to quickly jump to recent replies and activity.

tommoor:
closed by #5354

PR #5220, opened 2023-04-18
Notifications endpoints

apoorv-mishra:
Towards #5166 

- [x] Presenters
- [x] Policies
- [x] `notifications.list`
- [x] Add indexes against `createdAt` and `event` columns
- [x] Add partial index against `viewedAt`(`where viewedAt is null`)
- [x] Create column `archivedAt` and index it
- [x] Modify `notifications.list` for filtering archived notifications
- [x] `notifications.update`
- [x] `notifications.updateAll`


server/presenters/notification.ts
```diff
@@ -0,0 +1,20 @@
...
+import presentUser from "./user";
+
+export default function presentNotification(notification: Notification) {
+  return {
+    id: notification.id,
+    viewedAt: notification.viewedAt,
+    createdAt: notification.createdAt,
+    event: notification.event,
+    userId: notification.userId,
+    user: presentUser(notification.user),
```
tommoor, line 11: Probably not too useful as notifications will always be for yourself 😉

server/routes/api/notifications/notifications.ts
```diff
@@ -49,4 +56,84 @@ router.post(
...
+        order: [["createdAt", "DESC"]],
+        offset: ctx.state.pagination.offset,
+        limit: ctx.state.pagination.limit,
+      }),
+      Notification.count({
+        where: {
+          ...where,
+          viewedAt: {
+            [Op.is]: null,
+          },
```
apoorv-mishra, line 95: Note that there's a subtle assumption here. The count actually changes based on other `where` attributes. For example, if `archived` is received as `true`, then this count consists of notifications which are both archived _and_ unseen. On the contrary, the intended purpose may be to return the count of _all_ unseen notifications without extra filters. Let me know your thoughts @tommoor

server/routes/api/notifications/notifications.ts
```diff
@@ -49,4 +56,84 @@ router.post(
...
+      where = { ...where, event: eventType };
+    }
+    if (archived) {
+      where = {
+        ...where,
+        archivedAt: {
+          [Op.ne]: null,
+        },
+      };
+    }
```
apoorv-mishra, line 82: > Note that there's a subtle assumption here. The count actually changes based on other where attributes. For example, if archived is received as true, then this count consists of notifications which are both archived and unseen. On the contrary, the intended purpose may be to return the count of all unseen notifications without extra filters. Let me know your thoughts @tommoor

Similar assumption holds here too.

server/presenters/notification.ts
```diff
@@ -0,0 +1,20 @@
...
+    createdAt: notification.createdAt,
+    event: notification.event,
+    userId: notification.userId,
+    actorId: notification.actorId,
+    actor: notification.actor ? presentUser(notification.actor) : undefined,
+    commentId: notification.commentId,
+    documentId: notification.documentId,
+    revisionId: notification.revisionId,
+    collectionId: notification.collectionId,
+    teamId: notification.teamId,
```
tommoor, line 18:
```suggestion
```

server/models/Notification.ts
```diff
@@ -130,7 +134,7 @@ class Notification extends Model {
   @Column(DataType.UUID)
   teamId: string;
 
-  @AfterSave
+  @AfterCreate
```
tommoor, line 137: Why change? This was purposeful to ensure that we don't have events go through the system that haven't been saved to the db.

server/routes/api/notifications/notifications.ts
```diff
@@ -49,4 +56,124 @@ router.post(
...
+  }
+);
+
+router.post(
+  "notifications.update",
+  auth(),
+  validate(T.NotificationsUpdateSchema),
+  transaction(),
+  async (ctx: APIContext<T.NotificationsUpdateReq>) => {
+    const { id, markAsViewed, archive } = ctx.input.body;
```
tommoor, line 118: It may be easier and more forward-compatible to just allow the keys as they exist on the notification model in the request. In this case `viewedAt` and `archivedAt` – the frontend has to pass `new Date()` instead of a boolean but we get a lot in return.

server/routes/api/notifications/notifications.ts
```diff
@@ -49,4 +56,124 @@ router.post(
...
+    };
+  }
+);
+
+router.post(
+  "notifications.update_all",
+  auth(),
+  validate(T.NotificationsUpdateAllSchema),
+  async (ctx: APIContext<T.NotificationsUpdateAllReq>) => {
+    const { markAsViewed, archive } = ctx.input.body;
```
tommoor, line 144: Similar comment as the non-bulk endpoint

server/routes/api/notifications/notifications.ts
tommoor: The count changing on the where is okay, but I wouldn't call it `unseenCount` in that case, just call it `total` and add to the pagination object like we do for users…

```
pagination: { ...ctx.state.pagination, total },
```

server/models/Notification.ts
apoorv-mishra: The thinking was that `@AfterSave` also runs on _update_ besides _create_. Going forward _update_ is supposed to be handled by `notificationUpdater`, so keeping `@AfterSave` here would've resulted in a bug?

tommoor: Ahh, you're totally right

server/commands/notificationUpdater.ts
```diff
@@ -22,16 +23,16 @@ type Props = {
...
-  markAsViewed,
-  archive,
+  viewedAt,
+  archivedAt,
   ip,
   transaction,
 }: Props): Promise<Notification> {
-  if (markAsViewed) {
-    notification.viewedAt = new Date();
+  if (viewedAt) {
```
tommoor, line 31: I think this will need the same `isUndefined` logic to enable marking as unviewed?

server/routes/api/notifications/notifications.ts
```diff
@@ -49,4 +57,121 @@ router.post(
...
 
+router.post(
+  "notifications.list",
+  auth(),
+  pagination(),
+  validate(T.NotificationsListSchema),
+  transaction(),
+  async (ctx: APIContext<T.NotificationsListReq>) => {
+    const eventType = ctx.input.body.eventType ?? ctx.input.query.eventType;
+    const archived = ctx.input.body.archived ?? ctx.input.query.archived;
```
tommoor, line 68:
```suggestion
    const { eventType, archived } = ctx.input.body;
```

I don't think we need to support query params here, it's a POST endpoint

server/routes/api/notifications/notifications.ts
```diff
@@ -49,4 +57,121 @@ router.post(
...
+        offset: ctx.state.pagination.offset,
+        limit: ctx.state.pagination.limit,
+      }),
+      Notification.count({
+        where: {
+          ...where,
+          viewedAt: {
+            [Op.is]: null,
+          },
+        },
```
tommoor, line 97:
```suggestion
        where,
```

Unfortunately I suppose we need two count operations here, the count for total which should reflect the total for the filtered list for pagination and the unviewed total separately?

server/commands/notificationUpdater.ts
apoorv-mishra: If we want to allow that, yes. I assumed it's not supposed to be reverted.

tommoor: Yes, it should be revertable as you can mark notifications as unread like an email inbox 👍

server/routes/api/notifications/notifications.ts
apoorv-mishra: Sure.

server/commands/notificationUpdater.ts
apoorv-mishra: Enabling the same in bulk update as well.

tommoor:
Just an update - I’m building the UI off this branch so may have some requests as the functionality gets built out

apoorv-mishra:
Sure :+1:

tommoor:
Bundled into #5361

Issue #835, opened 2018-12-31
Add collection maintainers / owners


tommoor:
It would be great to be able to mark people as "maintainers" or "owners" of a collection so that there is some knowledge ownership.

Depending on internal culture these people may then be responsible for managing articles, creating new ones, or go-to folks for questions.

tommoor:
Depends on https://github.com/outline/outline/issues/989

matbrgz:
Hey, consider to implement document owner or owners.

UI Example
![captured (5)](https://user-images.githubusercontent.com/23284276/112705710-be9e4900-8e7e-11eb-8c79-d91ab97f423e.gif)

matbrgz:
up

rlljorge:
+1

PR #5273, opened 2023-04-28
feat: Collection admins (#5273

### LINKED ISSUE #3454: Sometimes private collections are visible to admins
I've experienced this issue twice: a non admin user creates a collection and removes read and write access to all (making the collection private). Now, me, as an admin user, logged into our self deployed Outline and was able to see this user's private collection and documents. After refreshing the page, the collection was no longer visible to me. 

**To Reproduce**
I don't know yet... We've only experienced this issue twice in the last month without any clear reproduction steps. I will keep a look out in our dev environment to try to catch it in the logs. 

**Expected behavior**
Private documents should not be visible to anyone, including admin users.

- Install: self hosted
- Version: v0.63.0

 - OS: Mac OS
 - Browser Chrome
 - Version 100.0.4896.127 (Official Build) (arm64)

tommoor:
This PR separates the permission for creating and editing documents in a collection from the ability to manage the collection (manage users/groups, change title, change sort order etc). As part of this it also separates viewing the collection's details from it's contents… this has a couple of really nice benefits:

- Give users access to write documents without the ability to change collection permissions
- We no longer download the structure of all collections as part of bootstrapping the app when opening a new tab, making loading the app sidebar much faster.
- Similarly this data is no longer downloaded when collection updates are made such as editing the title
- This allows for a future admin UI that can list all collections in the instance without also exposing their content, admins can manage permissions and add themselves as a member to view content.

A follow up here will be to create UI for admins to view and manage collections in settings.

closes #3454
related #835


server/policies/collection.ts
```diff
@@ -72,89 +75,76 @@ allow(User, "share", Collection, (user, collection) => {
...
+    return includesMembership(collection, Object.values(CollectionPermission));
   }
 
   return true;
 });
 
-allow(User, "delete", Collection, (user, collection) => {
+allow(
+  User,
+  ["updateDocument", "createDocument", "deleteDocument"],
```
tommoor, line 106: Note: All these new policies, separated from "read", "update" of collection itself.

server/routes/api/collections.ts
```diff
@@ -148,6 +148,21 @@ router.post("collections.info", auth(), async (ctx: APIContext) => {
   };
 });
 
+router.post("collections.documents", auth(), async (ctx: APIContext) => {
```
tommoor, line 151: New endpoint for getting the collection structure, it will no longer be returned from `collections.info` and requires the "readDocument" permission which admins do not have by default.

server/routes/api/documents/schema.ts
```diff
@@ -8,7 +8,9 @@ const DocumentsSortParamsSchema = z.object({
   /** Specifies the attributes by which documents will be sorted in the list */
   sort: z
     .string()
-    .refine((val) => ["createdAt", "updatedAt", "index", "title"].includes(val))
+    .refine((val) =>
+      ["createdAt", "updatedAt", "publishedAt", "index", "title"].includes(val)
```
tommoor, line 12: Unrelated bug I found, `publishedAt` is a valid sort option that was missing

app/components/WebsocketProvider.tsx
```diff
@@ -293,6 +293,10 @@ class WebsocketProvider extends React.Component<Props> {
       collections.add(event);
     });
 
+    this.socket.on("collections.update", (event: PartialWithId<Collection>) => {
```
tommoor, line 296: `collections.update` now includes the entire model

app/components/DocumentsLoader.tsx
```diff
@@ -0,0 +1,21 @@
...
+import * as React from "react";
+import Collection from "~/models/Collection";
+
+type Props = {
+  enabled: boolean;
+  collection: Collection;
+  children: React.ReactNode;
+};
+
+function DocumentsLoader({ collection, enabled, children }: Props) {
```
tommoor, line 11: Move to suspense if we ever get to React 18 🤷

app/models/Collection.ts
```diff
@@ -57,32 +57,34 @@ export default class Collection extends ParanoidModel {
...
+    }
+
     return (
       this.documents.length === 0 &&
       this.store.rootStore.documents.inCollection(this.id).length === 0
     );
   }
 
-  @computed
-  get documentIds(): string[] {
```
tommoor, line 75: No longer needed

server/models/Collection.ts
```diff
@@ -254,21 +254,17 @@ class Collection extends ParanoidModel {
...
-        defaults: {
-          permission: CollectionPermission.ReadWrite,
-          createdById: model.createdById,
-        },
-        transaction: options.transaction,
-      });
-    }
-
-    return undefined;
+    return CollectionUser.findOrCreate({
```
tommoor, line 257: Creating a collection now always adds yourself with admin permissions

PR #5302, opened 2023-05-04
Remove "millisecond" suffix from getSignedUrl() function

chr4:
The parameter actually expects seconds and not milliseconds, according to S3 documentation:

https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html

This was confusing me when debugging an issue with S3.


chr4:
Also adjusted the mentioned comment!

tommoor (approved):

PR #5307, opened 2023-05-05
Request validation for cron

apoorv-mishra:
Towards #4284


server/routes/api/cron/schema.ts
```diff
@@ -0,0 +1,44 @@
...
+    ctx.addIssue({
+      code: z.ZodIssueCode.custom,
+      message: "token is required",
+      fatal: true,
+    });
+
+    return z.NEVER;
+  }
+
+  const token = (req.body.token ?? req.query.token) as string;
```
tommoor, line 27: Interesting, not sure how I feel about having the authentication inside data validation, feels like this is mixing concerns?

server/routes/api/cron/cron.ts
```diff
@@ -0,0 +1,31 @@
...
+    success: true,
+  };
+};
+
+router.get("cron.:period", validate(T.CronSchema), cronHandler);
+router.post("cron.:period", validate(T.CronSchema), cronHandler);
+
+// For backwards compatibility
+router.get("utils.gc", cronHandler);
+router.post("utils.gc", cronHandler);
```
tommoor, line 29: Missing validate middleware on these

server/routes/api/cron/schema.ts
apoorv-mishra: You sound quite right. Perhaps we can retain `token` presence and `token.length` checks here and move the `crypto` checks inside the API handler?

apoorv-mishra: Separated concerns! Kept `token.length` check _with_ `crypto` check, made sense _semantically_ :+1:

tommoor (approved):

PR #5313, opened 2023-05-07
fix: use real boolean instead of aws's bool

libnewton:
Minor, obvious fix.


tommoor:
Thanks, bitten by auto-complete there I think

PR #5315, opened 2023-05-08
feat: Allow embeds to be used inside tables

tommoor:
Also cut the bottom of trello embeds so they look nicer 🙈


TaoChan1005:
There is a bug. When i input '/' in the cell, the menu wouldn't show. I need input'/{anything}' then delete'{anything}', the menu would show
Screen-Recording-2023-07-02-at-21 28 19

tommoor:
The menu only shows on Paragraphs currently, this is a known limitation

PR #5343, opened 2023-05-13
Update providerId to fallback to id in the absence of a sub field.

DandrewsDev:
This allows for the profile field used by ODIC to set the provider id to be overridden using an ENV Var. 

As discussed in [discussions#4032](https://github.com/outline/outline/discussions/4032)
Discord and Mattermost both provide slightly non standard OIDC data. This causes authentication using those services to fail.

This follows the same standard used already by the OIDC_DISPLAY_NAME env var. Allowing that field to be set to a different profile field.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -98,6 +98,10 @@ OIDC_USERINFO_URI=
 # Supports any valid JSON path with the JWT payload
 OIDC_USERNAME_CLAIM=preferred_username
 
+# Specify which claims to derive provider id information from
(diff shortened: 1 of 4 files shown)


tommoor:
I wonder if shouldn't just fallback to looking for `id` without extra config 🤔

tommoor (approved):
As the fallback is only used on non-OIDC compatible endpoints I think this is reasonably safe

PR #5366, opened 2023-05-21
chore: Upgrade all of prosemirror

tommoor:
- Fixes various editor bugs.
- In-built types


tommoor:
Note, currently blocked as table editing isn't functioning correctly.

tommoor:
- `/` still left in text after some commands
- Enter above table causes error
- Enter in table jumps selection to end

PR #5467, opened 2023-06-22
Request validation for `/api/shares.*`

apoorv-mishra:
Towards #4284

### DIFF
--- a/server/routes/api/shares/schema.ts
+++ b/server/routes/api/shares/schema.ts
@@ -2,6 +2,7 @@ import { isEmpty } from "lodash";
 import isUUID from "validator/lib/isUUID";
 import { z } from "zod";
 import { SHARE_URL_SLUG_REGEX, SLUG_URL_REGEX } from "@shared/utils/urlHelpers";
+import { Share } from "@server/models";
 import BaseSchema from "../BaseSchema";
 
 export const SharesInfoSchema = BaseSchema.extend({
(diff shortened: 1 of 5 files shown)


server/routes/api/shares/schema.ts
```diff
@@ -0,0 +1,97 @@
...
+});
+
+export type SharesInfoReq = z.infer<typeof SharesInfoSchema>;
+
+export const SharesListSchema = BaseSchema.extend({
+  body: z.object({
+    sort: z
+      .string()
+      .refine(
+        (val) =>
```
tommoor, line 33: Can we use the same method as assertSort rather than hard coding? 
https://github.com/outline/outline/blob/8e4270c3219097f7406a555e1cf5d56fabcafd4f/server/validation.ts#L56

server/routes/api/shares/schema.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export const SharesUpdateSchema = BaseSchema.extend({
+  body: z.object({
+    id: z.string().uuid(),
+    includeChildDocuments: z.boolean().optional(),
+    published: z.boolean().optional(),
+    urlId: z
+      .string()
+      .regex(SHARE_URL_SLUG_REGEX, {
+        message: "must contain only aphanumeric and dashes",
```
tommoor, line 71:
```suggestion
        message: "must contain only alphanumeric and dashes",
```

tommoor (approved):
Small nits

server/routes/api/shares/schema.ts
apoorv-mishra: My thinking was to only include those attributes which seem relevant for sorting, rather than including all. It's a decent tradeoff though :+1:

tommoor (approved):

PR #5474, opened 2023-06-24
Improve document revision creation

tommoor:
In two scenarios it feels as though document revisions are created too late and will now be created sooner.

- Write a document revision when all collaborators leave a document
- Write a document revision when 'Done editing' in edit mode


server/routes/api/documents/schema.ts
```diff
@@ -203,6 +203,9 @@ export const DocumentsUpdateSchema = BaseSchema.extend({
 
     /** Version of the API to be used */
     apiVersion: z.number().optional(),
+
+    /** Whether the editing session is complete */
+    done: z.boolean().optional(),
```
tommoor, line 208: This was already passed from the client, but not used

PR #5477, opened 2023-06-25
fix: Keyboard navigation around inline code marks

### LINKED ISSUE #5267: When editing a page cursor navigation with arrow keys is inconsistent with respect to inline code
Usually the arrow keys move the cursor one character, but for inline code the behavior is inconsistent.

**To Reproduce**
I made a small screenshot that contains written instructions on what is happening.

https://user-images.githubusercontent.com/17914/234877668-4d0556e1-de43-4b9a-a612-3685365740ec.mov

Similar issues show up when the cursor is at the beginning of inline code.

**Expected behavior**
Since moving around elsewhere is also possible with a single arrow key stroke, the expectation is that leaving inline code also takes only one keystroke. Or at least consistent behavior, or visual indication.

- Install: self hosted
- Version: 4f019b7

 - OS: macOS
 - Browser Firefox
 - Version 112.0.1

tommoor:
closes #5267


jannschu:
Sorry if this is spammy, but that's just great news for me! :-)

PR #5488, opened 2023-06-28
Handle promise linting

tommoor:
- Adds linting for misused and floating promises
- Fixes all the failing locations

Where a promise is used in a react hook for data loading I've just gone with `void`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,7 +40,7 @@
     "@typescript-eslint/no-floating-promises": "error",
     "@typescript-eslint/await-thenable": "error",
     "@typescript-eslint/no-misused-promises": [
-      "warn",
+      "error",
       {
         "checksVoidReturn": false
       }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
   "parserOptions": {
     "sourceType": "module",
     "extraFileExtensions": [".json"],
+    "project": "./tsconfig.json",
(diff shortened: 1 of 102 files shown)


server/commands/subscriptionCreator.ts
```diff
@@ -43,7 +43,7 @@ export default async function subscriptionCreator({
 
   // If the subscription was deleted, then just restore the existing row.
   if (subscription.deletedAt && resubscribe) {
-    subscription.restore({ transaction });
+    await subscription.restore({ transaction });
```
tommoor, line 46: This was bad

server/models/View.ts
```diff
@@ -64,7 +64,7 @@ class View extends IdModel {
 
     if (!created) {
       model.count += 1;
-      model.save(options);
+      await model.save(options);
```
tommoor, line 67: This was bad

.eslintrc
```diff
@@ -36,6 +37,14 @@
       "component": true,
       "html": true
     }],
+    "@typescript-eslint/no-floating-promises": "error",
+    "@typescript-eslint/await-thenable": "error",
+    "@typescript-eslint/no-misused-promises": [
+      "warn",
```
apoorv-mishra, line 43: Warnings are unfortunately ignored, as good as being absent :smile:

PR #5504, opened 2023-07-01
Edit datetime format

alephbang:
solve date, time formatting 
https://github.com/outline/outline/discussions/5480


tommoor:
Hi @alephbang – I'm afraid I can't see how this makes sense as an installation-level setting and can't be merged.

We already respect the user's browser language preference which seems the most correct way to go. I could entertain some argument for a separate user-level preference within Outline, but even that's a bit of a stretch.

PR #5553, opened 2023-07-11
JSON to client

### LINKED ISSUE #3000: Move to sending JSON rather than Markdown to client

tommoor:
See issue for what this unlocks in terms of future functionality. Makes JSON rather than markdown the canonical source of truth for reading documents.

- [x] Test public shares
- [x] Test restoring revisions
- [x] Test applying templates
- [x] Bump editor version to force reload
- [x] Images in public docs
- [x] Backwards compat API requests
- [x] Bug: Cycle detected in computation `asNavigationNode`

closes #3000


TaoChan1005:
How about this work?

server/routes/api/documents/documents.ts
```diff
@@ -1014,15 +1016,7 @@ router.post(
     document.collection = collection;
 
     ctx.body = {
-      data:
-        apiVersion === 2
```
tommoor, line 1018: Technically a breaking API change but it should only have been used internally in previous versions

server/routes/api/documents/documents.ts
```diff
@@ -1278,15 +1272,7 @@ router.post(
     });
 
     ctx.body = {
-      data:
-        apiVersion === 2
```
tommoor, line 1282: Technically a breaking API change but it should only have been used internally in previous versions

tommoor:
Planning on migration to storing JSON content before continuing here as the performance implications of loading state from DB are questionable, but not going to be good.

TaoChan1005:
Hi ,

I noticed that this pull request has not been updated for a while. Could you please let me know the current status? 

Thank you!

Issue #5109, opened 2023-03-28
Add hover cards for user mentions


tommoor:
Similar to what we have for document links, @mentions should show a hover card with more details. This may require some refactoring of the existing hover code to allow reuse.

PR #5571, opened 2023-07-17
Preview mentions

apoorv-mishra:
Towards #5109

- [x] `urls.unfurl` endpoint with tests
- [x] Refactor `HoverPreview` to consume `urls.unfurl` response and preview both documents and mentions

Fixes:
- [x] Switching hovers to adjacent elements shows same preview despite elements being different
- [x] Fix warning: Failed prop type: The prop `to` is marked as required in `Link`, but its value is `undefined`.
- [x] Hovering just over the bottom edge of document link or mention, results in infinite close-open-close preview cycle

https://github.com/outline/outline/assets/15694566/cd35aa74-7006-4c0f-bbd0-a8490cdc1f4b


app/components/HoverPreview.tsx
```diff
@@ -94,27 +118,65 @@ function HoverPreviewInternal({ element, id, onClose }: Props) {
...
-            }
-          </HoverPreviewDocument>
+          {isVisible ? (
+            <Animate>
+              <Card>
+                <Margin />
+                <CardContent>
+                  {data.type === "mention" ? (
+                    <Content to="">
+                      <Flex gap={12}>
```
tommoor, line 138: Can we split these into individual components? Otherwise the `React.Suspense` will be hit immediately, and better for the near future of supporting more than 2 types here.

app/components/HoverPreview.tsx
```diff
@@ -94,27 +118,65 @@ function HoverPreviewInternal({ element, id, onClose }: Props) {
     : anchorBounds.left;
   const leftOffset = anchorBounds.left - left;
 
-  return (
+  return !data || loading ? (
+    <LoadingIndicator />
```
tommoor, line 122: Nice

server/routes/api/urls/urls.ts
```diff
@@ -0,0 +1,165 @@
...
+      }
+      authorize(user, "read", mentionedUser);
+      authorize(user, "read", document);
+
+      const lastView = head(orderBy(document.views, ["updatedAt"], ["desc"]));
+      const lastViewedAt = lastView ? lastView.updatedAt : undefined;
+      const lastActiveAt = mentionedUser.lastActiveAt;
+      let description;
+      if (lastViewedAt && differenceInMinutes(new Date(), lastViewedAt) < 5) {
+        description = t("Currently viewing", { ...opts(user) });
```
tommoor, line 55: I don't think this is accurate unfortunately and will likely conflict with the avatars in the header, maybe we can just say "Viewed just now" or something like that.

server/routes/api/urls/urls.ts
```diff
@@ -0,0 +1,165 @@
...
+      } else if (
+        lastActiveAt &&
+        differenceInMinutes(new Date(), lastActiveAt) < 5
+      ) {
+        description = lastViewedAt
+          ? t("Online now • Viewed {{ lastViewedAt }}", {
```
tommoor, line 61: Please split all these translations into two strings so that the • is not included and "Online now" doesn't need to be translated multiple times

server/routes/api/urls/urls.ts
```diff
@@ -0,0 +1,165 @@
...
+  auth(),
+  validate(T.UrlsUnfurlSchema),
+  transaction(),
+  async (ctx: APIContext<T.UrlsUnfurlReq>) => {
+    const { url, documentId } = ctx.input.body;
+    const { user } = ctx.state.auth;
+    const { transaction } = ctx.state;
+    const urlObj = new URL(url);
+    const locale = dateLocale(user.language);
+    if (urlObj.protocol === "mention:") {
```
tommoor, line 31: It would be nice to create an `UnfurlHelper` and move this logic out of the controller, they shouldn't be so large.

`UnfurlHelper.parse(url)`

Individual private methods for:

`UnfurlHelper.document(url)`

`UnfurlHelper.mention(url)`

apoorv-mishra: I don't quite get this. Is `UnfurlHelper.parse(url)` just a wrapper around protocol check plus `parseMentionUrl()/parseDocumentSlug()` or is it supposed to wrap the `findByPk` calls as well?

tommoor: This is okay for now – now that the presenter logic is separate it's small enough 👍

tommoor (approved):

app/components/HoverPreview/HoverPreview.tsx
```diff
@@ -143,7 +141,7 @@ function HoverPreviewInternal({ element, id, onClose }: Props) {
                     />
                   ) : data.type === UnfurlType.Document ? (
                     <HoverPreviewDocument
-                      id={id}
+                      id={data.meta.id}
```
apoorv-mishra, line 144: Summary didn't change while hovering among adjacent elements. This change fixed the problem, but I hope it's the correct way to fix this problem :crossed_fingers: 

https://github.com/outline/outline/assets/15694566/f95f8130-b794-4322-be9d-cb81f98a4c59

cc: @tommoor

app/components/HoverPreview/HoverPreview.tsx
```diff
@@ -111,7 +111,7 @@ function HoverPreviewInternal({ element, onClose }: Props) {
     : anchorBounds.left;
   const leftOffset = anchorBounds.left - left;
 
-  if (!data && loading) {
+  if (loading) {
```
apoorv-mishra, line 114: Switching among adjacent elements flashes previously rendered card just before rendering the correct one. Preventing that here.

https://github.com/outline/outline/assets/15694566/4033d563-ab94-4e51-8e6a-6a888c8fcc57

PR #5595, opened 2023-07-23
Preview arbitrary urls using Iframely

apoorv-mishra:
Enhances #5109 

- [ ] Presenter named `presentUnfurl` or something similar
- [ ] Iframely [error handling](https://iframely.com/docs/result-codes)
- [ ] Tests
- [ ] Integrate with FE

Further improvements:

- [ ] Iframely API returns back [`cache_age`](https://oembed.com/) field which can be used to serve the cached response from our end rather than making calls to Iframely API


apoorv-mishra:
Closing in favor of #5598

PR #5598, opened 2023-07-23
Preview arbitrary urls within a document

apoorv-mishra:
Enhances #5109 

- [x] Presenter named `presentUnfurl` or something similar
- [x] Iframely [error handling](https://iframely.com/docs/result-codes)
- [x] Integrate with FE
- [x] Tests

Further improvements:

- [x] Iframely API returns back [`cache_age`](https://oembed.com/) field which can be used to serve the cached response from our end rather than making calls to Iframely API

### DIFF
--- a/app/components/HoverPreview/HoverPreviewLink.tsx
+++ b/app/components/HoverPreview/HoverPreviewLink.tsx
@@ -17,7 +17,7 @@ type Props = {
   url: string;
   /** Title for the preview card */
   title: string;
-  /** Url for thumbnail served by the link provider*/
+  /** Url for thumbnail served by the link provider */
   thumbnailUrl: string;
   /** Some description about the link provider */
   description: string;
--- a/app/components/HoverPreview/HoverPreviewLink.tsx
+++ b/app/components/HoverPreview/HoverPreviewLink.tsx
@@ -1,0 +1,31 @@
+import * as React from "react";
+import Img from "@shared/editor/components/Img";
+import Flex from "~/components/Flex";
+import { Preview, Title, Description } from "./Components";
+
+type Props = {
+  /** Link url */
+  url: string;
+  /** Title for the preview card */
+  title: string;
+  /** Url for thumbnail served by the link provider*/
+  thumbnailUrl: string;
+  /** Some description about the link provider */
+  description: string;
+};
+
+function HoverPreviewLink({ url, thumbnailUrl, title, description }: Props) {
+  return (
+    <Preview to={url}>
+      <Flex gap={12} column>
+        <Img src={thumbnailUrl} alt={""} />
+        <Flex column>
+          <Title>{title}</Title>
+          <Description>{description}</Description>
+        </Flex>
+      </Flex>
+    </Preview>
+  );
+}
+
+export default HoverPreviewLink;
--- a/app/components/HoverPreview/HoverPreviewLink.tsx
+++ b/app/components/HoverPreview/HoverPreviewLink.tsx
@@ -1,7 +1,16 @@
 import * as React from "react";
+import styled from "styled-components";
 import Img from "@shared/editor/components/Img";
 import Flex from "~/components/Flex";
-import { Preview, Title, Description } from "./Components";
+import {
+  Preview,
+  Title,
+  Description,
+  Card,
+  CardContent,
+  CARD_WIDTH,
+  THUMBNAIL_HEIGHT,
+} from "./Components";
 
 type Props = {
   /** Link url */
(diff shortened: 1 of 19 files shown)


apoorv-mishra:
Requested review specifically for UI improvements now that the FE integration is done.

tommoor:
Can we try making the image go all the way to the edges? Should also fix this issue where the last line of text is awkwardly chopped off. Also see if the size of the image can be reserved based on the thumbnail dimensions that come back from the API as when it loads in the size of the card changes.

image

server/presenters/unfurls/unfurl.ts
```diff
@@ -0,0 +1,19 @@
+import { IframelyErrorResponse, Unfurl } from "@shared/types";
+
+function presentUnfurl(data: any): Unfurl | IframelyErrorResponse {
+  return !data.error
```
tommoor, line 4: I don't like this overloading with the error response, but it won't be needed with the other suggestion

server/routes/api/urls/urls.ts
```diff
@@ -47,22 +49,24 @@ router.post(
     }
 
     const previewDocumentId = parseDocumentSlug(url);
-    if (!previewDocumentId) {
-      ctx.response.status = 204;
+    if (previewDocumentId) {
+      const document = previewDocumentId
+        ? await Document.findByPk(previewDocumentId)
```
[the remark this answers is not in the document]
apoorv-mishra: Uh-oh!

app/components/HoverPreview/HoverPreviewLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+  CARD_WIDTH,
+  THUMBNAIL_HEIGHT,
+} from "./Components";
+
+type Props = {
+  /** Link url */
+  url: string;
+  /** Title for the preview card */
+  title: string;
+  /** Url for thumbnail served by the link provider*/
```
tommoor, line 20:
```suggestion
  /** Url for thumbnail served by the link provider */
```

app/components/HoverPreview/HoverPreviewLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+  thumbnailUrl: string;
+  /** Some description about the link provider */
+  description: string;
+};
+
+function HoverPreviewLink({ url, thumbnailUrl, title, description }: Props) {
+  return (
+    <Preview to={{ pathname: url }} target="_blank" rel="noopener noreferrer">
```
tommoor, line 28:
```suggestion
    <Preview as="a" href={url} target="_blank" rel="noopener noreferrer">
```

Not tested, but this should work

plugins/iframely/server/iframely.ts
```diff
@@ -0,0 +1,56 @@
...
+    const val = await Redis.defaultClient.get(this.cacheKey(url));
+    if (val) {
+      return JSON.parse(val);
+    }
+  }
+
+  public static async get(url: string) {
+    try {
+      const cached = await this.cached(url);
+      return cached ? cached : this.cache(url);
```
tommoor, line 49: It's weird that the `fetch` method is inside `cache`, lets just split those apart. JSDoc comments appreciated on the publicly exported interface here

server/env.ts
```diff
@@ -601,6 +601,24 @@ export class Environment {
...
+    protocols: ["http", "https"],
+  })
+  public IFRAMELY_URL = process.env.IFRAMELY_URL ?? "https://iframe.ly";
+
+  /**
+   * Iframely API key
+   */
+  @IsOptional()
+  @CannotUseWithout("IFRAMELY_URL")
+  public IFRAMELY_API_KEY = process.env.IFRAMELY_API_KEY ?? "";
```
tommoor, line 620:
```suggestion
  public IFRAMELY_API_KEY = this.toOptionalString(process.env.IFRAMELY_API_KEY);
```

This _is_ optional, empty strings lead to bugs 😉

server/utils/unfurl.ts
```diff
@@ -0,0 +1,24 @@
...
+glob
+  .sync(path.join(rootDir, "plugins/*/server/unfurl.js"))
+  .forEach((filePath: string) => {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const resolver: UnfurlResolver = require(path.join(
+      process.cwd(),
+      filePath
+    ));
+    const name = startCase(filePath.split("/")[2]);
+    resolvers[name] = resolver;
```
tommoor, line 20: I believe we need to filter out resolvers based on `requiredEnvVars` here, it doesn't happen automatically. It is possible that there will be no valid resolvers and that case can then be handled

tommoor:
Looking good, my comments are mainly about handling the case of these environment variables not being set properly as all existing installations will be missing them.

app/components/HoverPreview/HoverPreviewLink.tsx
apoorv-mishra: Not as it is, but with some other modifications, yes :+1:

PR #5619, opened 2023-07-28
Request validation for `/api/collections.*`

apoorv-mishra:
Towards #4284


server/routes/api/collections/collections.test.ts
```diff
@@ -499,6 +499,31 @@ describe("#collections.add_group", () => {
     expect(res.status).toEqual(200);
   });
 
+  it("should fail with status 400 bad request when permission is null", async () => {
```
apoorv-mishra, line 502: Added this because `permission` can't be null as per `collection_groups` table.

server/routes/api/teams/teams.test.ts
```diff
@@ -287,7 +287,7 @@ describe("#team.update", () => {
       body: {
         token: admin.getJwtToken(),
         id: collection.id,
-        permission: "",
```
apoorv-mishra, line 290: I hope this is fine and nobody's sending `permission` as an empty string :crossed_fingers:

tommoor (approved):

PR #5638, opened 2023-08-01
Request validation for `/api/integrations.*`

### LINKED ISSUE #4284: Refine server side validations

apoorv-mishra:
Finally closes #4284 :pray:


server/routes/api/integrations/integrations.ts
```diff
@@ -127,15 +102,27 @@ router.post(
...
     const { user } = ctx.state.auth;
-    const integration = await Integration.findByPk(id);
+    const { transaction } = ctx.state;
+
+    const integration = await Integration.findByPk(id, { transaction });
     authorize(user, "delete", integration);
 
-    await integration.destroy();
+    await integration.destroy({ transaction });
+    // also remove the corresponding authentication if it exists
```
tommoor, line 116: This is already taken care of through delete cascade on the db relationship

server/routes/api/integrations/integrations.ts
```diff
@@ -127,15 +102,27 @@ router.post(
 router.post(
   "integrations.delete",
   auth({ admin: true }),
-  async (ctx: APIContext) => {
-    const { id } = ctx.request.body;
-    assertUuid(id, "id is required");
-
+  transaction(),
+  validate(T.IntegrationsDeleteSchema),
```
tommoor, line 106:
```suggestion
  validate(T.IntegrationsDeleteSchema),
  transaction(),
```

Lets make sure transaction is the last middleware

server/routes/api/integrations/integrations.ts
apoorv-mishra: That works the other way around, when you delete the `authetications` row. I added a test for this, comment these lines out and there fails the test 🙂

tommoor (approved):

PR #5655, opened 2023-08-07
feat: support self hosted grist

Ocarthon:
Add support for self hosted Grist instances.

This change adds the option to configure a grist instance url. With this you can embed urls from a self hosted grist instance.


tommoor:
Hey, you should be able to fix the linting errors largely with `yarn lint --fix` – but overall this looks good and will probably accept 👍

Ocarthon:
@tommoor Fixed the lint errors. I initially did not fix them as they were not related to my changes.

tommoor (approved):

PR #5673, opened 2023-08-09
feat: Add Google Maps Embed (#5667)

lampy255:
Add Google Maps Embed feature


shared/editor/embeds/GoogleMaps.test.ts
```diff
@@ -0,0 +1,22 @@
+import GoogleMaps from "./GoogleMaps";
+
+describe("GoogleMaps", () => {
+  const match = GoogleMaps.ENABLED[0];
+
+  test("to be enabled", () => {
+    expect(
+      '<iframe src="https://www.google.com/maps/embed?pb=!1m18!1m12!1m3!1d50977.036904273686!2d174.74383592605594!3d-37.00825027293197!2m3!1f0!2f0!3f0!3m2!1i1024!2i768!4f13.1!3m3!1m2!1s0x6d0d4fe87ef3d5bb%3A0xf00ef62249b7130!2sAuckland%20Airport!5e0!3m2!1sen!2snz!4v1691573100204!5m2!1sen!2snz" width="600" height="450" style="border:0;" allowfullscreen="" loading="lazy" referrerpolicy="no-referrer-when-downgrade"></iframe>'.match(
```
tommoor, line 8: Lets just match the url here the same as all of the other integrations, It's a good idea to accept pasting the iframe code but that change needs making in a different part of the system

shared/editor/embeds/GoogleMaps.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+function GoogleMaps(props: Props) {
+  const { matches } = props.attrs;
+  const source = matches[1];
+
+  return <Frame {...props} src={source} title={`GoogleMaps`} />;
+}
+
+GoogleMaps.ENABLED = [
+  /src="(https:\/\/www\.google\.com\/maps\/embed[^"]+)"/, // Extract src from google maps iframe embed code
```
tommoor, line 13: Just match the url here, see the other embeds for examples

lampy255:
@tommoor I've updated the regex and made sure it works as expected.

tommoor (approved):

tommoor:
This is the first one where there is no way to map from the public link that most people are likely to paste to an embed link 😢 .
Definitely pushes on that iframe parsing logic

PR #5695, opened 2023-08-16
fix: Incorrect error shown to user when connection limit is reached

tommoor:
Reported through support that the error was confusing in this condition.


server/collaboration/ConnectionLimitExtension.ts
```diff
@@ -6,28 +6,36 @@ import {
...
 import { trace } from "@server/logging/tracing";
+import { TooManyConnections } from "./CloseEvents";
 
 @trace()
 export class ConnectionLimitExtension implements Extension {
   /**
    * Map of documentId -> connection count
    */
-  connectionsByDocument: Map<string, number> = new Map();
+  connectionsByDocument: Map<string, Set<string>> = new Map();
```
tommoor, line 16: Updated to track `socketId` instead of a number

PR #5741, opened 2023-08-28
Use "Inter" as default typeface

tommoor:
As SF Pro and Inter are quite similar this is mainly being done for the non-Mac platforms out there.


shared/editor/components/Styles.ts
```diff
@@ -285,9 +285,6 @@ width: 100%;
   word-wrap: break-word;
   white-space: pre-wrap;
   white-space: break-spaces;
-  -webkit-font-variant-ligatures: none;
-  font-variant-ligatures: none;
-  font-feature-settings: "liga" 0; /* the above doesn't seem to work in Edge */
```
tommoor, line 290: Ligatures are on by default in Inter

tommoor:
Merging for now, I _may_ revert this on Mac and only use Inter on Windows/Linux as they have no decent system fonts.

PR #5753, opened 2023-08-31
Adding more features in "Security"

indrainsignia:
1. fix disable download button
2. feature to enable team members and admins to download documents (not just viewers)
3. feature to enable/disable copy-paste on documents

### DIFF
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -491,7 +491,11 @@ export const printDocument = createAction({
   analyticsName: "Print document",
   section: DocumentSection,
   icon: <PrintIcon />,
-  visible: ({ activeDocumentId }) => !!(activeDocumentId && window.print),
+  visible: ({ activeDocumentId, stores }) =>
+    !!(activeDocumentId && window.print) &&
+    stores.policies.abilities(activeDocumentId).download &&
+    (!!stores.auth?.team?.getPreference(TeamPreference.ViewersCanExport) ||
+      !!stores.auth?.team?.getPreference(TeamPreference.TeamCanExport)),
   perform: async () => {
     queueMicrotask(window.print);
   },
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -276,7 +276,10 @@ export const downloadDocumentAsHTML = createAction({
   icon: <DownloadIcon />,
   iconInContextMenu: false,
   visible: ({ activeDocumentId, stores }) =>
-    !!activeDocumentId && stores.policies.abilities(activeDocumentId).download,
+    !!activeDocumentId &&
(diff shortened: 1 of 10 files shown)


tommoor:
Hi, I don't think these changes make sense for the project right now – disabling copy and paste is silly on the web, it's far too easy to bypass it.

For the disabling of members and admins being able to download – for admins this makes not sense as they can simply change the setting. For members, I'd be open to the option but I'm not sure of the usecase where you'd have users that are allowed to read and edit documents but not download a copy?

indrainsignia:
By disabling copy and paste, I disable the right click as well. I just realized that it can be bypassed by select the text and then use browser copy and paste. But at least right now they can't use keyboard shortcut. I will change it to disable the text selection as well so they can't use browser native copy and paste as well. Will push again when I'm done.

For admin, yes they can change the settings, but it will affect the entire organization. Actually we can also limit it only for non admin (member and viewer), but I just want to be equal for entire team/organization.

For existing feature disable download for viewer, can actually be bypassed as well by save or print the web page, using menu available in browser. But I ignore this as of now.

Preventing download although they can open, create, or edit, is to protect the data. Let's say an organization keep confidential data inside, many people contribute to it. I want to make it hard for someone to steal the data. Let's say someone download it, store it in his local PC, and then the PC is stolen, or maybe he send the data outside organization.
Although they can still take a screenshot of it to do the same bad thing, but it will be hard for them to replicate the content. They need to type what is written in the image (or use OCR outside to help for real bad guys). Again, this assume I ignore the browser native save and print.

indrainsignia:
on my latest commit, it strengthen the security.
- If policy said viewers or team can't download, it will also disable print function and when they try to print using browser menu, it will just blank (media print display none)
- If policy said can't copy text, it will also disable the selection (so only tech guy who knows programming can hack it from browser)

routerino:
What you are suggesting isn't security related, it is DLP (data loss prevention) related. DLP is at its core a tug of war between usability and containing corporate knowledge.

I do not see any appetite (based on what @tommoor has said) to enter that tug of war, especially given the first step is to reduce usability by disabling fundamental browser features. At best its confusing, at worst it's actively hostile to the user.

stavros-k:
Just happened to see this PR (I don't need/care about this features), but I'll throw in an idea...

Instead of trying to disable printing and copying, in cases users should see but not download/export.
It might be a good idea to inject a watermark? Not sure how possible would that be given its a markdown.

But the idea is that they "can't" use the document after without "exposing" themselves that they got it without "permission".

tommoor:
Markdown is a plain text document, it's not possible to watermark.

I'm closing this out as unwanted as I mentioned above we're not in the business of disabling browser functionality for security theatre. As always, I'd recommend reading the contributing guide and discussing features with the team _before_ working on them an putting in a PR.

PR #5763, opened 2023-09-02
Local file storage

### LINKED ISSUE #5716: Add local file storage as alternative to S3
We've heard time and time again that setting up Minio is the most difficult part of hosting Outline, for many usecases folks will just end up having the files store locally on the same physical hardware after all that effort.

I have refactored the [S3 storage](https://github.com/outline/outline/tree/main/server/storage/files) so that it inherits from a base class that can be easily used to implement a local file storage alternative.

apoorv-mishra:
Closes #5716 

- [x] Guards and checks wherever appropriate(`env` vars etc.)
- [x] Tests
- [x] Self review


.env.sample
```diff
@@ -55,6 +55,12 @@ AWS_S3_UPLOAD_MAX_SIZE=26214400
 AWS_S3_FORCE_PATH_STYLE=true
 AWS_S3_ACL=private
 
+FILE_STORAGE=local
+FILE_STORAGE_LOCAL_ROOT=files
+# Generate a unique random key. The format is not important but you could still use
+# `openssl rand -hex 32` in your terminal to produce this.
+FILE_STORAGE_LOCAL_SECRET=generate_a_new_key
```
tommoor, line 62: Can use the existing `SECRET_KEY` for this, no need for another

plugins/storage/plugin.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Iframely",
```
tommoor, line 2: ...

plugins/storage/server/api/files.ts
```diff
@@ -0,0 +1,100 @@
...
+    file.size > env.FILE_STORAGE_UPLOAD_MAX_SIZE
+  ) {
+    throw InvalidRequestError(
+      `The selected file was larger than the ${bytesToHumanReadable(
+        env.FILE_STORAGE_UPLOAD_MAX_SIZE
+      )} maximum size`
+    );
+  }
+
+  const subdir = key.split("/").slice(0, -1).join("/");
```
tommoor, line 50: All this kind of stuff should be inside `LocalStorage` rather than the router.

plugins/storage/server/api/files.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Attachment } from "@server/models";
+import { authorize } from "@server/policies";
+import { APIContext } from "@server/types";
+import { getAttachmentForJWT } from "@server/utils/jwt";
+import { getFileFromRequest } from "@server/utils/koa";
+import { assertPresent, validateKey } from "@server/validation";
+
+const router = new Router();
+
+router.post("files.create", auth(), async (ctx: APIContext) => {
```
tommoor, line 29: I'll let you know when a fix is in to use validation middleware here

server/env.ts
```diff
@@ -611,6 +611,29 @@ export class Environment {
   @IsOptional()
   public AWS_S3_ACL = process.env.AWS_S3_ACL ?? "private";
 
+  /**
+   * Which file storage system to use
+   */
+  public FILE_STORAGE = process.env.FILE_STORAGE;
```
tommoor, line 617: Add `@IsOptional()` validation

server/index.ts
```diff
@@ -51,6 +52,12 @@ if (serviceNames.includes("collaboration")) {
 // This function will only be called once in the original process
 async function master() {
   await checkEnv();
+  if (env.FILE_STORAGE === "local") {
+    const rootDir = env.FILE_STORAGE_LOCAL_ROOT;
+    if (!existsSync(rootDir)) {
```
tommoor, line 57: Need error handling here and shut down the server

server/utils/jwt.ts
```diff
@@ -96,3 +97,30 @@ export async function getUserForEmailSigninToken(token: string): Promise<User> {
...
+
+export async function getAttachmentForJWT(token: string): Promise<Attachment> {
+  const payload = getJWTPayload(token);
+
+  if (payload.type !== "attachment") {
+    throw AuthenticationError("Invalid token");
+  }
+
+  // check the token is within it's expiration time
+  if (payload.expiresAt) {
```
tommoor, line 109: I'm 99% sure this check is part of `JWT.verify`

server/utils/jwt.ts
```diff
@@ -96,3 +97,30 @@ export async function getUserForEmailSigninToken(token: string): Promise<User> {
...
+    if (new Date(payload.expiresAt) < new Date()) {
+      throw AuthenticationError("Expired token");
+    }
+  }
+
+  const attachmentId = payload.key.split("/")[2];
+  const attachment = await Attachment.findByPk(attachmentId);
```
tommoor, line 116: Can use `rejectOnEmpty: true` instead of `invariant` here.

plugins/storage/server/api/files.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  if (attachment.isPrivate) {
+    if (!actor) {
+      throw AuthenticationError();
+    }
+    authorize(actor, "read", attachment);
+  }
+
+  const fileName = attachment.key.split("/")[3];
+  const contentType = mime.lookup(fileName) || "application/octet-stream";
```
tommoor, line 94: Content type is stored on the attachment record

plugins/storage/server/api/files.ts
apoorv-mishra: What if we expose this through a method on `Attachment`, something like `attachment.saveFile(file)` which internally calls the `FileStorage.upload`? Makes it consistent with other getters like `attachment.stream` and `attachment.buffer`.

server/env.ts
apoorv-mishra: And also default it to `s3` I suppose?

tommoor: I suppose we have to for now, in that case IsOptional not needed

plugins/storage/server/api/files.ts
tommoor: sounds good

server/test/support.ts
```diff
@@ -7,10 +9,80 @@ import onerror from "@server/onerror";
...
+      options.headers["Content-Type"] = "application/json";
+      options.body = JSON.stringify(options.body);
+    }
+
+    return nodeFetch(url, options);
+  });
+};
+
+// expose Promise
+TestServer.Promise = global.Promise;
```
apoorv-mishra, line 80: - [ ] Move to class-based syntax

server/storage/files/S3Storage.ts
```diff
@@ -47,7 +47,7 @@ export default class S3Storage extends BaseStorage {
     );
   }
 
-  public getPublicEndpoint(isServerUpload?: boolean) {
+  public getUploadUrl(isServerUpload?: boolean) {
```
tommoor, line 50: Remind me why this change? It seems in the end that it would still better be called `getPublicEndpoint` – you're using it as part of `getUrlForKey` after all?

server/test/factories.ts
```diff
@@ -444,6 +444,9 @@ export async function buildAttachment(overrides: Partial<Attachment> = {}) {
     updatedAt: new Date("2018-01-02T00:00:00.000Z"),
     ...overrides,
   });
+
+  attachment.key = `public/${overrides.userId}/${attachment.id}/images.docx`;
```
tommoor, line 448: Not loving that `images.docx` is hardcoded here, we should also remove `key` on L439 if it's just going to be overridden anyway.

Allow `buildAttachment` to accept a second argument with optional file name parameter?

server/test/support.ts
```diff
@@ -7,17 +11,95 @@ import onerror from "@server/onerror";
...
+        this.server
+          .listen(0, () => resolve())
+          .on("error", (err) => reject(err));
+      });
+    }
+
+    return this.listener;
+  }
+
+  fetch(path: string, opts: any) {
```
tommoor, line 39: It's actually quite nice having this helper internally now, will probably consider having another argument here for `user` and then change all the tests so that the jwtToken doesn't need to be injected into the headers everywhere?

(Obviously not in this PR)

server/test/support.ts
```diff
@@ -7,17 +11,95 @@ import onerror from "@server/onerror";
 import webService from "@server/services/web";
 import { sequelize } from "@server/storage/database";
 
+class TestServer {
```
tommoor, line 14: Lets move this to it's own file in `server/test` though.

server/utils/jwt.ts
```diff
@@ -96,3 +97,24 @@ export async function getUserForEmailSigninToken(token: string): Promise<User> {
 
   return user;
 }
+
+export async function getAttachmentForJWT(token: string): Promise<Attachment> {
```
tommoor, line 101: Now that I see this here it makes me want to move these finders into the relevant model as static methods, eg.

`User.findByEmailSigninToken` and `Attachment.findBySignature`

server/validation.ts
```diff
@@ -170,6 +172,30 @@ export const assertCollectionPermission = (
   assertIn(value, [...Object.values(CollectionPermission), null], message);
 };
 
+export class ValidateKey {
+  public static isValid = (key: string) => {
```
tommoor, line 176: Lets add direct tests for this validation

server/storage/files/LocalStorage.ts
```diff
@@ -0,0 +1,127 @@
...
+  public async getFileBuffer(key: string) {
+    const stream = this.getFileStream(key);
+    return new Promise<Buffer>((resolve, reject) => {
+      const chunks: Buffer[] = [];
+      if (!stream) {
+        return reject(new Error("No stream available"));
+      }
+
+      stream.on("data", function (d) {
+        chunks.push(d as Buffer);
```
tommoor, line 119:
```suggestion
      stream.on("data", function (d: Buffer) {
        chunks.push(d);
```

Does this pass? Preferable if so

server/storage/files/LocalStorage.ts
```diff
@@ -0,0 +1,127 @@
...
+    } else if (body instanceof Blob) {
+      src = Readable.from(Buffer.from(await body.arrayBuffer()));
+    } else {
+      src = Readable.from(body);
+    }
+
+    const destPath = path.join(env.FILE_STORAGE_LOCAL_ROOT_DIR, key);
+    closeSync(openSync(destPath, "w"));
+    const dest = createWriteStream(destPath);
+    src.pipe(dest);
```
tommoor, line 72: Need to wait until the stream has completed writing here

server/storage/files/S3Storage.ts
```diff
@@ -86,10 +90,10 @@ export default class S3Storage extends BaseStorage {
     acl,
   }: {
     body: S3.Body;
-    contentLength: number;
-    contentType: string;
+    contentLength?: number;
```
tommoor, line 93: Tempting to say these should remain as required, even if not used in all implementations

server/storage/files/S3Storage.ts
```diff
@@ -171,23 +175,21 @@ export default class S3Storage extends BaseStorage {
...
-      .getObject({
-        Bucket: env.AWS_S3_UPLOAD_BUCKET_NAME,
-        Key: key,
-      })
-      .promise();
-
-    if (response.Body) {
-      return response.Body as Blob;
-    }
+    const stream = this.getFileStream(key);
```
tommoor, line 178: There's nothing S3 specific here now, move to base class?

server/utils/jwt.ts
apoorv-mishra: `Attachment.findBySignature` makes sense particularly for the scope of this PR though.

server/test/factories.ts
apoorv-mishra: Sure, makes more sense.

apoorv-mishra: Can't skip `key` though. It's not nullable.

plugins/storage/server/api/files.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (
+      env.FILE_STORAGE_UPLOAD_MAX_SIZE &&
+      file.size > env.FILE_STORAGE_UPLOAD_MAX_SIZE
+    ) {
+      throw InvalidRequestError(
+        `The selected file was larger than the ${bytesToHumanReadable(
+          env.FILE_STORAGE_UPLOAD_MAX_SIZE
+        )} maximum size`
+      );
+    }
```
apoorv-mishra, line 44: Time to extract this into a separate middleware `multipart({ maxAllowedFileSize: number })`? The file can be received as `ctx.input.file`.

plugins/storage/server/api/files.ts
```diff
@@ -0,0 +1,86 @@
...
+import { getFileFromRequest } from "@server/utils/koa";
+import { createRootDirForLocalStorage } from "../utils";
+import * as T from "./schema";
+
+createRootDirForLocalStorage();
+
+const router = new Router();
+
+router.post(
+  "files.create",
```
apoorv-mishra, line 19: Rate limit?

tommoor: 10/min

plugins/storage/server/api/files.ts
tommoor: Great idea

apoorv-mishra: Created a separate issue [here](https://github.com/outline/outline/issues/5800).

plugins/storage/server/api/files.ts
```diff
@@ -49,17 +49,20 @@ router.get(
...
+        if (!actor) {
+          throw AuthenticationError();
+        }
+        authorize(actor, "read", attachment);
       }
-      authorize(actor, "read", attachment);
+    } else if (sig) {
+      attachment = await Attachment.findBySignature(sig);
+    } else {
+      throw ValidationError("Must provide either key or signature");
```
apoorv-mishra, line 65: Well, this has already been handled in the `schema`.

tommoor: Yea, I know but I don't like using the `!` syntax – if we can make the zod type better that would work too

tommoor:
Posted request for testers on the issue – https://github.com/outline/outline/issues/5716#issuecomment-1722312302

Needs testing "in the wild" this one, particularly with the migration from Minio.

Issue #5765, opened 2023-09-02
Missing sidebar contents in Safari


lampy255:
Right hand Sidebar contents fails to appear, or appears only partially in Safari.
Sometimes the sidebar animates out, but has no contents at all.

**To Reproduce**
1. View an Outline page in Safari
2. Trigger the sidebar in any way eg: click history

**Expected behavior**
The sidebar appears from the right hand side with all of its contents.

**Screenshots**
<img width="1840" alt="SideBarBug" src="https://github.com/outline/outline/assets/67034721/283ab1b1-5dd2-47cc-852f-eaf3580c836b">

**Outline:**
- Install: Self Hosted
- Version: 5e7ea165b4f76552a06f806fa163a1aab77070b4

**Desktop:**
 - OS: OSX Ventura 13.4.1
 - Browser Safari 
 - Version 16.5.1

tommoor:
Yea this has been a Safari bug for a while, weird one - it works fine in all other browsers 🤷‍♂️

lampy255:
I see framer-motion is used for sidebar animations and is quite far behind? Looks like they are up to 10.16.2

PR #5766, opened 2023-09-03
fix: Safari sidebar animation #5765

lampy255:
- Add Safari browser detect func in utils
- Apply display:block to right sidebar only on Safari Desktop  

Fixes bug that would cause sidebar contents to not display sometimes on Safari #5765


tommoor:
Nice, I don't really want browser detection hacks though, the maintenance burden is too high. If display block works then why not apply it across all browsers?

lampy255:
Yeah good point. I've removed the browser detection, and used 'Display: block' across the board. Seems to be working fine in Chrome and Safari.

tommoor (approved):
I'm surprised this works, but it does!

PR #5778, opened 2023-09-04
feat: Return attachments when exporting an individual file

### LINKED ISSUE #5293: images broken in markdown export
Exporting a page with images to markdown results in code like:

```
 ![](/api/attachments.redirect?id=$GUID)
```

Obviously that won't work basically anywhere else.

**To Reproduce**
1. Make a page
2. Drag an image into the page
3. Use the three dot menu to export to markdown
4. Open file, no image

**Expected behavior**
Either the URL should be converted to a full URL that points to the outline server itself so the image will still render, or the image itself should be included in the export and linked locally (preferred since this makes it more portable!)

HTML export currently does the former, though I suppose a second request would be the later behavior -- I'll start a discussion though about that as a new feature

- Install: self hosted
- Version: Image ID: 8f65321c7079 (from docker hub -- incidentally, is there an easier way to find out the commit hash from within the UI itself)

Don't think client info here is relevant since this is purely server behavior.

tommoor:
When export an individual document as Markdown or HTML, if there are attachments a zip file will be returned with the attachments inside and relative paths.

closes #5293

### DIFF
--- a/server/queues/tasks/ExportDocumentTreeTask.ts
+++ b/server/queues/tasks/ExportDocumentTreeTask.ts
@@ -1,5 +1,6 @@
 import path from "path";
 import JSZip from "jszip";
+import escapeRegExp from "lodash/escapeRegExp";
 import { FileOperationFormat, NavigationNode } from "@shared/types";
 import Logger from "@server/logging/Logger";
 import { Collection } from "@server/models";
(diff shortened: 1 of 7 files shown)


server/queues/tasks/ExportDocumentTreeTask.ts
```diff
@@ -73,14 +74,17 @@ export default abstract class ExportDocumentTreeTask extends ExportTask {
             date: attachment.updatedAt,
             createFolders: true,
           });
+
+          text = text.replace(
```
tommoor, line 78: Fix a bug where only the first instance of an attachment would be replaced in the output, subsequent identical images would not be.

server/utils/ZipHelper.ts
```diff
@@ -107,15 +110,21 @@ export default class ZipHelper {
...
+                compression: "DEFLATE",
+                compressionOptions: {
+                  level: 5,
+                },
+                ...options,
               },
               (metadata) => {
-                const percent = Math.round(metadata.percent);
-                if (percent !== previousMetadata.percent) {
+                if (metadata.currentFile !== previousMetadata.currentFile) {
```
tommoor, line 120: Just reducing the amount of debug logging here, a bit excessive.

image

PR #5785, opened 2023-09-06
chore: Improve perf of server tests

tommoor:
Removing the requirement to truncate db after each test. From ~11m down to ~3.5m 

image

- Remove `seed` helper
- Lots of updates to make hard-coded data unique (faker js)
- Remove old tests on scripts that have long ran
- Split parallelize server tests


server/routes/api/collections/collections.ts
```diff
@@ -807,7 +807,11 @@ router.post(
 
     authorize(user, "delete", collection);
 
-    const total = await Collection.count();
+    const total = await Collection.count({
```
tommoor, line 810: Found a bug here 🙈

Issue #5764, opened 2023-09-02
Document Incorrect View History data


lampy255:
A documents view history sometimes appears incorrect / out of sync with the Last Active value of a user on the Members page in Settings.

**To Reproduce**
- Create one or more documents
- Create two users
- Edit a document as User A
- After {x hours}, login as User B
- Observe that the relevant document claims User A "Viewed less then a minute ago"
- Observe that the last active value for User A is {x hours} on the Members Page in Settings

**Expected behavior**
If a user hasn't been online for 16 hours, then they can't have viewed a document less then 16 hours ago

**Screenshots**
User "Rob" viewed less then a minute ago
<img width="705" alt="ActivityCard" src="https://github.com/outline/outline/assets/67034721/63ce2491-4759-44b2-944b-edb3793c0419">

User "Rob" last active 16 hours ago
<img width="935" alt="MembersPage" src="https://github.com/outline/outline/assets/67034721/b7079bd6-9642-453d-9bec-ba59d6eacf30">

**Outline:**
- Install: Self Hosted
- Version: 5e7ea165b4f76552a06f806fa163a1aab77070b4

**Desktop:**
 - OS: OSX Ventura 13.4.1
 - Browser: Chrome
 - Version: 115.0.5790.170

tommoor:
I suspect the issue is that the viewed time is kept live while there is a browser tab open with the document, whereas the "Last active" time is only updated when an API request is made. 

Perhaps the viewed time should not be updated while a tab is open with no activity

PR #5804, opened 2023-09-10
fix: Only update views in collaborative server on data change

tommoor:
Using the `onAwarenessUpdate` hook resulted in the view being updated when the user comes on/offline and otherwise isn't necessarily viewing the document.  Also now updating  `user.lastActiveAt` to keep these in sync.


tommoor:
closes #5764

Issue #4802, opened 2023-01-30
Multiple Mermaid diagrams on a page make it deadly slow


chimit:
Multiple Mermaid diagrams on a page make it deadly slow. Scrolling lags, and any typing in the code has huge delays (a few seconds between a keyboard press and a letter appearing on the page).

**To Reproduce**
1. Create multiple Mermaid diagrams on a page

**Expected behavior**
It should be fast.

**Outline (please complete the following information):**
- Install: self-hosted
- Version: 0.66.2

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Safari
 - Version 16.2

tommoor:
I put 8 diagrams on a page and it seems just fine, keypresses are rendered within 10ms – perhaps some diagrams are much slower than others? If you can export the problematic document as Markdown then we can investigate from there

chimit:
@tommoor, here is an [example](https://gist.github.com/chimit/195c187d4c39f9a6e74fc00bc5f071aa). Just paste it into a new document, save, reopen, and try to edit any chart. You will see very long reaction on any key press.

cocoonkid:
I just tried this out and can confirm as soon as you edit the source the typing reaction times become incredible slow.

chimit:
This is still relevant.

PR #5850, opened 2023-09-19
Update Mermaid Rendering Flow & Update Mermaid to 10.4.0

antran22:
This PR addresses several issues:
- #4802: 
  - Every diagrams on a given page will be re-rendered when we change just one diagram. 
  - The re-rendering happens on every keystroke (not debounced)
  - The `mermaid.initialize()` function is called redundantly.
- #5043, #5168, #5457. For my team, there are many (subjectively) valuable functional addition to `mermaid@^10.0.0`. However, the bundle-size problem is not yet addressed, but I would like to bring this to discussion with the Outline community on such a trade-off.

### DIFF
--- a/shared/editor/extensions/Mermaid.ts
+++ b/shared/editor/extensions/Mermaid.ts
@@ -1,3 +1,6 @@
+import debounce from "lodash/debounce";
+import last from "lodash/last";
+import sortBy from "lodash/sortBy";
 import { Node } from "prosemirror-model";
 import {
   Plugin,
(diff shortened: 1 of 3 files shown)


shared/editor/extensions/Mermaid.ts
```diff
@@ -13,8 +16,82 @@ import { findBlockNodes } from "../queries/findChildren";
...
+  element?: HTMLElement;
+  readonly elementId: string;
+  private currentTextContent = "";
+  private _rendererFunc?: RendererFunc;
+  constructor() {
+    this.diagramId = uuidv4();
+    this.elementId = "mermaid-diagram-wrapper-" + this.diagramId;
+  }
+
+  initializeElement(): HTMLElement {
```
antran22, line 35: This ensures we only `createElement` at most once for a given diagram.

tommoor:
Seems interesting! The problem is the mermaid upgrade – as before, it adds 20% to initial download size and 33% to entire download size. It's ridiculous, and I'm not having this embed that most users don't use bloat the application so considerably.

Are you open to a PR without the upgrade but including the other improvements?

antran22:
@tommoor I split this PR into two others. #5852 will focus on updating the rendering flow.

PR #5852, opened 2023-09-20
feat: update mermaid rendering flow

antran22:
Split from #5850, this PR only concerns with addressing #4802.

Performance improvement:
- Each diagram now maintains its rendering state.
- Updating code of one diagram only re-render its output. Before we re-render the whole doc page on each change. This is done by each diagram maintaining its current text input, and only update when the text input changes.
- Add debounce for updating the diagram.

Usability improvement:
- Display MermaidJS render error to the user, instead of just "Error rendering Diagram"

PR #5853, opened 2023-09-20
Update MermaidJS to version 10.4.0

antran22:
Split from #5850, this PR only concerns with updating the MermaidJS library version.
This PR is based on top of #5852, and should be dealt with after the other PR.


antran22:
@tommoor an approach for this problem is to only load Mermaid.js through CDN. The script for mermaid.js will only be loaded if specified, or optionally added to `<head>` with a Server setting.

tommoor:
We have a requirement to operate in internal networks without internet access, so that restricts use of external script loading.

I think we could upgrade if can prevent it from changing the _initial_ js bundle size which increased by 20% with the upgrade. i.e. the bigger size can be workable if it only affects those actually using Mermaid

image

antran22:
@tommoor how can we go about with working on this. Do you have any pointer for me to investigate

tommoor:
I filed one related issue here, but it only accounts for 50kb…

https://github.com/mermaid-js/mermaid/issues/4872

tommoor:
Yes, some is loaded but relative-ci reports that the initial bundle size is still increased by 20% despite this 🤔

antran22:
@tommoor This may be a problem on `rollup-plugin-webpack-stats` reporting the Mermaid.js code as initial bundle.
The internal dependency of Mermaid.js is also lazily loaded, only requisites for a given diagram are loaded when Mermaid try to render the diagram type.

Another workaround for this to include the `Mermaid.min.js` file as a static assets and load it in header, which will skip this dependency from initial JS loading completely.

PR #5866, opened 2023-09-22
feat: Native video display

### LINKED ISSUE #3130: Native video embeds
After #859 all the pieces are in place to support this, we just need to add an inline player for uploaded videos.

tommoor:
- Adds support for displaying inline videos, and a video option to the slash menu.
- Refactors image resizing logic to be reusable with video node.
- Refactors image captions to be reusable with video node.

closes #3130


Limezy:
Many thanks for that great addition.
Would it be possible to use the same native player for audio .mp3 files ?
Would it be possible to make a link to a timestamp of that video / audio file ?

tommoor:
I plan to work on a custom video player but we'll see how much time is available for that ha

Limezy:
I am surprised that the native player component you have used can't read mp3 !

StarWhiz:
Thank you for making this feature which replicates what can be done on Confluence!

Since I'm a noob and don't know how to build the docker image to try this out, can I ask that a new built image with new tag be published eventually?

tommoor:
Available in https://github.com/outline/outline/releases/tag/v0.72.0

HsunGong:
@tommoor could you please add mp3/wav support as we also need to play it just like videos.

Meanwhile, it seems difficult for those who is not familiar with the code.

Thanks a lot!

Limezy:
Yes would love sound support too !
Thanks 😊

HsunGong:
I was wondering, if we can use the video component to play audios? Or just built-in html5 player?

PR #5876, opened 2023-09-24
Allow file access not in Attachments table

### LINKED ISSUE #5873: Avatars not loading when using local file system storage
image

### LINKED ISSUE #5875: Exports do not work with local file system storage
Related https://github.com/outline/outline/issues/5716 – seems to be because the exports are written to storage without an associated attachment record.

tommoor:
The file storage PR assumed that all files in storage had an associated attachment, but that's only the case for direct user uploads. Exports, avatars, and potentially other things are stored that don't have an attachment record.

closes #5875 
closes #5873


server/models/Attachment.ts
```diff
@@ -172,42 +168,6 @@ class Attachment extends IdModel {
...
   }
 
-  /**
-   * Find an attachment given a JWT signature.
-   *
-   * @param sign - The signature that uniquely identifies an attachment
-   * @returns A promise resolving to attachment corresponding to the signature
-   * @throws {AuthenticationError} Invalid signature if the signature verification fails
-   */
-  static async findBySignature(sign: string): Promise<Attachment> {
```
tommoor, line 182: Sad to remove but they didn't really make sense after the refactor

apoorv-mishra (commented):
I feel the implementation is a bit jumbled up inside the handler, hard to trace the logic. I propose the following:

- Change `getKeyFromContext` to something like `getDownloadConfigFromContext` to return `{ key: string; filename: string; skipAuth: boolean }` where `skipAuth` = `ctx.input.query.sig || key.bucket.isPublic`(can be done by utilizing the existing `parseKey` util)
- Inside the handler we can put up the implementation along the following lines:
```js
const { key, fileName, skipAuth } = getDownloadConfigFromContext(ctx);
if (skipAuth) {
      ctx.set(
        "Content-Type",
        (fileName ? mime.lookup(fileName) : undefined) ||
          "application/octet-stream"
      );
      ctx.attachment(fileName);
      ctx.body = FileStorage.getFileStream(key);
} else {
      const attachment = await Attachment.findOne({
        where: { key },
      });
      authorize(actor, "read", attachment);
      ctx.set("Content-Type", attachment.contentType);
      ctx.attachment(attachment.name);
      ctx.body = attachment.stream;
}
```

tommoor:
I like the ordering you have here but I don't think all the logic needs shuffling into `getDownloadConfigFromContext`… will update

plugins/storage/server/api/files.ts
```diff
@@ -58,34 +56,31 @@ router.get(
...
       );
       ctx.attachment(fileName);
       ctx.body = FileStorage.getFileStream(key);
+    } else {
+      const attachment = await Attachment.findOne({
+        where: { key },
+        rejectOnEmpty: true,
+      });
+
+      if (attachment.isPrivate) {
```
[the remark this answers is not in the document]
tommoor: In theory, but I'm not 100% certain there aren't attachments with public ACL in the `uploads` bucket 🤔 

Could probably remove

plugins/storage/server/api/files.test.ts
```diff
@@ -18,11 +21,33 @@ describe("#files.create", () => {
         key: "public/foo/bar/baz.png",
       },
     });
-    const body = await res.json();
     expect(res.status).toEqual(400);
-    expect(body.message).toEqual(
-      "key: Must be of the form uploads/<uuid>/<uuid>/<name> or public/<uuid>/<uuid>/<name>"
+  });
+
+  it("should fail with status 404 if existing file is requested with key", async () => {
```
apoorv-mishra, line 27: Misleading/incomplete text? This actually fails because of the missing attachment.

package.json
```diff
@@ -198,6 +198,7 @@
     "reflect-metadata": "^0.1.13",
     "refractor": "^3.6.0",
     "request-filtering-agent": "^1.1.2",
+    "resolve-path": "^1.4.0",
```
tommoor, line 201: This was already included as part of `koa-send`, so not an additional dep.

plugins/storage/server/api/files.test.ts
tommoor: I think it's ok, the file exists on disk and the key is valid but it returns 404 without any auth which is what I was aiming to test here.

PR #5890, opened 2023-09-26
fix:  MAXIMUM_IMPORT_SIZE do not work when importing documents

harlentan:
The validation size for imported documents  can not be specified by the MAXIMUM_IMPORT_SIZE which is intended to define the maximum size of document importing

### DIFF
--- a/server/commands/documentImporter.ts
+++ b/server/commands/documentImporter.ts
@@ -14,6 +14,7 @@ import DocumentHelper from "@server/models/helpers/DocumentHelper";
 import ProsemirrorHelper from "@server/models/helpers/ProsemirrorHelper";
 import turndownService from "@server/utils/turndown";
 import { FileImportError, InvalidRequestError } from "../errors";
+import env from "@server/env";
 
 interface ImportableFile {
   type: string;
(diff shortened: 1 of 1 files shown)


tommoor:
This value is an absolute upper bound on document size that isn't directly related to importing. We can't make this change as it will have all sorts of downstream affects to have documents this large.

If your raw text content is larger than 1Mb then consider splitting into multiple documents.

harlentan:
I had an exported markdown zip from other notes system. It was failed to import to Outline with an error said that the document is too large to import.  After searching through the issues and source code, I find there is no effect no matter what  the value of MAXIMUM_IMPORT_SIZE is.

YangWang0514:
> I had an exported markdown zip from other notes system. It was failed to import to Outline with an error said that the document is too large to import. After searching through the issues and source code, I find there is no effect no matter what the value of MAXIMUM_IMPORT_SIZE is.我有一个从其他笔记系统导出的markdown zip。无法导入到大纲，错误提示文档太大，无法导入。 在搜索了问题和源代码之后，我发现无论MAXIMUM_IMPORT_SIZE的值是多少都没有效果。

请问您解决了吗

tommoor:
See the available environment variables here:

https://docs.getoutline.com/doc/file-storage-N4M0T6Ypu7#h-environment-variables

PR #5909, opened 2023-10-02
Refactor of event insertion

tommoor:
Spiking on automatic insertion of events, using `ApiKey` model as it's the simplest.

Edit: Also `Attachment` and `FileOperation`. ~Needs more type-safety before continuing down this path~, but overall this will be a ton more reliable than the current manual hookup approach and remove some largely useless commands that only exist to wrap event creation in a transaction

Towards https://github.com/outline/outline/issues/7920


server/models/base/Model.ts
```diff
@@ -3,13 +3,158 @@ import isEqual from "fast-deep-equal";
...
 > extends SequelizeModel<TModelAttributes, TCreationAttributes> {
+  /**
+   * The namespace to use for events, if none is provided the table name is used.
+   */
+  static eventNamespace: string | undefined;
+
+  /**
+   * The properties to include in the event data when this model is mutated.
+   */
+  static eventData: string[] = [];
```
tommoor, line 36: `changeset` was added since this PR started, so I think now we can remove this and just record the full changeset for every event. Although there will be some exceptions, like with documents the full content/state should not be stored.

Maybe a method on model to `getEventData` that defaults to the changeset and can be overridden by other models is the way to go.

hmacr: Yup, adds the much needed type-safety as well!
Other than this, I didn't find anything of concern tbh.

PR #5936, opened 2023-10-05
Add registering of client-side relationship properties

### LINKED ISSUE #5903: Notification should be removed if associated document is deleted
Currently the notification ends up hanging in the UI but doesn't lead anywhere.

tommoor:
closes #5903 and creates a nice foundation for future improvements of model relationship definitions on client to match server.

Additional references can be added later to cleanup some custom logic, but keeping this small to land easily.

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -254,9 +254,7 @@ class WebsocketProvider extends React.Component<Props> {
     });
 
     this.socket.on("comments.delete", (event: WebsocketEntityDeletedEvent) => {
-      comments.inThread(event.modelId).forEach((comment) => {
-        comments.remove(comment.id);
-      });
+      comments.remove(event.modelId);
     });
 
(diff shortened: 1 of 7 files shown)


app/components/WebsocketProvider.tsx
```diff
@@ -254,9 +254,7 @@ class WebsocketProvider extends React.Component<Props> {
     });
 
     this.socket.on("comments.delete", (event: WebsocketEntityDeletedEvent) => {
-      comments.inThread(event.modelId).forEach((comment) => {
```
tommoor, line 257: This is the type of logic no longer needed once the relationships are defined on the model.

app/models/decorators/Relation.ts
```diff
@@ -31,6 +83,7 @@ export default function Relation<T extends typeof Model>(
           return undefined;
         }
 
+        const relationClassName = classResolver().name;
```
tommoor, line 86: This is needed now that the Comment -> Comment relationship was introduced

PR #6031, opened 2023-10-18
API : allow filter user list via emails

tut-tuuut:
Hello !

We are using Outline API to create users and groups in outline from an external tool. We miss a way to find users' outline IDs from their email address (since "invite" does not work for already existing/invited users). We added it in users.list: this way we can fix our data and easily use the rest of the API. Would it be an interesting feature for others?

Thank you!


tommoor (approved):
Looks good from here! You hit all the notes 👌

PR #6047, opened 2023-10-21
A general self-host web embed

TaoChan1005:
#6034 Add an embed named 'SelfHosted'. To use this embed, you need to set the self-host URL.  You can use 'reg::prefix to match all subdomains.

image


matbrgz:
Why you close this?

TaoChan1005:
> Why you close this?

I created a new PR. #6048

PR #6048, opened 2023-10-21
A general self-host web embed

TaoChan1005:
#6034 Add an embed named 'SelfHosted'. To use this embed, you need to set the self-host URL. You can use 'reg::prefix to match all subdomains.

image


superkelvint:
I think it'd be great to support a comma-separated list of domains/sub-domains. The regex syntax would allow this, but it'd be a little awkward.

tommoor:
Once #3000 is completed this won't be neccessary, we can make it so that any url can be switched between an embed or a link. The regex matching is just a limitation of having Markdown as a backing store and no way to differentiate between what is and isn't considered embedded.

I'm going to close this out because the long term solution is nearly complete

PR #6062, opened 2023-10-24
fix: remember prev selected lang when creating new code block

### LINKED ISSUE #6040: Previous/default code language is no longer used to start new code blocks
Not sure what happened here yet, but something regressed – starting a new code block should remember previous language selection.

### Discussed in https://github.com/outline/outline/discussions/6039

<sup>Originally posted by **cyr0nk0r** October 19, 2023</sup>
When creating a code block, either through the GUI directly, or if using markdown "```" the default language selected is "javascript".
Please add an option in the admin area to select the default language when code blocks are first created.

Not all of us want javascript to always be the first selection.</div>

Pranav2612000:
Previously, on creating a new code block, Javascript was always the default option. This commit fixes that by remembering user's last language used and creating a new block with that language. 

We already had `rme-code-language` key to store users language preference. We were also checking and using the store value from this key if it existed. But we were not updating the storage to update this key when user changes the language. This commit fixes that.

Fixes #6040.

### DIFF
--- a/shared/editor/nodes/CodeFence.ts
+++ b/shared/editor/nodes/CodeFence.ts
@@ -184,11 +184,15 @@ export default class CodeFence extends Node {
 
   commands({ type, schema }: { type: NodeType; schema: Schema }) {
     return {
-      code_block: (attrs: Record<string, Primitive>) =>
-        toggleBlockType(type, schema.nodes.paragraph, {
+      code_block: (attrs: Record<string, Primitive>) => {
+        if (attrs?.language) {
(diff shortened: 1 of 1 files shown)


tommoor (approved):
Yep, it was previously handled here and accidentally removed in a refactor – 

https://github.com/Pranav2612000/outline/commit/2427f4747a9f02fde25289332d64974f3ff03e6f#diff-3921dbba0e6d1ea6f8a69777453228141c3bd8c6419c375d9df062fcc8f30e63L283

tommoor (approved):

PR #6066, opened 2023-10-25
API - allow search of a group using its name

tut-tuuut:
Hello it's me again. :)

We are still using the outline API to sync our internal groups with outline groups. We have an issue: we cannot create two groups with the same name, but we have no easy way to find the existing group with that name to update internal data.

Currently, I loop in Python on every result of `groups.list` until I find the matching name. I propose a way to filter this endpoint to lighten the load on our outline instance.

### DIFF
--- a/server/routes/api/groups/groups.test.ts
+++ b/server/routes/api/groups/groups.test.ts
@@ -258,6 +258,39 @@ describe("#groups.list", () => {
         .includes(anotherUser.id)
     ).toBe(true);
   });
+
+  it("should allow to find a group by its name", async () => {
+    const user = await buildUser();
+    const group = await buildGroup({
+      teamId: user.teamId,
+    });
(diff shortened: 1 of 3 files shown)


tommoor (approved):

server/routes/api/groups/groups.test.ts
```diff
@@ -258,6 +258,39 @@ describe("#groups.list", () => {
...
+  it("should allow to find a group by its name", async () => {
+    const user = await buildUser();
+    const group = await buildGroup({
+      teamId: user.teamId,
+    });
+    const anotherGroup = await buildGroup({
+      teamId: user.teamId,
+    });
+
+    const unfilteredRes = await server.post("/api/groups.list", {
```
tommoor, line 271: It doesn't feel neccessary to include this portion of the test, as it's testing just the general list functionality

PR #6074, opened 2023-10-26
Add pinterest embed

PrikshitSingh24:
The pinterest embed looks like this:
Screenshot from 2023-10-25 20-10-38


tommoor:
I'm not sure what the usecase is for this one as it's never been requested… these look a bit ugly too with the huge border radius and white showing around the edges. Do you have more thoughts on why to include?

PR #6081, opened 2023-10-27
Feat/5870 list all comments

### LINKED ISSUE #5871: Add ability to list all comments via API endpoint
### Discussed in https://github.com/outline/outline/discussions/5870

<sup>Originally posted by **Shaddix** September 22, 2023</sup>
Other good option would be to have an API to retrieve all comments (currently I get the `"message": "documentId: Required"` when requesting `/api/comments.list` without `documentId`.</div>

Pranav2612000:
Previously documentId was a required field in the 'comments.list' api.
This commit changes that by making the documentId field optional, and
returning all comments in published documents if documentId is not
passed. Refer #5870
Fixes #5871.

### DIFF
--- a/server/routes/api/comments/comments.test.ts
+++ b/server/routes/api/comments/comments.test.ts
@@ -42,6 +42,37 @@ describe("#comments.list", () => {
     expect(body.policies.length).toEqual(1);
     expect(body.policies[0].abilities.read).toEqual(true);
   });
+  it("should return all comments for a collection", async () => {
+    const team = await buildTeam();
+    const user = await buildUser({ teamId: team.id });
+    const collection = await buildCollection({
+      userId: user.id,
+      teamId: team.id,
+    });
+    const document = await buildDocument({
+      userId: user.id,
+      teamId: user.teamId,
+      collectionId: collection.id,
+    });
+    const comment = await buildComment({
+      userId: user.id,
+      teamId: team.id,
+      documentId: document.id,
+    });
+    const res = await server.post("/api/comments.list", {
+      body: {
+        token: user.getJwtToken(),
+        collectionId: collection.id,
+      },
+    });
+    const body = await res.json();
+
+    expect(res.status).toEqual(200);
+    expect(body.data.length).toEqual(1);
+    expect(body.data[0].id).toEqual(comment.id);
+    expect(body.policies.length).toEqual(1);
+    expect(body.policies[0].abilities.read).toEqual(true);
+  });
   it("should return all comments", async () => {
     const team = await buildTeam();
     const user = await buildUser({ teamId: team.id });
--- a/server/routes/api/comments/comments.test.ts
+++ b/server/routes/api/comments/comments.test.ts
@@ -1,0 +1,92 @@
+import {
+  buildCollection,
+  buildComment,
+  buildDocument,
+  buildTeam,
+  buildUser,
+} from "@server/test/factories";
+import { getTestServer } from "@server/test/support";
+
+const server = getTestServer();
+
+describe("#comments.list", () => {
+  it("should require authentication", async () => {
+    const res = await server.post("/api/comments.list");
+    const body = await res.json();
+    expect(res.status).toEqual(401);
+    expect(body).toMatchSnapshot();
+  });
+  it("should return all comments for a document", async () => {
+    const team = await buildTeam();
+    const user = await buildUser({ teamId: team.id });
+    const document = await buildDocument({
+      userId: user.id,
+      teamId: user.teamId,
+    });
+    const comment = await buildComment({
+      userId: user.id,
+      teamId: team.id,
+      documentId: document.id,
+    });
+    const res = await server.post("/api/comments.list", {
+      body: {
+        token: user.getJwtToken(),
+        documentId: document.id,
+      },
+    });
+    const body = await res.json();
+
+    expect(res.status).toEqual(200);
+    expect(body.data.length).toEqual(1);
+    expect(body.data[0].id).toEqual(comment.id);
+    expect(body.policies.length).toEqual(1);
+    expect(body.policies[0].abilities.read).toEqual(true);
+  });
+  it("should return all comments", async () => {
+    const team = await buildTeam();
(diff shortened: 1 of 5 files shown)


Pranav2612000:
@tommoor This is ready for review now. Can you take a look please?

server/routes/api/comments/comments.ts
```diff
@@ -64,15 +64,26 @@ router.post(
...
+      comments = await Comment.findAll({
+        where: { documentId },
+        order: [[sort, direction]],
+        offset: ctx.state.pagination.offset,
+        limit: ctx.state.pagination.limit,
+      });
+    } else {
+      comments = await Comment.scope(["defaultScope", "withDocument"]).findAll({
+        where: {
+          "$document.teamId$": user.teamId,
```
tommoor, line 80: I've not seen this syntax before, probably better to use `include` here instead of `withDocument` scope and then you can add the where clause directly on the include.

One other problem here is that although it's limiting to the team, it's not limiting to the collections that the user has access to, so currently comments from private collections will be returned. You can use `user.collectionIds()` to get a list of `collectionIds` to filter to.

tommoor (changes requested):

server/routes/api/comments/comments.ts
Pranav2612000: Sure @tommoor. Thank you. I've done those changes now. Could you take another look when you have time?

tommoor: Looks bood, while we're here you could allow passing a specific `collectionId` to get comments for a specific collection.

Are you adding tests?

Pranav2612000: Great suggestion @tommoor . I've added the logic to get comments for a specific collection using `collectionId`.

I couldn't find any tests for the `comments.list` event. I think I missed them. So I haven't added them. Do you know where I can find the tests? ( I tried global search for `comments.list` but couldn't find any occurrences in a test file )

tommoor: The comments routes are currently missing a test file, but pretty much every other api route has one so there are lots of equivalent tests for `list` endpoints to look at

server/routes/api/comments/comments.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  buildComment,
+  buildDocument,
+  buildTeam,
+  buildUser,
+} from "@server/test/factories";
+import { getTestServer } from "@server/test/support";
+
+const server = getTestServer();
+
+describe("#comments.list", () => {
```
Pranav2612000, line 12: @tommoor Can you review the PR again? I've added the tests here. Thank you

server/routes/api/comments/comments.ts
```diff
@@ -61,18 +62,53 @@ router.post(
...
-      order: [[sort, direction]],
-      offset: ctx.state.pagination.offset,
-      limit: ctx.state.pagination.limit,
-    });
+    let comments;
+    if (documentId) {
+      const document = await Document.findByPk(documentId, { userId: user.id });
+      authorize(user, "read", document);
+      comments = await Comment.findAll({
+        where: { documentId },
```
tommoor, line 73:
```suggestion
        where: {
          teamId: user.teamId,
          documentId: document.id,
        },
```

Lets use `document.id` here, it will allow `documentId` to be in the slug format and still work.

server/routes/api/comments/comments.ts
```diff
@@ -61,18 +62,53 @@ router.post(
...
+        limit: ctx.state.pagination.limit,
+      });
+    } else if (collectionId) {
+      const collection = await Collection.findByPk(collectionId);
+      authorize(user, "read", collection);
+      comments = await Comment.findAll({
+        include: [
+          {
+            model: Document,
+            where: {
```
tommoor, line 85:
```suggestion
            where: {
              required: true,
```

server/routes/api/comments/comments.ts
```diff
@@ -61,18 +62,53 @@ router.post(
...
+        offset: ctx.state.pagination.offset,
+        limit: ctx.state.pagination.limit,
+      });
+    } else {
+      const accessibleCollectionIds = await user.collectionIds();
+      comments = await Comment.findAll({
+        include: [
+          {
+            model: Document,
+            where: {
```
tommoor, line 101:
```suggestion
            where: {
              required: true,
```

server/routes/api/comments/schema.ts
```diff
@@ -57,7 +57,8 @@ export type CommentsDeleteReq = z.infer<typeof CommentsDeleteSchema>;
 export const CollectionsListSchema = BaseSchema.extend({
   body: CollectionsSortParamsSchema.extend({
     /** Id of a document to list comments for */
-    documentId: z.string(),
+    documentId: z.string().optional(),
+    collectionId: z.string().optional(),
```
tommoor, line 61:
```suggestion
    collectionId: z.string().uuid().optional(),
```

server/routes/api/comments/comments.ts
Pranav2612000: On it. Also, do you want the `teamId` condition as well here ( as suggested in your suggestion ). Asking because `Comment` does not have teamId so we'll need to join with the `Document` to filter by the teamId. Previously also, there was no teamId filter.

tommoor: Ah, good point you can leave it off.

server/routes/api/comments/comments.ts
Pranav2612000: Also here, you've suggested the `required: true` to be inside the `where` clause. Is that the right place, or should it be outside `where`? Like mentioned [here](https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#required-eager-loading)

tommoor: outside yes, apologies

Pranav2612000:
Up for review again @tommoor 
I've also rebased the comments to make the commit history more readable.
Thank you for you reviews.

tommoor (approved):

PR #6089, opened 2023-10-29
Add template section in collection description page

PrikshitSingh24:
Have added template section in the collection page which contains template associated with that collection:

[Screencast from 29-10-23 05:43:47 PM IST.webm](https://github.com/outline/outline/assets/99068054/f281dad1-306c-4f65-8e56-51e10a05b22b)

### DIFF
--- a/app/scenes/Collection.tsx
+++ b/app/scenes/Collection.tsx
@@ -182,6 +182,9 @@ function CollectionScene() {
                   <Tab to={collectionPath(collection.url)} exact>
                     {t("Documents")}
                   </Tab>
+                  <Tab to={collectionPath(collection.url, "templates")} exact>
+                    {t("Templates")}
+                  </Tab>
                   <Tab to={collectionPath(collection.url, "updated")} exact>
                     {t("Recently updated")}
                   </Tab>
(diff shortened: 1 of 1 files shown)


tommoor:
With the recent move to templates being managed and edited in the settings area I don't think this makes sense – the jump to settings when clicking on a template is particularly jarring.

It would also be the only tab that's a filter instead of a sorting option.

zeekhuge:
Hey @tommoor 

I think Templates are an integral part of the Outline workflow. They are what allows the many documents created by the team to be in a well-defined structure.

Furthermore, I think they should be treated as first-class element. Almost as important as the Document itself (maybe just slightly less)

And therefore should be extremely easily accessible within the collection. This PR is just meant for that.

zeekhuge:
I like the idea of having a central place to manage all templates, but I think it would rather be more usefulnto have them in collections itself, since they are, just like a document, associated with Collection.

In fact, if this feature conflicts with having settings in the template, maybe we should give-up the template-section all together and have them in the collections itself.

Maybe the community can chime in here and add examples of their workflows/usecases.

Is there any discussion which triggered moving Templates to Settings ?

tommoor:
I think there's an argument to be made that templates should be available for easy collection on the collection screen but a tab isn't the right place as it confuses the metaphors. One thing we could do is convert this "New doc" button in the top right into a segmented button with a disclosure to directly create a new doc from a template

image

PR #6134, opened 2023-11-10
Fix optional authentication

NextFire:
Related to #6061 

When accessing a public link with a stale `accessToken` in cookies, the `/documents.info` endpoint returns a 401 `Invalid token` and the public document can't be accessed (you need to go to the homepage re-login the instance first or clear your cookies/use a private browsing session).
This PR tries to fix this issue by enforcing the `optional` field in the auth middleware even when an `accessToken` is present.

(Auth source: Keycloak OpenID Connect)

### DIFF
--- a/server/middlewares/authentication.ts
+++ b/server/middlewares/authentication.ts
@@ -57,14 +57,14 @@ export default function auth(options: AuthenticationOptions = {}) {
       token = ctx.cookies.get("accessToken");
     }
 
-    if (!token && options.optional !== true) {
-      throw AuthenticationError("Authentication required");
-    }
+    try {
+      if (!token) {
+        throw AuthenticationError("Authentication required");
+      }
 
-    let user: User | null;
-    let type: AuthenticationType;
+      let user: User | null;
+      let type: AuthenticationType;
 
-    if (token) {
       if (ApiKey.match(String(token))) {
         type = AuthenticationType.API;
         let apiKey;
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks for taking a stab at this problem! I'm thinking it might be clearer to surround most of `authMiddleware` with a try/catch and then re-throw the error only if `!options.optional`.

This would move the handling of optional to a single location rather than scattered checks and make it more maintainable?

NextFire:
Yes, sure!

server/middlewares/authentication.ts
```diff
@@ -146,8 +146,12 @@ export default function auth(options: AuthenticationOptions = {}) {
           getRootSpanFromRequestContext(ctx)
         );
       }
-    } else {
```
tommoor, line 149: I believe we need to keep this otherwise auth will only be set to `{}` in the event of an error

NextFire: I think the current behavior is quite the same as previously:
- If `user` is defined and has the correct permissions, `ctx.auth` will be fully set at L133-L137
- If `options.optional` is true, one exception will be thrown at some point, then caught so `ctx.auth` can be assigned to `{}` at L151
- In other cases (insufficient perms/not optional), an exception will be thrown, caught but thrown again. `ctx.auth` has indeed no value here but the API call is interrupted anyway…

tommoor: I was thinking the case where `token` is set but invalid, I suppose there the JWT parsing will fail

NextFire: Yes, that's the case!

tommoor (approved):

NextFire:
Oh I forgot to update the JSDoc

https://github.com/outline/outline/blob/e76dcc0bafe317fede1ad80e494635b5670cf7df/server/middlewares/authentication.ts#L21-L25

tommoor:
Updated 👍

PR #6181, opened 2023-11-20
Add column `id` to `user_permissions` and `group_permissions`

### LINKED ISSUE #6180: `id` column missing in `user_permissions` and `group_permissions`
It's become incumbent to add the `id` column to `user_permissions` and `group_permissions` tables, to avoid weird `sequelize` issues disrupting dev further down the line.

apoorv-mishra:
Closes #6180 

Ported from [here](https://github.com/sequelize/sequelize/issues/2263#issuecomment-287467495).

### DIFF
--- a/server/migrations/20231120074257-add-column-id-to-user-permissions.js
+++ b/server/migrations/20231120074257-add-column-id-to-user-permissions.js
@@ -34,14 +34,6 @@ module.exports = {
   },
 
   async down(queryInterface, Sequelize) {
-    await queryInterface.sequelize.transaction(async (transaction) => {
-      await queryInterface.removeColumn("user_permissions", "id", {
-        transaction,
-      });
-      await queryInterface.sequelize.query(
-        `DROP EXTENSION IF EXISTS "uuid-ossp";`,
-        { transaction }
-      );
-    });
+    await queryInterface.removeColumn("user_permissions", "id");
   },
 };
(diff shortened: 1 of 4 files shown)


server/migrations/20231120142213-add-column-id-to-group-permissions.js
```diff
@@ -0,0 +1,47 @@
...
+
+  async down(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.removeColumn("group_permissions", "id", {
+        transaction,
+      });
+      await queryInterface.sequelize.query(
+        `DROP EXTENSION IF EXISTS "uuid-ossp";`,
+        { transaction }
+      );
```
tommoor, line 44:
```suggestion
```

I don't think we need to fully revert here, if one of these migrations needs to be reverted we don't want to lose that extension

server/migrations/20231120074257-add-column-id-to-user-permissions.js
```diff
@@ -0,0 +1,47 @@
...
+
+  async down(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.removeColumn("user_permissions", "id", {
+        transaction,
+      });
+      await queryInterface.sequelize.query(
+        `DROP EXTENSION IF EXISTS "uuid-ossp";`,
+        { transaction }
+      );
```
tommoor, line 44:
```suggestion
```

I don't think we need to fully revert here, if one of these migrations needs to be reverted we don't want to lose that extension

tommoor (approved):

PR #6208, opened 2023-11-23
Document.findByPk() with `userId` and `rejectOnEmpty`

### LINKED ISSUE #6207: `Document.findByPk()` throws with `rejectOnEmpty: true` even when the document exists
```js
Document.findByPk(document.id, {
  userId: user.id,
  rejectOnEmpty: true,
})
```

The above call throws `EmptyResultError` even when the queried document exists.

Happens due to the [`separate`](https://github.com/outline/outline/blob/7c319c17c6c1aaa3010d06a9dd1c750a7cd59766/server/models/Collection.ts#L80) flag in one of the associations. Looks like the `rejectOnEmpty` flag cascades and gets applied to the associations too. Related upstream issue [here](https://github.com/sequelize/sequelize/issues/13133).

apoorv-mishra:
Closes #6207

### DIFF
--- a/server/models/Document.ts
+++ b/server/models/Document.ts
@@ -11,6 +11,7 @@ import {
   FindOptions,
   ScopeOptions,
   WhereOptions,
+  EmptyResultError,
 } from "sequelize";
 import {
   ForeignKey,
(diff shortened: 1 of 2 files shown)


server/models/Document.ts
```diff
@@ -58,6 +59,7 @@ export const DOCUMENT_VERSION = 2;
 type AdditionalFindOptions = {
   userId?: string;
   includeState?: boolean;
+  rejectOnEmpty?: boolean | Error;
```
apoorv-mishra, line 62: Had to keep `boolean | Error` instead of just `boolean` in order to please `tsc` :see_no_evil:

tommoor (approved):

PR #6210, opened 2023-11-24
feat: add support for tldraw snapshot links

### LINKED ISSUE #6188: embed tldraw feature not work when readonly
tldraw embed function not work on page when readonly mode

this [code](https://github.com/outline/outline/blob/e2ad3f9e73bc8a7a383532660aaf3c6c2a85be63/shared/editor/embeds/Tldraw.tsx#L17C1-L18C1) only accept start with `https://www.tldraw.com/v/` but read only url started `https://www.tldraw.com/r` 

Is this an intended feature? or bugs?

Pranav2612000:
Fixes #6188

### DIFF
--- a/shared/editor/embeds/Tldraw.test.ts
+++ b/shared/editor/embeds/Tldraw.test.ts
@@ -9,9 +9,26 @@ describe("Tldraw", () => {
     ).toBeTruthy();
   });
 
+  test("to be enabled for snapshot link", () => {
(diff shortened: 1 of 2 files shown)


tommoor (approved):
I'm also not sure if this solves the original issue entirely but seems like a fan change 👍

tommoor (approved):

PR #6217, opened 2023-11-27
Allow embedding of GitLab snippets

EspadaV8:
Similar to embedding a GitHub Gist, this allows embedding of a GitLab snippet. Supports embedding of personal snippets and snippets belonging to a group/project.

image


tommoor (approved):
Great, this all seems quite reasonable

EspadaV8:
Thanks @tommoor :slightly_smiling_face:

tommoor:
After merging I tried with this [random snippet](https://gitlab.com/-/snippets/3625157) from GitLab and the result isn't too great as it ends up with a margin and scrollable, not like your screenshot at all.

tommoor:
What it looks like selected:

image

this kind of scrolling isn't really acceptable… 😢 

image

EspadaV8:
@tommoor Ah. That does look and function how I built it. What would your recommendation be to fix this then? I think without the scrolling it wouldn't really that useful since for nearly all snippets you won't be able to see enough of them to be useful withing a documentation page.

EspadaV8:
I wonder if instead of iframing the embed script and can just embed the script tag and have the snippet expand as large as needed within the document itself.

tommoor:
In hindsight I see that in the screenshots now. The GitHub embed is a fixed height and allows scrolling of the content – the GitLab snippet cannot work like this?

I'd rather not be injecting scripts into the main page if possible as it becomes a security risk.

EspadaV8:
@tommoor Just tried embedding this Gist - https://gist.github.com/trpubz/e61ef0537ed6ea7b4c8fba9f649ccb22 - and I don't see any scrolling of that content locally. Is there something I need to change somewhere to enable the scrolling?

EspadaV8:
Ah, I tried a different Gist - https://gist.github.com/ASkyeye/1e92771f287b0c6a8977edd348c5a1b1 - and that does have scrolling. Not sure why one does and the other doesn't.

...

Trying another Gist with 2 files - https://gist.github.com/user2023cpu/48601160a9626f1d475af806f131d23d - and the embed isn't really useful. You can only see 2 lines of the second file (and I'm guessing if there were more than 2 files you wouldn't see any of the other files).

image

tommoor:
I don't think it's ever been tested with these ones that have multiple files in a single gist. It certainly doesn't look very good either – at least there's not a bunch of extra margin?

Ideally they would show the right height upto some reasonable upper limit?

EspadaV8:
Removing the margin is straightforward with something like this (can open an MR if you'd like)

```
diff --git a/shared/editor/embeds/GitLabSnippet.tsx b/shared/editor/embeds/GitLabSnippet.tsx
index 05bb38883..50f2585a4 100644
--- a/shared/editor/embeds/GitLabSnippet.tsx
+++ b/shared/editor/embeds/GitLabSnippet.tsx
@@ -13,7 +13,8 @@ function GitLabSnippet(props: Props) {
const id = snippetUrl.pathname.split("/").pop();
const snippetLink = `${snippetUrl}.js`;
const snippetScript = `<script type="text/javascript" src="${snippetLink}"></script>`;
-  const styles = "<style>body { margin: 0; }</style>";
+  const styles =
+    "<style>body, body .gitlab-embed-snippets { margin: 0; }</style>";
const iframeHtml = `<html><head><base target="_parent">${styles}</head><body>${snippetScript}</body></html>`;

return (
```

There are some JS hacks around that will resize the iframe based on its content - for example https://stackoverflow.com/questions/819416/adjust-width-and-height-of-iframe-to-fit-with-content-in-it - but not sure how best to implement this within the application.

I think if the auto-resizing iframe has some max-height where it stops, the issue would still happen at some point, unless scrolling is allowed past that max-height. There are some really long Gists out there - https://gist.github.com/choco-bot/4d2d06a7d6112822ff477ae0491b8575 - and when that is embedded all you can see right now are the first 6 bullet points.

Testing that long Gist embed on Notion, and it looks like they might run the embed directly within the page and wrap the content there, since you get the full length of the Gist shown as a block.

PR #6258, opened 2023-12-06
Add GitHub OAuth as a login option

Southpaw1496:
I help run an open-source project, and Outline would be a great place for us to store things like process documents and drafts of public website pages. We don't pay for Google Workspace, Azure, or Slack, so we currently don't have a way to log in short of hosting our own identity server. Since we all use GitHub, GitHub login would be a great way for us to authenticate. It's not quite OIDC-compliant, so it needs its own authentication provider. It's functional right now so far as I've tested, but there are still a couple of issues that need to be discussed with the maintainers before it's 100% ready.

## Restricting users based on organisation
The PR that this one is loosely based on (#2399) included an option for restricting users based on their GitHub organisation. In the context of self-hosted instances, this functionality is not necessary, as each user can create a GitHub app that only they or their GitHub Organisation are able to use. However, if this login method is used in Outline cloud, there will need to be a check, which can be accomplished in a couple of ways, each with disadvantages:

If a GitHub App is used, which is the more modern way to interact with the GitHub API, you can use the ["Check Organization Membership for a User"](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#check-organization-membership-for-a-user) API, but this would require each team to authorise an API application from Outline in their GitHub organisation so that Outline could access this information. This endpoint also communicates with HTTP status codes as opposed to a response body, and I didn't find a way to do that using the `request()` helper function that some other auth providers use.

If an OAuth app is used, you can use the ["List Organisations for an authenticated user"](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations-for-the-authenticated-user) API, but this requires the `user` scope, which allows full read-write access over the user's account, which I think is too intrusive. The `read:org` scope can also be used, but that requires authorisation by the GitHub organisation, since the endpoint only returns the organisations that the application requesting it can perform actions on.

## Other flaws
- On line 37 of `github.ts`, I use `req`, instead of `ctx` like the other auth providers. If I change it, everything goes red and I receive the error "No overload matches this call". This also happens if I try to assign the type `Context` to `req`, which is why I've left it as type `any`.
- On line 52, `subdomain` is currently set to `slugify.domain` like in the OIDC provider. This is because unlike Google, Slack, or Microsoft, a user can be part of multiple Organisations and so there's no way to know which one should be used for the subdomain. Using `profile.name` or `profile.username` may be more appropriate.
- On line 77, I've hardcoded `expiresIn` to 86400, or 24h. When I try to use the `params` option like the other auth providers, I receive the error "No overload matches this call".


tommoor:
Hi @Southpaw1496 – we're not going to add additional authentication providers into the main project at this time, one of the reasons for implementing OIDC was to support as many SSO providers as possible without custom implementations for each and the associated maintenance.

Appreciate the work here, and hopefully the structure of the codebase makes it easy to maintain your fork with this work.

PR #6297, opened 2023-12-16
Fix Suspended User Login Error Redirect URL

ruizehung:
Redirect suspended user to `/?notice=user-suspended` in `email.ts` and `passport.ts`. 

Probably better to set redirect notice as constant somewhere and import it in both client and server. But this would be a much bigger refactoring.

### Before
Previous behavior. Notice that the `notice` query parameter in `http://127.0.0.1:3000/?notice=suspended` is `suspended`
image

### After 
Current behavior. The error message is now properly displayed
image


tommoor (approved):
Yea I'll follow up with converting these to shared const

PR #6315, opened 2023-12-26
Handle `users.demote` event

### LINKED ISSUE #6314: Changing a member to a viewer doesn't revoke their permissions
If a user is demoted to a viewer through _Settings -> Members_, their permissions aren't immediately revoked.

apoorv-mishra:
Closes #6314

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -6,7 +6,6 @@ import * as React from "react";
 import { withTranslation, WithTranslation } from "react-i18next";
 import { io, Socket } from "socket.io-client";
 import { toast } from "sonner";
-import { Pagination } from "@shared/constants";
 import { FileOperationState, FileOperationType } from "@shared/types";
 import RootStore from "~/stores/RootStore";
 import Collection from "~/models/Collection";
@@ -464,10 +463,7 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on("users.demote", async (event: PartialWithId<User>) => {
       if (auth.user && event.id === auth.user.id) {
-        await Promise.all([
-          documents.fetchPage({ limit: Pagination.defaultLimit }),
-          collections.fetchPage({ limit: Pagination.defaultLimit }),
-        ]);
+        await Promise.all([documents.fetchAll(), collections.fetchAll()]);
       }
     });
 
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -463,7 +463,8 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on("users.demote", async (event: PartialWithId<User>) => {
       if (auth.user && event.id === auth.user.id) {
-        await Promise.all([documents.fetchAll(), collections.fetchAll()]);
+        documents.all.forEach((document) => policies.remove(document.id));
+        await collections.fetchAll();
       }
     });
 
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -6,6 +6,7 @@ import * as React from "react";
 import { withTranslation, WithTranslation } from "react-i18next";
 import { io, Socket } from "socket.io-client";
 import { toast } from "sonner";
+import { Pagination } from "@shared/constants";
 import { FileOperationState, FileOperationType } from "@shared/types";
 import RootStore from "~/stores/RootStore";
 import Collection from "~/models/Collection";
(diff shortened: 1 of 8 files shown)


app/components/WebsocketProvider.tsx
```diff
@@ -460,6 +462,15 @@ class WebsocketProvider extends React.Component<Props> {
       }
     );
 
+    this.socket.on("users.demote", async (event: PartialWithId<User>) => {
+      if (auth.user && event.id === auth.user.id) {
+        await Promise.all([
+          documents.fetchPage({ limit: Pagination.defaultLimit }),
+          collections.fetchPage({ limit: Pagination.defaultLimit }),
+        ]);
```
apoorv-mishra, line 470: Logically, we need updated policies regarding documents and collections for that user. I hope this is seems to be the right way to achieve that? Another tactic could be `policies.remove(document.id)` for all accessible documents but the similar doesn't seem to work well for collections.

tommoor: Lets do that for documents (remove), I'm working on a more generic system for loading relationships but policies aren't defined as a real relationship yet. For collections refetching the page makes sense but we need to remove all the previous collection policies still as there may be more than a page of collections loaded in memory.

apoorv-mishra: Still, the same problem remains if the user has a collection opened which is returned in next page and the permissions are revoked(collection policies are removed). The user would see the Not Found page. But yeah, we've somewhat delegated the problem, to later pages.

tommoor: Maybe we should introduce a `fetchAll` method that will paginate through until done

apoorv-mishra: Seems more reliable. Can give it a try.

apoorv-mishra: Done!

app/stores/DocumentsStore.ts
```diff
@@ -359,6 +361,19 @@ export default class DocumentsStore extends Store<Document> {
   fetchOwned = (options?: PaginationParams): Promise<Document[]> =>
     this.fetchNamedPage("list", options);
 
+  @action
+  fetchAll = async (): Promise<Document[]> => {
```
tommoor, line 365: I would just move this to the BaseStore, and pages to fetch should be based on the response not what happens to be cached in store already – the pagination data can be accessed with the symbol, eg `response[PAGINATION_SYMBOL].total`

app/components/WebsocketProvider.tsx
```diff
@@ -464,10 +463,7 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on("users.demote", async (event: PartialWithId<User>) => {
       if (auth.user && event.id === auth.user.id) {
-        await Promise.all([
-          documents.fetchPage({ limit: Pagination.defaultLimit }),
-          collections.fetchPage({ limit: Pagination.defaultLimit }),
-        ]);
+        await Promise.all([documents.fetchAll(), collections.fetchAll()]);
```
tommoor, line 466: Lets just remove the document policies and treat it differently for now, doesn't seem like a good idea to pull all documents down

app/stores/DocumentsStore.ts
apoorv-mishra: Sorry, shouldn't it be based on the cached items? How on `response`? It doesn't even exist in the scope of the method. What am I missing... :thinking:

apoorv-mishra: Are you suggesting we make an initial `fetchPage` request to get the value of `total` and then subsequently iterate and fetch pages _based on_ total?

tommoor: That or looping until there's no results left

app/components/WebsocketProvider.tsx
apoorv-mishra: You know, that's a problem when we change back from viewer to member. They'd have to refresh in this case to be able to edit the document again. But, in a way this might be _ok_.

tommoor: Yea, I think you could easily add the same frontend handling for `users.promote`

tommoor (approved):

app/components/WebsocketProvider.tsx
apoorv-mishra: Looks like the role change from _viewer_ to _member_ still happens through `users.demote`, despite it being a _promotion_ semantically. `users.promote` is hit when the permission is upgraded to _admin_. Can tackle this separately later.

PR #6336, opened 2024-01-01
chore: Replace `css-inline` with `@css-inline/css-inline-wasm`

Stranger6667:
Hi! I am the author of `css-inline` and I've recently added `napi.rs` bindings to `css-inline` and now it is published under a different name. There is also a WASM module available under `@css-inline/css-inline-wasm` which is more or less the same as the package used here (smaller in size though). The option names are now camel-cased.

Let me know if there is a need for native modules for other platforms (I see that `@rollup/rollup` has more platforms in `yarn.lock`)


tommoor:
Thanks, great module!

tommoor:
Currently supported platforms we build for are listed here:

https://github.com/outline/outline/blob/d3e0b19202f29dcd0b6d6b0e1130c71c68976f66/.circleci/config.yml#L150

```
linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/ppc64le,linux/s390x
```

Seems like there is probably a mismatch but honestly I'm not sure which of these actually see use… it's not really possible to know.

Stranger6667:
Thanks! 

`css-inline` is limited by `napi.rs` which [supports](https://napi.rs/docs/cross-build/summary):

> Windows x64
> Windows x86
> Windows arm64
> macOS x64
> macOS arm64
> Linux x64 gnu (glibc 2.17)
> Linux x64 musl
> Linux arm64 gnu (glibc 2.17)
> Linux arm64 musl
> Linux armv7 gnueabihf
> Android arm64
> Android armv7

It seems that `linux/ppc64le,linux/s390x` are missing from this list (I am also not sure about some `arm` details there). Technically, the Rust compiler supports these platforms, but we still need support from the `napi.rs` side. 

In any event, the WASM module should work if an appropriate runtime is in place, which I guess should be the case as the previous wasm-based `css-inline` version is used now. Let me know what you think - maybe I can update the PR with the WASM-based version.

tommoor:
Unfortunately It's probably best to stick to the wasm module for now, given this architecture dependency – if I drop the platforms someone will come out the woodwork 😆

Stranger6667:
Done! :)

tommoor:
Unfortunately the new module is failing, clearly we need better test coverage here

```
 TypeError: Cannot read properties of undefined (reading '__wbindgen_add_to_stack_pointer')
     at inline (/Users/tom/Projects/outline/node_modules/@css-inline/css-inline-wasm/index.js:213:10)
     at inline2 (/Users/tom/Projects/outline/node_modules/@css-inline/css-inline-wasm/index.js:412:10)
     at HTMLHelper.inlineCSS (/Users/tom/Projects/outline/build/server/models/helpers/HTMLHelper.js:18:38)
     at DocumentPublishedOrUpdatedEmail.beforeSend (/Users/tom/Projects/outline/build/server/emails/templates/DocumentPublishedOrUpdatedEmail.js:60:46)
     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
     at async DocumentPublishedOrUpdatedEmail.send (/Users/tom/Projects/outline/build/server/emails/templates/BaseEmail.js:67:24)
     at async Queue.<anonymous> (/Users/tom/Projects/outline/build/server/services/worker.js:114:7)
```

Stranger6667:
Oh, I forgot that it should be initialised first. Will check how the tests are done and will try to fix this

tommoor (changes requested):
As discussed

Stranger6667:
thanks for bumping this issue :) I'll get back to it soon, meanwhile, there is `0.13.0` that fixes some important issues with `nth-child` selectors

server/models/helpers/HTMLHelper.ts
```diff
@@ -1,20 +1,28 @@
...
   /**
    * Move CSS styles from <style> tags to inline styles with default settings.
    *
    * @param html The HTML to inline CSS styles for.
    * @returns The HTML with CSS styles inlined.
    */
-  public static inlineCSS(html: string): string {
+  public static async inlineCSS(html: string): Promise<string> {
+    if (!initialized) {
+      const path = require.resolve("@css-inline/css-inline-wasm/index_bg.wasm");
```
tommoor, line 16: Why doesn't the module just deal with this implementation detail by the way? 🤔

tommoor (approved):
Okay, got everything passing here and testing locally looks good.

server/models/helpers/HTMLHelper.ts
Stranger6667: I think it would be nice to do this way indeed. I’ve recently switched the build implementation and mostly followed an example from [resvg-js](https://github.com/yisibl/resvg-js)

PR #6361, opened 2024-01-09
fix: Improve logic for word import

tommoor:
Solid refactor of `documentImporter` logic that has been due for a while
Adds support for embedded images in `.doc` files exported from Confluence


tommoor:
Love it when completely unrelated tests start failing with syntax errors

tommoor:
Unbelievable, the dependency cache as culprit?!

PR #6414, opened 2024-01-22
Github integration

### LINKED ISSUE #6406: GitHub integration
We're planning on making GitHub the first authenticated editor integration, the initial version will allow connecting GitHub to the workspace and then show authenticated details on link hover.

After that we'll work on embedding link details into in-page widgets, once #3000 finally lands which would currently block that.

apoorv-mishra:
Closes #6406


plugins/github/client/Icon.tsx
apoorv-mishra, line 1: Replace with GitHub's icon.

server/routes/api/auth/auth.test.ts
```diff
@@ -93,10 +93,11 @@ describe("#auth.config", () => {
     const res = await server.post("/api/auth.config");
     const body = await res.json();
     expect(res.status).toEqual(200);
-    expect(body.data.providers.length).toBe(3);
+    expect(body.data.providers.length).toBe(4);
     expect(body.data.providers[0].name).toBe("Slack");
     expect(body.data.providers[1].name).toBe("OpenID Connect");
     expect(body.data.providers[2].name).toBe("Google");
+    expect(body.data.providers[3].name).toBe("Github");
```
apoorv-mishra, line 100: There's a subtle distinction between GitHub and others - others are also used for user login but GitHub is expected to be used only as an integration.

tommoor: For this reason let's keep /auth routes for login options only. It's ok in the main api namespace

tommoor: Alternatively can register the route at the /auth/ path but don't out the plugin logic in the auth folder. 

Just don't want it picked up as a login provider

apoorv-mishra: Oops! I think we ought to exclude `github` here as we've done for `email`...

apoorv-mishra: > Alternatively can register the route at the /auth/ path but move the plugin logic out of the auth subdirectory.

Uff, this would require modifying the dynamic import logic. Moving under `/api` is easier 😅

server/routes/api/integrations/integrations.ts
```diff
@@ -129,9 +141,32 @@ router.post(
...
+      transaction,
+    });
     authorize(user, "delete", integration);
 
+    // If it's GitHub, trigger its API for deletion of this installation
+    if (integration.service === IntegrationService.GitHub) {
+      const installationId = (integration as Integration<IntegrationType.Embed>)
+        .settings?.github?.installation.id;
+
+      if (env.GITHUB_APP_ID && env.GITHUB_APP_PRIVATE_KEY && installationId) {
```
apoorv-mishra, line 155: Yes, uninstall requires app auth as opposed to user auth 🙈

tommoor: Makes sense

.env.sample
```diff
@@ -114,6 +114,18 @@ OIDC_DISPLAY_NAME=OpenID Connect
 # Space separated auth scopes.
 OIDC_SCOPES=openid profile email
 
+# To configure Github auth, you'll need to create a GitHub App at
```
tommoor, line 117:
```suggestion
# To configure the GitHub integration, you'll need to create a GitHub App at
```

app/utils/PluginLoader.ts
```diff
@@ -24,6 +25,9 @@ export default class PluginLoader {
     function importAll(r: any, property: string) {
       Object.keys(r).forEach((key: string) => {
         const id = key.split("/")[3];
+        if (id === "github" && env.GITHUB_INTEGRATION_DISABLED) {
```
tommoor, line 28: Could just make this a generic `PLUGINS_DISABLED` – won't need to remove it later then?

plugins/github/plugin.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "Github",
```
tommoor, line 2:
```suggestion
  "name": "GitHub",
```

plugins/github/plugin.json
```diff
@@ -0,0 +1,9 @@
...
+  "description": "Adds a Github authentication provider, support for link unfurling.",
```
tommoor, line 3:
```suggestion
  "description": "Adds a GitHub integration for link unfurling.",
```

plugins/github/client/Settings.tsx
```diff
@@ -0,0 +1,179 @@
...
+import List from "~/components/List";
+import ListItem from "~/components/List/Item";
+import Notice from "~/components/Notice";
+import Scene from "~/components/Scene";
+import Text from "~/components/Text";
+import env from "~/env";
+import useCurrentTeam from "~/hooks/useCurrentTeam";
+import useQuery from "~/hooks/useQuery";
+import useStores from "~/hooks/useStores";
+import GithubIcon from "./Icon";
```
tommoor, line 21:
```suggestion
import GitHubIcon from "./Icon";
```

Please capitalize correctly, sorry going to be a stickler about this.

plugins/github/client/Settings.tsx
```diff
@@ -0,0 +1,179 @@
...
+import GithubButton from "./components/GithubButton";
```
tommoor, line 22:
```suggestion
import GitHubButton from "./components/GitHubButton";
```

plugins/github/client/Settings.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+function Github() {
```
tommoor, line 24:
```suggestion
function GitHub() {
```

etc...

plugins/github/server/api/github.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+      }
+
+      const github = new GitHub({ code: code!, state: teamId });
+
+      let installation;
+      try {
+        installation = await github.getInstallation(installationId);
+      } catch (err) {
+        return ctx.redirect(GitHub.errorUrl("unauthenticated"));
```
tommoor, line 77: Lets use `Logger.error` for debugging

plugins/github/server/api/github.ts
```diff
@@ -0,0 +1,117 @@
...
+            });
+            return redirectOnClient(
+              ctx,
+              GitHub.callbackUrl({
+                baseUrl: team.url,
+                params: ctx.request.querystring,
+              })
+            );
+          } catch (err) {
+            return ctx.redirect(GitHub.errorUrl("unauthenticated"));
```
tommoor, line 64: Similar, can we `Logger.error` for debugging please.

plugins/github/server/api/github.ts
```diff
@@ -0,0 +1,117 @@
...
+          teamId: user.teamId,
+          authenticationId: authentication.id,
+          settings: {
+            github: {
+              installation: {
+                id: installationId,
+                account: {
+                  id: installation.account?.id,
+                  name:
+                    // @ts-expect-error Property 'login' does not exist on type
```
tommoor, line 102: Weird, and worrying

shared/types.ts
```diff
@@ -105,15 +109,31 @@ export enum DocumentPermission {
...
   : T extends IntegrationType.Analytics
   ? { measurementId: string }
   : T extends IntegrationType.Post
   ? { url: string; channel: string; channelId: string }
   : T extends IntegrationType.Command
   ? { serviceTeamId: string }
   :
-      | { url: string }
+      | {
+          url?: string;
```
tommoor, line 129:
```suggestion
```

Wrong, yea?

server/routes/api/integrations/integrations.ts
```diff
@@ -129,9 +141,32 @@ router.post(
...
 
-    const integration = await Integration.findByPk(id, { transaction });
+    const integration = await Integration.findByPk(id, {
+      rejectOnEmpty: true,
+      transaction,
+    });
     authorize(user, "delete", integration);
 
+    // If it's GitHub, trigger its API for deletion of this installation
+    if (integration.service === IntegrationService.GitHub) {
```
tommoor, line 151: Hmm, this sucks inline – it should be in an async processor, but in order to do that you need the details of the deleted integration, which means it should be soft-deleted or can add the details on the `integrations.delete` event. Thoughts?

apoorv-mishra: Async would be better, yea. Question is whether we should delete the record _after_ uninstall is successful from GitHub's end.

tommoor: I would delete our record immediately so it can reflect in the UI and then async delete github (the background job will retry). If it eventually fails for some reason it's not the end of the world as we deleted the auth on our side, it's just cleanup

apoorv-mishra: Maybe, we can soft-delete it inside `integrations.delete` and then permanently delete in the async processor after uninstall is successful.

tommoor: No soft delete on this table currently 🙈 . Does also leave the question of how do all the other integrations get hard deleted

apoorv-mishra: Yeah, I can add it, no issues.

apoorv-mishra: Note that the slack ones will accumulate then since no async delete for them 🙈

tommoor: The processor can just hard delete anything that isn't github 👍

plugins/github/server/api/github.ts
apoorv-mishra: Can't help it I guess 😢

.env.sample
```diff
@@ -114,6 +114,19 @@ OIDC_DISPLAY_NAME=OpenID Connect
 # Space separated auth scopes.
 OIDC_SCOPES=openid profile email
 
+# To configure GitHub auth, you'll need to create a GitHub App at
```
tommoor, line 117:
```suggestion
# To configure the GitHub integration, you'll need to create a GitHub App at
```

plugins/github/client/Settings.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const { integrations } = useStores();
+  const { t } = useTranslation();
+  const query = useQuery();
+  const error = query.get("error");
+  const [relationsLoaded, setLoaded] = React.useState(false);
+
+  React.useEffect(() => {
+    void integrations.fetchPage({
+      service: IntegrationService.GitHub,
+      limit: 100,
```
tommoor, line 35: I know this is copied from the other screens, but you implemented `fetchAll` now so can use that here

plugins/github/plugin.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "GitHub",
+  "description": "Adds a GitHub authentication provider, support for link unfurling.",
```
tommoor, line 3:
```suggestion
  "description": "Adds a GitHub integration for link unfurling.",
```

I want to remove "auth provider" type language, otherwise people will think you can login with it

server/env.ts
```diff
@@ -397,6 +397,36 @@ export class Environment {
     process.env.SLACK_CLIENT_SECRET ?? process.env.SLACK_SECRET
   );
 
+  /**
+   * GitHub OAuth2 client credentials. To enable authentication with GitHub.
```
tommoor, line 401:
```suggestion
   * GitHub OAuth2 client credentials. To enable integration with GitHub.
```

shared/editor/embeds/Gist.tsx
```diff
@@ -24,7 +24,7 @@ function Gist(props: Props) {
       height="200px"
       scrolling="no"
       id={`gist-${id}`}
-      title="Github Gist"
+      title="GitHub Gist"
```
tommoor, line 27: 😉

plugins/github/server/processors/GitHubProcessor.ts
```diff
@@ -0,0 +1,73 @@
+import { App, Octokit } from "octokit";
+import { IntegrationService, IntegrationType } from "@shared/types";
+import env from "@server/env";
+import { Integration, IntegrationAuthentication } from "@server/models";
+import BaseProcessor from "@server/queues/processors/BaseProcessor";
+import { sequelize } from "@server/storage/database";
+import { IntegrationEvent, Event } from "@server/types";
+
+export default class GitHubProcessor extends BaseProcessor {
```
tommoor, line 9:
```suggestion
export default class IntegrationDeletedProcessor extends BaseProcessor {
```

Can we make this more generic, catch github as a special case and then _any other_ integration that gets deleted will be hard deleted here too. I don't want to be in a situation where we have to remember to create a processor for all other integration types

plugins/github/server/processors/GitHubProcessor.ts
```diff
@@ -0,0 +1,73 @@
...
+          installation_id: installationId,
+        }
+      );
+    }
+
+    await sequelize.transaction(async (transaction) => {
+      await integration.destroy({ transaction, force: true });
+      // also remove the corresponding authentication if it exists
+      if (integration.authenticationId) {
+        await IntegrationAuthentication.destroy({
```
tommoor, line 64: We should probably move this into an `AfterDelete` hook on the integration model so it doesn't need to be remembered everywhere

plugins/slack/server/processors/SlackProcessor.ts
```diff
@@ -141,4 +151,30 @@ export default class SlackProcessor extends BaseProcessor {
       }),
     });
   }
+
+  async integrationDeleted(event: IntegrationEvent) {
```
tommoor, line 155: This won't be needed once `IntegrationDeletedProcessor` exists

server/index.ts
```diff
@@ -83,6 +83,15 @@ async function start(id: number, disconnect: () => void) {
   // Apply default rate limit to all routes
   app.use(defaultRateLimiter());
 
+  app.context.redirectOnClient = function (url: string) {
```
tommoor, line 86:
```suggestion
  /** Perform a redirect on the browser so that the users auth cookies are included in the request. */
  app.context.redirectOnClient = function (url: string) {
```

server/models/Integration.ts
```diff
@@ -80,6 +85,22 @@ class Integration<T = unknown> extends IdModel<
...
+  static async destoryIntegrationAuthentications(
+    model: Integration,
+    options?: InstanceDestroyOptions
+  ) {
+    if (options?.force && model.authenticationId) {
+      await IntegrationAuthentication.destroy({
+        where: {
+          id: model.authenticationId,
+        },
+      });
```
tommoor, line 101: Need to spread `options` here to ensure it's using the same db transaction

plugins/github/server/processors/IntegrationDeletedProcessor.ts
```diff
@@ -0,0 +1,64 @@
...
+import { IntegrationEvent, Event } from "@server/types";
+
+export default class IntegrationDeletedProcessor extends BaseProcessor {
+  static applicableEvents: Event["name"][] = ["integrations.delete"];
+
+  private githubClient?: Octokit;
+
+  constructor() {
+    super();
+    if (!env.GITHUB_APP_ID || !env.GITHUB_APP_PRIVATE_KEY) {
```
tommoor, line 15: This is going to prevent the deletion logic running entirely for eg Slack when the GitHub integration is not enabled

apoorv-mishra: The deletion seems to be working fine for Slack when I comment out `GITHUB_APP_ID` and `GITHUB_APP_PRIVATE_KEY`. What am I missing?

tommoor: You're right, sorry this was pre-coffee.

.env.sample
```diff
@@ -122,6 +122,19 @@ OIDC_DISPLAY_NAME=OpenID Connect
...
+# To configure the GitHub integration, you'll need to create a GitHub App at
+# => https://github.com/settings/apps
+#
+# When configuring the Client ID, add a redirect URL under "Permissions & events":
+# https://<URL>/api/github.callback
+GITHUB_CLIENT_ID=get_a_key_from_slack
+GITHUB_CLIENT_SECRET=get_the_secret_of_above_key
+GITHUB_APP_NAME=get_the_name_of_registered_github_app
+GITHUB_APP_ID=get_the_id_of_registered_github_app
+GITHUB_APP_PRIVATE_KEY=get_the_private_key_of_registered_github_app
```
tommoor, line 134:
```suggestion
GITHUB_CLIENT_ID=
GITHUB_CLIENT_SECRET=
GITHUB_APP_NAME=
GITHUB_APP_ID=
GITHUB_APP_PRIVATE_KEY=
```

Will document setup separately, these placeholders aren't terribly useful

.env.sample
```diff
@@ -122,6 +122,19 @@ OIDC_DISPLAY_NAME=OpenID Connect
...
+
+# Disabled plugins, values should be comma separated
+PLUGINS_DISABLED=github
```
tommoor, line 137:
```suggestion
```

Remove from sample

plugins/github/server/processors/IntegrationDeletedProcessor.ts
```diff
@@ -0,0 +1,62 @@
...
+import { GitHub } from "../github";
+
+export default class IntegrationDeletedProcessor extends BaseProcessor {
+  static applicableEvents: Event["name"][] = ["integrations.delete"];
+
+  private githubClient?: Octokit;
+
+  constructor() {
+    super();
+    if (!GitHub.appId || !GitHub.appPrivateKey) {
```
tommoor, line 15: This is going to stop integrations from deleting properly if GitHub isn't setup, seems like a mistake?

apoorv-mishra: Tried Slack after disabling GitHub. Deletion seems to be working fine. Sorry, how would it fail?

plugins/github/server/processors/IntegrationDeletedProcessor.ts
```diff
@@ -48,12 +32,8 @@ export default class IntegrationDeletedProcessor extends BaseProcessor {
...
 
       if (installationId) {
-        await this.githubClient?.request(
-          "DELETE /app/installations/{installation_id}",
-          {
-            installation_id: installationId,
-          }
-        );
+        const github = new GitHub();
+        await github.deleteInstallation(installationId);
```
apoorv-mishra, line 36: >This is going to stop integrations from deleting properly if GitHub isn't setup, seems like a mistake?

Better now? I hope 🤞

tommoor: I think it's clearer, yep thanks

plugins/github/server/processors/IntegrationDeletedProcessor.ts
apoorv-mishra, line 1: I'm not sure if this processor should really be inside `plugins/github`, it applies well to all integrations per se. The only reason it's kept here is because of deletion of github app installation 🙈

tommoor (approved):

PR #6462, opened 2024-01-31
Refactor and regroup `urlHelpers` utils

### LINKED ISSUE #6461: Refactor `urlHelpers`
It'd be good to regroup `urlHelpers` utils into different providers.

apoorv-mishra:
Closes #6461, #6464


shared/utils/UrlHelper.ts
```diff
@@ -0,0 +1,16 @@
+class Outline {
```
tommoor, line 1: Honestly lets lose the `Outline` namespace – we can assume it's the default…

tommoor:
Oh looks like you took on https://github.com/outline/outline/issues/6464 already, cool 😄

shared/utils/UrlHelper.ts
apoorv-mishra: Sorry, you mean these should be accessible for instance, like `UrlHelper.twitter` instead of `UrlHelper.Outline.twitter`?

tommoor: yep

apoorv-mishra:
Note that this needs to be merged before the GitHub Integration so that I can have `GitHubUtils`(similar to `SlackUtils`) there. For that to actually work, I need the `build.js` changes made in this PR.

cc: @tommoor

app/components/Sharing/PublicAccess.tsx
```diff
@@ -78,7 +78,7 @@ function PublicAccess({ document, share, sharedParent }: Props) {
 
         const val = ev.target.value;
         setUrlId(val);
-        if (val && !SHARE_URL_SLUG_REGEX.test(val)) {
+        if (val && !UrlHelper.SHARE_URL_SLUG_REGEX.test(val)) {
```
tommoor, line 81: I think we should expose some methods on `UrlHelper` next as there is a lot of repetition still, eg.

`UrlHelper.isShareUrl(val)` rather than `UrlHelper.SHARE_URL_SLUG_REGEX.test(val)`

tommoor (approved):

PR #6479, opened 2024-02-02
feat: replace (--) with emdash

L11R:
Quick PR, based on discussion here:
https://github.com/outline/outline/discussions/6477


L11R:
@tommoor do I need to do something extra other than that?

tommoor:
I think that's it! Will test and deploy today

tommoor (approved):

L11R:
I don't want this feature off but rather disable any automatic character conversion inside pre-formatted blocks.

PR #6493, opened 2024-02-05
Remove excessive `.babelrc`

tommoor:
Lets see if tests pass here

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,46 +1,46 @@
 {
   "presets": [
     "@babel/preset-react",
-    "@babel/preset-typescript",
     [
       "@babel/preset-env",
       {
-        "corejs": {
-          "version": "3",
-          "proposals": true
-        },
-        "useBuiltIns": "usage"
+        "targets": {
+          "node": "18"
+        }
       }
-    ]
+    ],
+    "@babel/preset-typescript"
   ],
   "plugins": [
-    "styled-components",
+    "babel-plugin-transform-typescript-metadata",
     [
       "@babel/plugin-proposal-decorators",
       {
         "legacy": true
       }
     ],
-    "@babel/plugin-transform-destructuring",
-    "@babel/plugin-transform-regenerator",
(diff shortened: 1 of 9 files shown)


.babelrc
```diff
@@ -11,19 +10,28 @@
         },
         "useBuiltIns": "usage"
       }
-    ]
+    ],
+    "@babel/preset-typescript"
   ],
   "plugins": [
-    "styled-components",
+    "babel-plugin-transform-typescript-metadata",
```
apoorv-mishra, line 17: Where would this be put to use, currently?

.babelrc
```diff
@@ -11,19 +10,28 @@
...
-    "@babel/plugin-transform-destructuring",
-    "@babel/plugin-transform-regenerator",
-    "transform-class-properties"
+    "@babel/plugin-proposal-class-properties",
+    [
+      "transform-inline-environment-variables",
+      {
+        "include": [
+          "SOURCE_COMMIT",
+          "SOURCE_VERSION"
```
apoorv-mishra, line 30: I don't see these env vars in codebase.

tommoor: They come from the deployment process

.babelrc
tommoor: This plugin enables anything using `Reflect.metadata` to work correctly, which is a number of things – including eg inferring Sequelize datatypes from typescript types. .e.g without it you immediately get

```
Specified type of property 'providerId'
            cannot be automatically resolved to a sequelize data type. Please
            define the data type manually
```

.babelrc
apoorv-mishra: Yeah, my bad, had case-sensitivity toggled while searching 🙈

.babelrc
apoorv-mishra: Hmm, I still get this error upon `make up` in this branch.

tommoor: I believe I have the magic incantation now

PR #6520, opened 2024-02-10
fix: URL RegExp for embeds

### LINKED ISSUE #6451: Grist embed does not work
After entering the domain of a self-hosted Grist instance at the `/settings/integrations/self-hosted` route, it is not possible to add a publicly visible Grist document with its URL to a document.

**To Reproduce**

1. Create a new document with a title.
2. Type `/gri` at the beginning of the line and press Return when the suggestion for Grist appears.
3. Ctrl + V the URL and press Return
4. Nothing happens, neither in the UI, nor the console or in the network tab of the browser developer tools.

**Expected behavior**

Some kind of embed for the sheet is displayed.

**Screenshots**

grafik

grafik

grafik

**Outline:**

- Install: self hosted
- Version: e62c734c419dcca9cb998fed5584d08f5fa82afe

**Desktop:**

- OS: Ubuntu
- Browser Firefox
- Version 123b

The two services are served from different second-level domains and no specific CORS settings were made.

The integrations section of the guide and hosting documentation also miss a page for the Grist case.

* [Integrations - Outline](https://docs.getoutline.com/s/guide/doc/integrations-eUbAF2WAjF)
* [Integrations - Outline](https://docs.getoutline.com/s/hosting/doc/integrations-Oav5MXNDJk)

almereyda:
This is a simple fix that allows embedded URLs of integrations with a `url` parameter set in their preferences (currently Draw.io and Grist) to be matched again.

Closes #6451

### DIFF
--- a/shared/utils/urls.ts
+++ b/shared/utils/urls.ts
@@ -152,5 +152,7 @@ export function urlRegex(url: string | null | undefined): RegExp | undefined {
 
   const urlObj = new URL(sanitizeUrl(url) as string);
 
-  return new RegExp(escapeRegExp(`${urlObj.protocol}//${urlObj.host}`));
+  return new RegExp(
+    "^" + escapeRegExp(`${urlObj.protocol}//${urlObj.host}`) + "/(.+)$"
+  );
 }


almereyda:
@tommoor I couldn't wait and have filed this PR, because the change seemed trivial. Feel free to edit, refactor or change this PR to meet your style guides.

It may be that the change is a bit intrusive at where it was made, but it only felt right, because URL regexes should always match whole URLs, and not just the scheme and the host, if I'm not mistaken.

tommoor:
I'm writing some tests for this method right now, as I'm still unclear as to why this change would fix the problem

almereyda:
Thank you. Things started working, when `settingsDomainRegex` was equipped with a `^` in the beginning and with `\/(.+)$` at the end.

Inspecting the variables in `matcher` during runtime showed that the RegExp that was generated from `this.settings?.url` was different in form from the ones that are prefilled in `RegexMatch`.

tommoor:
I decided to go a different route as the bug is actually in the newly introduced `transformMatch` method on embed, but thanks for the pointers.

closed in 2d6a7ed24

almereyda:
Thanks, glad you chose the `matcher`. As I mentioned, I had also considered it, but didn't know how. Your detailed knowledge of the code base certainly helps here.

PR #6537, opened 2024-02-14
improv: use statusFilter instead of includeArchive,includeDrafts for document search

Pranav2612000:
Refer: https://github.com/outline/outline/discussions/6425


server/routes/api/documents/schema.ts
```diff
@@ -25,6 +25,10 @@ const DocumentsSortParamsSchema = z.object({
 
 const DateFilterSchema = z.object({
   /** Date filter */
+  statusFilter: z
+    .union([z.literal("published"), z.literal("archived"), z.literal("draft")])
```
tommoor, line 29: Should be able to use `z.nativeEnum` to share the type definition

server/models/helpers/SearchHelper.ts
```diff
@@ -354,36 +352,51 @@ export default class SearchHelper {
...
             },
+            archivedAt: {
+              [Op.eq]: null,
+            },
           },
+        ],
+      });
+    }
+
+    if (options.statusFilter && options.statusFilter.includes("draft")) {
```
tommoor, line 371:
```suggestion
    if (options.statusFilter?.includes("draft")) {
```

Collapse these clauses

tommoor (commented):
Looks reasonable so far, thanks!

server/models/helpers/SearchHelper.test.ts
```diff
@@ -207,48 +206,119 @@ describe("SearchHelper", () => {
...
+      });
       const { results } = await SearchHelper.searchForUser(user, "test", {
-        includeDrafts: false,
+        statusFilter: [StatusFilter.Published],
       });
-      expect(results.length).toBe(0);
+      expect(results.length).toBe(1);
     });
 
-    test("should not include drafts with user permission", async () => {
```
Pranav2612000, line 230: @tommoor I'm not sure what this test does and if it still applies. Can you help me with this? Do we still want to keep this condition?

Pranav2612000:
@tommoor Addressed your feedback and updated the tests.  Had some questions about the `membership.id` and `createdById` check we had when search in drafts. ( https://github.com/outline/outline/pull/6537#discussion_r1491894403 )
Do we still need those checks? What do they do?
Thank you.

Pranav2612000:
Also will start with the updates to the UI. I can see the current `FilterOption` component we have for choosing the status only allows choosing one option at a time.
image

Do you have a different component for "FilterOption" which allows choosing multiple items at a time? Or do you want me to build it out?

server/models/helpers/SearchHelper.test.ts
tommoor: The test is checking that `includeDrafts=false` filters correctly when there is a user membership to the document. Some equivalent should be easy enough to keep

tommoor:
> Do you have a different component for "FilterOption" which allows choosing multiple items at a time? Or do you want me to build it out?

There isn't anything yet, this is what I was referring to in [this comment](https://github.com/outline/outline/discussions/6425#discussioncomment-8436940) – changing `activeKey` to a plural `activeKeys` would do the trick. Or honestly in hindsight i dislike this naming, could be `selectedItems` or something more clear.

server/models/helpers/SearchHelper.test.ts
Pranav2612000: Yup. I've re-added the tests with some minor changes. Can you verify this once when you have time?

Pranav2612000:
Thank you for the reviews @tommoor 
I've done the updates. Can you take another look.

Updated the `FilterOptions` to add support for Multi selecting values

server/models/helpers/SearchHelper.ts
```diff
@@ -354,36 +352,57 @@ export default class SearchHelper {
...
+
+    if (options.statusFilter?.includes(StatusFilter.Archived)) {
+      statusQuery.push({
+        archivedAt: {
           [Op.ne]: null,
         },
       });
     }
 
+    if (options.statusFilter?.length) {
```
tommoor, line 400: Previously archived documents were filtered by default, we should retain this behavior. Given the new changes it probably makes sense to do this on the frontend by making `published,draft` statusFilter enabled in the client by default.

Pranav2612000:
Thank you for the review @tommoor .
I've removed the notes, updated defaultLabel and enabled `published` & `draft` by default.

tommoor:
I will checkout and test properly tomorrow, but I think this is ok. A good follow up issue will be to add the option to also search trashed documents.

tommoor (approved):

tommoor:
Technically this is a breaking API change but we'll get away with it

Pranav2612000:
Thank you for the opportunity @tommoor and for the final finishing touches.

PR #6539, opened 2024-02-15
Add ability to logout from OIDC

### LINKED ISSUE #3954: Logout via OIDC not working/not implemented?
When I logout from Wiki it doesnt remove the OIDC(in this case Keycloak) Token -> I can just click sign in again and be logged in automatically.

**To Reproduce**
1. Click on Sign in
2. Click on Sign out
3. Click on Sign in again
4. You are signed in without logging in

**Expected behavior**
I expect Outline to log me out via the logout endpoint of the OIDC provider.
This could be implemented via the logout endpoint provided by the OIDC provider which can be found by querying the .wellknown endpoint which is defined in the OIDC spec. 

**Outline:**
- Install: self hosted
- Version: 
https://github.com/outline/outline/tree/v0.65.2
https://github.com/outline/outline/commit/039cc8a3b3ed2a0a2b28e96797b51b372f794234

 - OS: Windows
 - Browser: Chrome

Shuttleu:
Adds ability to add OIDC_LOGOUT_URI to .env to enable logout redirect. 

Closes #3954 

May need to be adjusted to check how the user was logged in? Currently whenever the user logs out, if OIDC_LOGOUT_URL is set, it will redirect to the OIDC logout endpoint, regardless of how the user logged in.
If the app only has OIDC authentication, this won't be an issue, but could be if the app has multiple authentication methods.

### DIFF
--- a/app/scenes/Logout.tsx
+++ b/app/scenes/Logout.tsx
@@ -8,6 +8,7 @@ const Logout = () => {
   void auth.logout();
   if (env.OIDC_LOGOUT_URI) {
     window.location.replace(env.OIDC_LOGOUT_URI);
+    return null;
   }
   return <Redirect to="/" />;
 };
--- a/app/scenes/Logout.tsx
+++ b/app/scenes/Logout.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { Redirect } from "react-router-dom";
-import env from "@shared/env";
+import env from "~/env";
 import useStores from "~/hooks/useStores";
 
 const Logout = () => {
--- a/app/scenes/Logout.tsx
+++ b/app/scenes/Logout.tsx
@@ -1,3 +1,4 @@
+import env from "@shared/env";
 import * as React from "react";
 import { Redirect } from "react-router-dom";
 import useStores from "~/hooks/useStores";
@@ -5,6 +6,8 @@ import useStores from "~/hooks/useStores";
 const Logout = () => {
   const { auth } = useStores();
   void auth.logout();
+  if (env.OIDC_LOGOUT_URI)
+    window.location.replace(env.OIDC_LOGOUT_URI);
   return <Redirect to="/" />;
 };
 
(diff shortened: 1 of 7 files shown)


app/scenes/Logout.tsx
```diff
@@ -1,10 +1,14 @@
...
 import { Redirect } from "react-router-dom";
+import env from "@shared/env";
 import useStores from "~/hooks/useStores";
 
 const Logout = () => {
   const { auth } = useStores();
   void auth.logout();
+  if (env.OIDC_LOGOUT_URI) {
+    window.location.replace(env.OIDC_LOGOUT_URI);
+  }
```
tommoor, line 11: Probably wise to `return null;` here to avoid a competing redirect

Shuttleu: Good point, added that

app/scenes/Logout.tsx
```diff
@@ -1,10 +1,15 @@
 import * as React from "react";
 import { Redirect } from "react-router-dom";
+import env from "@shared/env";
```
tommoor, line 3:
```suggestion
import env from "~/env";
```

tommoor (approved):
Looks good, thanks for helping out

Shuttleu:
No worries. I feel it's only fair to contribute where I can.

hamzadis:
There seems to be a slight bug here, (at least with Firefox) where the Outline `auth.delete` request is interrupted by the page redirect (error `NS_BINDING_ABORTED`).  Seems to work in Chrome as it just happens to also send out the request in time, but I doubt it's guaranteed to.

The consequence is that the Outline session is never invalidated, even if the OIDC logout is complete, so ressing the back button just brings me back to a working Outline interface.

I'm guessing that the code should wait for the logout to complete fully before doing the redirect.

tommoor:
@hamzadis good thought, as you mention this issue does not show up on Chrome but it makes sense – added a fix for that here.

PR #6556, opened 2024-02-19
Propagate Enter key events in SuggestionsMenu so that handleLinkInputKeydown() can handle it.

superkelvint:
In SuggestionsMenu, Enter key events are not being propagated to handleLinkInputKeydown().

### DIFF
--- a/app/editor/components/SuggestionsMenu.tsx
+++ b/app/editor/components/SuggestionsMenu.tsx
@@ -467,7 +467,6 @@ function SuggestionsMenu<T extends MenuItem>(props: Props<T>) {
 
       if (event.key === "Enter") {
         event.preventDefault();
-        event.stopPropagation();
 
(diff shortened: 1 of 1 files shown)


tommoor:
Hi, thanks for the contribution – could you describe specifically the bug this is fixing?

superkelvint:
Hi, apologies for not giving more details.

If the user types out an embed URL and presses enter, nothing happens. 

To repro: 
1. /youtube
2. Select the Youtube embed
3. Type (do not paste): https://www.youtube.com/embed/Fel4WTp7cTc
4. Hit enter
5. Nothing happens.

There is code in handleLinkInputKeydown() to handle the Enter event, but it is never received because the Enter event is not propagated to it, hence the PR.

tommoor (approved):

Issue #3470, opened 2022-04-29
OpenSearch search redirects to loging instead of search results


pstuifzand:
I have added the OpenSearch in Firefox and when I search I get a login page even when I'm already logged in.

**To Reproduce**
1. Add opensearch to search by right clicking on address bar on Outline webpage
2. Click Add "Outline"
3. Search with Outline in address bar
4. You see a login page

**Expected behavior**
You should see the login page only when not logged in. When logged in you should see the search results.

**Screenshots**
![screenshot-outline-after-opensearch](https://user-images.githubusercontent.com/64524/165903887-aaa8f08e-06a2-4244-bdc8-5fc44cf3722f.png)

**Outline (please complete the following information):**
- Install: getoutline.com

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 99

PR #6567, opened 2024-02-21
Do not load opensearch.xml from CDN, so it uses the correct team subdomain

waf:
Fixes a bug that results in a broken opensearch.xml for team subdomains.

See https://github.com/outline/outline/issues/3470 -- that issue, with the same reproduction steps, is happening again due to a change in 3eabb309493a88473932ddabefa1fdff93c11bde. It changed the opensearch URL from `/opensearch.xml` to `{cdn-url}/opensearch.xml`.

The team subdomain's `/opensearch.xml` is correct, but because the `<link rel="search" ...` points to the CDN URL, this correct opensearch.xml is not used.


tommoor (approved):

tommoor:
Good call, thanks

PR #6576, opened 2024-02-22
fix: Conditionally show collections tooltip on invite screen

hmacr:
When inviting a user, an empty tooltip is shown if there are no collections.

This PR aims to address that by displaying an alternate text, instead of an empty tooltip.

### DIFF
--- a/app/scenes/Invite.tsx
+++ b/app/scenes/Invite.tsx
@@ -137,29 +137,25 @@ function Invite({ onSubmit }: Props) {
   );
 
   const collectionCount = collections.nonPrivate.length;
-  const collectionAccessNote = (
+  const collectionAccessNote = collectionCount ? (
     <span>
       <Trans>Invited members will receive access to</Trans>{" "}
-      {collectionCount > 0 ? (
-        <Tooltip
-          content={
-            <>
-              {collections.nonPrivate.map((collection) => (
-                <li key={collection.id}>{collection.name}</li>
-              ))}
-            </>
-          }
-        >
-          <Def>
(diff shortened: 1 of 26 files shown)


shared/i18n/locales/cs_CZ/translation.json
```diff
@@ -655,6 +655,7 @@
   "Sorry, you can only send {{MAX_INVITES}} invites at a time": "Litujeme, můžete odeslat pouze {{MAX_INVITES}} pozvánek najednou",
   "Invited members will receive access to": "Pozvaní členové obdrží přístup k",
   "{{collectionCount}} collections": "{{collectionCount}} collections",
+  "non-private collections when added": "non-private collections when added",
```
hmacr, line 658: What's the process for translating the newly added texts?

tommoor: Did you manually update these files? These translation json other than `en_US` are handled through the translation provider (CrowdIn)

app/scenes/Invite.tsx
```diff
@@ -140,19 +140,23 @@ function Invite({ onSubmit }: Props) {
...
-          </>
-        }
-      >
-        <Def>
-          <Trans>{{ collectionCount }} collections</Trans>
-        </Def>
-      </Tooltip>
+      {collectionCount > 0 ? (
+        <Tooltip
+          content={
```
tommoor, line 145: It's probably fine to just return `undefined` here in this scenario and suppress the tooltip as a simple fix

tommoor (approved):

PR #6581, opened 2024-02-22
Add ability to sync groups with OIDC provider

Shuttleu:
Currently there is no way to sync groups between OIDC and outline groups.

Adding a env var `OIDC_GROUPS_CLAIM` that, if set, will add any missing groups and add the user to them, then remove them from any groups they should no longer be in.

If the env var is not set, nothing will happen, so this entirely opt in.

### DIFF
--- a/plugins/oidc/server/auth/oidc.ts
+++ b/plugins/oidc/server/auth/oidc.ts
@@ -126,7 +126,7 @@ if (
                 "The groups claim in the profile parameter that was returned must be an array."
               );
             }
-            groups = (profile[env.OIDC_GROUPS_CLAIM] as unknown) as string[];
+            groups = profile[env.OIDC_GROUPS_CLAIM] as unknown as string[];
           }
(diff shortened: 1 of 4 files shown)


tommoor:
Hi @Shuttleu – thank you for the contribution, unfortunately this functionality won't be merged into the community edition as we are planning on adding group syncing for all SSO providers as part of the licensed edition. You can, of course, maintain your fork with this functionality.

Again, appreciate the work but this is a good reason why following the [contribution guidelines](https://github.com/outline/outline#contributing) and creating an issue for discussion first is important.

Shuttleu:
👌

TobiasGrether:
@tommoor Is it known when we can expect this to be added to the licensed edition?

tommoor:
In terms of roadmap, we're working on collection sharing first, then this :) 

https://docs.getoutline.com/s/roadmap

PR #6597, opened 2024-02-25
Separate environment configs

### LINKED ISSUE #6447: Create an interface to register env variables within a plugin
<https://github.com/outline/outline/pull/6414/files#diff-8fc679f0e0aca6c603438006e40e4c28123404884e276e4f87bc600f8581f104>

This makes me think we should have some way to register env variables within the plugin, this is one of the last things left centralized

[Slack Message](https://getoutline.slack.com/archives/D03TAQ9AEJE/p1706586862647599)

tommoor:
Currently development and sample env variables are somewhat mixed together – development should work "out of the box"

closes #6447


plugins/azure/server/env.ts
```diff
@@ -0,0 +1,27 @@
+import { IsOptional } from "class-validator";
+import { Environment } from "@server/env";
+import environment from "@server/utils/environment";
+import { CannotUseWithout } from "@server/utils/validators";
+
+class AzurePluginEnvironment extends Environment {
```
tommoor, line 6: @apoorv-mishra what do you think? Each plugin just creates it's own `env` with the variables it needs.

package.json
```diff
@@ -99,7 +99,7 @@
     "date-fns": "^2.30.0",
     "dd-trace": "^3.33.0",
     "diff": "^5.1.0",
-    "dotenv": "^4.0.0",
```
apoorv-mishra, line 102: Looks like we're on stone age `dotenv`.

.env.development
```diff
@@ -0,0 +1,10 @@
+URL=https://app.outline.dev:3000
```
apoorv-mishra, line 1: Hmm, had to update this to run the app successfully.

tommoor: Can you be more specific – in what way did you have to update it? You're using a different URL?

apoorv-mishra: Yeah, mine's `https://local.outline.dev:3000`.

tommoor: I'll update the default here to that 👍

apoorv-mishra:
Upon `make test`
```
Sequelize CLI [Node: 18.18.2, CLI: 6.6.1, ORM: 6.35.2]

Loaded configuration file "server/config/database.json".
Using environment "test".

ERROR: database "outline" is being accessed by other users
ERROR DETAIL: There is 1 other session using the database.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [test] Error 1
```

Not sure why it's trying to access `outline` when it's expected to hit the `outline-test` db.

.env.development
apoorv-mishra: How about leaving it empty and throwing instead? Letting the user know they need to update this...But, there may be many others I guess 🤔

tommoor: I want a standardized default that works for development here

tommoor:
Merging soon, will have a small conflict with your GitHub branch @apoorv-mishra as you'll need to move those new env variables into `plugins/github/server/env.ts` and import the plugin env file.

plugins/azure/server/env.ts
apoorv-mishra: Now I may have some thoughts after #6627 . It'd be convenient to apply `@Public` decorator if all envs are kept in a single file, preferably `server/env.ts` - same limitation that exists now 😞

tommoor: I think the decorator works well with this setup, no? I would try moving the Slack env variables and moving to the plugin as a proof of concept first.

apoorv-mishra: Kind of tricky. For example, moving the Slack env vars would only include public vars from its own class(`SlackPluginEnvironment`) plus parent class(`Environment`). So we need a way to _merge_ the public vars from all the classes.

apoorv-mishra: Keeping all the vars in a single file, flat structure like previously would simplify all of it. I'm now in favor of reverting this 😅

apoorv-mishra: I mean after all these are env vars, feels _ok_ to keep them centralized?

tommoor: Store the public env inside the module context of the decorator, it will effectively merge them all

apoorv-mishra: Well, we can only keep the var _names_, not values because we only receive `target` object and `propertyKey` as params in the decorator func. In order to retrieve their values we need `env[propertyKey]`, `slackEnv[propertyKey]`, `azureEnv[propertyKey]` etc...

tommoor: We can definitely get the value, `defineMetadata` should enable this – I'm 99% sure

PR #6599, opened 2024-02-26
feat: Archive all notifications

### LINKED ISSUE #6463: Add option to archive all notifications
The API already exists, `notifications.update_all` with `archivedAt` timestamp. I suggest that the settings cog here is converted to a menu with the options:

* Archive all notifications
* Notification settings

image

hmacr:
Closes #6463 

### Changelog
- Add an action "Archive all notifications".
- Add a menu inside `Notifications` popover which contains the `Archive` and `Settings` options.
- Filter the archived / non-archived notifications in the `notifications.list` API.
- Fetch only the non-archived notifications in the `Notifications` popover.

### DIFF
--- a/app/components/ContextMenu/index.tsx
+++ b/app/components/ContextMenu/index.tsx
@@ -38,6 +38,8 @@ export type Placement =
 
 type Props = MenuStateReturn & {
   "aria-label"?: string;
+  /** Reference to the rendered menu div element */
+  menuRef?: React.RefObject<HTMLDivElement>;
   /** The parent menu state if this is a submenu. */
   parentMenuState?: Omit<MenuStateReturn, "items">;
   /** Called when the context menu is opened. */
@@ -52,6 +54,7 @@ type Props = MenuStateReturn & {
 };
 
 const ContextMenu: React.FC<Props> = ({
+  menuRef,
   children,
   onOpen,
   onClose,
@@ -96,16 +99,21 @@ const ContextMenu: React.FC<Props> = ({
     t,
   ]);
 
-  // Perf win – don't render anything until the menu has been opened
+  // Perf win – render an empty menu until the menu has been opened
   if (!rest.visible && !previousVisible) {
-    return null;
+    return <Menu {...rest} />;
   }
 
   // sets the menu height based on the available space between the disclosure/
   // trigger and the bottom of the window
   return (
     <>
-      <Menu hideOnClickOutside={!isMobile} preventBodyScroll={false} {...rest}>
+      <Menu
+        ref={menuRef}
+        hideOnClickOutside={!isMobile}
+        preventBodyScroll={false}
+        {...rest}
+      >
         {(props) => (
           <InnerContextMenu
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/app/components/ContextMenu/index.tsx
+++ b/app/components/ContextMenu/index.tsx
@@ -99,9 +99,9 @@ const ContextMenu: React.FC<Props> = ({
     t,
   ]);
 
-  // Perf win – render an empty menu until the menu has been opened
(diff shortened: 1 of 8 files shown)


app/components/ContextMenu/index.tsx
```diff
@@ -96,16 +99,21 @@ const ContextMenu: React.FC<Props> = ({
     t,
   ]);
 
-  // Perf win – don't render anything until the menu has been opened
+  // Perf win – render an empty menu until the menu has been opened
   if (!rest.visible && !previousVisible) {
-    return null;
+    return <Menu {...rest} />;
```
tommoor, line 104: I'd like to keep this perf gain of rendering nothing here, even rendering lots of "empty" menus adds up with all the hooks etc that are contained within.

Why was this needed? Lets think of another way

hmacr: Initially, the menu was rendered as a `modal` dialog on top of a popover.
In this setup, anytime the menu is opened, the popover was closed which ultimately closed the menu as well. By trial & error, I found that rendering an empty menu did not trigger the popover `hide` method.

Per your suggestion, to keep rendering nothing until the menu is opened, the menu is now rendered as a non-modal dialog which fixes the issue.

app/menus/NotificationMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+        title: "Notification settings",
+        visible: true,
+        onClick: () => performAction(navigateToNotificationSettings, context),
+        icon: <SettingsIcon />,
+      },
+    ],
+    [context]
+  );
+
+  useOnClickOutside(
```
tommoor, line 36: I'm not sure this is achieving anything to be honest

tommoor (commented):
Honestly, seems pretty good – you've found all the patterns! I'd like to not have the icons in the menu in this case, they don't really add anything – but that requires a way to tell the `ContextMenu` not to render icons which we don't have here.

Could be a separate PR?

app/menus/NotificationMenu.tsx
hmacr: Without this, when the menu is open, clicking outside will close both the popup & the menu. Ideally, the user would have hoped to close just the menu.

Is this not working? 🤔

tommoor: That would be ideal, but it didn't seem to work when I tried

hmacr: That's weird, it worked when I tried now.

One case I've observed is, if we click inside the document, it closes all the popovers and menus. Maybe that's what is happening?

**Screencast for reference**
https://github.com/outline/outline/assets/132386067/16e986f8-4881-43f3-8f18-d749f381a3f9

hmacr:
> Honestly, seems pretty good – you've found all the patterns! I'd like to not have the icons in the menu in this case, they don't really add anything – but that requires a way to tell the `ContextMenu` not to render icons which we don't have here.
> 
> Could be a separate PR?

This is done!

Throughout the app, we alternate between using the utility `actionToMenuItem` method and defining a separate menu config from actions. This PR uses both patterns. I can modify it to use one if there is a preferred pattern if needed.

app/menus/NotificationMenu.tsx
tommoor: That must be it, it's ok for now

tommoor:
> Throughout the app, we alternate between using the utility actionToMenuItem method and defining a separate menu config from actions. This PR uses both patterns. I can modify it to use one if there is a preferred pattern if needed.

TLDR: We're slowly moving towards actions as the "standard", so ideally new functionality is created in an action though it's not a strict rule. `actionToMenuItem` is a shim to enable the new actions to work with the format expected by the older `ContextMenu` until enough are converted.

tommoor (approved):

PR #6623, opened 2024-03-02
chore: Plugin registration

tommoor:
Changing from a convention -> configuration approach. For a couple of reasons, 

1) We need plugins to have priority/ordering
1) Remove `requiredEnvVars` and instead of have plugins decide if they are enabled internally
1) Convention is mainly useful in frameworks, here where very few people know the codebase the clarity of configuration is preferred
1) Reducing the amount of directory traversal, which was building up with each "type" of plugin 
1) Strong typing of plugin interfaces


server/models/helpers/AuthenticationHelper.ts
```diff
@@ -1,70 +1,18 @@
...
-            name: name ?? config.name,
-            enabled,
-            router: authProvider,
-          });
-        }
-      });
-
-    this.providersCache = sortBy(authenticationProviderConfigs, "id");
-    return this.providersCache;
+    return PluginManager.getEnabledPlugins(PluginType.AuthProvider);
```
tommoor, line 15: A good example of where this makes consumption much simpler

tommoor (commented):
I believe it makes sense to merge this first and potentially have a similar PR for the frontend components as well.

apoorv-mishra (commented):
Need to fix the build 👇 
```
[backend] Compiling...
[backend] Copying static files...
[backend] Error: Command failed: mkdir -p ./build/plugins/iframely && cp ./plugins/iframely/plugin.json ./build/plugins/iframely/plugin.json
[backend] cp: ./plugins/iframely/plugin.json: No such file or directory
```

server/utils/PluginManager.ts
```diff
@@ -0,0 +1,146 @@
...
+      value,
+      priority: PluginPriority.Normal,
+      ...options,
+    };
+
+    Logger.debug(
+      "plugins",
+      `Plugin ${options.enabled === false ? "disabled" : "enabled"} "${
+        options.id
+      }" (${options.description})`
```
apoorv-mishra, line 78: nit:
```suggestion
      }" ${options.description ? `(${options.description})` : ""}`
```

apoorv-mishra:
Remember to get rid of `glob.sync...` from [here](https://github.com/outline/outline/blob/820e4839d51d8d5f65502bb7ed67c27e29841016/server/emails/templates/index.ts) @tommoor

PR #6635, opened 2024-03-05
fix: show complete page without scrollbar in print view

### LINKED ISSUE #6615: Print wiki page shows a scroll
If I want to print a page using Edge/Chrome it shows a scroll.

image

However in Firefox works well

image

Pranav2612000:
Fixes #6615

### DIFF
--- a/app/hooks/useMobile.ts
+++ b/app/hooks/useMobile.ts
@@ -2,5 +2,7 @@ import { breakpoints } from "@shared/styles";
 import useMediaQuery from "~/hooks/useMediaQuery";
 
 export default function useMobile(): boolean {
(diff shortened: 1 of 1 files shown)


Pranav2612000:
As a temporary fix, I've unset the height when trying to print the document. But a more holistic solution can be to check if we need to set `height: 100vh` for the MobileScrollWrapper

Pranav2612000:
Or we can also update the `useMobile` hook to return false when opening the document for print by updating the condition in `useMobile.ts`

tommoor:
I would do that later one yea, check print in useMobile

Pranav2612000:
@tommoor updated the PR. can you take a look now?

tommoor (approved):

PR #6645, opened 2024-03-06
fix: HTML import escapes dollar signs in code (#6638)

### LINKED ISSUE #6638: HTML import breaks code by escaping dollar signs
The HTML import performs dollar sign escaping over the whole document. This also escapes dollar signs in code (inline and blocks). But in code, other than in regular text, these escapes are preserved when rendering the document.

This also affects the Confluence import, effectively halting our migration effort.

**To Reproduce**
Save the following HTML code as file and import using the *Import document* action.

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Code Block Dollar Signs</title>
        <META http-equiv="Content-Type" content="text/html; charset=UTF-8">
    </head>

    <body>

<p>Here is a shell script:</p>
<pre><code>foo=($BASH $BASH_VERSION)
for i in &quot;${foo[@]}&quot;; do
    printf &#39;%s=%s\n&#39; &#39;\$i&#39; &quot;$i&quot;
done</code></pre>
<p>And this is inline code: <code>i=3; echo "\$i=$i"</code></p>
<p>And this is just an amount of money: $100</p>
<p>And this is an escaped dollar sign: \$</p>

        </body>
</html>
```

This results in the following Markdown:

````markdown
Here is a shell script:

```
foo=(\$BASH \$BASH_VERSION)
for i in "\${foo[@]}"; do
    printf '%s=%s\n' '\\$i' "\$i"
done
```

And this is inline code: `i=3; echo "\\$i=\$i"`

And this is just a an amount of money: \$100

And this is an escaped dollar sign: \\\$
````

Note that all the dollar signs in code are escaped.

**Expected behavior**
Instead, it should be converted into this Markdown:

````markdown
Here is a shell script:

```
foo=($BASH $BASH_VERSION)
for i in "${foo[@]}"; do
    printf '%s=%s\n' '\$i' "$i"
done
```

And this is inline code: `i=3; echo "\$i=$i"`

And this is just a an amount of money: \$100

And this is an escaped dollar sign: \\\$
````

**Screenshots**
n/a

- Install: self hosted
- Version: outlinewiki/outline-enterprise:0.75.3

n/a

**Mobile (please complete the following information):**
n/a

panos--:
Closes #6638


tommoor (approved):
👌 excellent, thankyou

PR #6649, opened 2024-03-07
fix: Restore env validation for plugins

### LINKED ISSUE #6637: Env vars under `plugins` aren't validated anymore
Validation decorators(`IsBoolean`, `IsNotEmpty()` etc...) seemed to have stopped working for env vars under `plugins` after #6597 . Although ones in `server/env.ts` are working fine, reason being [this line](https://github.com/outline/outline/blob/fd34a6d19c6c9190c583f0594339eab1ef81e39c/server/utils/startup.ts#L62).

tommoor:
closes #6637

### DIFF
--- a/server/env.ts
+++ b/server/env.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 // eslint-disable-next-line import/order
 import environment from "./utils/environment";
 import os from "os";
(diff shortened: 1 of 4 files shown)


server/env.ts
```diff
@@ -19,19 +20,20 @@ import Deprecated from "./models/decorators/Deprecated";
...
-    return this.validationPromise;
+    process.nextTick(() => {
+      void validate(this).then((errors) => {
+        if (errors.length > 0) {
+          let output =
+            "Environment configuration is invalid, please check the following:\n\n";
+          output += errors.map(
+            (error) => "- " + Object.values(error.constraints ?? {}).join(", ")
+          );
+          console.warn(output);
```
tommoor, line 32: Unfortunately cannot use `Logger` here as it will result in a dependency loop

PR #6670, opened 2024-03-11
feat: support custom db schema

axelrindle:
As mentioned in #6669, I need support for specifying a custom Postgres schema. This PR seeks to implement this requirement.


server/storage/database.ts
```diff
@@ -11,6 +11,7 @@ const isSSLDisabled = env.PGSSLMODE === "disable";
 const poolMax = env.DATABASE_CONNECTION_POOL_MAX ?? 5;
 const poolMin = env.DATABASE_CONNECTION_POOL_MIN ?? 0;
 const url = env.DATABASE_CONNECTION_POOL_URL || env.DATABASE_URL;
+const schema = env.DATABASE_SCHEMA ?? null;
```
tommoor, line 14:
```suggestion
const schema = env.DATABASE_SCHEMA;
```

Seems like types are complaining that `undefined` is expected here

tommoor:
image

The linter needs satisfying I'm afraid

tommoor (approved):

PR #6677, opened 2024-03-14
`PluginManager` refactor

### LINKED ISSUE #6676: Refactor `PluginManager`
New interface 👇 
```js
PluginManager.add([
  {
    id: "slack",
    type: Hook.AuthProvider,
    value: router,
  },
  {
    type: Hook.API,
    value: hooks,
  },
  {
    type: Hook.Processor,
    value: SlackProcessor,
  },
]);
```

apoorv-mishra:
Closes #6676

### DIFF
--- a/plugins/azure/server/index.ts
+++ b/plugins/azure/server/index.ts
@@ -7,6 +7,10 @@ const enabled = !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET;
 
 if (enabled) {
   PluginManager.add([
-    { ...config, type: PluginType.AuthProvider, value: router },
+    {
+      ...config,
+      type: PluginType.AuthProvider,
+      value: { router, id: config.id },
+    },
   ]);
 }
--- a/plugins/azure/server/index.ts
+++ b/plugins/azure/server/index.ts
@@ -6,11 +6,9 @@ import env from "./env";
 const enabled = !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET;
 
 if (enabled) {
-  PluginManager.add([
-    {
-      ...config,
-      type: PluginType.AuthProvider,
-      value: { router, id: config.id },
-    },
-  ]);
+  PluginManager.add({
+    ...config,
+    type: PluginType.AuthProvider,
+    value: { router, id: config.id },
+  });
 }
--- a/plugins/azure/server/index.ts
+++ b/plugins/azure/server/index.ts
@@ -1,4 +1,4 @@
-import { PluginManager, PluginType } from "@server/utils/PluginManager";
+import { Hook, PluginManager } from "@server/utils/PluginManager";
 import config from "../plugin.json";
 import router from "./auth/azure";
 import env from "./env";
@@ -8,7 +8,7 @@ const enabled = !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET;
 if (enabled) {
   PluginManager.add({
     ...config,
-    type: PluginType.AuthProvider,
+    type: Hook.AuthProvider,
     value: { router, id: config.id },
   });
 }
--- a/plugins/azure/server/index.ts
+++ b/plugins/azure/server/index.ts
@@ -3,7 +3,10 @@ import config from "../plugin.json";
 import router from "./auth/azure";
 import env from "./env";
 
-PluginManager.register(PluginType.AuthProvider, router, {
-  ...config,
-  enabled: !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET,
-});
+const enabled = !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET;
+
+if (enabled) {
+  PluginManager.add([
+    { ...config, type: PluginType.AuthProvider, value: router },
+  ]);
+}
--- a/plugins/webhooks/server/index.ts
+++ b/plugins/webhooks/server/index.ts
@@ -5,7 +5,25 @@ import WebhookProcessor from "./processors/WebhookProcessor";
 import CleanupWebhookDeliveriesTask from "./tasks/CleanupWebhookDeliveriesTask";
 import DeliverWebhookTask from "./tasks/DeliverWebhookTask";
 
-PluginManager.register(PluginType.API, webhookSubscriptions, config)
-  .registerProcessor(WebhookProcessor)
-  .registerTask(DeliverWebhookTask)
-  .registerTask(CleanupWebhookDeliveriesTask);
+PluginManager.add([
+  {
+    ...config,
+    type: PluginType.API,
+    value: webhookSubscriptions,
+  },
+  {
+    ...config,
+    type: PluginType.Processor,
+    value: WebhookProcessor,
+  },
+  {
+    ...config,
+    type: PluginType.Task,
+    value: DeliverWebhookTask,
+  },
+  {
+    ...config,
+    type: PluginType.Task,
+    value: CleanupWebhookDeliveriesTask,
+  },
+]);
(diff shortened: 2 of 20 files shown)


apoorv-mishra:
Keeping this as a draft for now. Exploring some more ideas to improve upon the existing abstractions, for example, how about 👇 
```js
const plugin = new Plugin(config)
plugin.registerHook(Hook.API, apiHook)
plugin.registerHook(Hook.Processor, processorHook)
plugin.registerHook(Hook.Task, taskHook)

// PluginManager
PluginManager.register(plugin)

PluginManager.filterPluginsContaining(Hook.EmailTemplate).forEach((plugin) => {
  emails[plugin.id] = plugin.getHook(Hook.EmailTemplate);
});
```

Plugin object looks something like 👇 
```js
{
  id: "slack",
  name: "Slack",
  description: "Slack...",
  enabled: true,
  priority: 20,
  hooks: {
    [Hook.API]: apiHook,
    [Hook.Auth]: authHook,
    [Hook.Processor]: processorHook,
    [Hook.Uninstall]: uninstallHook,
  }
}
```

Let me know your thoughts @tommoor

tommoor:
On the idea above – not sure introducing the additional level of a "plugin" gives us much, it just makes the hooks more annoying to access :P. I think we should just double down on these things being called "hooks" – plugins are the folder, plugins can register "hooks".

server/utils/PluginManager.ts
```diff
@@ -52,83 +53,33 @@ export type Plugin<T extends PluginType> = {
...
-  public static register<T extends PluginType>(
-    type: T,
-    value: PluginValueMap[T],
-    options: Omit<Plugin<T>, "value"> = {
-      id: uuid(),
-    }
-  ) {
-    if (!this.plugins.has(type)) {
-      this.plugins.set(type, []);
+  public static add(plugins: Array<Plugin<PluginType>>) {
```
tommoor, line 66: Suggest we make this so that it can accept an array or singular `Plugin` as singular is a very common case.

tommoor, line 66: I'm pretty set on `PluginType` being `Hook` or `HookType` now

plugins/webhooks/server/index.ts
```diff
@@ -5,7 +5,25 @@ import WebhookProcessor from "./processors/WebhookProcessor";
 import CleanupWebhookDeliveriesTask from "./tasks/CleanupWebhookDeliveriesTask";
 import DeliverWebhookTask from "./tasks/DeliverWebhookTask";
 
-PluginManager.register(PluginType.API, webhookSubscriptions, config)
-  .registerProcessor(WebhookProcessor)
-  .registerTask(DeliverWebhookTask)
-  .registerTask(CleanupWebhookDeliveriesTask);
+PluginManager.add([
+  {
+    ...config,
```
tommoor, line 10: Do we even need config spread on all these? 🤔

server/queues/tasks/index.ts
```diff
@@ -13,7 +13,7 @@ requireDirectory<{ default: BaseTask<any> }>(__dirname).forEach(
   }
 );
 
-PluginManager.getEnabledPlugins(PluginType.Task).forEach((plugin) => {
+PluginManager.getPlugins(PluginType.Task).forEach((plugin) => {
   tasks[plugin.id] = plugin.value;
```
tommoor, line 17:
```suggestion
  tasks[plugin.value.name] = plugin.value;
```

This was done before in `registerTask`

server/queues/processors/index.ts
```diff
@@ -13,7 +13,7 @@ requireDirectory<{ default: BaseProcessor }>(__dirname).forEach(
   }
 );
 
-PluginManager.getEnabledPlugins(PluginType.Processor).forEach((plugin) => {
+PluginManager.getPlugins(PluginType.Processor).forEach((plugin) => {
   processors[plugin.id] = plugin.value;
```
tommoor, line 17:
```suggestion
  processors[plugin.value.name] = plugin.value;
```

This was done before in `registerProcessor`

server/utils/PluginManager.ts
```diff
@@ -142,16 +93,6 @@ export class PluginManager {
     return sortBy(this.plugins.get(type) || [], "priority") as Plugin<T>[];
   }
 
-  /**
-   * Returns all the enabled plugins of a given type in order of priority.
-   *
-   * @param type The type of plugin to filter by
-   * @returns A list of plugins
-   */
-  public static getEnabledPlugins<T extends PluginType>(type: T) {
```
tommoor, line 151: `getPlugins` -> `getHooks`

plugins/webhooks/server/index.ts
apoorv-mishra: We shouldn't, but since `id` is required in type `Plugin`, it's to please the `tsc`. Agree that it's confusing though. By the way, the aforementioned idea eliminates this problem without an off-putting hack 😉

plugins/azure/server/index.ts
```diff
@@ -1,9 +1,14 @@
...
 
-PluginManager.register(PluginType.AuthProvider, router, {
-  ...config,
-  enabled: !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET,
-});
+const enabled = !!env.AZURE_CLIENT_ID && !!env.AZURE_CLIENT_SECRET;
+
+if (enabled) {
+  PluginManager.add({
+    ...config,
```
tommoor, line 10: So we should be able to finally get rid of these spreads now?

apoorv-mishra: We might still need `priority`, `name`, `description` etc to pass as params. Sorry, is the end goal to completely get rid of the JSON config and instead pass these explicitly wherever required?

tommoor: True on priority. My thinking with the `.json` is that it's a good safe way to share config when both the client and server need to share information. In the cases where a plugin doesn't have any client component then it's not _required_ to have a json

apoorv-mishra: >In the cases where a plugin doesn't have any client component then it's not required to have a json

I suppose then the `...config` doesn't hurt? Keep it as a convention?

server/utils/PluginManager.ts
```diff
@@ -35,100 +35,58 @@ export enum PluginType {
...
+  [Hook.UnfurlProvider]: UnfurlSignature;
 };
 
-export type Plugin<T extends PluginType> = {
+export type Plugin<T extends Hook> = {
+  /** Plugin type */
+  type: T;
   /** A unique ID for the plugin */
-  id: string;
+  id?: string;
```
tommoor, line 50: I think I would remove `id` from the type now, to avoid confusion with the `id` in auth provider value

tommoor (approved):
Looks good, modulo that one comment

server/utils/PluginManager.ts
apoorv-mishra: Screenshot 2024-03-16 at 19 34 56 1

Phew! Looks like `id` is deeply integrated...

apoorv-mishra: (not posed as a problem, now it's just `config.value.id`)

apoorv-mishra: Screenshot 2024-03-16 at 19 39 33

`id` then has to be removed from many other places, like ☝️

tommoor: > Phew! Looks like id is deeply integrated...

Actually it's a good job tsc caught this.

> id then has to be removed from many other places, like ☝️

Do it

server/routes/api/authenticationProviders/authenticationProviders.ts
```diff
@@ -89,13 +89,15 @@ router.post(
...
-        const row = teamAuthenticationProviders.find((t) => t.name === p.id);
+        const row = teamAuthenticationProviders.find(
+          (t) => t.name === p.value.id
+        );
 
         return {
-          id: p.id,
-          name: p.id,
+          id: p.value.id,
+          name: p.value.id,
```
apoorv-mishra, line 100: Thanks for `id` removal suggestion 🙏

tommoor (approved):

PR #6696, opened 2024-03-21
Helper for cache related utilities

### LINKED ISSUE #6695: Move unfurl caching out of iframely plugin
We're about to have more than one unfurl provider, at which point it no longer makes sense to have the caching layer in the iframely plugin – instead it should be at the next level up, within the router – or more likely a helper called from the router.

apoorv-mishra:
Closes #6695

### DIFF
--- a/server/routes/api/urls/urls.ts
+++ b/server/routes/api/urls/urls.ts
@@ -76,7 +76,9 @@ router.post(
     }
 
     // External resources
-    const cachedData = await CacheHelper.getData(CacheHelper.getUnfurlKey(url));
+    const cachedData = await CacheHelper.getData(
+      CacheHelper.getUnfurlKey(url, actor.teamId)
+    );
     if (cachedData) {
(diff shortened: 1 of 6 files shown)


server/routes/api/urls/urls.ts
```diff
@@ -75,12 +76,24 @@ router.post(
...
+      const data = await plugin.value.unfurl(url);
       if (data) {
-        return "error" in data
-          ? (ctx.response.status = 204)
-          : (ctx.body = presentUnfurl(data));
+        if ("error" in data) {
+          return (ctx.response.status = 204);
+        } else {
+          await CacheHelper.setData(
+            CacheHelper.getUnfurlKey(url),
```
tommoor, line 91: As the cache now includes authenticated integrations the cache key must include the team ID to stop the cache being shared between workspaces.

Another advantage of this would be that we can clear the cache related to a specific team when they enable new unfurl integrations.

tommoor (approved):

PR #6713, opened 2024-03-25
Refactor GitHub Integration

### LINKED ISSUE #6712: Refactor GitHub Integration
Some problems surfacing here

- Inconsistent contracts exposed due to `GitHubUser` & `GitHubApp` classes. Would be better to expose something like,
   ```js
   GitHub.authenticateAsUser(code, state)
   GitHub.authenticateAsInstallation(installationId)
   ```
   This goes well with the [documentation](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/about-authentication-with-a-github-app).
- It'd be good to have transformers to transform GitHub API responses to data format we need. Currently, this is all being done ad-hoc within `unfurl`. This would yield itself unmaintainable if we decide to support unfurling for more resources(commits, comments etc.) going forward.

apoorv-mishra:
Closes #6712


plugins/github/server/uninstall.ts
```diff
@@ -9,7 +9,8 @@ export async function uninstall(
     const installationId = integration.settings?.github?.installation.id;
 
     if (installationId) {
-      return githubApp.deleteInstallation(installationId);
+      const client = await GitHub.authenticateAsInstallation(installationId);
+      await GitHub.requestAppUninstall(installationId, client);
```
tommoor, line 13: This wrapper has ended up kind of unnecessary, and it's a bit of an awkward DX to pass `client` as second param – could just use octokit directly here?

plugins/github/server/github.ts
```diff
@@ -1,73 +1,295 @@
...
-  /** Required to authenticate as GitHub App */
-  private static id = env.GITHUB_APP_ID;
-  private static key = env.GITHUB_APP_PRIVATE_KEY
+enum Resource {
+  PR = "pull",
+  Issue = "issue",
+}
+
+type PreviewData = {
+  [Resource.PR]: {
```
tommoor, line 21: You can use the `UnfurlResponse` type for these, it's generic so the shape of `meta` can be passed in

plugins/github/server/api/github.ts
```diff
@@ -65,13 +66,14 @@ router.get(
       }
     }
 
-    const githubUser = new GitHubUser({ code: code!, state: teamId });
+    const client = await GitHub.authenticateAsUser(code!, teamId);
+    const installationsByUser = await GitHub.requestAppInstallations(client);
```
tommoor, line 70: Isn't this less performant than before? Why change to loading all installations 🤔

plugins/github/server/uninstall.ts
apoorv-mishra: Can use directly, yes. I just stuck to keeping all `octokit.request()` calls within the class with the intent of not having to import Octokit in other places. Just trying to set a convention but flexible, yes.

plugins/github/server/api/github.ts
apoorv-mishra: We were always loading all the installations. There doesn't seem to be an API to load a single installation for a given user.

plugins/github/server/uninstall.ts
tommoor: Lets use this pattern and extend the client perhaps:

https://github.com/octokit/core.js/#defaults

Will allow doing `await client.requestAppUninstall()`

tommoor (approved):

plugins/github/server/github.ts
apoorv-mishra: Resolving types separately altogether in another pass.

PR #6719, opened 2024-03-26
build: Vite - set rollup output filenames

vio:
This PR explicitly sets the format of the filenames to `[name].[hash][ext]`.

----

Rollup 4 (Vite 5)  changed the default asset/chunk/entry filenames format to `[name]-[hash][ext]`, and the hashing algorithm is now base64. Since a base64 hash can contain `-`, bundle-stats is not able to safely extract the hash and compare the assets by filename:

image

- [More info about bundle-stats hash extraction for Vite/Rollup](https://relative-ci.com/documentation/guides/vite-config/#use-a-common-pattern-with-content-hashes-for-filenames)
- [Rollup 5 proposed changes](https://github.com/rollup/rollup/issues/5389#issuecomment-1938251936)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -172,6 +172,11 @@ export default () =>
         input: {
           index: "./app/index.tsx",
         },
+        output: {
+          assetFileNames: 'assets/[name].[hash][extname]',
+          chunkFileNames: 'assets/[name].[hash].js',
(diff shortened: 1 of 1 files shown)


tommoor (approved):
Thanks 🙏

PR #6721, opened 2024-03-27
chore: Migrate user `role` to new column

tommoor:
In order to support a new `guest` role first we need to transition the existing booleans to an enum.

### This PR
- [x] Add new database column
- [x] Double write to existing and new field
- [x] Backfill rows

### Future PR's
- [x] Change `isViewer` and `isAdmin` to getters
- [ ] Remove old columns
- [ ] Make `role` not nullable

towards #2946

Issue #2946, opened 2022-01-17
Guest role with restricted permissions


M6Ka7:
If I invite several external users to my outline installation, everyone could see other users who have viewed / edited an article. When you invite external users from company A and from company B they can see each other 
 - on articles viewed in the past
 - on articles edited
 - on articles viewed at the moment
 - in the settings menu (groups/members)
 - maybe other pages

which would be an GDPR incident.

**To Reproduce**
1. Create an external user by inviting per mail 
2. Give the user minimal access (viewer)
3. Login with "magic link"
4. Visit settings page and then members or groups
5. Visit an arbitrary page and click the avatar button to see
    - who is online (now)
    - who viewed this page
    - who edited this page

**Expected behavior**
External users must not see others to avoid GDPR penalties. There should also be an feature to disable the avatar - button on 
- user
- role
- login type
- global

level and forbid access to settings menu. 

If the disabling is not possible, the information should be anonymized e.g. 
- 3 people are watching this article now (instead of User A, User B, User C)
- 24 people viewed/edited this article.

**Screenshots**
<img width="634" alt="image" src="https://user-images.githubusercontent.com/97874931/149734507-ce97a07a-ea7c-4e0b-a0f8-ed7cb3b3cfb5.png">
The screenshot is made from an external user which can see other (external users) marked with red rectangle.

<img width="655" alt="image" src="https://user-images.githubusercontent.com/97874931/149735615-f7e1b9a0-2337-49e8-bf33-7f78d2343c90.png">

**Outline (please complete the following information):**
- Install: self hosted
- Version: v0.61.1

Shide:
I was testing Outline for my company and I've found the same issue here.

MT-167 @moduon

michaelkarrer81:
+1 This may be a blocking issue to use outline in my company.

rafaelbn:
👍 This is a blocking issue

tommoor:
It sounds like the ask here is for a "guest" role which is being conflated with viewer.

Guests would have much more restricted data access, potentially only being able to access collections that they are explicitly added to and not have access to settings screens. Does this sound right?

M6Ka7:
The guest role seems to be the right direction for me. But does not cover it 100%.

> The background is: The **G**eneral **D**ata **P**rotection **R**egulation, which applies across the EU, threatens up to €20 million (about $22 million USD) for exposure of sensitive (identifying) data. 

[...]

PR #6724, opened 2024-03-27
Convert `isViewer` and `isAdmin` to getters

tommoor:
Continuation of https://github.com/outline/outline/pull/6721

towards https://github.com/outline/outline/issues/2946

### DIFF
--- a/server/models/User.ts
+++ b/server/models/User.ts
@@ -137,14 +137,6 @@ class User extends ParanoidModel<
   @Column
   name: string;
 
-  @Default(false)
-  @Column
-  isAdmin: boolean;
-
-  @Default(false)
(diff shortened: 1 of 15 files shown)


server/models/User.ts
```diff
@@ -562,21 +566,9 @@ class User extends ParanoidModel<
...
     if (res.count >= 1) {
       if (to === UserRole.Member) {
-        await this.update(
-          {
-            isAdmin: false,
-            isViewer: false,
-          },
-          options
-        );
+        await this.update({ role: to }, options);
```
tommoor, line 569: I'm sure this can all be simplified further now, but doing the minimum to move this forward

Issue #6650, opened 2024-03-07
Incompatible LaTeX syntax


dreamerblue:
**To Reproduce**
1. input regular inline/block LaTeX or copy markdown text containing LaTeX from other editors.
2. Outline doesn't use the general LaTeX syntax (`$` for inline, `$$` for block) to render.

**Expected behavior**
use `$` for inline, `$$` for block. This is the widely used LaTeX syntax. Please **DO NOT** use custom syntax like `$$` or `$$$`, which is not compatible.

Outline's custom syntax is not recognized by most other markdown editors when copying from other editors or pre-existing markdown documents for import, or export to markdown. This caused a lot of problems for us.

The biggest issues at hand are:

1. LaTeX texts not recognized correctly if imported from a document source of another specification, and Outlint doesn't have any automatic conversion to fix it
<img width="944" alt="image" src="https://github.com/outline/outline/assets/19774268/a791e568-7589-4162-91cf-0a43e535f503">

2. if working on Outline and using custom syntax `$$` and `$$$` to get a superficially acceptable rendering, when exporting as markdown and distributing it on other platforms, the exported markdown still retains the `$$` and `$$$`, which cannot render correctly on other platforms.
<img width="1128" alt="image" src="https://github.com/outline/outline/assets/19774268/5e824cf5-8837-4515-b43a-a9f65788e0c5">

**Outline (please complete the following information):**
- Install: [self hosted]
- Version: [7685637729b167701c3bd8921ceeb144e29e6e71] (0.75.2)

**Expected solution**

Deprecate the custom syntax and use the general `$`/`$$` to render LaTeX, including typing, copying and pasting, importing documents and exporting to markdown.

**Test cases**

```markdown
### Inline

$x = y$

### Block

$$
y = x
$$
```

Wrong render result: copy and paste to Outline

<img width="539" alt="image" src="https://github.com/outline/outline/assets/19774268/6f436101-cf25-4b74-bfdc-11e8366717fa">

Correct result: from popular markdown editors like Typora

<img width="573" alt="image" src="https://github.com/outline/outline/assets/19774268/b1ca90aa-a792-4b2c-9267-53fc76d0cb03">

Another correct result: GitHub's built-in markdown renderer

### Inline

$x = y$

### Block

$$
y = x
$$

tommoor:
> Deprecate the custom syntax and use the general $/$$ to render LaTeX, including typing, copying and pasting, importing documents and exporting to markdown.

[...]

PR #6727, opened 2024-03-27
fix: Adjust math delimiter to fix markdown import/export compatibility issue

dreamerblue:
This PR is expected to fix the markdown import/export compatibility issue in #6650.

### Case 1: Import

Here is the test case, save as `.md` and use it to import document:

```plain
## Inline

Here is inline $x=y$

## Block

$$
y=x
$$
```

Excepted result in Outline editor:

image

All general LaTeX texts should be converted to Outline `math_inline` or `math_block`.

### Case 2: Input and view in editor

This PR won't modify the input shortcut and trigger key. So the original trigger points (`$$` for inline, ```$$$ + Space``` for block) are preserved. And all imported or manaually inputed math nodes is still correct in `state`, only the generated `text` is changed.

### Case 3: Export

Export an Outline document which contains math inline/block to markdown.

The result `.md` shoule be like (All formulas can be rendered correctly in any other editors):

image

## Other notes

I'm new to contributing to this project, and after reading some of the source code and trying to modify it, I verified these test cases above locally.

This modification does not automatically fix documents that already have custom text cached in the Outline database (e.g. `$$$`), it only works for new incremental documents. Perhaps the core maintainer could help add a batch function for auto-fixing. The old text cache may only be affected if the export function is used. Viewing and editing in Outline still works fine for older documents.

### DIFF
--- a/shared/editor/nodes/Math.ts
+++ b/shared/editor/nodes/Math.ts
@@ -69,9 +69,9 @@ export default class Math extends Node {
   }
 
   toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
-    state.write("$$");
+    state.write("$");
     state.text(node.textContent, false);
-    state.write("$$");
+    state.write("$");
   }
 
   parseMarkdown() {
--- a/shared/editor/nodes/MathBlock.ts
+++ b/shared/editor/nodes/MathBlock.ts
@@ -41,10 +41,10 @@ export default class MathBlock extends Node {
   }
 
   toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
-    state.write("$$$\n");
+    state.write("$$\n");
     state.text(node.textContent, false);
     state.ensureNewLine();
-    state.write("$$$");
+    state.write("$$");
     state.closeBlock(node);
   }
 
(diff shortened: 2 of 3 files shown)


dreamerblue:
Hi, any suggestions or reviews?

tommoor:
As I kind of mentioned on the one problem is that the `$` trigger character does not work well in a WYSIWYG editor, it causes too many false positives that convert standard paragraphs with dollar symbols in them into inline latex. This was the original reason that the _trigger_ was changed in the first place, so I do stand by that remaining the same.

I think now that #3000 is closed we could make a change that allows parsing the standard syntax on paste, and exporting the standard syntax on export?

dreamerblue:
> As I kind of mentioned on the one problem is that the `$` trigger character does not work well in a WYSIWYG editor, it causes too many false positives that convert standard paragraphs with dollar symbols in them into inline latex. This was the original reason that the _trigger_ was changed in the first place, so I do stand by that remaining the same.
> 
> I think now that #3000 is closed we could make a change that allows parsing the standard syntax on paste, and exporting the standard syntax on export?

Yes, I agree. Due to the #3000 reason, we should only make changes for importing, exporting and pasting to support standard markdown syntax. As I mentioned at the case 2 of this PR, it's better to keep the current behaviour for inputing.

dreamerblue:
any review?

dreamerblue:
Oh, any updates or suggestions? It's indeed a serious bug for LaTeX users.

tommoor (approved):

PR #6730, opened 2024-03-28
Refactor unfurling related types

### LINKED ISSUE #6729: Refactor types related to unfurling
Some of these types are getting hard to work with all the generics stuff. Hoping to prevent the same fate as that of integration settings 🤞

apoorv-mishra:
Closes #6729


server/routes/api/urls/urls.ts
```diff
@@ -53,7 +53,11 @@ router.post(
       authorize(actor, "read", user);
       authorize(actor, "read", document);
 
-      ctx.body = await presentMention(user, document);
+      ctx.body = await presentUnfurl({
```
tommoor, line 56: I'm not sure if moving these into a single presenter is beneficial, you've lost type safety here that would be retained simply by using `presentMention` directly?

server/presenters/unfurl.ts
```diff
@@ -0,0 +1,209 @@
...
+  url: data.html_url,
+  type: UnfurlResourceType.Issue,
+  id: `#${data.number}`,
+  title: data.title,
+  description: data.body_text,
+  author: {
+    name: data.user.login,
+    avatarUrl: data.user.avatar_url,
+  },
+  labels: data.labels.map((label: { name: string; color: string }) => ({
```
tommoor, line 98: While overall I like the changes in this PR it's disappointing that the "custom" metadata fields are now all mixed in with the standard ones. I did like the clarity of separation between those fields that are expected on all responses

server/routes/api/urls/urls.ts
apoorv-mishra: Well, the long-term goal is to eliminate need for treating "unfurling" separately for mentions, documents and external resources. The following pattern seems to be emerging:

- Parse URL
- Fetch resource(Database/REST)
- Present

Collating into a single presenter was an initial step.

server/presenters/unfurl.ts
apoorv-mishra: Rationale here was to eliminate the barrier of decision-making - what should go inside `meta` and what not. Should `author` be inside `meta`? What about `createdAt`?

To be honest, `meta` just seemed to be a "hack" attribute to dump in resource specific attributes. Now that we've come with different types altogether, it perhaps eliminates the need for it? We still have the flexibility to put it for upcoming integrations if we've good reasons.

server/routes/api/urls/urls.ts
tommoor: > Well, the long-term goal is to eliminate need for treating "unfurling" separately for mentions, documents and external resources.

True enough,

tommoor (approved):

PR #6731, opened 2024-03-28
Migrate from s3 sdk v2 to v3

### LINKED ISSUE #5683: Upgrade AWS SDK to v3
```
[backend] [api] (node:8956) NOTE: We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023.
[backend] [api]
[backend] [api] Please migrate your code to use AWS SDK for JavaScript (v3).
[backend] [api] For more information, check the migration guide at https://a.co/7PzMCcy
```

lng2020:
close #5683. Nothing breaking, test with my local setup and everything works.
Noticeable change:
The s3 SDK v3 method `PutObjectCommand` doesn't support stream input and it requires a `content-length` field to be set, which is discussed in https://github.com/aws/aws-sdk-js-v3/issues/2348. 
So I didn't use `PutObjectCommand` because I think loading whole files into the buffer may cause a burden. As the [issue comment](https://github.com/aws/aws-sdk-js-v3/issues/2348#issuecomment-841002347) above suggests, I use the `Upload` class from `@aws-sdk/lib-storage` which is compatible with v2.
> Just in case anyone needs a clear answer: putObject can't write a stream to S3, you have to use the Upload class in @aws-sdk/lib-storage


lng2020:
Need to polish about `signature-v4-crt`

tommoor:
Appreciate the help here 🙏

package.json
```diff
@@ -47,6 +47,11 @@
     "> 0.25%, not dead"
   ],
   "dependencies": {
+    "@aws-sdk/client-s3": "3.535.0",
+    "@aws-sdk/lib-storage": "3.535.0",
+    "@aws-sdk/s3-presigned-post": "3.535.0",
```
lng2020, line 52: My local env can work without importing as expected. But CI cannot find this module so I imported it here. Don't know the root cause and I found a similar issue with `aws-sdk/middleware-endpoint` before(https://github.com/aws/aws-sdk-js-v3/issues/3983).

server/storage/files/S3Storage.ts
```diff
@@ -1,6 +1,18 @@
...
-import AWS, { S3 } from "aws-sdk";
+import { Readable } from "stream";
+import {
+  S3Client,
+  DeleteObjectCommand,
+  GetObjectCommand,
+  ObjectCannedACL,
+} from "@aws-sdk/client-s3";
+import { Upload } from "@aws-sdk/lib-storage";
+import "@aws-sdk/signature-v4-crt";
```
lng2020, line 10:
```suggestion
import "@aws-sdk/signature-v4-crt"; // https://github.com/aws/aws-sdk-js-v3#functionality-requiring-aws-common-runtime-crt
```

server/storage/files/S3Storage.ts
```diff
@@ -13,14 +25,14 @@ export default class S3Storage extends BaseStorage {
...
-      secretAccessKey: env.AWS_SECRET_ACCESS_KEY,
+    this.client = new S3Client({
+      forcePathStyle: env.AWS_S3_FORCE_PATH_STYLE,
+      credentials: {
+        accessKeyId: env.AWS_ACCESS_KEY_ID || "",
+        secretAccessKey: env.AWS_SECRET_ACCESS_KEY || "",
+      },
       region: env.AWS_REGION,
       endpoint: this.getEndpoint(),
-      signatureVersion: "v4",
```
lng2020, line 23: This condition is deprecated and now is replaced by `import "@aws-sdk/signature-v4-crt";` according to https://github.com/aws/aws-sdk-js-v3#functionality-requiring-aws-common-runtime-crt

lng2020:
ready for review

tommoor:
Thanks, I plan to test this week

tommoor:
So, there is a bug here – when downloading a document that includes an embedded image (Document -> ... -> Download -> Markdown), the image in the generated zip file is 0 bytes.

This would suggest that [getFileStream](https://github.com/outline/outline/blob/cf6a946c9c5597bb197fdba35f65d548ade8342d/server/storage/files/S3Storage.ts#L196) is not working correctly. I checked out `main` to test it continues to work correctly there.

server/storage/files/S3Storage.ts
```diff
@@ -193,29 +210,29 @@ export default class S3Storage extends BaseStorage {
...
+  public getFileStream(key: string): NodeJS.ReadableStream | null {
     invariant(
       env.AWS_S3_UPLOAD_BUCKET_NAME,
       "AWS_S3_UPLOAD_BUCKET_NAME is required"
     );
 
-    try {
-      return this.client
-        .getObject({
+    this.client
```
tommoor, line 219: The response is no longer returned, can we restore this to async/await style?

tommoor:
I really don't want to change the signature of the storage interface here, it's weird to have a method called `getFileStream` that returns a promise :/ It feels like it was very close before the last round of changes with the exception of that one issue

lng2020:
> can we restore this to async/await style?

I assume this means using async/await in `getFileStream` and results in a promise returned so I change the signature. Could u kindly specify what the right approach is?

tommoor:
I think you're right, this is a terrible limitation of the new version – how annoying.

server/storage/files/S3Storage.ts
```diff
@@ -194,28 +194,29 @@ export default class S3Storage extends BaseStorage {
...
 
-  public getFileStream(key: string): NodeJS.ReadableStream | null {
+  public getFileStream(key: string): Promise<NodeJS.ReadableStream | null> {
     invariant(
       env.AWS_S3_UPLOAD_BUCKET_NAME,
       "AWS_S3_UPLOAD_BUCKET_NAME is required"
     );
 
-    this.client
+    return this.client
```
tommoor, line 219: This seemed to be a bug – nothing returned here, but with this change I think it's ready to merge

PR #6741, opened 2024-03-30
Policies refactor

tommoor:
Policies are impossible to work on, this makes them nicer?

### DIFF
--- a/server/policies/group.ts
+++ b/server/policies/group.ts
@@ -10,6 +10,14 @@ allow(User, "createGroup", Team, (actor, team) =>
   )
 );
 
+allow(User, "listGroups", Team, (actor, team) =>
+  and(
+    //
+    isTeamModel(actor, team),
+    !actor.isGuest
+  )
+);
+
 allow(User, "read", Group, (actor, team) =>
   and(
     //
--- a/server/policies/group.ts
+++ b/server/policies/group.ts
@@ -3,7 +3,11 @@ import { allow } from "./cancan";
 import { and, isTeamAdmin, isTeamModel, isTeamMutable } from "./utils";
 
 allow(User, "createGroup", Team, (actor, team) =>
-  and(isTeamAdmin(actor, team), isTeamMutable(actor))
+  and(
+    //
+    isTeamAdmin(actor, team),
+    isTeamMutable(actor)
+  )
 );
 
 allow(User, "listGroups", Team, (actor, team) =>
@@ -23,5 +27,9 @@ allow(User, "read", Group, (actor, team) =>
 );
 
 allow(User, ["update", "delete"], Group, (actor, team) =>
-  and(isTeamAdmin(actor, team), isTeamMutable(actor))
+  and(
+    //
+    isTeamAdmin(actor, team),
+    isTeamMutable(actor)
+  )
 );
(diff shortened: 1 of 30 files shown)


server/policies/group.ts
```diff
@@ -1,37 +1,27 @@
...
-allow(User, "createGroup", Team, (actor, team) => {
-  if (!team || actor.isGuest || actor.isViewer || actor.teamId !== team.id) {
-    return false;
-  }
-  if (actor.isAdmin) {
-    return true;
-  }
+allow(User, "createGroup", Team, (actor, team) =>
+  and(
+    //
```
tommoor, line 7: These comments are just to force vertical formatting otherwise prettier will place on a single line which is undesirable here.

server/policies/user.ts
```diff
@@ -1,114 +1,67 @@
...
+allow(User, "read", User, isTeamModel);
 
-  throw AdminRequiredError();
-});
-
-allow(User, "readDetails", User, (actor, user) => {
-  if (!user || user.teamId !== actor.teamId) {
-    return false;
-  }
-  if (user === actor) {
```
tommoor, line 68: Found a bug here, these will never have equality

server/policies/user.ts
```diff
@@ -1,114 +1,67 @@
...
-  }
-  if (actor.isAdmin) {
-    return true;
-  }
+allow(User, "inviteUser", Team, (actor, team) =>
+  and(
+    isTeamModel(actor, team),
+    !actor.isGuest,
+    !actor.isViewer,
+    actor.isAdmin || !!team?.getPreference(TeamPreference.MembersCanInvite)
```
apoorv-mishra, line 21: `or` can replace `||`?

server/policies/user.ts
```diff
@@ -1,114 +1,67 @@
...
+  )
+);
 
-  throw AdminRequiredError();
-});
+allow(User, ["update", "delete", "readDetails"], User, (actor, user) =>
+  and(
+    //
+    isTeamModel(actor, user),
+    actor.isAdmin || actor.id === user?.id
```
apoorv-mishra, line 29: `or` can replace `||`?

server/policies/user.ts
tommoor: Yea, I'm not sure I need to be religious about it if it's a simple condition but definitely works best if there are a bunch of OR statements

server/policies/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { User, Team } from "@server/models";
+import Model from "@server/models/base/Model";
+
+export function and(...args: boolean[]) {
+  return args.every(Boolean);
+}
+
+export function or(...args: boolean[]) {
+  return args.some(Boolean);
+}
```
apoorv-mishra, line 10: Much simpler than I expected 👍

server/policies/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Check if the actor is present in the same team as the model.
+ *
+ * @param actor The actor to check
+ * @param model The model to check
+ * @returns True if the actor is in the same team as the model
+ */
+export function isTeamModel(actor: User, model: Team | Model | null) {
```
apoorv-mishra: Don't quite get the name here. Could have been `belongsTo` or something similar?

PR #6772, opened 2024-04-06
feat: Add button to empty trash

### LINKED ISSUE #6215: Add button to empty trash
- This functionality should only be available to admins
- Should include a confirmation prompt
- Needs a new API endpoint so we don't need to make hundreds of API requests

hmacr:
Closes #6215

**Changelog**
- Add an API `documents.empty_trash`.
- Add a button in the Trash screen. This will trigger a confirmation prompt.
- Add a new section `Trash` in cmd bar. This includes the `Empty` action.


server/routes/api/documents/documents.ts
```diff
@@ -1701,4 +1701,48 @@ router.post(
   }
 );
 
+router.post(
+  "documents.empty_trash",
+  auth({ admin: true }),
+  validate(T.DocumentsEmptyTrashSchema),
+  async (ctx: APIContext<T.DocumentsEmptyTrashReq>) => {
+    const { user } = ctx.state.auth;
+    const { ids } = ctx.input.body;
```
tommoor, line 1710: We don't need to send `ids`, it's not particularly likely that the client will have the ids of every document in the trash in fact. Better to have the server find deleted docs that the user has access to

server/routes/api/documents/documents.ts
```diff
@@ -1701,4 +1701,48 @@ router.post(
   }
 );
 
+router.post(
+  "documents.empty_trash",
+  auth({ admin: true }),
```
tommoor, line 1706: Just changed this to

```suggestion
  auth({ role: UserRole.Admin }),
```

server/routes/api/documents/documents.ts
```diff
@@ -1701,4 +1701,48 @@ router.post(
...
+      where: {
+        id: {
+          [Op.in]: ids,
+        },
+      },
+      paranoid: false,
+    });
+
+    const documentIds = documents.map((document) => document.id);
+    await Document.update(
```
tommoor, line 1721: If this is needed perhaps it should probably be moved into `documentPermanentDeleter` but I'm not sure it its… presumably you saw an issue?

app/actions/definitions/trash.tsx
```diff
@@ -0,0 +1,28 @@
+import { TrashIcon } from "outline-icons";
+import React from "react";
+import DeleteDocumentsInTrash from "~/scenes/Trash/components/DeleteDocumentsInTrash";
+import { createAction } from "~/actions";
+import { TrashSection } from "~/actions/sections";
+import { trashPath } from "~/utils/routeHelpers";
+
+export const permanentlyDeleteDocumentsInTrash = createAction({
+  name: ({ t }) => t("Empty"),
+  analyticsName: "Permanently delete all documents in Trash",
```
tommoor, line 10:
```suggestion
  analyticsName: "Empty trash",
```

server/routes/api/documents/documents.ts
hmacr: Oh yeah, pagination.

So, the API would need to send the deleted ids back to the UI (to remove those docs from the store). Otherwise, the UI has to refresh all collections & docs to fetch the latest state (which seems un-necessary).
Is this an acceptable approach?

server/routes/api/documents/documents.ts
hmacr: I didn't really.
This is present in the individual document permanent delete API  ([code ref](https://github.com/outline/outline/blob/main/server/routes/api/documents/documents.ts#L1216)); so used it here in case it was needed for a special case.
Is this not needed henceforth?

server/routes/api/documents/documents.ts
tommoor: I'm not sure if it needs to send the ids back as the client can just remove any deleted documents it has?

server/routes/api/documents/documents.ts
tommoor: Managed to trace this back to the original implementation (https://github.com/outline/outline/commit/9fccc280d72e91e5cc8efbed73c04095f9be493a#diff-ce4bb28c7d77e61b5a4348fcda078c27194ddb25b68df1d0b2e34fdc3ef26ad9R1189), I think it's needed – but can be moved into the `documentPermanentDeleter` command so that it is shared

hmacr:
@tommoor I've addressed the review; please check.

app/actions/definitions/trash.tsx
```diff
@@ -0,0 +1,28 @@
+import { TrashIcon } from "outline-icons";
+import React from "react";
+import DeleteDocumentsInTrash from "~/scenes/Trash/components/DeleteDocumentsInTrash";
+import { createAction } from "~/actions";
+import { TrashSection } from "~/actions/sections";
+import { trashPath } from "~/utils/routeHelpers";
+
+export const permanentlyDeleteDocumentsInTrash = createAction({
```
tommoor, line 8: Sorry, I didn't notice the first round of review – this belongs in `actions/definitions/documents` just fine as we're acting on documents here.

app/actions/definitions/trash.tsx
```diff
@@ -0,0 +1,28 @@
...
+  name: ({ t }) => t("Empty"),
+  analyticsName: "Empty Trash",
```
tommoor, line 10:
```suggestion
  analyticsName: "Empty trash",
```

consistency

app/actions/definitions/trash.tsx
```diff
@@ -0,0 +1,28 @@
...
+  section: TrashSection,
+  icon: <TrashIcon />,
+  dangerous: true,
+  visible: ({ stores }) => stores.documents.deleted.length > 0,
+  perform: ({ stores, t, location }) => {
+    stores.dialogs.openModal({
+      title: t("Permanently delete all documents"),
```
tommoor, line 17:
```suggestion
      title: t("Permanently delete documents in trash"),
```

app/actions/sections.ts
```diff
@@ -20,3 +20,5 @@ export const TeamSection = ({ t }: ActionContext) => t("Workspace");
 
 export const RecentSearchesSection = ({ t }: ActionContext) =>
   t("Recent searches");
+
+export const TrashSection = ({ t }: ActionContext) => t("Trash");
```
tommoor, line 24: I don't think this is needed either, it's fine in the DocumentsSection?

app/scenes/Trash/components/DeleteDocumentsInTrash.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+function DeleteDocumentsInTrash({ onSubmit, shouldRedirect }: Props) {
+  const { t } = useTranslation();
+  const { documents } = useStores();
+  const history = useHistory();
+
+  const handleSubmit = async () => {
+    await documents.emptyTrash();
+    toast.success(t("Deleted all the documents in Trash"));
```
tommoor, line 22:
```suggestion
    toast.success(t("Trash emptied"));
```

app/scenes/Trash/index.tsx
```diff
@@ -2,19 +2,35 @@ import { observer } from "mobx-react";
...
 function Trash() {
   const { t } = useTranslation();
   const { documents } = useStores();
+  const context = useActionContext();
   return (
     <Scene icon={<TrashIcon />} title={t("Trash")}>
-      <Heading>{t("Trash")}</Heading>
+      <Flex align="baseline" justify="space-between">
+        <Heading>{t("Trash")}</Heading>
+        {documents.deleted.length > 0 && (
```
tommoor, line 24: There is an `actions` property on the `<Scene` component to place this button in the correct place, no need for the flex wrapper

server/routes/api/documents/documents.ts
```diff
@@ -1701,4 +1690,59 @@ router.post(
...
+  async (ctx: APIContext) => {
+    const { user } = ctx.state.auth;
+
+    const collectionIds = await user.collectionIds({
+      paranoid: false,
+    });
+    const collectionScope: Readonly<ScopeOptions> = {
+      method: ["withCollectionPermissions", user.id],
+    };
+    const viewScope: Readonly<ScopeOptions> = {
```
tommoor, line 1705: No need to include the `viewScope` here

app/actions/sections.ts
hmacr: Since we are acting on Trash, I thought creating a new section with its associated action(s) provided a clean separation (and also a clean UI in the cmd-bar).
I'm not sure `Empty Trash` is something that needs to be associated to a single document (and hence in the Documents section of cmd-bar).

tommoor: Okay, sure – lets keep the section but I'd still move the action into `definitions/documents` as definitely acting on documents fundamentally.

hmacr: Sure, will do.

app/scenes/Trash/index.tsx
hmacr: Screenshot 2024-04-12 at 8 26 47 AM

The `action` on the `Scene` looks disassociated from the Header at the far-right. Is this alright?
I'd prefer the button to live right next to the Header. What do you think?

tommoor: Yes, it's ok for consistency with other scenes. Might turn into a "..." if other actions get added here.

tommoor (approved):

app/actions/definitions/documents.tsx
```diff
@@ -828,6 +830,26 @@ export const permanentlyDeleteDocument = createAction({
   },
 });
 
+export const permanentlyDeleteDocumentsInTrash = createAction({
+  name: ({ t }) => t("Empty"),
+  analyticsName: "Empty trash",
+  section: TrashSection,
+  icon: <TrashIcon />,
+  dangerous: true,
+  visible: ({ stores }) => stores.documents.deleted.length > 0,
```
tommoor, line 839: I realized one thing here, the route is admin only – so this needs to be visible to admins only too, can check `user.isAdmin`

PR #6776, opened 2024-04-07
Fixes hover preview going out of window bounds

### LINKED ISSUE #6760: Hover preview going out of bounds again
Was previously fixed in #5712 but looks like it regressed somewhere post that.

Well, the problem seems to be that of `cardRef` being passed as a `ref` to the `Position` component. The height and width of the `Position` component always remains zero and it actually appears so in the `useHoverPosition` hook as `card.height` and `card.width`.

The initial attempt to fix was to simply pass `cardRef` as `ref` to the individual hover preview components as was happening before but that doesn't work anymore and needs further investigation.

apoorv-mishra:
Closes #6760

### DIFF
--- a/app/editor/extensions/HoverPreviews.tsx
+++ b/app/editor/extensions/HoverPreviews.tsx
@@ -16,14 +16,16 @@ export default class HoverPreviews extends Extension {
   state: {
     activeLinkElement: HTMLElement | null;
     data: Record<string, any> | null;
+    dataLoading: boolean;
   } = observable({
     activeLinkElement: null,
     data: null,
+    dataLoading: false,
   });
 
   get defaultOptions(): HoverPreviewsOptions {
     return {
-      delay: 500,
+      delay: 600,
     };
   }
 
@@ -59,6 +61,7 @@ export default class HoverPreviews extends Extension {
                       .pop();
 
                     if (url) {
+                      this.state.dataLoading = true;
                       try {
                         const data = await client.post("/urls.unfurl", {
                           url: url.startsWith("/") ? env.URL + url : url,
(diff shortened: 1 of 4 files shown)


app/editor/extensions/HoverPreviews.tsx
```diff
@@ -45,8 +49,27 @@ export default class HoverPreviews extends Extension {
...
+                      .split("/")
+                      .pop();
+
+                    if (url) {
+                      try {
+                        const data = await client.post("/urls.unfurl", {
+                          url: url.startsWith("/") ? env.URL + url : url,
+                          documentId,
+                        });
+                        this.state.activeLinkElement = element;
```
tommoor, line 67: I haven't ran this yet but setting the `activeLinkElement` state only after the async network request has come back seems like a recipe for buggy behavior…?

app/editor/extensions/HoverPreviews.tsx
```diff
@@ -13,13 +15,17 @@ interface HoverPreviewsOptions {
...
   } = observable({
     activeLinkElement: null,
+    data: null,
+    dataLoading: false,
   });
 
   get defaultOptions(): HoverPreviewsOptions {
     return {
-      delay: 500,
+      delay: 600,
```
apoorv-mishra, line 28: Purposely set opening delay to be greater than closing delay(500ms) to prevent picking of stale dimensions for card positioning 👇 

https://github.com/outline/outline/assets/15694566/7b351b22-6d8a-4bb1-8a25-ceaf0427b915

Although a consistent solution for this would be to move the closing delay logic from the component to the extension(in `mouseout` handler). But we're bound to keep that logic within the component because we have to prevent the preview card from closing if hovered 😬

tommoor:
Works well 👌 

image

– just one thing I found, which is kind of a problem from before that is made worse now by the above positioning. Some of the hover cards increase in width once the image loads, causing them to reduce in height and leave a gap. I think we can fix this just by fixing the width to 375 always.

tommoor (approved):
With the caveat of changing to a fixed width preview we're good to go

PR #6782, opened 2024-04-09
feat: duplicate a document as draft

### LINKED ISSUE #6778: Allow duplicating a document as draft
It would be nice if one could decide to copy a page as "Unpublished". This would especially be useful if the page has mentions on it, as they get notified immediately after the duplication.
I think this could be done relatively easy in the Dialog that already pops up when duplicating a page.

_Originally posted by @nliechti in https://github.com/outline/outline/discussions/6766_

hmacr:
Closes #6778


tommoor (commented):
Works well, lets flip the UI though so it shows a toggle with "Published" label enabled by default that you toggle off to duplicate as a draft.

tommoor (approved):

PR #6791, opened 2024-04-11
add H3 to editor formatting toolbar

hmacr:
This PR adds H3 option to the editor formatting toolbar.

As of now, the only way to change an existing heading level from H1/H2 to H3 is to remove the content and use the `slash directive` to insert a H3 node.


tommoor (approved):
It's not actually the _only_ way, you can also type three `###` at the beginning of any node to turn it into a H3.

PR #6864, opened 2024-05-02
next-video player integrated

sajidali2444:
In this PR default video player replaced with Next-Video player.


sajidali2444:
wrong github account used

PR #6866, opened 2024-05-02
next-video player integrated

MuhammadAkif:
What is completed in this PR

1. Next Video Player is added
2. Globally change default video player with Next Video Player
3. Changed Vite config to add environment for Next Video Player
4.  Dynamically loaded Next Video Player import inside useEffect

Demo loom
https://www.loom.com/share/be040e63844d448498274a27ecd8e7b2

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,6 +150,7 @@
     "mobx-react": "^6.3.1",
     "mobx-utils": "^4.0.1",
     "natural-sort": "^1.0.0",
+    "next-video": "^1.0.4",
     "node-fetch": "2.7.0",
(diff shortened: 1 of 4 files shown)


tommoor:
While I agree that the video player can be improved through custom styling (it's a shame the default browser players are so poor), this doesn't seem like the way to go. This particular library is optimized for next and adds hundreds of unnecessary extra dependencies to the project, all we really need is a nice simple React component.

Really the best way forward is probably to build something custom, _maybe_ use `video.js` or a similar library to make interacting with the player easier?

PR #6867, opened 2024-05-03
fix: Simplify redirect logic for self-hosted

### LINKED ISSUE #6859: Invalid redirect on OAuth errors when using reverse proxies
Using OAuth mecanism, using a reverse proxy, and upon OAuth errors, the existing logic will fail to replace the inner hosting port (usually 3000 in a docker container), and will end up redirecting to `https://external.address.org:3000` instead of `https://external.address.org`. 

This can be traced back to: https://github.com/outline/outline/blob/main/server/middlewares/passport.ts#L43

Indeed, the serving host string in `ctx.get("host")` will be the host value received by the inner request made by the proxy... targetting the inner container and port (usually 3000). The link given to the source show it tries then to replace the internal host with a provided correct external hostname, but forgets about the port number : effectively overwriting only the host name and leaving the port number intact.

Hence the issue.

tommoor:
closes #6859

### DIFF
--- a/server/middlewares/passport.ts
+++ b/server/middlewares/passport.ts
@@ -25,8 +25,8 @@ export default function createMiddleware(providerName: string) {
 
           if (err.id) {
             const notice = err.id.replace(/_/g, "-");
-            const redirectUrl = err.redirectUrl ?? "/";
-            const hasQueryString = redirectUrl?.includes("?");
+            const redirectPath = err.redirectPath ?? "/";
+            const hasQueryString = redirectPath?.includes("?");
 
             // Every authentication action is routed through the apex domain.
             // But when there is an error, we want to redirect the user on the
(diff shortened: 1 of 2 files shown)


server/middlewares/passport.ts
```diff
@@ -35,18 +35,17 @@ export default function createMiddleware(providerName: string) {
...
-            const requestHost = ctx.get("host");
+            const requestHost = state?.host ?? ctx.hostname;
             const url = new URL(
-              `${reqProtocol}://${requestHost}${redirectUrl}`
+              env.isCloudHosted
+                ? `${reqProtocol}://${requestHost}${redirectPath}`
+                : `${env.URL}${redirectPath}`
             );
 
-            url.host = host;
```
tommoor, line 48: This seems vestigial – why construct the URL above and then override the host immediately?

Issue #6683, opened 2024-03-18
the bold text appears as **text** on the reading page


jochne:
Bug Report:

Bug Description:
In a Chinese language environment, while using Outline self-hosted version 0.75.2, when a sentence includes text formatted in bold within the editor, this formatting does not carry over to the shared reading page. Instead, the bold text appears as `**text**` on the reading page.

To Reproduce:

1. Access the Outline self-hosted platform (version 0.75.2).
2. Create or edit an article containing a sentence with Chinese characters in bold format.
3. Save the changes and preview or share the article to view the read-only version.
4. Notice how the bold formatting is not preserved; it's displayed as `**text**`.

Expected Behavior:
When a user formats text as bold within the editor, this formatting should be consistently maintained across all views, including the shared reading page.

Desktop Environment Details:

* OS: macOS 14.2.1
* Browser: Chrome
* Version: 121.0.6167.184 (Official Build) (arm64)

On shared page with view mode

![CleanShot 2024-03-18 at 4  10 38@2x](https://uploads.linear.app/47519575-84a7-4b8b-8a29-c645d1bb39ee/c62228f3-53d1-4902-9001-07ed73ba4b66/b40803cf-16b7-4ae7-bfc3-7d4bbfdf87f6?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzQ3NTE5NTc1LTg0YTctNGI4Yi04YTI5LWM2NDVkMWJiMzllZS9jNjIyMjhmMy01M2QxLTQ5MDItOTAwMS0wN2VkNzNiYTRiNjYvYjQwODAzY2YtMTZiNy00YWU3LWJmYzMtN2Q0YmJmZGY4N2Y2IiwiaWF0IjoxNzExNTAyMTQyLCJleHAiOjMzMjgyMDYyMTQyfQ.I0KeH2k-aSi5OnKUhtIkTZiovTkJSfGCGIinlqGWS0M)

On the page after login or in edit mode

[...]

PR #6871, opened 2024-05-05
fix: markdown-it bug for incorrect strong tag closing

shepherdwind:
## Problem

By default, markdown-it identifies whether `**` can be closed based on the following rules:

- If there is a space on the left, then `open = true`
- If there is a space on the right, then `close = true`
- If neither side has a space, it checks for special characters such as `+-=` or other language separators for example Chinese semicolons `：`. If these symbols are present, they are treated as spaces.

The last rule leads to the following parsing results:

```js
render('**==foo==**') === '<p>**<mark>foo</mark>**</p>';
render('**中文：**中文') === '<p>**中文：**中文</p>';
```

In these two examples, the second `**` is preceded by `=` and `：`, which are recognized as delimiters, identified as an open tag, causing the preceding `**` to not close properly, so it is output directly.

In reality, we expect both cases to output a strong tag.

> Similar issues #6683.

## Fix logic

The implementation logic is as follows: insert a function before `balance_pairs`. After completing inline rule parsing, adjust the delimiters. After adjustment, whether `**` can be closed depends on whether there was an open `**` before it.

After using this plugin, the above examples will output as follows:

```js
render('**==foo==**') === '<p><strong><mark>foo</mark></strong></p>';
render(' **Chinese:** Chinese ') === ' < p > < strong > Chinese: </ strong > Chinese </ p > ';
```

> For more details, please refer to the implementation of the [match-pairs](https://github.com/shepherdwind/match-pairs) package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,6 +144,7 @@
     "markdown-it": "^13.0.1",
     "markdown-it-container": "^3.0.0",
     "markdown-it-emoji": "^2.0.0",
+    "markdown-match-pairs": "^1.0.2",
     "mermaid": "9.3.0",
     "mime-types": "^2.1.35",
     "mobx": "^4.15.4",
--- a/shared/editor/marks/Bold.ts
+++ b/shared/editor/marks/Bold.ts
@@ -1,3 +1,4 @@
+import matchPairs from "markdown-match-pairs";
 import { toggleMark } from "prosemirror-commands";
 import { InputRule } from "prosemirror-inputrules";
 import { MarkSpec, MarkType } from "prosemirror-model";
@@ -30,6 +31,10 @@ export default class Bold extends Mark {
       ],
       toDOM: () => ["strong"],
     };
+  }
+
+  get rulePlugins() {
+    return [matchPairs];
   }
 
   inputRules({ type }: { type: MarkType }): InputRule[] {
--- a/yarn.lock
+++ b/yarn.lock
@@ -9892,6 +9892,11 @@ markdown-it@^14.0.0:
     punycode.js "^2.3.1"
     uc.micro "^2.0.0"
 
+markdown-match-pairs@^1.0.2:
+  version "1.0.2"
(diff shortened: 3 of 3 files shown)


shepherdwind:
The plugin logic here is placed in a separate repository for the convenience of integrating old test cases with the markdown-it package, ensuring that the changes does not affect existing logic.

tommoor:
Hey, thanks for the fix – I've seen this issue once or twice.

Can I ask why not publish this fix upstream to the https://github.com/markdown-it/markdown-it repo and fix for everyone? It's well maintained, only has one outstanding pull request, and it seems like you have done a thorough job of creating test cases.

shepherdwind:
> Hey, thanks for the fix – I've seen this issue once or twice.
> 
> Can I ask why not publish this fix upstream to the https://github.com/markdown-it/markdown-it repo and fix for everyone? It's well maintained, only has one outstanding pull request, and it seems like you have done a thorough job of creating test cases.

I initially looked at the narkdown-it repository and found several similar issues(markdown-it/markdown-it/issues/1010, markdown-it/markdown-it/issues/1013), but they were all rejected for not complying with the commonmark specification. I feel that they may not be willing to merge such changes, so I created a separate plugin myself.

However, I can raise an issue there to see if it is deemed appropriate, and if so, I will merge the changes over there.

tommoor:
Thanks for that link, I'll read through the context and make a decision – I think it's likely we're merge this fix. Hopefully this will become less important in the near future as we'll be moving rendering away from parsing Markdown, so it only needs to happen on import/export of external data.

shepherdwind:
Today in that issue I replied, and only then did I realize that commonmark supports nesting of em and strong. I need to make some modifications and add compatibility tests for commonmark.

tommoor:
FWIW, this will be much less of an issue once #3000 is finally merged as we'll then never be rendering from markdown, it will only be considered on _import_.

PR #6874, opened 2024-05-06
feat: copy entity id's

axelrindle:
Add's a new context menu entry to collections and documents allowing to copy the internal uuid for usage e.g. in webhooks.

grafik


tommoor:
I don't see this making sense to add to the menus, we already have this functionality in `cmd+k` which is more appropriate for a developer-type action – perhaps you missed it there?

For most users of Outline the database ID is at best meaningless and at worst confusing.

axelrindle:
I did indeed miss this functionality in `cmd+k`, thanks for pointing out!

PR #6942, opened 2024-05-24
feat preview: add support for youtube clips

### LINKED ISSUE #6941: Add support for previewing Youtube clips
### Discussed in https://github.com/outline/outline/discussions/6940

<sup>Originally posted by **Pranav2612000** May 24, 2024</sup>
At the moment we have support for previewing and playing Youtube videos ( using iframe ). It would be great if we could add support for Youtube clips as well.

Youtube clips are of the following format
`https://www.youtube.com/embed/<id>?clip=<clipId>&clipt=<clipTimeIdentifier>`

Users can get this id by using the "Share" button for the clip, choosing "Embed" and copying the `src` from the iframe
image

A small caveat is that the URL is `HTML encoded` so the user may need to change `&amp;` to `&`.

cc: @clearlysid</div>

Pranav2612000:
Fixes #6941


Pranav2612000:
@tommoor ready for review

Should we also update https://github.com/outline/website/blob/main/integrations/youtube.md to mention that user can also include Youtube clips by obtaining the embed url from "Share" > "Embed" and changing the `&amp;` to `&`?

tommoor:
Pasting the Iframe should "just work" did you try it?

tommoor:
Yea, you need to also change the regex match so it will trigger for `https://youtube.com/clip/...` urls

Pranav2612000:
> Yea, you need to also change the regex match so it will trigger for `https://youtube.com/clip/...` urls

Oh. We don't need that because, like I mentioned in the issue, the URLs for the clips are of the form
```
https://www.youtube.com/embed/<id>?clip=<clipId>&clipt=<clipTimeIdentifier>
```
which will be caught by the current regex.


The `https://youtube.com/clip/<clipId>` won't work because it also needs the `clipt` parameter.

tommoor:
Hmm ok, well the thing I want to avoid is manually changing any links – if you have to tell users to switch encoded characters around it's a dead end.

Pranav2612000:
Makes sense? Are you okay if I add custom logic to encode the characters in code?

tommoor:
Yea, lets just explain why – honestly it seems like a bug on YouTube's side that these characters are encoded, so lets just account for that getting fixed

Pranav2612000:
> Yea, lets just explain why in a code comment – honestly it seems like a bug on YouTube's side that these characters are encoded, so lets just account for that getting fixed

@tommoor added support for correctly fetching the query params even if the string contains `&amp;`s instead of `&`. Can you take another look?

tommoor (approved):
It's a shame there isn't enough info in the main clip url, but an improvement from before 👍

clearlysid:
Thanks for this @Pranav2612000 @tommoor! I send a bunch fo memo's using Outline and this will make em look much nicer 🙌

PR #6786, opened 2024-04-10
feat: allow user to set TOC display preference

hmacr:
Closes #6383

PR #6943, opened 2024-05-24
feat: allow user to set TOC display preference

### LINKED ISSUE #6383: Add display preference to move document TOC to right hand side
Currently thinking this would be a workspace level preference.

hmacr:
Closes #6383 

Creating a new PR since I messed up the git tree locally in #6786 

This PR builds on top of #6786, and used grid to place the TOC and editor.
I haven't touched the sidebar (left & right) since it impacts the layout of the document header and editor; It requires a separate effort to render it properly across all screen-sizes as well.

**How does the alignment work?**
- The editor is centered as long as the container has extra space. Otherwise, the editor takes up the full-width.
- TOC is placed on either side of the editor depending on the user's preference.


tommoor:
Hey, I didn't realize you'd actually updated this PR to address some of the original feedback – I found a couple of things:

- "Full width" document rendering no longer works correctly, seems like related classes were removed?
- The top of the TOC is no longer aligned with the document title

Otherwise this looks good, I think the option would be appreciated

hmacr:
I've fixed the issues.

Also, the document emoji is now always displayed on the `left` side (`top` seemed odd).
Let me know if this needs to be reverted.

tommoor:
> Also, the document emoji is now always displayed on the left side (top seemed odd).

Yea, that's totally fine – i didn't love this either.

One other regression I just found is that there is no background on the TOC anymore and the z-index isn't height enough, this is important because for example tables and images can now spread outside the bounds of the editor meaning the TOC has to be displayed ontop.

See example: https://docs.getoutline.com/s/420d1c13-bf65-4613-92b2-14d2e0981285

hmacr:
Damn.. I didn't know that full-width tables and images are possible :(
I've set the z-index for the TOC which fixed both the regressions.

Something that I noticed while testing (not a regression).. full-width nodes are not rendered properly when the document is set to use the full width. I'll debug what the issue and fix it in a separate PR

https://github.com/outline/outline/assets/132386067/7a1c2664-1c79-415f-8e10-c3cdc372cf06

tommoor:
Yes, i noticed this too :( but didn't want to bother you with it as it's an existing bug

tommoor (approved):

tommoor:
Ah crap… approved and then saw this

image

tommoor:
I think in this case z-index can just be `1` to bump above other relative content

app/scenes/Document/components/Document.tsx
```diff
@@ -617,6 +616,8 @@ type ContentsContainerProps = {
 };
 
 const ContentsContainer = styled.div<ContentsContainerProps>`
+  margin-top: calc(44px + 6vh);
```
hmacr, line 619: @tommoor Curious about this coz it breaks the TOC alignment with the title 🤔 

**Normal Page**
Screenshot 2024-06-16 at 10 18 13 PM

**Deleted Page**
Screenshot 2024-06-16 at 10 18 26 PM

tommoor: I want it always aligned with the metadata beneath the title, sorry should have been clearer

hmacr: Ahhh cool! :)

PR #6945, opened 2024-05-25
feat: support direction and text align

bahmannejati:
https://github.com/outline/outline/pull/6817


bahmannejati:
Sorry for that, my mistake

PR #6947, opened 2024-05-26
Feature: Add iframe Insertion Capability to the Editor

RenderCoder:
### Description

This PR introduces a new feature that allows users to insert iframes into the editor. Based on the discussions in #5137, various adjustments have been made to enhance the implementation. The iframe is now inserted directly into the document's source code as an `<iframe>` HTML tag. This ensures that the content is rendered correctly even when the document is accessed via API for embedding in other pages.

### Changes

- Added a new toolbar button for inserting iframes.
- Updated the editor's parsing and rendering logic to handle direct iframe HTML tags.
- Added a ﻿`urlTransformer` function configuration to the `﻿EmbedDescriptor` to support transforming user-entered URLs into predefined text content, such as wrapping them with HTML tags. This approach avoids matching all external content by domain and allows for future enhancements to embed more diverse content types.
- Ensured iframes remain functional and correctly displayed when the document content is fetched and rendered in other applications via API.

### Screenshots

Bevis_2024-05-26_at_23 35 33@2x

### How to Test

1. Open the editor.
2. Press "/"
3. Select the new iframe button in the menu.
4. Enter a valid iframe URL and any additional attributes in the dialog.
5. Insert the iframe and verify it is correctly embedded and displayed.
6. Save the document and access its content via the API to ensure the iframe is properly rendered in other applications.

### Others

I believe that solely relying on URL matching presents several issues. For scenarios where external API requests are needed to render document content, it adds some pressure. It would be best to consider wrapping all external content with iframe HTML tags in future updates, reflected in the document's original Markdown content.

---

Feel free to adjust details according to the specific implementation and requirements of the project!


tommoor:
We can support arbitrary iframes now, but it's not neccessary to output the iframe html into the Markdown – I'd rather keep this as a standard link. This removes the need for the new `urlTransformer` too.

tommoor:
Let me know if you simplify this to make the changes above

tommoor:
> I'll give an example. The current design is that as long as it is a single-line URL and meets certain specific rules, the system will insert it as an iframe. But what if the user wants to display the original link? Is this unified conversion rule unreasonable?

This design exists because Markdown was previously used as the canonical storage prior to v0.77.0 – this is no longer needed as Markdown is just an export format like HTML, plain text, or PDF. The fact that something is an "embed" node in the data structure is enough now :)

Issue #6957, opened 2024-05-27
Table improvements


tommoor:
With #3000 out the way we can now begin on finally bringing much-needed improvements to tables.

- [x] Allow more content types in cells
- [x] Control to toggle header columns / rows
- [x] Add option to make table full-width
- [x] Resizable columns
- [ ] #6977

As requested in #6910, #6606, #6514, #6412, #6282, #5651, #5379, #4816, #3712, #4720, #3066, #4536, #4383, #4361, and about 100 support tickets…

PR #6958, opened 2024-05-27
Table improvements

### LINKED ISSUE #6722: Large images overflow table cells
When inserting large images into table cells, the image will completely overflow the cell (and sometimes the whole viewport).
It is then very difficult for the user to locate the resize handles and eventually drag the image back to a reasonable size.

**To Reproduce**
1. Create a table
2. Drag and drop a large image into any of the cells
3. Observe that the image has overflown the boundaries of the page

**Expected behavior**
Perhaps a sanity check or maximum initial width should be applied to avoid this.

**Outline:**
- Install: Self Hosted
- Version: 7685637 (0.75.2)

**Desktop:**
 - OS: Mac OS Sonoma 14.4
 - Browser Chrome & Safari

tommoor:
Resolves all issues in #6957 except cell merging. 

 - [x] Allow more content types in cells
 - [x] Control to toggle header columns / rows
 - [x] Add option to make table full-width
 - [x] Resizable columns
 - [x] Fix missing "Add row before" menu item on first row
 - [x] Improvements to table overflow/scrolling
 - [x] Inline controls to add rows/columns

image
image

closes #6722


tommoor:
I'm pretty sure with this PR we've gone from having the worst tables to the best 😆

PR #6972, opened 2024-05-31
Devcontainer

IamTaoChen:
Create a  Devcontainer for quick development.


almereyda:
I've had a quick look at this. The initiative is much appreciated.

While we cannot address here that the Node version 18 has left active support and entered maintenance phase (https://nodejs.org/en/about/previous-releases), I like the idea to support developers with a declarative, immutable and reproducible development environment, by building on standard development containers.

The whole setup looks very sane and well thought through.

My comments address the Dockerfile, which adds some opinionated components, which may not be useful to everyone:

- Would it be possible to install mkcert from a binary tarball, instead of installing Brew to install it?
- Could all `vscode` related settings be removed, so this is just a regular development container and not one bound to the Microsoft implementation?

Here we're using development containers without VS Code, by leveraging the available CLI package `@devcontainers/cli` (which slightly differs from the one that is installed by Microsoft).

I'm also asking myself, if it would be possible to add some kind of integration test, to check that the setup works as expected.

A little documentation about how to use this in a `.devcontainer/README.md` file could potentially also be helpful to some.

Issue #6456, opened 2024-01-31
Move components from `reakit` to `radix-ui`


tommoor:
`Reakit` is dead, `Radix` won.

Sometimes we bet on the wrong frameworks and that's definitely what happened here – [shadcn](https://ui.shadcn.com/) provides some great boilerplate to make moving across easier, combining the parts of Radix into usable UI components. We should move Outline's components one by one to radix, in particular:

- [x] `InputSelect`
- [x] `VisuallyHidden`
- [x] `Tooltip`
- [x] `Switch`
- [x] `Modal`
- [x] `Popover`
- [x] `ViewReactionsDialog`
- [x] `Guide`
- [x] `SearchPopover`
- [x] `SkinTonePicker`
- [x] `ContextMenu`
- [x] `FilterOptions`
- [x] `ToolbarMenu`

---

- [x] Remove old Radix components and lib
- [ ] Rename `V2` menu variants

The context menu will be a relatively large undertaking and as such, should be left until last.

tommoor:
### InputSelect

- [x] DefaultCollectionInputSelect (#8541)
- [x] Details (Settings)(ToC preference) (#8541)
- [x] CommentSortMenu (#8541)
- [x] Preferences (Settings)(Language and theme preferences) (#8711)
- [x] Security (Settings)(Default role for new users) (#8711)
- [x] InputSelectPermission (#9585)
- [x] InputMemberPermissionSelect (#9599)
- [x] SelectLocation (TemplatizeDialog) (#9599)
- [x] ApiKeyNew (#9599)
- [x] Invite (#9599)

PR #5987, opened 2023-10-11
Add the keyboard operation on recent search items

ktmouk:
## Description
- This PR adds the keyboard operation on recent search items, which helps users to access items quickly. I used `StyledArrowKeyNavigation` component to implement this feature.
- Close: https://github.com/outline/outline/issues/4286

## Recording

https://github.com/outline/outline/assets/17381979/b882762d-8b73-4115-a839-31878f6b5cf5

PR #6985, opened 2024-06-03
Replace `reakit/composite` with `react-roving-tabindex`

### LINKED ISSUE #6984: Replace `reakit/Composite` with `react-roving-tabindex`
As a first step towards #6456 , `reakit/Composite` needs to eliminated. Replacing it with [`react-roving-tabindex`](https://github.com/stevejay/react-roving-tabindex)

apoorv-mishra:
Closes #6984
Towards #6456


tommoor:
If you could make arrow navigation work in the share popover user list at the same time that would be GREAT 😄 

image

apoorv-mishra:
While testing for search, a bug/limitation was uncovered. If you arrow down to any result except first, then click on search input and arrow down again from there, you're stuck at the first result. Apparently, a bug/limitation of the library 🙈

apoorv-mishra:
Cloned and linked the lib on local. Will try and see if we can patch it!

apoorv-mishra:
Okay, here we go... 🚗 😉 

https://github.com/outline/outline/assets/15694566/4be7edbf-2b86-4854-9918-74440aa1b86c

app/scenes/Search/Search.tsx
```diff
@@ -179,13 +179,13 @@ function Search(props: Props) {
...
       const firstResultItem = (
-        resultListCompositeRef.current?.querySelectorAll(
+        resultListRef.current?.querySelectorAll(
           "[href]"
         ) as NodeListOf<HTMLAnchorElement>
       )?.[0];
 
       const firstRecentSearchItem = (
-        recentSearchesCompositeRef.current?.querySelectorAll(
+        recentSearchesRef.current?.querySelectorAll(
```
tommoor, line 188: This doesn't seem to be working – it was here: https://github.com/outline/outline/pull/5987

app/components/DocumentListItem.tsx
```diff
@@ -49,6 +52,17 @@ function DocumentListItem(
   const user = useCurrentUser();
   const [menuOpen, handleMenuOpen, handleMenuClose] = useBoolean();
 
+  let itemRef: React.Ref<HTMLAnchorElement> =
+    React.useRef<HTMLAnchorElement>(null);
+  if (ref) {
+    itemRef = ref;
+  }
+
+  const [tabIndex, focused, handleKeyDown, handleClick, handleFocus] =
```
tommoor, line 61: Why would this library return a massive array instead of an object? Very tempted to change that seeing as we've forked it already.

tommoor (commented):
All looks good apart from the recent searches list – I can't get past the top item

app/scenes/Search/Search.tsx
apoorv-mishra: Thanks for the spot 🙏

app/components/DocumentListItem.tsx
apoorv-mishra: Makes sense 👍

tommoor (approved):

tommoor (approved):

PR #7005, opened 2024-06-07
Enhanced Discord Support

sebasptsch:
This PR adds enhanced Discord support during the authentication process.

## Description 💬
In several community groups I've recently adopted Outline as Internal Knowledgebases after trying different pieces of software like Wiki.JS, static git-based documentation and Notion. Outline strikes a great balance of collaborative features and editor integrations.

The groups that I've started hosting Outline for both use Discord as their primary means of organisation. Because of the small technical footprint of these groups having authentication self-contained rather than through an external OIDC provider such as Authentik means that the configuration is much simplified and easier to replicate should there be any issues.

Because the Wiki is only being used internally but our server is open to the public it's a requirement that we verify the roles of those logging into the wiki (the primary reason we need to overlay another OAuth provider like Authentik.

## Implementation 🛠️
The current plan for implementation is to use the plugin system that other authentication methods are hooked into to add a new plugin (Discord) with support for OAuth and additional queries to fetch data about the ascociated Discord server like: 
- Server icon and name (for the workspace)
- User nickname and profile image in server (for new member creation)
- User roles (for wiki membership whitelist)

This PR introduces a single new (unbundled) dev dependancy
- Discord API Types

## Checklist ✅
- [x] Discord Login and Scopes
- [x] Discord Server Support (Server Icon, Server Name)
- [x] Discord Profile Support (Grabbing username and profile icon) 
- [x] Discord Server Roles Whitelist
- [x] Env File Template
- [x] Documentation


sebasptsch:
Currently auth isn't working due to an oauth code error and the logo is the wrong size.

Todo: double check team code

tommoor:
Hey @sebasptsch – Discord authentication is actually already supported through the OIDC adapter – 

https://github.com/outline/outline/discussions/6355

sebasptsch:
> Hey @sebasptsch – Discord authentication is actually already supported through the OIDC adapter –
> 
> See example: #6355

Hey @tommoor the primary reason I haven't used the OIDC auth method is because if used with discord it will allow anyone to sign in. By using a discord auth provider I can limit the wiki to approved members.

If there's a way to do this without a new provider I would be very grateful.

sebasptsch:
No more changes planned

.env.sample
```diff
@@ -127,6 +127,16 @@ GITHUB_APP_NAME=
...
 
+# To configure Discord auth, you'll need to create a Discord Application at
+# => https://discord.com/developers/applications/
+#
+# When configuring the Client ID, add a redirect URL under "OAuth2":
+# https://<URL>/api/discord.callback
+DISCORD_CLIENT_ID=
+DISCORD_CLIENT_SECRET=
+DISCORD_SERVER_ID=
+DISCORD_SERVER_ROLES=
```
ashleylamont, line 138: nitpick: The usage of this env property seems unclear here? Could the comment in the env sample possibly be expanded slightly to clarify that this should be a comma separated list of role ids that are required to authenticate?

server/errors.ts
```diff
@@ -208,3 +208,19 @@ export function ClientClosedRequestError(
...
+export function DiscordGuildError(
+  message = "User is not a member of the required server"
+) {
+  return httpErrors(400, message, {
+    id: "discord_guild_error",
+  });
+}
+
+export function DiscordGuildRoleError(
+  message = "User does not have the required role"
```
ashleylamont, line 221: nit: Can the message on this be clarified to specify that it refers to discord roles, rather than outline roles/groups?

server/errors.ts
```diff
@@ -208,3 +208,19 @@ export function ClientClosedRequestError(
     id: "client_closed_request",
   });
 }
+
+export function DiscordGuildError(
+  message = "User is not a member of the required server"
```
ashleylamont, line 213: nit: Can the message on this be clarified to specify that it refers to discord servers?

plugins/oidc/server/auth/oidc.ts
ashleylamont, line 1: question: Could we get some clarity around what's being changed here, and why? Does this need to be changed for the new discord plugin to work? Have these changes been tested against other OIDC configurations, especially those using this plugin that don't have a dedicated plugin?

plugins/discord/server/auth/discord.ts
```diff
@@ -0,0 +1,218 @@
...
+              );
+
+            /** If the user has a nickname in the server, use that as the name */
+            if (guildMember.nick) {
+              userName = guildMember.nick;
+            }
+
+            /** If the user has a custom avatar in the server, use that as the avatar */
+            if (guildMember.avatar) {
+              userAvatarUrl = `https://cdn.discordapp.com/guilds/${guild.id}/users/${profile.id}/avatars/${guildMember.avatar}.png`;
```
ashleylamont, line 153: question: Will this override the avatar in outline if that is set manually after authenticating the first time?

ashleylamont (changes requested):
Love this PR! I reckon this would be a great change, particularly with the deeper integration against discord for things like roles. Just a few minor nits and a few questions :)
Otherwise looks good!

ashleylamont:
> Hey @sebasptsch – Discord authentication is actually already supported through the OIDC adapter –
> 
> See example: #6355

Looks like this adds a handful of new features and deeper integrations over the existing OIDC plugin, as is mentioned in the discussions thread you've linked :)

plugins/oidc/server/auth/oidc.ts
sebasptsch: image

The reason I removed this was because it overrode the passport-oauth2 default handling of user profiles, breaking the passport-discord implementation.
I'm unsure whether this breaks the other oAuth implementations. I'll do some testing today.

plugins/discord/server/auth/discord.ts
sebasptsch: Based on the code in `server/commands/userProvisioner.ts` it seems that the only thing that get's updated is the email.

plugins/oidc/server/auth/oidc.ts
sebasptsch: I've been able to test this with Google and it seems this change doesn't affect it. It looks like the implementation of user profile urls is left up to whomever builds on top of the `passport-oauth2` package.

https://github.com/jaredhanson/passport-oauth2/blob/master/lib/strategy.js#L318

As a result this change only affects code which directly uses `passport-oauth2` which is limited to the `oidc` and `discord` authentication providers. By removing this line and moving the fetch to a later line it prevents cross-contamination between the OIDC plugin and the Discord plugin.

server/errors.ts
sebasptsch: Changed to `User is not a member of the required Discord server`

server/errors.ts
sebasptsch: Changed to `User does not have the required role from the Discord server`

.env.sample
sebasptsch: Changed to
```shell
# To configure Discord auth, you'll need to create a Discord Application at
# => https://discord.com/developers/applications/
#
# When configuring the Client ID, add a redirect URL under "OAuth2":
# https://<URL>/api/discord.callback
DISCORD_CLIENT_ID=
DISCORD_CLIENT_SECRET=

# DISCORD_SERVER_ID should be the ID of the Discord server that Outline is
# integrated with. 
# Used to verify that the user is a member of the server as well as server
# metadata such as nicknames, server icon and name.
DISCORD_SERVER_ID=

# DISCORD_SERVER_ROLES should be a comma separated list of role IDs that are
# allowed to access Outline. If this is not set, all members of the server
# will be allowed to access Outline.
# DISCORD_SERVER_ID and DISCORD_SERVER_ROLES must be set together.
DISCORD_SERVER_ROLES=
```

ashleylamont (approved):

tommoor:
Hey, working on other things with deadlines but I'm willing to land this once it's tested etc.

sebasptsch:
Thanks @tommoor let me know if there's anything you want me to do regarding testing.

plugins/discord/client/Icon.tsx
```diff
@@ -0,0 +1,30 @@
...
+type Props = {
+  /** The size of the icon, 24px is default to match standard icons */
+  size?: number;
+  /** The color of the icon, defaults to the current text color */
+  fill?: string;
+  className?: string;
+};
+
+function DiscordLogo({ size = 24, fill = "currentColor", className }: Props) {
+  return (
```
tommoor, line 12: The icon is too big compared to others, it should be 16x16 within a 24x24 space

image

plugins/discord/server/auth/discord.ts
```diff
@@ -0,0 +1,218 @@
...
+            },
+            user: {
+              email,
+              name: userName,
+              language,
+              avatarUrl: userAvatarUrl,
+            },
+            authenticationProvider: {
+              name: config.id,
+              providerId: domain ?? "",
```
tommoor, line 190: This ends up storing the email domain of the first login as the provider ID. This is somewhat of an issue for OIDC already as it was a bad assumption – for Discord we have the server ID and should use that here.

server/errors.ts
```diff
@@ -208,3 +208,19 @@ export function ClientClosedRequestError(
     id: "client_closed_request",
   });
 }
+
+export function DiscordGuildError(
```
tommoor, line 212: Could we move these to `/plugins/discord/server/errors.ts` – thanks!

plugins/discord/server/auth/discord.ts
```diff
@@ -0,0 +1,218 @@
...
+
+            /** Find the guild that matches the configured server ID */
+            const guild = guilds?.find((g) => g.id === env.DISCORD_SERVER_ID);
+
+            /** If the user is not in the server, throw an error */
+            if (!guild) {
+              throw DiscordGuildError();
+            }
+
+            /** Get the guild's icon
```
tommoor, line 123: Fix the comment formatting here

tommoor (commented):
Works well, just a few small comments

sebasptsch:
Resolved comments.

tommoor (approved):

PR #7022, opened 2024-06-12
[docs] Remove dead link for Node Security Project

bkrausz:
The Node Security Project seems to be dead? The link leads to a squatted domain.


tommoor (approved):
Urgh, good find.

PR #7028, opened 2024-06-12
Add AZURE_TENANT_ID enviroment variable for specific azure tenant.

airbenich:
As seen here https://github.com/outline/passport-azure-ad-oauth2 the Azure plugin already has the needed code to specify an specific azure tenant when using the azure tenant login. Therefore i added an ENV variable for this.

### DIFF
--- a/plugins/azure/server/auth/azure.ts
+++ b/plugins/azure/server/auth/azure.ts
@@ -28,7 +28,8 @@ if (env.AZURE_CLIENT_ID && env.AZURE_CLIENT_SECRET) {
       clientID: env.AZURE_CLIENT_ID,
       clientSecret: env.AZURE_CLIENT_SECRET,
       callbackURL: `${env.URL}/auth/azure.callback`,
-      useCommonEndpoint: true,
+      useCommonEndpoint: env.AZURE_TENANT_ID ? false : true,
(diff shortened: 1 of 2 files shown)


tommoor:
Open to this – it would be good to add the validation decorator to check the value is a valid FQDN

tommoor (approved):

gnagy:
The docs ought to be updated as well, I couldn't get Azure login working without specifying AZURE_TENANT_ID.

https://docs.getoutline.com/s/hosting/doc/microsoft-entra-UVz6jsIOcv

tommoor:
Yea, it should be added but honestly I'm not sure under which scenarios this is needed – we went 4 years without it. Let me know if you have insight

gnagy:
I'm also not clear on the scenarios, but login didn't work without specifying tenant ID and there was no error either in app logs or in Azure, I just happened to find the Github issue after a good deal of googling.

airbenich:
I'm sorry about that. The parameter should actually be optional. It also looks like this in the code. The AZURE_TENANT_ID is required in the context of a company that solves its SSO via Azure. When logging in, it should be directly visible that it is the professional login & not a private Microsoft login or any other login. This helps to make the user experience more pleasant when logging in.

PR #7037, opened 2024-06-14
fix: Scroll does not reset when navigating shared docs on mobile

### LINKED ISSUE #6968: Scroll does not reset when navigating between publicly shared pages on mobile
**To Reproduce**
1. Go to a publicly shared document
2. Scroll down
3. Navigate to another
4. Scroll is not reset

tommoor:
closes #6968


app/components/ScrollToTop.ts
```diff
@@ -25,8 +27,9 @@ export default function ScrollToTop({ children }: Props) {
     ) {
       return;
     }
-    window.scrollTo(0, 0);
+    (scrollContainerRef?.current || window).scrollTo(0, 0);
```
tommoor, line 30: The problem was this hardcoded to resetting `window` scroll, but on mobile we're in a wrapper.

PR #7038, opened 2024-06-14
feat: Unified icon picker

### LINKED ISSUE #6475: Add option to use emoji for collection icons
Related https://github.com/outline/outline/issues/6469 and #2356 

Should be relatively straight forward, all the pieces exist.

hmacr:
Closes #6475 

### What's implemented?
- Unified `IconPicker` to support both Outline icons and emojis.
- Allow icons and emojis to be used as collection and documents icons.

### What's changed?
- Custom `IconPicker` (with virtual rendering) that matches the Outline theme; supports tracking frequently used icons.
- Removed dependency on `emoji-mart` picker component; we depend only on their emoji data now (useful for search).
 
### Screen Recording

https://github.com/outline/outline/assets/132386067/dc6d09ec-f10e-4c1a-b1f4-31a77c64e271

 <br>
Sorry, this turned out to be a bigger PR than what I anticipated.


apoorv-mishra:
Definitely appreciate the effort but I don't think we're aligned with the approach here. All we want is to facilitate adding emojis to collections, similar in manner to documents. As far as I understand the scope of issue, the work here stretches, in fact diverges much beyond that. Particularly, replacing `emoji-mart` altogether isn't expected.

tommoor:
Wow, there's certainly a lot here and first impressions on the UI you've built are excellent. While replacing `emoji-mart` wasn't really on the cards as Apoorv said, I do like what you've got here and It would be great to take it to production after a more thorough review.

As there is a _lot_ here I need to ask if any parts of this code are taken from anywhere and what design inspiration you used as well – the picker looks _very_ similar to Linear's to the point of perhaps including some of the poorer design decisions there.

Thanks!

hmacr:
> Definitely appreciate the effort but I don't think we're aligned with the approach here. All we want is to facilitate adding emojis to collections, similar in manner to documents. As far as I understand the scope of issue, the work here stretches, in fact diverges much beyond that. Particularly, replacing `emoji-mart` altogether isn't expected.

@apoorv-mishra I started with just plugging the existing `IconPicker` and `EmojiPicker` together into a single component (with tabs like now), but soon found them to have different UI. So, I thought of building a new picker which closely aligns with the design of Outline. Hence, the effort expanded from what was expected from #6475 

 <br>

> Wow, there's certainly a lot here and first impressions on the UI you've built are excellent. While replacing `emoji-mart` wasn't really on the cards as Apoorv said, I do like what you've got here and It would be great to take it to production after a more thorough review.
> 
> As there is a _lot_ here I need to ask if any parts of this code are taken from anywhere and what design inspiration you used as well – the picker looks _very_ similar to Linear's to the point of perhaps including some of the poorer design decisions there.
> 
> Thanks!

Yeah, the design is very much inspired from Linear and Notion.
The `color` for the color-wheel displayed in the picker is taken from Linear; rest all were built from scratch adhering to Outline design.

I'd love for this to be integrated as it provides a more consistent icon-picker UI (compared to `emoji-mart`) to the user.
Let me know your thoughts.

tommoor:
> The color for the color-wheel displayed in the picker is taken from Linear; rest all were built from scratch adhering to Outline design.

Okay, good to know. We might just tweak that aspect a bit and make it a back arrow instead – the colorwheel is not actually very clear and is too much of a direct copy :) 

Otherwise, will run through this code/UX in the next day or so.

hmacr:
> > The color for the color-wheel displayed in the picker is taken from Linear; rest all were built from scratch adhering to Outline design.
> 
> Okay, good to know. We might just tweak that aspect a bit and make it a back arrow instead – the colorwheel is not actually very clear and is too much of a direct copy :)
> 
> Otherwise, will run through this code/UX in the next day or so.

Cool!
I wasn't a fan of the color wheel either; but I haven't decided what to put there (so stuck with it for now).

This is a draft version, since there are a couple of things needed to be done.
- i18n for the icon/emoji categories
- emoji alignment in some places
- skip showing country flags in `Windows`

app/components/IconPicker/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  EmojiSkin: "emoji-skin",
+  IconsFrequency: "icons-freq",
+  EmojisFrequency: "emojis-freq",
+  LastIcon: "last-icon",
+  LastEmoji: "last-emoji",
+};
+
+const getStorageKey = (key: string) => `${STORAGE_KEYS.Base}.${key}`;
+
+export const getEmojiSkinKey = () => getStorageKey(STORAGE_KEYS.EmojiSkin);
```
tommoor, line 23: Might as well change all these to be `const` instead of function, avoid recalcluating more than once…

```suggestion
export const emojiSkinToneKey = getStorageKey(STORAGE_KEYS.EmojiSkin);
```

plugins/webhooks/server/tasks/DeliverWebhookTask.ts
```diff
@@ -423,12 +424,18 @@ export default class DeliverWebhookTask extends BaseTask<Props> {
       paranoid: false,
     });
 
+    const collection = model && (await presentCollection(undefined, model));
+    if (collection) {
+      // For backward compatibility, set a default color.
+      collection.color = collection.color ?? colorPalette[0];
```
tommoor, line 430: Perhaps we just change `color` to a getter on the server model and handle the fallback there once?

app/components/IconPicker/components/GridTemplate.tsx
```diff
@@ -0,0 +1,119 @@
...
+import Grid from "./Grid";
+import { IconButton } from "./IconButton";
+
+/**
+ * icon/emoji size is 24px; and we add 4px padding on all sides,
+ */
+const BUTTON_SIZE = 32;
+
+type OutlineNode = {
+  type: "outline";
```
tommoor, line 17: Should use `IconType`?

app/components/IconPicker/components/GridTemplate.tsx
```diff
@@ -0,0 +1,119 @@
...
+  name: string;
+  color: string;
+  initial: string;
+  onClick: (icon: string) => void;
+  delay: number;
+};
+
+type EmojiNode = {
+  type: "emoji";
```
tommoor, line 26: Should use `IconType`

app/components/IconPicker/components/EmojiPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+  return concat(
+    getFrequentEmojis(),
+    getCategoryData(EmojiCategory.People),
+    getCategoryData(EmojiCategory.Nature),
+    getCategoryData(EmojiCategory.Foods),
+    getCategoryData(EmojiCategory.Activity),
+    getCategoryData(EmojiCategory.Places),
+    getCategoryData(EmojiCategory.Objects),
+    getCategoryData(EmojiCategory.Symbols),
+    getCategoryData(EmojiCategory.Flags)
```
tommoor, line 238: As you mentioned, should skip this on non-mac

app/components/DocumentCard.tsx
```diff
@@ -109,12 +113,18 @@ function DocumentCard(props: Props) {
...
+              <DocumentSquircle
+                icon={document.icon}
+                color={document.color ?? undefined}
+              />
             ) : (
-              <Squircle color={collection?.color}>
+              <Squircle
+                color={
+                  collection?.color ??
+                  (!pin?.collectionId ? theme.slateLight : theme.slateDark)
```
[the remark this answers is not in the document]
hmacr: Inside this block, we render a fallback icon for the document depending on collection/home pin.
For a collection pin, we want a darker background with a white foreground.
For home pin, we use the collection icon.

app/components/IconPicker/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+import useMobile from "~/hooks/useMobile";
+import useOnClickOutside from "~/hooks/useOnClickOutside";
+import useWindowSize from "~/hooks/useWindowSize";
+import { hover } from "~/styles";
+import EmojiPanel from "./components/EmojiPanel";
+import IconPanel from "./components/IconPanel";
+import { PopoverButton } from "./components/PopoverButton";
+
+const TAB_NAMES = {
+  Outline: "outline",
```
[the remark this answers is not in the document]
hmacr: I was contemplating it as well, but equality checks like `iconType === IconType.Icon` looked unnatural.
Do you think `Icon` is better?

app/components/IconPicker/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+    popover.hide();
+    onChange(null, null);
+  }, [popover, onChange]);
+
+  const handleQueryChange = React.useCallback(
+    (q: string) => setQuery(q),
+    [setQuery]
+  );
+
+  const handlePopoverButtonClick = React.useCallback(
```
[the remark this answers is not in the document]
hmacr: Oh right!
I noticed it too; missed fixing it before creating this PR.

plugins/webhooks/server/tasks/DeliverWebhookTask.ts
hmacr: The frontend now owns the `color` logic and setting a default color, if necessary; there's just too many state transitions to keep track of in the icon picker.
So the icon picker would send the final icon & color (null for an emoji) to the backend.

We're setting a default color here to ensure the user's webhooks do not fail since we changed the model's color field to accept `null` values now.

app/components/IconPicker/components/EmojiPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const handleEmojiClick = React.useCallback(
+    ({ id, value }: { id: string; value: string }) => {
+      onEmojiChange(value);
+      incrementEmojiCount(id);
+    },
+    [onEmojiChange, incrementEmojiCount]
+  );
+
+  const isSearch = query !== "";
+  const templateData: DataNode[] = isSearch
```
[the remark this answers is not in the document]
hmacr: `getAllEmojis` can be memoized. I don't think search can be.. 🤔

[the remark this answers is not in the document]
hmacr: That'd just be unnecessary, I think.
The search flow gets executed only when query is non-empty, and the component has to re-render anyway whenever the query changes.
Am I missing something? 🤔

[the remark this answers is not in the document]
tommoor: In the case of changing query would bust yea so that part isn't needed to be within memo, just thought it might make dx easier to wrap the whole of `templateData`. Whatever you choose 👍

app/components/Icons/EmojiIcon.tsx
```diff
@@ -17,18 +17,28 @@ type Props = {
...
-  justify-content: center;
-  flex-shrink: 0;
+  display: inline-block;
   width: ${(props) => props.$size}px;
   height: ${(props) => props.$size}px;
-  font-size: ${(props) => props.$size - 10}px;
 `;
+
+const SVG = ({ size, emoji }: { size: number; emoji: string }) => (
+  <svg width={size} height={size} xmlns="http://www.w3.org/2000/svg">
```
hmacr, line 33: Rendering the emoji inside an `svg` for a consistent alignment across retina & non-retina displays.

app/components/IconPicker/components/EmojiPanel.tsx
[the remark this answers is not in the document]
hmacr: On second thought, we do not `memo` for either of the cases since they are mutually exclusive (and needs component re-render).

tommoor:
Looking great, I'm just pondering on how best to roll this out now– it's somewhat backwards incompatible as it stands.

- Folks that have the web app loaded already will stop receiving `emoji` in responses, we can mitigate this with a force reload.
- But, also… between the migration running and servers being updated queries will fail

hmacr:
We can split this PR into two deployments.

**Deployment #1**
1. Instead of renaming the column, we can add a new `icon` column in the document.
2. Run a script to copy `emoji` to `icon` .

**Deployment #2**
1. Roll out this change.

If we want to prevent force reload,
1. BE will continue populating the `emoji` field in the fetch API. FE will use `icon` and fallback to `emoji`.
2. FE will use the `icon` field to update.

hmacr:
@tommoor I've rebased the PR to resolve merge conflicts.

tommoor:
Yes, I think the split deployment is the correct way to go – if you're open to creating a PR that adds the `icon` columns and the double writing I can add the backfill script and deploy in prod before the rest of this merges.

hmacr:
Sure, let me create one.

tommoor:
@hmacr I messed up here and I think overwrote your last few fixes, could you force-push the latest from your machine again when you have a chance?

hmacr:
Got it. I'll rebase and update the PR with the server impl needed to populate emoji and icon.

tommoor:
Data backfill is complete in prod 👏

hmacr:
@tommoor Last three commits are added now. This adds the emoji and icon fields in the server response (also removed the redundant migration file).

I'm not aware of the deployment strategy... can we ensure the server from this build gets deployed before the app?
This way we don't need fallback in the app (and the force-reload).

server/models/Document.ts
```diff
@@ -255,14 +255,18 @@ class Document extends ParanoidModel<
...
 
-  /** An emoji to use as the document icon. */
+  /**
+   * An emoji to use as the document icon,
+   * This is used as fallback (for backward compat) when icon is not set.
+   */
   @Length({
-    max: 1,
-    msg: `Emoji must be a single character`,
+    max: 50,
```
hmacr, line 263: This matches the `icon` validation (double write).

server/models/Revision.ts
```diff
@@ -71,13 +71,18 @@ class Revision extends IdModel<
...
   content: ProsemirrorData;
 
+  /**
+   * An emoji to use as the document icon,
+   * This is used as fallback (for backward compat) when icon is not set.
+   */
   @Length({
-    max: 1,
-    msg: `Emoji must be a single character`,
+    max: 50,
```
hmacr, line 79: Same as `Document`.

tommoor:
> can we ensure the server from this build gets deployed before the app?

Absolutely!

app/components/Icons/CollectionIcon.tsx
```diff
@@ -16,42 +17,49 @@ type Props = {
...
-  // If the chosen icon color is very dark then we invert it in dark mode
-  // otherwise it will be impossible to see against the dark background.
-  const color =
-    inputColor ||
-    (ui.resolvedTheme === "dark" && collection.color !== "currentColor"
-      ? getLuminance(collection.color) > 0.09
-        ? collection.color
-        : "currentColor"
-      : collection.color);
+  if (!collection.icon || collection.icon === "collection") {
```
hmacr, line 32: We need the `negation` check since collections whose icon is `collection` have `null` set in the DB currently.

hmacr:
I've tested on my end.
Please check if everything looks good.

tommoor (approved):

server/models/Document.ts
tommoor: We can have a follow up PR to remove everything related to `emoji` column in DB

PR #7057, opened 2024-06-16
feat: TOC position for publicly shared docs

### LINKED ISSUE #7055: Respect TOC position preference on shared docs
Currently only on internal docs but should be easy enough to expose preference to shared pages.

Related https://github.com/outline/outline/issues/6383
Related https://github.com/outline/outline/pull/6943

hmacr:
Closes #7055

### DIFF
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -7,7 +7,12 @@ import { RouteComponentProps, useLocation } from "react-router-dom";
 import styled, { ThemeProvider } from "styled-components";
 import { setCookie } from "tiny-cookie";
 import { s } from "@shared/styles";
-import { NavigationNode, PublicTeam } from "@shared/types";
+import {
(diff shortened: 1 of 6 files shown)


tommoor:
That was fast, ha. Lets just use the existing `presentPublicTeam` which is there to expose a selection of team preferences rather than add another object type

hmacr:
Since `PublicTeam` holds the custom team branding info, I added a new object (doesn't have a link with custom branding). This could have other misc preferences we want to support in the future as well.

Do you think we should stick with the `PublicTeam` object?

tommoor:
I can see the logic, I think it's fine to use `PublicTeam` here – it's maybe a bit badly named in retrospect but can consider the toc position to be part of the "branding" in a way.

hmacr:
With this change, the toc-position is customized when the team opts into branding.

tommoor (approved):

PR #7064, opened 2024-06-17
feat: add API key expiry options

### LINKED ISSUE #6998: Add API key expiry options
It would be nice to have some options for automatic API key expiry, GitHub is a good model here.

image

image

hmacr:
Closes #6998 

`react-day-picker` library is added to support custom expiry dates.

**Screen Recording**

https://github.com/outline/outline/assets/132386067/53a8df02-5e2f-490d-a1a1-1d1402550739


app/models/ApiKey.ts
```diff
@@ -13,6 +13,10 @@ class ApiKey extends Model {
   @observable
   name: string;
 
+  @Field
+  @observable
+  expiryAt?: string;
```
tommoor, line 18:
```suggestion
  expiresAt?: string;
```

Sorry to be a pain, can we rename this property throughout? There are a few other places the similar concept exists in the codebase and it's always "expiresAt"

app/scenes/ApiKeyNew/components/ExpiryDatePicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+        )}
+      </PopoverDisclosure>
+      <Popover {...popover} ref={popoverContentRef} width={280} shrink>
+        <DayPicker
+          mode="single"
+          required={true}
+          selected={selectedDate}
+          onSelect={handleSelect}
+          style={styles}
+          footer={<ErrorFooter error={error} />}
```
tommoor, line 77: Can we use the `matcher` prop to disable all options before today? That's preferable to letting the user select and then showing an error

https://react-day-picker.js.org/api/type-aliases/Matcher

app/scenes/ApiKeyNew/components/ExpiryDatePicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+            {selectedDate
+              ? formatDate(selectedDate, "MMM dd, yyyy", { locale })
+              : t("Choose a date")}
+          </StyledPopoverButton>
+        )}
+      </PopoverDisclosure>
+      <Popover {...popover} ref={popoverContentRef} width={280} shrink>
+        <DayPicker
+          mode="single"
+          required={true}
```
tommoor, line 73:
```suggestion
          required
          mode="single"
```

app/scenes/ApiKeyNew/components/ExpiryDatePicker.tsx
```diff
@@ -0,0 +1,116 @@
+import { faCalendar } from "@fortawesome/free-solid-svg-icons";
```
tommoor, line 1:
```suggestion
import { CalendarIcon } from "outline-icons";
```

Please use `outline-icons` for UI, only using FA for the user-selectable icons. If new icons are needed that aren't in `outline-icons` we'll make them.

app/scenes/ApiKeyNew/components/ExpiryDatePicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+  selectedDate?: Date;
+  onSelect: (date: Date) => void;
+};
+
+const ExpiryDatePicker = ({ selectedDate, onSelect }: Props) => {
+  const { t } = useTranslation();
+  const theme = useTheme();
+
+  const userLocale = useUserLocale();
+  const locale = React.useMemo(() => dateLocale(userLocale), [userLocale]);
```
tommoor, line 26:
```suggestion
  const locale = dateLocale(userLocale);
```

No memo needed it's a simple lookup

app/scenes/ApiKeyNew/components/ExpiryDatePicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+  );
+};
+
+const Icon = () => (
+  <IconWrapper>
+    <FontAwesomeIcon icon={faCalendar} />
+  </IconWrapper>
+);
+
+const ErrorFooter = ({ error }: { error?: string }) => {
```
tommoor, line 90: Might not need this if the before days are disabled

app/scenes/ApiKeyNew/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+          value={name}
+          minLength={ApiKeyValidation.minNameLength}
+          maxLength={ApiKeyValidation.maxNameLength}
+          required
+          autoFocus
+          flex
+        />
+        <Flex align="center" gap={16}>
+          <StyledExpirySelect
+            ariaLabel={t("API key expiry")}
```
tommoor, line 106:
```suggestion
            ariaLabel={t("Expiration")}
```

One less string to translate 🤷

app/scenes/ApiKeyNew/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+          {expiryType === ExpiryType.Custom ? (
+            <ExpiryDatePicker
+              selectedDate={expiryAt}
+              onSelect={handleSelectCustomDate}
+            />
+          ) : (
+            <StyledExpiryText type="secondary" size="small">
+              {expiryAt
+                ? `${dateToExpiry(expiryAt.toString(), t, userLocale)}.`
+                : t("Never expires!")}
```
tommoor, line 122:
```suggestion
                : t("Never expires")}
```

app/scenes/ApiKeyNew/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { t } = useTranslation();
+  const userLocale = useUserLocale();
+
+  const submitDisabled =
+    isSaving || !name || (!expiryAt && expiryType !== ExpiryType.NoExpiration);
+
+  const expiryOptions = React.useMemo<Option[]>(
+    () =>
+      Object.values(ExpiryType).map((value) => ({
+        label: value,
```
tommoor, line 44: Label needs translation hook. I think you could simplify this to avoid the `ExpiryType` enum and just have all of the values be in days 🤔 

Could inline the config inside this memo then too.

app/scenes/Settings/ApiKeys.tsx
```diff
@@ -23,6 +24,23 @@ function ApiKeys() {
...
+  const handleCopy = React.useCallback(
+    (keyId: string) => {
+      if (copyTimeoutIdRef.current) {
+        clearTimeout(copyTimeoutIdRef.current);
+      }
+      setCopiedKeyId(keyId);
+      copyTimeoutIdRef.current = setTimeout(() => {
+        setCopiedKeyId(null);
+      }, 3000);
+      toast.message(t("API token copied to clipboard"));
```
tommoor, line 39:
```suggestion
      toast.message(t("API key copied to clipboard"));
```

tommoor (approved):
Another hit PR, great first pass

app/models/ApiKey.ts
hmacr: No worries :) `expiresAt` looks cleaner.

app/scenes/Settings/ApiKeys.tsx
hmacr: This breaks the existing translations.
Do we want to do this?

tommoor: Yes it's ok, thanks

app/scenes/ApiKeyNew/components/ExpiryDatePicker.tsx
hmacr: Ah! This is missing in the [IconLibrary](https://github.com/outline/outline/blob/main/shared/utils/IconLibrary.tsx#L220).
It'd be nice to add the missing icons so they can be used as collection/document icons.

tommoor: Yes, not all available icons are in there as many don't make sense – Calendar could be though

app/scenes/ApiKeyNew/index.tsx
hmacr: Without `ExpiryType` enum, I had to use hardcoded strings in places where it is used for comparison.
I figured it's cleaner to keep the enum and refactored the label to use translated strings.

tommoor (approved):

PR #7066, opened 2024-06-17
Add icon column to document

hmacr:
Pre-requisite for #7038 

- Add `icon` and `color` column to `documents` and `revisions` table.
- On document creation and updation, set `icon` from `emoji` and `color` to `null`.
- On revision creation, set `icon` from `emoji` and `color` to `null`.

### DIFF
--- a/server/commands/documentCreator.ts
+++ b/server/commands/documentCreator.ts
@@ -97,6 +97,8 @@ export default async function documentCreator({
       sourceMetadata,
       fullWidth: templateDocument ? templateDocument.fullWidth : fullWidth,
       emoji: templateDocument ? templateDocument.emoji : emoji,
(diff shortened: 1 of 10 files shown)


hmacr:
This is kept fairly simple to handle just the double write.
Post this deployment and the backfill execution, #7038 needs to be rebased and will need some tweaks to 
- populate both `emoji` and `icon` in the server response.
- use `icon` and fallback to `emoji` in the app.

tommoor (approved):

tommoor:
For self-hosted I like to add the backfill script into the migration itself as the data is always under a few tens of thousands of docs it completes in ms. 

I'll tag it into the PR here, just a heads up

PR #7078, opened 2024-06-18
Update Docker compose and add docker.env

Its4Nik:
Latest Docker compose version doesn't need to have the "version: 3.X" part at the top

+ Some people may find it confusing without a docker.env directly on the GitHub so maybe add a default docker.env


tommoor (changes requested):
We don't want these in the repo –

`docker-compose.yml` is setup for development here
`docker.env` contains secrets and should not be committed to source control

PR #7082, opened 2024-06-18
feat: Add lastUsedAt to API keys

### LINKED ISSUE #7080: Add `lastAccessedAt` to API keys
It struck me working on https://github.com/outline/outline/pull/7064 that it would be very useful information to know when a specific API key was last used on the list.

Similar to `User.lastActiveAt` model we can have a column that is updated at a max of every few minutes (not on every request).

hmacr:
Closes #7080 

The column is named `lastUsedAt` (keys are _used_ than _accessed_).
Let me know if this needs to be changed.

Marking the PR as draft to merge after #7064, so the list-item subtitle can be tweaked to display both the expiry and last-used info.

### Screenshot
Screenshot 2024-06-19 at 2 37 51 AM


app/scenes/Settings/components/ApiKeyListItem.tsx
```diff
@@ -17,6 +19,12 @@ const ApiKeyListItem = ({ apiKey }: Props) => {
   const { t } = useTranslation();
   const [linkCopied, setLinkCopied] = React.useState<boolean>(false);
 
+  const subtitle = apiKey.lastUsedAt ? (
```
hmacr, line 22: For the existing keys, we do not have info on its last use; so we render a `null` subtitle.

tommoor:
We can also just match lastActiveAt on users? Maybe not 100% correct terminology but this might be preferable for consistency?

Not sure, happy to be persuaded to keep "used"

hmacr:
I agree - "active" can be used for consistency.

app/scenes/Settings/components/ApiKeyListItem.tsx
```diff
@@ -24,6 +24,12 @@ const ApiKeyListItem = ({ apiKey, isCopied, onCopy }: Props) => {
 
   const subtitle = (
     <>
+      {apiKey.lastActiveAt && (
```
tommoor, line 27: Lets just switch these around, so it's **created ⋅ used ⋅ expiry**

Makes more logical sense

tommoor (approved):

PR #7100, opened 2024-06-21
fix: Include drafts when loading templates

### LINKED ISSUE #7099: Drafts are not included in the Templates page.
Draft templates are not displayed after refreshing the Templates page.

**To Reproduce**
1. Go to `Home` and then navigate to `Settings > Templates`.
2. Verify that the draft templates are displayed.
3. Refresh the page.
4. **Bug:** Draft templates are not displayed.

hmacr:
Closes #7099 

When the app is initially loaded, drafts are loaded into the store by the `App` sidebar.
This gets lost when the user navigates to templates settings & refreshes the page.

### DIFF
--- a/server/routes/api/documents/documents.ts
+++ b/server/routes/api/documents/documents.ts
@@ -100,8 +100,6 @@ router.post(
       where = {
         ...where,
         template: true,
-        // override publishedAt filter from Document defaultScope to include drafts.
-        publishedAt: { [Op.or]: { [Op.eq]: null, [Op.ne]: null } },
       };
     }
 
--- a/server/routes/api/documents/documents.ts
+++ b/server/routes/api/documents/documents.ts
@@ -97,7 +97,12 @@ router.post(
     };
 
     if (template) {
-      where = { ...where, template: true };
+      where = {
+        ...where,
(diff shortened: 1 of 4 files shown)


server/routes/api/documents/documents.ts
```diff
@@ -97,7 +97,12 @@ router.post(
     };
 
     if (template) {
-      where = { ...where, template: true };
+      where = {
+        ...where,
+        template: true,
+        // override publishedAt filter from Document defaultScope to include drafts.
+        publishedAt: { [Op.or]: { [Op.eq]: null, [Op.ne]: null } },
```
tommoor, line 104: I think this will pull other users drafts, also it's a shame that it's so "hacky" of course. It might actually be better to just load drafts from the templates page and wait for both requests to complete. 🤔

hmacr: I wanted to avoid an additional API call; but if that's alright, cool :) Let me update the PR.

tommoor: Understandable, It'll be okay here given it's a settings page.

tommoor (approved):

PR #7108, opened 2024-06-21
fix: set document's collection for determining policies

### LINKED ISSUE #7107: Document becomes uneditable after changing the title (or) emoji.
After changing the title (or) emoji in a document, it becomes uneditable.
Share, publish, and other actions are not displayed.

**To Reproduce**
1. Change the title / emoji of a document.
2. Verify the document is uneditable and actions are hidden.

hmacr:
Closes #7107 

Document permissions are set to `false` since the collection is not loaded for the document.

Fails here -> https://github.com/outline/outline/blob/main/server/policies/document.ts#L32

As an action item, we should add `policy` checks to all the API tests to prevent such regressions.

### DIFF
--- a/server/routes/api/documents/documents.ts
+++ b/server/routes/api/documents/documents.ts
@@ -1049,6 +1049,13 @@ router.post(
       ip: ctx.request.ip,
     });
 
+    collection = document.collectionId
+      ? await Collection.scope({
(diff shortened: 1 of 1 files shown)


hmacr:
Closing in favor of #7111

PR #7120, opened 2024-06-23
Allow direct editing of collection icon

tommoor:
Follow on to #7038

### DIFF
--- a/app/components/Icons/CollectionIcon.tsx
+++ b/app/components/Icons/CollectionIcon.tsx
@@ -2,7 +2,6 @@ import { observer } from "mobx-react";
 import { CollectionIcon } from "outline-icons";
 import { getLuminance } from "polished";
 import * as React from "react";
-import { randomElement } from "@shared/random";
 import { colorPalette } from "@shared/utils/collections";
 import Collection from "~/models/Collection";
 import Icon from "~/components/Icon";
@@ -32,7 +31,7 @@ function ResolvedCollectionIcon({
   if (!collection.icon || collection.icon === "collection") {
     // If the chosen icon color is very dark then we invert it in dark mode
     // otherwise it will be impossible to see against the dark background.
-    const collectionColor = collection.color ?? randomElement(colorPalette);
+    const collectionColor = collection.color ?? colorPalette[0];
     const color =
       inputColor ||
       (ui.resolvedTheme === "dark" && collectionColor !== "currentColor"
--- a/app/scenes/Collection/index.tsx
+++ b/app/scenes/Collection/index.tsx
@@ -187,7 +187,6 @@ function CollectionScene() {
                       size={40}
                       popoverPosition="bottom-start"
                       onChange={handleIconChange}
-                      allowDelete
                       borderOnHover
                     />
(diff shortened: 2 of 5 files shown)


app/components/Icons/CollectionIcon.tsx
```diff
@@ -32,7 +31,7 @@ function ResolvedCollectionIcon({
   if (!collection.icon || collection.icon === "collection") {
     // If the chosen icon color is very dark then we invert it in dark mode
     // otherwise it will be impossible to see against the dark background.
-    const collectionColor = collection.color ?? randomElement(colorPalette);
+    const collectionColor = collection.color ?? colorPalette[0];
```
tommoor, line 34: @hmacr random here seemed like a bad default, we still have the "random" color picking logic on a new collection elsewhere

app/scenes/Document/components/Document.tsx
```diff
@@ -659,7 +659,7 @@ type EditorContainerProps = {
 
 const EditorContainer = styled.div<EditorContainerProps>`
   // Adds space to the gutter to make room for icon & heading annotations
-  padding: 0 44px;
+  padding: 0 40px;
```
tommoor, line 662: Just trying to better align on 8px grid and between collection/document titles

app/scenes/Document/components/DocumentTitle.tsx
```diff
@@ -346,10 +346,6 @@ const Title = styled(ContentEditable)<TitleProps>`
   }
 `;
 
-const StyledIconPicker = styled(IconPicker)`
-  ${extraArea(8)}
```
tommoor, line 350: Not really needed now the button is so large

app/components/Icons/CollectionIcon.tsx
hmacr: random-color shouldn't kick in, no? because a collection icon will have a color unless an emoji was used? 🤔 

When it's an emoji, it follows a similar treatment to "new" collection to use a different color on each render.

app/scenes/Collection/index.tsx
```diff
@@ -164,7 +178,22 @@ function CollectionScene() {
           ) : (
             <>
               <HeadingWithIcon>
-                <HeadingIcon collection={collection} size={40} expanded />
+                {can.update ? (
+                  <React.Suspense fallback={fallbackIcon}>
+                    <Icon
+                      icon={collection.icon ?? "collection"}
+                      color={collection.color ?? colorPalette[0]}
```
hmacr, line 185: The color fallback is encapsulated inside the `Icon` component. Maybe we should let it handle instead of providing one here?

app/scenes/Collection/index.tsx
```diff
@@ -164,7 +178,22 @@ function CollectionScene() {
...
+                      initial={collection.name[0]}
+                      size={40}
+                      popoverPosition="bottom-start"
+                      onChange={handleIconChange}
+                      allowDelete
```
hmacr, line 190: We shouldn't allow the user to remove a collection icon.

Issue #7140, opened 2024-06-24
CIDR IP Addresses get converted into fractions


lampy255:
**To Reproduce**
1. Type out a CIDR IP Address in a document EG: "10.0.101.1/24"
2. Outline overrides this to become "10.0.101.½4"

**Expected behavior**
IP Addresses should be recognised and not converted into fractions

**Outline:**
- Install: Self Hosted
- Version: [f4c7a99](https://github.com/outline/outline/commit/f4c7a99a36889b11a1af8dccd045d4d2eebb0d7f)

tommoor:
The responsible code is here…

https://github.com/outline/outline/blob/main/app/editor/extensions/SmartText.ts#L5-L11

PR #7141, opened 2024-06-24
fix: Improve SmartText fraction regex pattern

lampy255:
Modify fraction regex statements to not match if directly preceded by a character


lampy255:
Fixes #7140

tommoor (approved):

PR #7150, opened 2024-06-25
Workspace templates

### LINKED ISSUE #6978: Workspace templates
Currently templates need to belong to a collection, this is somewhat restrictive as many templates would work well at a workspace level - for example "Meeting notes". Recently we've seen folks creating collections just to create templates and this is a pattern that has a lot of downsides, instead we should just support templates without a `collectionId` being visible to all workspace members.

hmacr:
Closes #6978 

When an existing document is templatized, the user now has an option to make it a draft. I think this will be helpful in some scenarios - lmk if this is not needed.

This PR also has fixes for a couple of bugs.
- Collection members can templatize a document.
- Filter draft templates in the templates menu.
- Hide templates menu after a new document is published.
- Show templates settings page to members of a team (per-template access restricted).

### New Templates
Screenshot 2024-06-26 at 3 30 27 AM

### Templates (_displayed in doc header_)
Screenshot 2024-06-26 at 3 30 10 AM

### Templatize Dialog
Screenshot 2024-06-26 at 3 31 45 AM


server/routes/api/documents/documents.ts
```diff
@@ -923,7 +932,7 @@ router.post(
   validate(T.DocumentsTemplatizeSchema),
   transaction(),
   async (ctx: APIContext<T.DocumentsTemplatizeReq>) => {
-    const { id } = ctx.input.body;
+    const { id, publish, workspace } = ctx.input.body;
```
tommoor, line 935: In terms of an interface I think it would be better to allow `collectionId` in this payload – you could pass any `collectionId` to templatize into that specific collection (irregardless of the current collection), or pass `collectionId=null` to represent no collection (workspace)

This would also be better aligned with how draft documents work

app/menus/TemplatesMenu.tsx
```diff
@@ -25,36 +25,83 @@ function TemplatesMenu({ onSelectTemplate, document }: Props) {
...
-      {...menu}
-    >
-      {replaceTitleVariables(template.titleWithDefault, user)}
-    </MenuItem>
+  const otherItemsWithHeader: MenuItem[] = React.useMemo(
+    () =>
+      otherTemplates.length
+        ? [
+            { type: "separator" },
+            { type: "heading", title: t("Other collections") },
```
tommoor, line 88: I think as part of this we should just remove this section, once you can share templates across collections at the workspace level it becomes unnecessary. This menu starts to look pretty messy with three headers. With this section gone it could be:

```
Collection templates
-- Workspace templates header --
Workspace templates
```

server/routes/api/documents/schema.ts
```diff
@@ -273,17 +278,24 @@ export type DocumentsUpdateReq = z.infer<typeof DocumentsUpdateSchema>;
...
+    collectionId: z.string().uuid().optional(),
 
     /** Parent Id, in case if the doc is moved to a new parent */
     parentDocumentId: z.string().uuid().nullish(),
 
     /** Helps evaluate the new index in collection structure upon move */
     index: z.number().gte(0).optional(),
+
+    /** Whether this should be considered a template */
+    template: z.boolean().optional(),
```
tommoor, line 290: I don't understand this option or how it relates to moving

hmacr: This provides an additional validation to ensure only templates can be moved into a collection-less space.

Umm.. It's not used in the route impl; perhaps this is an overkill..?

server/routes/api/documents/documents.ts
hmacr: Along with this, the templatize dialog can have a "collection" selection instead of defaulting to the current one. Your thoughts?

server/routes/api/documents/schema.ts
hmacr: I think this is not needed. The correct way to perform this validation is after fetching the doc from the database.

shared/i18n/locales/en_US/translation.json
```diff
@@ -200,12 +202,14 @@
   "{{ completed }} of {{ total }} tasks": "{{ completed }} of {{ total }} tasks",
   "Template created, go ahead and customize it": "Template created, go ahead and customize it",
   "Creating a template from <em>{{titleWithDefault}}</em> is a non-destructive action – we'll make a copy of the document and turn it into a template that can be used as a starting point for new documents.": "Creating a template from <em>{{titleWithDefault}}</em> is a non-destructive action – we'll make a copy of the document and turn it into a template that can be used as a starting point for new documents.",
+  "Published": "Published",
+  "Create a published template that's available for use immediately.": "Create a published template that's available for use immediately.",
+  "Save the template within the workspace to enable sharing across collections.": "Save the template within the workspace to enable sharing across collections.",
```
hmacr, line 207: I'm kinda uncertain between using "within the workspace" and "at the workspace level".
Or perhaps, a better one? 🤔

server/routes/api/documents/documents.ts
tommoor: Yes, I think so – It can default to the current collection and allow selection of another or "Workspace"

server/routes/api/documents/schema.ts
tommoor: > The correct way to perform this validation is after fetching the doc from the database

👍

shared/i18n/locales/en_US/translation.json
tommoor: All templates are technically "_within_ the workspace", so that's probably not the best choice. I think with the UI changes discussed this sentence will go away.

server/routes/api/documents/documents.ts
```diff
@@ -129,7 +130,15 @@ router.post(
       } // otherwise, filter by all collections the user has access to
     } else {
       const collectionIds = await user.collectionIds();
-      where = { ...where, collectionId: collectionIds };
+      where = {
+        ...where,
+        collectionId:
+          template && can(user, "readDocument", user.team)
+            ? {
+                [Op.or]: [{ [Op.in]: collectionIds }, { [Op.is]: null }],
```
tommoor, line 138: Lets combine this logic with L103 and guard this else with `!template` to keep the template querying in one place

server/routes/api/documents/documents.ts
```diff
@@ -934,12 +943,21 @@ router.post(
...
+    } else {
+      authorize(user, "createDocument", user.team);
+    }
+
     const document = await Document.create(
       {
         editorVersion: original.editorVersion,
-        collectionId: original.collectionId,
+        collectionId,
         teamId: original.teamId,
```
tommoor, line 959:
```suggestion
        teamId: user.teamId,
```

Can we change this while we're here, just safer.

server/routes/api/documents/documents.ts
```diff
@@ -1034,6 +1052,9 @@ router.post(
           }
         );
         authorize(user, "createChildDocument", parentDocument, { collection });
+      } else if (document.isWorkspaceTemplate) {
+        const team = await Team.findByPk(document.teamId);
+        authorize(user, "createDocument", team);
```
tommoor, line 1057:
```suggestion
        authorize(user, "createDocument", user.team);
```

Combined with suggested change on L940

server/routes/api/documents/documents.ts
```diff
@@ -1136,10 +1159,21 @@ router.post(
...
+    );
+
+    if (collectionId) {
+      const collection = await Collection.scope({
+        method: ["withMembership", user.id],
+      }).findByPk(collectionId, { transaction });
+      authorize(user, "updateDocument", collection);
+    }
+
+    if (document.isWorkspaceTemplate) {
```
tommoor, line 1174: I don't like these separate `if` statements, too easy for things to fall between the cracks. Can we change it to something like

```
if (collectionId) {
 // ...
} else if (document.template) {
 // ...
} else {
  throw
}
```

tommoor (commented):
Thanks 🙏

app/components/TemplatizeDialog/SelectLocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+            t("Collections could not be loaded, please reload the app")
+          );
+          setFetchError(error);
+        } finally {
+          setFetching(false);
+        }
+      }
+    }
+    void fetchData();
+  }, [fetchError, t, fetching, collections]);
```
apoorv-mishra, line 103: Can use [`useRequest`](https://github.com/outline/outline/blob/main/app/hooks/useRequest.ts) hook here.

server/routes/api/documents/documents.ts
hmacr: This keeps the `collectionId` logic in one place based on collection-id param from request.
If we move the template condition, we'll have to duplicate the collection auth checks from L117.
I think the current impl provides a single place for `collectionId` related query. Thoughts? 🤔

tommoor: Fair, will reevaluate – I really just dislike the amount of template exceptions that are around now – open to other ideas to make this more long-term maintainable

app/scenes/Document/components/Header.tsx
```diff
@@ -264,14 +264,17 @@ function DocumentHeader({
...
-                <TemplatesMenu
-                  document={document}
-                  onSelectTemplate={onSelectTemplate}
-                />
-              </Action>
-            )}
+            {(isEditing || !user?.separateEditMode) &&
+              !isTemplate &&
+              isNew &&
+              !document.publishedAt && (
```
hmacr, line 270: This hides the templates selection menu after a new document is published in the same screen.

This should've been part of the initial PR; I honestly don't know how I missed it 😞

tommoor: Reverting this change actually, as documents are now auto-published in collections and this prevented applying a template.

server/policies/team.ts
```diff
@@ -32,3 +32,20 @@ allow(User, ["delete", "audit"], Team, (actor, team) =>
...
   )
 );
+
+allow(User, ["createDocument", "readDocument"], Team, (actor, team) =>
+  and(
+    //
+    !actor.isGuest,
+    !actor.isViewer,
+    isTeamModel(actor, team)
+  )
```
tommoor, line 42: We lost the `isTeamMutable` check here when you moved it from `policies/document`

In general I'm worried these could be confusing, thoughts on renaming to `createTemplate` etc?

hmacr: Yeah, that was intentional because `isTeamMutable` defaults to `true`. Is the `isTeamMutable` implementation a bug?

re: renaming, yes `template` is much better. I'll update the PR.

tommoor: Not a bug, it is used in private fork to control access

hmacr: Ah right!!

tommoor (approved):

server/policies/team.ts
```diff
@@ -32,3 +39,22 @@ allow(User, ["delete", "audit"], Team, (actor, team) =>
...
+  and(
+    //
+    !actor.isGuest,
+    !actor.isViewer,
+    isTeamModel(actor, team),
+    isTeamMutable(actor)
+  )
+);
+
+allow(User, "updateTemplate", Team, (actor, team) =>
```
tommoor, line 53: Trying to remember why `updateTemplate` is different here to create and read 🤔

hmacr: We wanted admins to be able to edit all workspace templates irrespective of the owner.
This policy is used within document policies like [update](https://github.com/hmacr/outline/blob/b1de4641115cca3d80bed1263693eaffa25f9435/server/policies/document.ts#L104-L110) for e.g. that checks whether the user is the template's owner (or) team admin.

Is this causing any bugs? Are we planning to change the behavior?

Issue #7039, opened 2024-06-14
Dropping large image in table cell causes overflow


lampy255:
**To Reproduce**
Steps to reproduce the behaviour:
1. Create a table
2. Drop a large image into any table cell
3. Observe that the image overflows the viewport
4. In order to resize the image to something more reasonable, the user has to horizontally scroll until they can find the resize handle.

**Expected behaviour**
The user will expect the same behaviour as when a large image is dropped anywhere else on the document. IE: Regardless of the image resolution, it should never render wider than the document body.

**Outline:**
- Install: Self Hosted
- Version: v0.77.0 [fd17364](https://github.com/outline/outline/commit/fd17364ebf0012cdf3e9a325a0f528d27b7aa638)

**Desktop:**
 - OS: macOS Sonoma 14.5
 - Browser: Safari
 - Version: 17.5

tommoor:
You can also resize from the left handle, although it might feel unnatural to do so.

tusharagrawall:
Give me the honour of doing this issue. I'll fix this within minutes.

PR #7151, opened 2024-06-25
fix: Add max natural width logic to Images

lampy255:
If an exceptionally large image is added to a document, cap the natural width at 400 until the user overrides it with resize handles

### DIFF
--- a/shared/editor/components/Image.tsx
+++ b/shared/editor/components/Image.tsx
@@ -81,20 +81,8 @@ const Image = (props: Props) => {
               // For some SVG's Firefox does not provide the naturalWidth, in this
               // rare case we need to provide a default so that the image can be
               // seen and is not sized to 0px
-              let nw = (ev.target as HTMLImageElement).naturalWidth || 300;
-              let nh = (ev.target as HTMLImageElement).naturalHeight;
-
-              // If the natural image width is greater than nwMax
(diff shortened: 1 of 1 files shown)


lampy255:
Open to feedback on this one.
My only concern is that existing documents with existing large images that have never been resized would have their width changed to 400 with this update..

lampy255:
fixes #7039

tommoor:
I think it's best to keep natural width to the correct value, to cap it at an arbitrary number for the purposes of display does feel wrong. I think the "correct" approach is probably a little more involved where we [modify transactions](https://prosemirror.net/docs/ref/version/0.17.0.html#state.Plugin.constructor^options.appendTransaction) that add an image to a table and write the width attribute to a value – something like half the document width?

lampy255:
> I think it's best to keep natural width to the correct value, to cap it at an arbitrary number for the purposes of display does feel wrong. I think the "correct" approach is probably a little more involved where we [modify transactions](https://prosemirror.net/docs/ref/version/0.17.0.html#state.Plugin.constructor%5Eoptions.appendTransaction) that add an image to a table and write the width attribute to a value – something like half the document width?

Yeah this sounds like a better approach.
I'll have another go at this.

lampy255:
Ok unfortunately this approach is a little too far outside my React skill level. Will close this PR for now.

PR #7156, opened 2024-06-26
Disallow empty comments and comments with only whitespaces

### LINKED ISSUE #7148: Users can post "empty" comments
Turns out there's a way to create empty comment. Steps 👇 

1. Post a comment
2. Click on "Edit", erase the comment and click on "Save".

Screenshot 2024-06-25 at 21 13 59

apoorv-mishra:
Closes #7148

### DIFF
--- a/server/routes/api/schema.ts
+++ b/server/routes/api/schema.ts
@@ -15,7 +15,7 @@ export const ProsemirrorSchema = z.custom<TProsemirrorData>((val) => {
   try {
     const node = Node.fromJSON(schema, val);
     node.check();
-    return ProsemirrorHelper.isEmptyData(val as TProsemirrorData, schema);
+    return !ProsemirrorHelper.isEmpty(node, schema);
   } catch (_e) {
     return false;
   }
--- a/server/routes/api/schema.ts
+++ b/server/routes/api/schema.ts
@@ -2,6 +2,7 @@ import formidable from "formidable";
 import { Node } from "prosemirror-model";
 import { z } from "zod";
 import { ProsemirrorData as TProsemirrorData } from "@shared/types";
+import { ProsemirrorHelper } from "@shared/utils/ProsemirrorHelper";
 import { schema } from "@server/editor";
 
 export const BaseSchema = z.object({
@@ -14,7 +15,7 @@ export const ProsemirrorSchema = z.custom<TProsemirrorData>((val) => {
   try {
     const node = Node.fromJSON(schema, val);
     node.check();
-    return true;
+    return !!ProsemirrorHelper.toPlainText(node, schema).trim();
   } catch (_e) {
     return false;
   }
--- a/server/routes/api/schema.ts
+++ b/server/routes/api/schema.ts
@@ -15,7 +15,7 @@ export const ProsemirrorSchema = z.custom<TProsemirrorData>((val) => {
   try {
     const node = Node.fromJSON(schema, val);
     node.check();
-    return !!ProsemirrorHelper.toPlainText(node, schema).trim();
+    return ProsemirrorHelper.isEmptyData(val as TProsemirrorData, schema);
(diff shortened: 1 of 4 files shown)


server/routes/api/schema.ts
```diff
@@ -14,7 +15,7 @@ export const ProsemirrorSchema = z.custom<TProsemirrorData>((val) => {
   try {
     const node = Node.fromJSON(schema, val);
     node.check();
-    return true;
+    return !!ProsemirrorHelper.toPlainText(node, schema).trim();
```
tommoor, line 18: You can use `isEmptyData` method for this which will avoid having to convert the entire document to text

apoorv-mishra: Well, that actually allows whitespaces, so opted for `toPlainText`.

apoorv-mishra: I'm not sure if there's a better way than to traverse the entire doc 🤔 

Well, maybe we can halp and break out upon spotting the first non-empty "plain-textified" node?

tommoor: > Well, that actually allows whitespaces

Are you sure? If so it would be fine to update that method behavior.

tommoor: `isEmptyData` is almost there already, if the document isn't a single completely empty paragraph node then it's not empty basically. No need to traverse further

apoorv-mishra: >if the document isn't a single completely empty paragraph node then it's not empty basically

```
{
  "data": {
    "type": "doc",
    "content": [
      {
        "type": "paragraph"
      },
      {
        "type": "paragraph"
      },
      {
        "type": "paragraph"
      },
      {
        "type": "paragraph",
        "content": [
          {
            "type": "text",
            "text": "              "
          }
        ]
      }
    ]
  }
}
```

☝️ _Technically_ non-empty, but would be an empty comment _practically_. Therefore, need to traverse till we find a non-empty(trimmed) "plain-textified" node.

server/routes/api/schema.ts
```diff
@@ -14,7 +15,7 @@ export const ProsemirrorSchema = z.custom<TProsemirrorData>((val) => {
   try {
     const node = Node.fromJSON(schema, val);
     node.check();
-    return true;
+    return ProsemirrorHelper.isEmptyData(val as TProsemirrorData, schema);
```
tommoor, line 18: Because you're running `Node.fromJSON` inside of `isEmptyData` this is still parsing the document twice – once above on L16 and then again.

Seeing as we're already parsing to be able to validate we can use the `node` value here and pass it into `isEmpty`, something like:

```suggestion
    return !ProsemirrorHelper.isEmpty(node);
```

shared/utils/ProsemirrorHelper.ts
```diff
@@ -176,8 +142,35 @@ export class ProsemirrorHelper {
    *
    * @returns True if the editor is empty
    */
-  static isEmpty(doc: Node) {
-    return !doc || doc.textContent.trim() === "";
+  static isEmpty(doc: Node, schema?: Schema) {
```
apoorv-mishra, line 145: Requires passing `schema` as an arg since the method is under `shared/` and schema is defined under `server/`.

shared/editor/nodes/Image.tsx
```diff
@@ -166,6 +166,8 @@ export default class Image extends SimpleImage {
           ["p", { class: "caption" }, 0],
         ];
       },
+      toPlainText: (node) =>
+        node.attrs.alt ? `(image:${node.attrs.alt})` : "(image)",
```
tommoor, line 170:
```suggestion
        node.attrs.alt ? `(image: ${node.attrs.alt})` : "(image)",
```

tommoor (approved):

PR #5389, opened 2023-05-25
feat: Add kroki.io Embed Service, add self-hosted kroki

WarlockFish:
Add the kroki.io embed 
image

PR #7163, opened 2024-06-29
Add Kroki Support

### LINKED ISSUE #6798: Not all mermaid diagrams available
The lazy-loaded mermaid diagrams cannot be used, which doesn't make much sense. As a user, I'd expect to be able to use all the diagrams.

**To Reproduce**
Try to create a mermaid mindmap diagram in a document using the following mermaid markdown:
```
mindmap
Root
    A
      B
      C
```

You will see that outline doesn't include any of the lazy-loaded diagrams.

**Expected behavior**
I can use any mermaid diagram.

**Screenshots**
image

- Install: self hosted
- Version: 7685637 or v0.75.2

**Desktop:**
 - OS: Linux - KDE Neon
 - Browser Firefox
 - Version 124.0.1

op8867555:
Hi,
This is a follow up PR of #6680 which add Kroki support and fixes #4242 (potentially fixes #6798).

This PR does:
* Add a new `Kroki Diagram` block and a command to create them,
* Add a new environment variable `KROKI_URL` for supporting self-hosted Kroki instance.


app/editor/menus/block.tsx
```diff
@@ -205,5 +205,11 @@ export default function blockMenuItems(dictionary: Dictionary): MenuItem[] {
       keywords: "diagram flowchart",
       attrs: { language: "mermaidjs" },
     },
+    {
+      name: "code_block",
+      title: "Kroki Diagrams",
+      keywords: "diagram flowchart",
+      attrs: { language: "kroki", diagram: "c4plantuml" },
+    },
```
gabschne, line 213: I think the [official favicon](https://github.com/yuzutech/kroki.io/blob/master/assets/favicon-128.png) would be a suitable icon for the block entry. However, it should be noted that while [the Kroki (repo) itself is under the MIT licence](https://github.com/yuzutech/kroki/blob/main/LICENSE), the "kroki.io" [website repository](https://github.com/yuzutech/kroki.io) does not have a licence file, so it might be necessary - or rather legally correct - to create an issue and ask for a usage permission.

gabschne:
@tommoor Is there anything that can be done to help with this, or is there something that needs to be done before this can be reviewed (by you)?

Supporting Kroki would bring in many new diagram types.

And in our case, we heavily depend on some of these and need it before we can do a full switch to Outline.

gabschne:
I still think this would be a nice feature for Outline and based on the comments and upvotes in [the old PR](https://github.com/outline/outline/pull/5389), some others folks seem to be interested in it as well.

Also: Setting the URL of the Kroki instance with the env var is fine, but it might be more consistent to make it configurable on the self-hosted integrations settings page (`/settings/integrations/self-hosted`).

app/editor/menus/block.tsx
mulder999: The [license](https://github.com/yuzutech/kroki.io/blob/master/LICENSE) for kroki.io is Mozilla Public License 2.0

yboujraf:
Dear @gabschne ,
Any news when it will be PR ?

Best Regards,

PR #7167, opened 2024-07-01
Prevent modal top from shrinking

### LINKED ISSUE #7166: Collection share dialog shrinking
Collection share dialog shrinks vertically to remain center aligned as search query is typed. It shouldn't shrink from top.

https://github.com/outline/outline/assets/15694566/afda0f93-48f6-4e4b-a98c-d1985db5598f

apoorv-mishra:
Closes #7166

### DIFF
--- a/app/components/List/Item.tsx
+++ b/app/components/List/Item.tsx
@@ -4,6 +4,7 @@ import {
 } from "@getoutline/react-roving-tabindex";
 import { LocationDescriptor } from "history";
 import * as React from "react";
+import scrollIntoView from "smooth-scroll-into-view-if-needed";
 import styled, { useTheme } from "styled-components";
 import { s, ellipsis } from "@shared/styles";
 import Flex from "~/components/Flex";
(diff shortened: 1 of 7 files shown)


app/components/Modal.tsx
```diff
@@ -254,7 +254,7 @@ const Header = styled(Flex)`
 const Small = styled.div`
   animation: ${fadeAndScaleIn} 250ms ease;
 
-  margin: auto auto;
+  margin: 25vh auto auto auto;
```
apoorv-mishra, line 257: Note that this affects dialogs across the board.

tommoor: I believe we need to define a max height too, maybe 70vh?

image

apoorv-mishra: Just a note on this - setting max height leads us to manage overflow which provokes to manage scroll 😅 

Looked into all this yesterday, almost there...

apoorv-mishra: 🤞 


https://github.com/outline/outline/assets/15694566/68b03146-af78-4dc5-a3a6-992146a28710

app/components/List/Item.tsx
```diff
@@ -47,22 +48,33 @@ const ListItem = (
...
-  useFocusEffect(focused, itemRef as React.RefObject<HTMLAnchorElement>);
+  useFocusEffect(focused, itemRef);
+
+  const handleFocus = React.useCallback(() => {
+    if (itemRef.current) {
+      scrollIntoView(itemRef.current, {
+        scrollMode: "if-needed",
+        behavior: "auto",
+        block: "center",
+        boundary: window.document.body,
```
tommoor, line 74: Isn't this boundary the default?

app/hooks/useMaxHeight.ts
```diff
@@ -0,0 +1,43 @@
+import * as React from "react";
+import useMobile from "./useMobile";
+import useWindowSize from "./useWindowSize";
+
+const useMaxHeight = ({
+  elementRef,
+  maxViewportHeight = 90,
+  margin = 8,
```
tommoor, line 8:
```suggestion
  margin = 16,
```

app/hooks/useMaxHeight.ts
```diff
@@ -0,0 +1,43 @@
+import * as React from "react";
+import useMobile from "./useMobile";
+import useWindowSize from "./useWindowSize";
+
+const useMaxHeight = ({
+  elementRef,
+  maxViewportHeight = 90,
```
tommoor, line 7: I don't love this naming, might be better to pass as a fraction or… `maxViewportPercentage` for clarity

app/components/List/Item.tsx
apoorv-mishra: Doesn't look like. The scroll isn't prevented upon keyboard navigation if `boundary` is absent. This fixes the problem.

tommoor (approved):

PR #7204, opened 2024-07-08
fix: TOC active indicator for last heading

hmacr:
This fixes a bug where the last heading in TOC does not get the active indicator when it is scrolled past.


tommoor (approved):
💯

PR #7222, opened 2024-07-10
feat: show frequent languages in code block menu

hmacr:
Implementation for the feature discussed in #7211

### DIFF
--- a/app/editor/menus/code.tsx
+++ b/app/editor/menus/code.tsx
@@ -1,7 +1,9 @@
 import { CopyIcon, ExpandedIcon } from "outline-icons";
+import { Node as ProseMirrorNode } from "prosemirror-model";
 import { EditorState } from "prosemirror-state";
 import * as React from "react";
 import { LANGUAGES } from "@shared/editor/extensions/Prism";
+import { getFrequentCodeLanguages } from "@shared/editor/lib/code";
 import { MenuItem } from "@shared/editor/types";
 import { Dictionary } from "~/hooks/useDictionary";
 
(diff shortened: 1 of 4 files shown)


app/editor/menus/code.tsx
```diff
@@ -12,6 +14,26 @@ export default function codeMenuItems(
 ): MenuItem[] {
   const node = state.selection.$from.node();
 
+  const allLanguages = Object.entries(LANGUAGES);
+  const frequentLanguages = getFrequentCodeLanguages();
+
+  const frequentLangMenuItems = frequentLanguages.map((value) => {
```
hmacr, line 20: This is sorted by frequency.
Perhaps, alphabetical sort is better?

tommoor: Hard to say which is better to be honest, trying it out I think most recent works fine – I would start with this and we can change if there is any confusion

shared/editor/lib/code.ts
```diff
@@ -0,0 +1,65 @@
+import Storage from "../../utils/Storage";
+
+const RecentStorageKey = "rme-code-language";
```
tommoor, line 3:
```suggestion
const LegacyStorageKey = "rme-code-language";
```

I think it would be preferable to explicitly mark this as legacy, just use it to fill out the new key on instantiation and delete. `getRecentCodeLanguage` can then just use the first item in frequent code languages going forward?

hmacr: I'm afraid the frequent code languages is an object of language to its frequencies - it's not possible to find the recent (last used) language from them.
The idea of maintaining a frequency object is to show the user their most probable preferred languages based on their usage history, rather than just the recent (say last 5) selections.

Do you think we should simplify this and maintain just the recent langs in an array?

tommoor: Gotcha, so the most used might not be the most recent – okay that seems fine

tommoor (approved):

PR #7223, opened 2024-07-11
fix: Outdated `updatedBy` returned in document mutations

### LINKED ISSUE #7220: Outdated "lastModified" user returned in `documents.archive` response
Had a report of the incorrect user showing up in the archived banner on documents after archiving and before a page reload. 

Should also check trashing to see if it has the same issue.

tommoor:
closes #7220

### DIFF
--- a/server/routes/api/documents/documents.test.ts
+++ b/server/routes/api/documents/documents.test.ts
@@ -1955,7 +1955,7 @@ describe("#documents.archived", () => {
       userId: user.id,
       teamId: user.teamId,
     });
-    await document.delete(user.id);
+    await document.delete(user);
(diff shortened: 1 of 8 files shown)


server/routes/api/documents/documents.test.ts
```diff
@@ -3800,6 +3801,8 @@ describe("#documents.unpublish", () => {
     expect(res.status).toEqual(200);
     expect(body.data.id).toEqual(document.id);
     expect(body.data.publishedAt).toBeNull();
+    expect(body.data.createdBy.id).toEqual(user.id);
```
tommoor, line 3804: Regression test

Issue #7142, opened 2024-06-25
Contents tab blocking a full width table


Eliza-j:
When creating a table in the first part on the document, the Contents tab will block the table when it became a full width table.

**To Reproduce**
1. Create a new document
2. Add title
3. Insert some big heading (4 to 5)
4. Create a table
5. Add some columns
6. Toggle "Full-Width" on the table
7. If haven't been toggled, toggle the Contents tab

**Expected behavior**
The contents tab is moved either upward or sideway, anywhere as long as it doesn't block the table

**Screenshots**
![image](https://github.com/outline/outline/assets/82260666/b5d89a93-f88b-45bd-93b8-af7bb922dc7a)

**Outline (please complete the following information):**
- Install: Self hosted
- Version: https://github.com/outline/outline/commit/f4c7a99a36889b11a1af8dccd045d4d2eebb0d7f

**Desktop (please complete the following information):**
 - OS: IOS
    - Browser: Chrome
       - Version: 124.0.6367.208 (Official Build) (arm64)
 - OS: Windows 10
    - Browser: Firefox
    - Version: 127.0.1

PR #7233, opened 2024-07-13
fix: move TOC to free space

hmacr:
Possible fix for #7142

### DIFF
--- a/app/components/Editor.tsx
+++ b/app/components/Editor.tsx
@@ -47,6 +47,7 @@ export type Props = Optional<
   onSynced?: () => Promise<void>;
   onPublish?: (event: React.MouseEvent) => void;
   editorStyle?: React.CSSProperties;
+  onFullWidthElemsChange?: (fullWidthElems: HTMLElement[]) => void;
 };
 
 function Editor(props: Props, ref: React.RefObject<SharedEditor> | null) {
@@ -57,6 +58,7 @@ function Editor(props: Props, ref: React.RefObject<SharedEditor> | null) {
     onHeadingsChange,
     onCreateCommentMark,
     onDeleteCommentMark,
+    onFullWidthElemsChange,
   } = props;
   const userLocale = useUserLocale();
   const locale = dateLocale(userLocale);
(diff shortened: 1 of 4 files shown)


app/components/Editor.tsx
```diff
@@ -258,23 +261,68 @@ function Editor(props: Props, ref: React.RefObject<SharedEditor> | null) {
...
+      }
+    }, 0);
+  }, [onFullWidthElemsChange]);
+
   const handleChange = React.useCallback(
     (event) => {
       onChange?.(event);
       updateHeadings();
       updateComments();
+      updateFullWidthElems();
```
tommoor, line 312: It's tough, because I can see why this is needed but I kind of hate the existing pattern with `updateComments` and `updateHeadings` that exists here so I'm quite remiss to extend it further. There is also the problem that this method is querying the dom instead of the editor data which makes it weirder too, did you try getting the data out of the editor?

hmacr: > did you try getting the data out of the editor
Actually, I did not.
I just found the `nodeDOM` and `domAtPos` methods in prosemirror that allow us to get the DOM node - let me try this.

hmacr: >  I'm quite remiss to extend it further

Makes sense. 
Feels like the editor is doing too many things - setting up the necessary prosemirror objects, rendering auxiliary components / toolbars, handling events. I think it'd be cleaner to extract the editor event handling into a separate dispatcher and plug into the editor.

By having an event dispatcher, it becomes possible to register the interested events for document changes. Example of events would be headings_modified, comment_created, etc.
When the document changes, the event dispatcher would receive the prosemirror node and transactions -> figure out what has happened -> call the necessary event handlers with the transformed data.

An example of the change flow would be
- User adds a heading.
- Editor invokes EventDispatcher with the doc node and transactions.
- EventDispatcher determines that a heading was added and checks if any handler is registered for this event.
- From the doc node, EventHandler builds an internal representation of headings that's understood by the rest of the codebase.
- EventHandler dispatches / calls the registered event handler.

This abstracts the prosemirror -> internal transformation from the rest of the codebase. It also provides a cleaner interface to create / remove event handlers into the editor.
The work involved and the touch points for this implementation would be quite large - probably best to progressively support events in the dispatcher instead of a complete re-write.

I'd love to know your thoughts on this. Or perhaps, suggestions for a different approach.

hmacr: @tommoor Did you get a chance to look at the above query?

tommoor: Do you think it's plausible to reapproach this now with the new document context?

hmacr: Yes - It's in my todo list.
I'll update the PR today 👍

hmacr:
Closed in favour of #7737 (branch created in the main repo).

PR #7243, opened 2024-07-14
If a document is created within a collection, publish it automatically.

feingold-ant:
If a document is created within a collection, publish it automatically. Drafts can still be created outside of collections (and subsequently added to a collection when published).

### DIFF
--- a/app/menus/NewDocumentMenu.tsx
+++ b/app/menus/NewDocumentMenu.tsx
@@ -26,7 +26,7 @@ function NewDocumentMenu() {
       placement="bottom"
     >
       <Button as={Link} to={newDocumentPath()} icon={<PlusIcon />}>
-        {t("New draft")}
+        {t("New doc")}
       </Button>
     </Tooltip>
   );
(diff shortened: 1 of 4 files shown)


server/commands/documentCreator.ts
```diff
@@ -147,6 +147,10 @@ export default async function documentCreator({
     }
   );
 
+  if (collectionId) {
+    publish = true;
+  }
```
bkrausz, line 152: Should we do this on the client instead? Silently turning non-publish into publish in the app may be a bit subtle.

tommoor: Yea agreed, it would be nice to keep the API backwards compatible as well

feingold-ant: Sure thing

app/menus/NewDocumentMenu.tsx
```diff
@@ -26,7 +26,7 @@ function NewDocumentMenu() {
       placement="bottom"
     >
       <Button as={Link} to={newDocumentPath()} icon={<PlusIcon />}>
-        {t("New doc")}
+        {t("New draft")}
```
feingold-ant, line 29: @tommoor should translation have been updated here?

tommoor: It will be displayed on the home screen in top right, might be a bit confusing… I would leave as-is for now

tommoor (approved):

Issue #6174, opened 2023-11-18
`JSONImportTask` should not go through Markdown converter


tommoor:
Follow on from https://github.com/outline/outline/pull/5553 – we need to avoid the markdown translation to avoid lossyness going forward.

tommoor:
After #3000 this has gotten more important

hmacr:
@tommoor I'm so sorry I didn't realize this issue was assigned to you 😓

tommoor:
Okay in this case, it was only assigned to me as I thought noone else would be willing to work on it

hmacr:
Got it. I usually pick issues that are unassigned - somehow missed it here.

PR #7259, opened 2024-07-16
fix: replace document links when importing markdown

### LINKED ISSUE #7089: Markdown internal links after import
The Import function doesn't generate properly Markdown internal links between documents. This happens both with an external markdown library and with an Outline Markdown export. Tried without success with different internal link syntax in the imported document (e.g. `[My File](/mypath/myfile.md)` , `[My File](./mypath/myfile.md)`). 
Link is generated as an external link.

**To Reproduce**
Issue can be reproduced by exporting in Markdown format a test collection with internal links and then importing it back.

**Expected behavior**
Markdown internal links should be reproduced in Outline

**Screenshots**
image
Example of a test collection with an internal link in `Document referencing` pointing to `Document referenced`

-----

image
How `Document referencing.md` appears in the markdown zip export

----

image
How `Document referencing.md` is generated after importing it back in outline. The link is to the address `https://./Document%20referenced.md`

**Outline:**
- Install: self hosted
- Version: 0.77.2

**Desktop:**
 - OS: Windows
 - Browser: Chrome
 - Version: 125.0.6422.176

hmacr:
Closes #7089

### DIFF
--- a/server/queues/tasks/ImportMarkdownZipTask.ts
+++ b/server/queues/tasks/ImportMarkdownZipTask.ts
@@ -1,3 +1,4 @@
+import path from "path";
 import fs from "fs-extra";
 import escapeRegExp from "lodash/escapeRegExp";
 import mime from "mime-types";
@@ -42,6 +43,8 @@ export default class ImportMarkdownZipTask extends ImportTask {
       documents: [],
       attachments: [],
     };
+
+    const docPathToIdMap = new Map<string, string>();
 
     async function parseNodeChildren(
       children: FileTreeNode[],
@@ -102,6 +105,8 @@ export default class ImportMarkdownZipTask extends ImportTask {
           // When there is a file and a folder with the same name this handles
           // the case by combining the two into one document with nested children
           if (existingDocument) {
+            docPathToIdMap.set(child.path, existingDocument.id);
+
             if (existingDocument.text === "") {
               output.documents[existingDocumentIndex].text = text;
             }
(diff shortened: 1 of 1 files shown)


server/queues/tasks/ImportMarkdownZipTask.ts
```diff
@@ -169,6 +176,29 @@ export default class ImportMarkdownZipTask extends ImportTask {
...
+        const referredDocPath = match[1];
+        const normalizedDocPath = decodeURI(
+          path.normalize(`${basePath}/${referredDocPath}`)
+        );
+
+        const referredDocId = docPathToIdMap.get(normalizedDocPath);
+        if (referredDocId) {
+          document.text = document.text.replace(
+            referredDocPath,
+            `<<${referredDocId}>>`
```
hmacr, line 198: This is handled during persistence [here](https://github.com/outline/outline/blob/main/server/queues/tasks/ImportTask.ts#L450) (gets replaced with `doc/{docId}`).
This gets handled in the UI as the app takes care of redirection based on id / slug.
However, when collections with this kind of docs are exported, the internal doc links are not replaced with the corresponding relative URLs.

I'm working on another PR that essentially removes `doc/{docId}` trick and sets the proper doc slug during persistence itself. For backward compatibility, the exporters will handle the `docId` replacement as well.

hmacr:
#7089 is a problem for JSON importer as well - I figured it can be fixed as part of #6174 (avoids re-work).

However, I can add the JSON fix to this PR if it's of high priority.

server/queues/tasks/ImportMarkdownZipTask.ts
tommoor: > I'm working on another PR that essentially removes doc/{docId} trick and sets the proper doc slug

👏

tommoor (approved):

PR #7266, opened 2024-07-17
Archive collections

### LINKED ISSUE #3141: Archive collections
### Discussed in https://github.com/outline/outline/discussions/3138

<sup>Originally posted by **michaelkarrer81** February 16, 2022</sup>
We use collections often as a "Project Wiki". After a project is done we would like to archive the full collectrion with all of it's content. It might be used as a reference in the future but only in very rare occasions. </div>

apoorv-mishra:
Closes #3141 

## Backend
- [x] Migration
- [x] Permissions
- [x] Endpoints:
  - [x] `collections.archive`
  - [x] `collections.restore`
  - [x] `collections.archived`
  - [x] `collections.list`
  - [x] `documents.restore`
- [x] Events
  - [x] `collections.archive`
  - [x] `collections.restore`

## Frontend
- [x] Revamp "Archive" section in sidebar to display the collection structure
- [x] Collection menu options _Archive_ & _Restore_
- [x] Banner for archived collection
- [x] Collection should be archived upon dragging and dropping onto the "Archive" section
- [x] Events
  - [x] `collections.archive`
  - [x] `collections.restore`

### On going fixes
- [x] ~"Archive" section isn't active when a collection under it _is_ active~ Prevent highlighting the "Archive" section in sidebar if any doc/collection under it _is_ active
- [x] Disable drag & drop upon any document under "Archive" section
- [x] Have "Restore to" menu option for a document under archived collection


server/models/User.ts
```diff
@@ -308,6 +308,30 @@ class User extends ParanoidModel<
...
+    return this.teamId === teamId;
+  }
+
+  /**
+   * Whether the user is a given team's admin.
+   *
+   * @param team The team to check
+   * @returns Whether the user is an admin of the team
+   */
+  public isTeamAdmin(team?: Team | string | null) {
```
apoorv-mishra, line 331: Aware of another `isTeamAdmin` under `policies/utils`. Open to thoughts on which one should we stick to...

server/policies/collection.test.ts
```diff
@@ -40,7 +40,11 @@ describe("admin", () => {
       teamId: team.id,
       permission: CollectionPermission.Read,
     });
-    const abilities = serialize(user, collection);
+    // reload to get membership
+    const reloaded = await Collection.scope({
+      method: ["withMembership", user.id],
+    }).findByPk(collection.id);
```
apoorv-mishra, line 46: Reload wasn't required earlier because of early return within policy check, preventing control to invoke `includeMemberships(...)`. Now, with `and...or` pattern it becomes mandatory.

server/routes/api/collections/collections.ts
```diff
@@ -871,6 +879,167 @@ router.post(
...
+    // Restore the entire document structure within collection
+    for (const document of documents) {
+      collection = await collection!.addDocumentToStructure(
+        document,
+        undefined,
+        {
+          transaction,
+        }
+      );
+    }
```
apoorv-mishra, line 983: Should be async perhaps...

server/models/Document.ts
```diff
@@ -928,68 +930,69 @@ class Document extends ParanoidModel<
...
       }
+    }
 
-      if (!this.template && this.publishedAt && collection) {
-        await collection.addDocumentToStructure(this, undefined, {
-          transaction,
-        });
-        if (this.collection) {
-          this.collection.documentStructure = collection.documentStructure;
-        }
```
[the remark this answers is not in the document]
tommoor:
```
          this.collection.documentStructure = collection.documentStructure;
```

Is there so that the collection in memory is updated and consistent, we do this everywhere that `addDocumentToStructure` is called, so perhaps the same work could just be done inside that method?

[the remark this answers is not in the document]
apoorv-mishra: Although I can't find a place where it's later accessed(checked follow up logic in handler and hooks), still, putting it back since I'm not completely sure 🤔

app/components/Sidebar/components/ArchiveLink.tsx
```diff
@@ -61,7 +61,6 @@ function ArchiveLink() {
           icon={<ArchiveIcon open={isDocumentDropping} />}
           exact={false}
           label={t("Archive")}
-          active={documents.active?.isArchived && !documents.active?.isDeleted}
```
apoorv-mishra: Made this choice to prevent scenarios like 👇

Screenshot 2024-09-11 at 3 54 15 PM

Doesn't look good!

That being said, same behavior holds for the current collection structure in sidebar. If you make any doc under it _active_, the highlight is stolen from the collection.

tommoor: Yes, it should match collection->document highlight behavior 👍

apoorv-mishra:
I think it's ready for an initial review with some minor work left 👇 

- [ ] When an archived collection page is reloaded, it doesn't show any documents under "Documents" tab
- [ ] More thorough unit tests on backend
- [ ] Hide behind a feature flag

cc: @tommoor

apoorv-mishra:
> - [x] When an archived collection's page is reloaded, it doesn't show any documents under "Documents" tab

Got rid of _tabs_ in archived collection page as part of ☝️ , let me know if we've an argument to keep them, will restore back if so

Screenshot 2024-09-20 at 5 01 56 PM

server/routes/api/collections/collections.ts
```diff
@@ -881,6 +892,188 @@ router.post(
...
+
+    ctx.body = {
+      data: await presentCollection(ctx, collection!),
+      policies: presentPolicies(user, [collection]),
+    };
+  }
+);
+
+router.post(
+  "collections.archived",
```
apoorv-mishra, line 1027: Maybe we can get rid of this in favor or `collections.list` with `StatusFilter`, similar to `documents.list`?

cc: @tommoor

apoorv-mishra: Don't mind taking this as a follow up though. Helps cut down further work here – lots of small tasks really add up 🙈

tommoor: As it's brand new though… it should reduce code overall

server/types.ts
```diff
@@ -184,6 +183,16 @@ export type DocumentEvent = BaseEvent<Document> &
...
         };
       }
+    | {
+        name: "documents.unarchive";
+        documentId: string;
+        collectionId: string;
+        data: {
+          title: string;
+          /** Id of collection from which the document is unarchived */
+          srcCollectionId: string;
```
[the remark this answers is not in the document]
apoorv-mishra: Needed because of https://github.com/outline/outline/commit/c5fad7587f4b3e481602cd7f5c560c5a8649b9b1.

server/routes/api/documents/documents.ts
```diff
@@ -694,81 +693,100 @@ router.post(
...
+      await document.restoreTo(destCollectionId!, { transaction, user }); // destCollectionId is guaranteed to be defined here
+      await Event.createFromContext(
+        ctx,
+        {
+          name: "documents.unarchive",
+          documentId: document.id,
+          collectionId: document.collectionId,
+          data: {
+            title: document.title,
+            srcCollectionId,
```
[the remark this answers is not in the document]
apoorv-mishra: >Needed because of https://github.com/outline/outline/commit/c5fad7587f4b3e481602cd7f5c560c5a8649b9b1.

Same!

server/routes/api/collections/collections.ts
```diff
@@ -881,6 +892,187 @@ router.post(
...
+    await Document.update(
+      {
+        lastModifiedById: user.id,
+        archivedAt: null,
+      },
+      {
+        where: {
+          collectionId: collection.id,
+          teamId: user.teamId,
+          archivedAt: collection.archivedAt,
```
apoorv-mishra, line 995: Wanted to especially highlight this! The condition here is put with the intention that _only_ the docs which were archived as part of collection archival will be restored. Although there's a catch in a _rare_ scenario where a doc within a collection was _separately_ archived at the _same time_ the collection was archived – in this case, both might end up having the _same_ `archivedAt`. Then, if the collection is restored, the doc which was separately archived, would be restored as part of collection restoration, which would be an unintended behavior.

apoorv-mishra: Maybe _locking_ rows and then setting `archivedAt` to `transaction_timestamp()` could be a solution.

tommoor: >  there's a catch in a rare scenario where a doc within a collection was separately archived at the same time the collection was archived

I suppose I don't understand this, how would the timestamps be identical?

tommoor:
Overall this works really well 👍 , a couple of quick UX nits:

image

On the collection it would be nice if this is replaced with the "Archive" breadcrumb, especially as search doesn't work by default for archived docs anyway.

image

Similarly "Search in collection" can be removed for archived collections

image

Can you move "Archive…" below "Export".

server/routes/api/collections/collections.ts
apoorv-mishra: Suppose a user triggered `documents.archive` for a document with id = `docId` and _concurrently_ `collections.archive` for the collection to which `docId` belongs. In a rare scenario, this might end up setting `archivedAt`, for both the doc and collection, to same timestamp. Then 👇 
>if the collection is restored, the doc which was separately archived, would be restored as part of collection restoration, which would be an unintended behavior.

tommoor: But it would have to be the identical millisecond? Or am I missing something there

server/routes/api/collections/collections.ts
```diff
@@ -801,40 +807,80 @@ router.post(
   auth(),
   validate(T.CollectionsListSchema),
   pagination(),
+  transaction(),
```
tommoor, line 810: No need for transaction here as we're querying, not mutating

server/models/User.ts
tommoor: Would prefer to keep these methods under `policies`

tommoor (approved):

tommoor (approved):

server/routes/api/collections/collections.ts
apoorv-mishra: Yeah, exact same `new Date()` value for both. Hence, extremely rare.

server/routes/api/collections/collections.ts
```diff
@@ -1008,12 +1004,9 @@ router.post(
       method: ["withMembership", user.id],
     }).findByPk(id, {
       transaction,
+      rejectOnEmpty: true,
```
apoorv-mishra, line 1007: I think we might need to revert this back. It's kind of [problematic](https://github.com/outline/outline/issues/6207) with the `"withMembership"` scope.

tommoor: The behavior is overridden in Collection model, same as Document to get around this bug :)

pesaventofilippo:
Hi, I tried building a Docker image against `main` with these changes, and I still can't see the archive option from the UI. Is this something we have to enable manually? I'm a workspace admin and admin of the collection I'm trying to archive

PR #7270, opened 2024-07-18
fix: Set doc slug on importing collections

hmacr:
This PR removes the `doc/{docId}` format and sets the doc slug when collections are imported.

Documents which use `doc/{docId}` to link other documents do not have these references replaced with relative URLs when exported. This PR fixes this as well.

Bug mentioned [here](https://github.com/outline/outline/pull/7259#discussion_r1679128685)

Related #7259

### DIFF
--- a/server/queues/tasks/ImportTask.ts
+++ b/server/queues/tasks/ImportTask.ts
@@ -25,7 +25,7 @@ import {
 } from "@server/models";
 import { sequelize } from "@server/storage/database";
 import ZipHelper from "@server/utils/ZipHelper";
-import { generateDocUrlId } from "@server/utils/url";
+import { generateUrlId } from "@server/utils/url";
 import BaseTask, { TaskPriority } from "./BaseTask";
 
 type Props = {
@@ -563,7 +563,7 @@ export default abstract class ImportTask extends BaseTask<Props> {
         }
       }
 
-      doc.urlId = generateDocUrlId();
+      doc.urlId = generateUrlId();
     }
   }
 }
--- a/server/queues/tasks/ImportTask.ts
+++ b/server/queues/tasks/ImportTask.ts
@@ -25,6 +25,7 @@ import {
 } from "@server/models";
 import { sequelize } from "@server/storage/database";
 import ZipHelper from "@server/utils/ZipHelper";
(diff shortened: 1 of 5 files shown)


server/queues/tasks/ImportTask.ts
```diff
@@ -562,4 +545,25 @@ export default abstract class ImportTask extends BaseTask<Props> {
...
+            urlId: doc.urlId,
+          },
+        });
+
+        if (!existing) {
+          continue;
+        }
+      }
+
+      doc.urlId = generateDocUrlId();
```
hmacr, line 566: The chances of a collision with `{slugifiedTitle}-{urlId}` should be relatively low within a single machine.
To make it more resilient, retry can be added here; but I think it might be an overkill.

tommoor: Looks like the chances of collision are 1 in ~70millon, I can live with that

https://www.wolframalpha.com/input?i=1%2F2%2Bsqrt%281%2F4-2*%2836%5E10%29*ln%280.5%29%29

server/utils/url.ts
```diff
@@ -0,0 +1,5 @@
+import randomstring from "randomstring";
+
+const DocumentUrlIdLen = 10;
+
+export const generateDocUrlId = () => randomstring.generate(DocumentUrlIdLen);
```
tommoor, line 5:
```suggestion
const UrlIdLength = 10;

export const generateUrlId = () => randomstring.generate(UrlIdLength);
```

Lets just make it generic, the same method exists for collection – so could use this there too

tommoor (commented):
Nice, great improvement and also avoids a redirect when navigating to these imported documents

server/queues/tasks/ImportTask.ts
hmacr: _nit_:  The collision chances are 1 in ~1 billion, I believe (62 characters instead of 36).
https://www.wolframalpha.com/input?i=1%2F2+%2B+sqrt%281%2F4+%2B+2*%2862%5E10%29*ln%282%29%29

The actual collisions that we're worried about (_{title}-{urlId}_) will be even lower than `urlId` on its own.

tommoor (approved):

PR #7274, opened 2024-07-18
feat: Lossless JSON import

### LINKED ISSUE #6174: `JSONImportTask` should not go through Markdown converter

hmacr:
Closes #6174


server/queues/tasks/ImportJSONTask.ts
```diff
@@ -152,21 +153,79 @@ export default class ImportJSONTask extends ImportTask {
...
+      });
+
+      return Fragment.fromArray(nodes);
+    };
+
+    for (const document of output.documents) {
+      const node = Node.fromJSON(schema, document.data);
+      const transformedNode = node.copy(transformFragment(node.content));
+      document.data = transformedNode;
+      document.text = serializer.serialize(transformedNode);
```
tommoor, line 228: Do we end up serializing the node twice here, once on L81 and then again?

hmacr: Umm.. yes.
I think this `for` loop should be moved inside a `if (output.attachments.length) { /* logic to replace attachment urls */ }`.

hmacr: I was thinking that I could make this change once the initial review is over for this PR.
Please let me know when you've reviewed it.

tommoor: Yep, reviewed – this is the only thing that stood out.

hmacr:
@tommoor Merge conflicts have been resolved, please check.

PR #7275, opened 2024-07-18
feat: Invite groups to documents

### LINKED ISSUE #6639: Add the ability to add groups to individual documents
### Discussed in https://github.com/outline/outline/discussions/6555

<sup>Originally posted by **AlexShku** February 19, 2024</sup>
Hello! 
Thank you for the wonderful 0.75.0 update. 
It would be great if it was possible to add predefined groups and not just users individually to Documents. </div>

tommoor:
This PR adds all the required endpoints and UI to enable adding groups to documents.

closes #6639


server/models/Document.ts
```diff
@@ -563,22 +623,28 @@ class Document extends ParanoidModel<
   views: View[];
 
   /**
-   * Returns an array of unique userIds that are members of a document via direct membership
+   * Returns an array of unique userIds that are members of a document
```
tommoor, line 626: Perf here will be a bottleneck, these needs to be paginated soon

PR #7282, opened 2024-07-21
fix: Error when passed too many user invites

bkrausz:
Previous code was silently dropping invites beyond 20.


tommoor (approved):

server/routes/api/users/users.ts
```diff
@@ -433,13 +433,20 @@ router.post(
...
+      );
+    }
+
     const { user } = ctx.state.auth;
     const team = await Team.findByPk(user.teamId);
     authorize(user, "inviteUser", team);
 
     const response = await userInviter({
       user,
-      invites: invites.slice(0, UserValidation.maxInvitesPerRequest),
```
tommoor, line 442: Embarrassing this was like this 🙈

PR #7294, opened 2024-07-23
fix: Index calculation does not take into account user scope

### LINKED ISSUE #7283: Collection index calculation is creating increasingly-long names
It appears that the length of `Collection` index keeps growing such that eventually it exceeds 100 chars, causing an error.

**To Reproduce**
Make 800 collections (not sure if fully reproducible).

Here's our collections sorted by `createdAt` ascending. I included length and base64 so you can see the preceeding spaces:

```
index	base64	length
1	ICAgICAx	6
8	ICAgICAgIDg=	8
t	dA==	1
8	OA==	1
8	OA==	1
,	LA==	1
8	OA==	1
,	LA==	1
8	OA==	1
t	dA==	1
8	OA==	1
,	LA==	1
&	Jg==	1
#	Iw==	1
"	Ig==	1
!	IQ==	1
P	IFA=	2
8	IDg=	2
,	ICw=	2
&	ICY=	2
#	ICM=	2
"	ICI=	2
!	ICE=	2
P	ICBQ	3
8	ICA4	3
,	ICAs	3
&	ICAm	3
#	ICAj	3
"	ICAi	3
!	ICAh	3
P	ICAgUA==	4
8	ICAgOA==	4
,	ICAgLA==	4
&	ICAgJg==	4
#	ICAgIw==	4
"	ICAgIg==	4
!	ICAgIQ==	4
P	ICAgIFA=	5
8	ICAgIDg=	5
,	ICAgICw=	5
&	ICAgICY=	5
#	ICAgICM=	5
"	ICAgICI=	5
/	ICAgICAv	6
P	ICAgICBQ	6
8	ICAgICA4	6
,	ICAgICAs	6
&	ICAgICAm	6
!	ICAgICAh	6
D	ICAgICAgICBE	9
P	ICAgICAgUA==	7
P	ICAgICAgUA==	7
8	ICAgICAgOA==	7
,	ICAgICAgLA==	7
&	ICAgICAgJg==	7
#	ICAgICAgIw==	7
"	ICAgICAgIg==	7
!P	ICAgICAgIVA=	8
P	ICAgICAgIFA=	8
,	ICAgICAgICw=	8
&	ICAgICAgICY=	8
#	ICAgICAgICM=	8
"	ICAgICAgICI=	8
!	ICAgICAgICE=	8
P	ICAgICAgICBQ	9
8	ICAgICAgICA4	9
,	ICAgICAgICAs	9
&	ICAgICAgICAm	9
#	ICAgICAgICAj	9
"	ICAgICAgICAi	9
!	ICAgICAgICAh	9
P	ICAgICAgICAgUA==	10
8	ICAgICAgICAgOA==	10
,	ICAgICAgICAgLA==	10
&	ICAgICAgICAgJg==	10
#	ICAgICAgICAgIw==	10
"	ICAgICAgICAgIg==	10
!	ICAgICAgICAgIQ==	10
P	ICAgICAgICAgIFA=	11
8	ICAgICAgICAgIDg=	11
,	ICAgICAgICAgICw=	11
&	ICAgICAgICAgICY=	11
#	ICAgICAgICAgICM=	11
"	ICAgICAgICAgICI=	11
!	ICAgICAgICAgICE=	11
P	ICAgICAgICAgICBQ	12
8	ICAgICAgICAgICA4	12
,	ICAgICAgICAgICAs	12
&	ICAgICAgICAgICAm	12
#	ICAgICAgICAgICAj	12
"	ICAgICAgICAgICAi	12
!	ICAgICAgICAgICAh	12
P	ICAgICAgICAgICAgUA==	13
8	ICAgICAgICAgICAgOA==	13
,	ICAgICAgICAgICAgLA==	13
&	ICAgICAgICAgICAgJg==	13
#	ICAgICAgICAgICAgIw==	13
"	ICAgICAgICAgICAgIg==	13
!	ICAgICAgICAgICAgIQ==	13
P	ICAgICAgICAgICAgIFA=	14
```

You'll see that this keeps growing rather quickly, so not surprising that they'd exceed 100.

These were created via API call with no explicitly specified `index`

tommoor:
In a workspace with hundreds of collections the calculation of `index` became larger than the column limit

closes #7283

### DIFF
--- a/server/routes/api/collections/collections.ts
+++ b/server/routes/api/collections/collections.ts
@@ -60,29 +60,16 @@ router.post(
     const { user } = ctx.state.auth;
     authorize(user, "createCollection", user.team);
 
-    if (!index) {
-      const collections = await Collection.findAll({
-        where: {
-          teamId: user.teamId,
-          deletedAt: null,
-        },
-        attributes: ["id", "index", "updatedAt"],
-        limit: 1,
-        order: [
-          // using LC_COLLATE:"C" because we need byte order to drive the sorting
-          Sequelize.literal('"collection"."index" collate "C"'),
-          ["updatedAt", "DESC"],
-        ],
+    if (index) {
+      index = await removeIndexCollision(user.teamId, index, { transaction });
+    } else {
+      const first = await Collection.findFirstCollectionForUser(user, {
+        attributes: ["id", "index"],
         transaction,
       });
-
-      index = fractionalIndex(
-        null,
-        collections.length ? collections[0].index : null
-      );
+      index = fractionalIndex(null, first ? first.index : null);
     }
 
-    index = await removeIndexCollision(user.teamId, index);
     const collection = Collection.build({
       name,
       content: data,
@@ -891,7 +878,7 @@ router.post(
     });
     authorize(user, "move", collection);
 
-    index = await removeIndexCollision(user.teamId, index);
+    index = await removeIndexCollision(user.teamId, index, { transaction });
     await collection.update(
       {
         index,
--- a/server/models/Collection.ts
+++ b/server/models/Collection.ts
@@ -197,8 +197,8 @@ class Collection extends ParanoidModel<
   color: string | null;
 
   @Length({
-    max: 100,
-    msg: `index must be 100 characters or less`,
+    max: 256,
+    msg: `index must be 256 characters or less`,
   })
   @Column
   index: string | null;
--- a/server/models/Collection.ts
+++ b/server/models/Collection.ts
@@ -425,13 +425,18 @@ class Collection extends ParanoidModel<
   /**
    * Find the first collection that the specified user has access to.
    *
-   * @param user User object
+   * @param user User to find the collection for
+   * @param options Additional options for the query
    * @returns collection First collection in the sidebar order
    */
-  static async findFirstCollectionForUser(user: User) {
+  static async findFirstCollectionForUser(
+    user: User,
+    options: FindOptions = {}
+  ) {
     const id = await user.collectionIds();
     return this.findOne({
       where: {
+        teamId: user.teamId,
         id,
       },
       order: [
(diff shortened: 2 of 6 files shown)


server/routes/api/collections/collections.ts
```diff
@@ -60,29 +60,16 @@ router.post(
...
       });
-
-      index = fractionalIndex(
-        null,
-        collections.length ? collections[0].index : null
-      );
+      index = fractionalIndex(null, first ? first.index : null);
     }
 
-    index = await removeIndexCollision(user.teamId, index);
```
tommoor, line 85: No point checking for index collision if we just calculated it

apoorv-mishra: Related: I'm wondering if `collectionIndexing` could be removed from `collections.list` endpoint 🤔 
```js
    const nullIndex = collections.findIndex(
      (collection) => collection.index === null
    );

    if (nullIndex !== -1) {
      const indexedCollections = await collectionIndexing(user.teamId);
      collections.forEach((collection) => {
        collection.index = indexedCollections[collection.id];
      });
    }
```

I don't see an obvious place where collection `index` is set to `null`(checked `collections.create` & `collections.update`). Perhaps, it's never set to `null` 🤔

I'm wondering if this indexing part was added as an [_in-place migration_ strategy](https://github.com/outline/outline/pull/2953/commits/e9789b95bd5595b5843752cbe1c2be4a9887caa2). In that case we can safely remove this if migration is achieved.

tommoor: > I'm wondering if this indexing part was added as an [in-place migration strategy](https://github.com/outline/outline/pull/2953/commits/e9789b95bd5595b5843752cbe1c2be4a9887caa2).

Yep, I think it doesn't do any harm to keep? We occasionally get workspaces that become active again after a year or two of no usage

Issue #2018, opened 2021-04-08
Add support for Dropbox embeds


tommoor:
This will be super useful for embedding a single file or folder in a doc.

- https://dropbox.tech/developers/new-file-and-folder-embedder-launched-in-preview
- https://dropbox.tech/developers/now-available--dropbox-embedder

tommoor:
I don't remember unfortunately, presumably it didn't work very well

LouisKottmann:
Hello again.

I have a working branch, and my client OK'd upstreaming it. Expect a PR in the coming days.

PR #7299, opened 2024-07-24
add Dropbox embeddings support

LouisKottmann:
This PR adds support for [Dropbox Embedder](https://www.dropbox.com/developers/embedder) as promised in #2018 

Images, PDF, Slides, Excel, GIFs and even Dropbox Folders will show up whenever a dropbox link is shared, as is expected from outline embeds.

Since Dropbox Embedder uses an HTML anchor element with a magic CSS class applied to it, and a JS script to be loaded, which is unlike most other embeds/plugins, I had to improvise some bits but nothing crazy I believe (I have never touched React before though).

I have successfully built a Docker image by following `.circleci/config.yml` from this repo, and deployed it on my self-host instance, and it works as far as I manually tested (CRUD operations on documents).


tommoor:
Hey, thanks for taking the time – this is a good first pass.

For rendering it doesn't make sense to inject the Dropbox script into the top level of every page – it would be better if this third party script is completely isolated. Take a look at the implementation for GitHub and GitLab which I think this should closely follow:

https://github.com/outline/outline/blob/e033b20d6fa1d3e67a18e77e77219f9c639c91b0/server/routes/embeds.ts#L39

We'll also want to disable the Dropbox item in the menu if `DROPBOX_APP_KEY` is not set

LouisKottmann:
I have disabled the Dropbox item in the menu if DROPBOX_APP_KEY is not set and refined the height a bit.

However for the script injection it's not clear to me how to proceed.
Github and Gitlab embeds seem to loop around to outline calling itself on `/embeds/(github|gitlab)` in order to refine their respective iframes settings.

But Dropbox embedder works differently as it is not an iframe. The dropbox script (dropins.js) transforms the anchor element into other nested HTML elements.

It feels fine and clean to me to have the dropbox script load if it is configured for the outline instance, as it should be cached anyway?

Otherwise, can you give my another pointer on where to implement the dropbox script injection please?

tommoor:
> It feels fine and clean to me to have the dropbox script load if it is configured for the outline instance

This makes more sense in self-hosted, but this code is also ran in multi-tenant cloud where it would be loaded for everyone unneccessarily.

> But Dropbox embedder works differently as it is not an iframe. The dropbox script (dropins.js) transforms the anchor element into other nested HTML elements.

This is similar to how the GitHub embed works actually, there are a couple of reasons for this "loop around" – but the main is complete isolation of the injected content for security, the Dropbox script will run in that iframe which has it's own CSP rules that do not affect the rest of the platform.

tommoor (approved):

tommoor:
Works great, thanks for working through those changes

PR #7301, opened 2024-07-24
Suppress document publish notification upon document creation

### LINKED ISSUE #7300: Document title is missing from 'new document published' notification emails
After upgrading from version 0.75.0 to 0.78.0 we noticed that document titles are now missing from 'new document published' notification emails. (not sure if it's technically a bug, as it might be related to the new 'auto-publish' feature introduced in https://github.com/outline/outline/pull/7243, but it looks weird when I get a bunch of these emails)

**To Reproduce**
1. Go to a published collection
2. Create a new document with the title 'Test'
3. Click on 'Done editing'
4. Check the inbox of a user subscribed to the collection

**Expected behavior**
Document title is present in the email (?)

**Screenshots**
image

**Outline**
- Install: self-hosted
- Version: outlinewiki/outline:0.78.0 image from DockerHub as of 07/23/2024

apoorv-mishra:
Closes #7300

### DIFF
--- a/server/commands/documentCreator.ts
+++ b/server/commands/documentCreator.ts
@@ -156,23 +156,25 @@ export default async function documentCreator({
       silent: true,
       transaction,
     });
-    await Event.create(
-      {
-        name: "documents.publish",
-        documentId: document.id,
-        collectionId: document.collectionId,
-        teamId: document.teamId,
-        actorId: user.id,
-        data: {
-          source: importId ? "import" : undefined,
-          title: document.title,
+    if (document.title) {
+      await Event.create(
+        {
+          name: "documents.publish",
+          documentId: document.id,
+          collectionId: document.collectionId,
+          teamId: document.teamId,
+          actorId: user.id,
+          data: {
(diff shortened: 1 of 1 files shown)


apoorv-mishra (commented):
The following processors handle `documents.publish` event:
1. `BacklinksProcessor` - No need to handle the event when there are no backlinks in it because a newly created document is empty.
2. `NotificationsProcessor` - No need since it's a newly created document. Should instead have a notification for `documents.create` in this case.
3. `RevisionsProcessor` - No need because no new revision created here.
4. `SlackProcessor` - Same rationale as for `NotificationsProcessor` above.

tommoor:
Definitely need to prioritize a fix for this, I'm not sure if this is it though as there are folks that are subscribed to "new document in collection" notifications, `documents.publish` webhooks, slack integration etc that will now no longer post.

I'm tempted to merge for a temporary fix to remove these empty notifs while we think about how to restore these in a useful way.

tommoor (approved):

PR #7318, opened 2024-07-27
fix: use all properties from zip when importing a collection

### LINKED ISSUE #7317: Importing a collection sometimes does not set the right properties from zip
When a collection (that does not have a "name" clash with the existing collections) is imported, some properties like icon, color, permission, etc. are not set with the right values from the zip.

**Cause:** https://github.com/outline/outline/blob/main/server/queues/tasks/ImportTask.ts#L367

hmacr:
Closes #7317

### DIFF
--- a/server/queues/tasks/ImportTask.ts
+++ b/server/queues/tasks/ImportTask.ts
@@ -2,6 +2,7 @@ import path from "path";
 import fs from "fs-extra";
 import chunk from "lodash/chunk";
 import truncate from "lodash/truncate";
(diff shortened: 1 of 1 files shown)


tommoor (approved):
I might consider refactoring this into a `sharedDefaults` object that both code paths can use to avoid similar future mistakes 🤔

hmacr:
True - makes it infallible to a certain extent.

tommoor (approved):

PR #7319, opened 2024-07-28
Feature: Add generic Iframe support

Blendman974:
Add the command /iframe.
outline-iframe

Simpler version of #6947 :  
Added a generic `EmbedDescriptor` that matches all links.  

To prevent making all links Iframes (when pasting for example), there is a  `noAutoMatch` boolean (default to false) on the `EmbedDescriptor` that prevents it from being processed as an embed on paste.

Icon is from #5137


shared/editor/embeds/index.tsx
```diff
@@ -621,6 +624,17 @@ const embeds: EmbedDescriptor[] = [
     icon: <Img src="/images/youtube.png" alt="YouTube" />,
     component: YouTube,
   }),
+  /* The generic iframe embed should always be the last one */
+  new EmbedDescriptor({
+    title: "Iframe",
```
tommoor, line 629:
```suggestion
    title: "Embed",
```

This seems like better language for the average user, iframe can remain as a search keyword.

shared/editor/embeds/index.tsx
```diff
@@ -55,6 +55,8 @@ export class EmbedDescriptor {
   defaultHidden?: boolean;
   /** Whether the bottom toolbar should be hidden – use this when the embed itself includes a footer */
   hideToolbar?: boolean;
+  /** Whether the embed should be auto matched when pasting a URL */
+  noAutoMatch?: boolean;
```
tommoor, line 59:
```suggestion
  matchOnPaste?: boolean;
```

Maybe we just get explicit with this?

Blendman974:
Renamed "Iframe" to "Embed" (also changed the asset name for consistency)
And changed the boolean to matchOnPaste. (defaults to true so we don't need to add it on every existing Embeds)
Thx @tommoor

rcronin:
Any idea when this will get merged into the code base? Really excited about this feature

tommoor:
Needs testing and probably will choose another image from our existing libraries, but other than that it's likely to be in v0.79.0

airbenich:
How is the size (width & height) of the embeding calculated? Can the width & height be set manually?

Geequlim:
This feature is simply amazing! Please merge this PR as soon as possible and release a new version—I can't wait any longer!

tommoor (approved):

florian-obradovic:
Awesome - I get a refused to connect:
CleanShot 2024-09-05 at 21 07 06@2x

tommoor:
In that case Plane is blocking the frame from rendering, there's nothing we can do on Outline side.

florian-obradovic:
@tommoor thanks.
Any suggestions what causes the block?
So, is it a web server setting or in the application code?
It happens with all sites I tried.

tommoor:
If you open developer tools in the browser, the error will be visible in the network tab.

PR #7344, opened 2024-08-02
Make share dialog scrollable

### LINKED ISSUE #7328: Share modal doesn't scroll when it gets too long
On latest Chrome...you can see the modal flows below the window (look for the border radius):

CleanShot 2024-07-30 at 16 58 26@2x

Flexbox doesn't seem to play nice with overflow: scroll. Also seeing this on mobile. I got a fix working for desktop, but it's somewhat ugly so decided to punt to a ticket:

```
diff --git a/app/components/Modal.tsx b/app/components/Modal.tsx
index 5cb396b51..10017c9d8 100644
--- a/app/components/Modal.tsx
+++ b/app/components/Modal.tsx
@@ -120,7 +120,7 @@ const Modal: React.FC<Props> = ({
                     <SmallContent style={style} shadow>
                       <ErrorBoundary component="div">{children}</ErrorBoundary>
                     </SmallContent>
-                    <Header>
+                    <Header shrink={false}>
                       {title && <Text size="large">{title}</Text>}
                       <NudeButton onClick={onRequestClose}>
                         <CloseIcon />
@@ -256,7 +256,8 @@ const Header = styled(Flex)`
 const Small = styled.div`
 
-  margin: 25vh auto auto auto;
+  margin: 20vh auto auto auto;
+  max-height: 75vh;
   width: 75vw;
   min-width: 350px;
   max-width: 450px;
@@ -285,6 +286,7 @@ const Small = styled.div`
 
 const SmallContent = styled(Scrollable)`
   padding: 12px 24px 24px;
+  overflow: hidden auto;
 `;
 
 export default observer(Modal);
```

I think the right approach is to move off of flexbox for the modal, then `max-height`, etc, can work better, but not positive.

apoorv-mishra:
Closes #7328


apoorv-mishra:
Screenshot 2024-08-02 at 23 20 49

We should perhaps keep this part out of the scrollable container and make it "sticky" at the bottom of the popover.

tommoor:
> We should perhaps keep this part out of the scrollable container and make it "sticky" at the bottom of the popover.

There's definitely good arguments for it, I wouldn't increase the scope of this even more if we can help it though.

tommoor:
There's a lot of refactoring here, can you note the spots where changes _matter_?

apoorv-mishra:
Yeah, most important changes are the new `AccessControlList.tsx` components. The idea is ported from the exiting `Suggestions.tsx` components where the list items are wrapped in a scrollable container.

app/components/Sharing/Collection/SharePopover.tsx
```diff
@@ -367,35 +362,7 @@ function SharePopover({ collection, visible, onRequestClose }: Props) {
...
-                  });
-                }}
-                disabled={!can.update}
-                value={collection?.permission}
-                labelHidden
-                nude
-              />
-            </div>
-          }
-        />
```
apoorv-mishra, line 396: Note that these dispersed list items had to be colocated within a single component so as to wrap everything up inside a scrollable container.

PR #7358, opened 2024-08-06
Fetch collections upon initial load

### LINKED ISSUE #7357: Sidebar hidden
Collapsing the "Collections" section in sidebar, followed by a reload, makes the sidebar hidden 👇 

https://github.com/user-attachments/assets/b944280e-d67d-40f7-920d-c985ca555d03

apoorv-mishra:
Closes #7357

### DIFF
--- a/app/components/Sidebar/App.tsx
+++ b/app/components/Sidebar/App.tsx
@@ -5,7 +5,6 @@ import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 import { useTranslation } from "react-i18next";
 import styled from "styled-components";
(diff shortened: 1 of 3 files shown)


tommoor:
But now we'll be showing a flash of collection loading state again :( I think it's better to always load collections, even if that sidebar item is collapsed.

Sidenote: I've been thinking it probably makes sense soon to implement some sort of init endpoint that returns a lot of the data needed on load in a single request.

tommoor (changes requested):
Now `collections.list` is called twice when collections is open. How about we change this to `fetchAll` and remove the `fetch` from the `sidebar/Collections`?

tommoor (approved):

PR #7366, opened 2024-08-09
feat: add Umami

QYG2297248353:
#7280 

Finish Add Umami Support

> 但是存在一些小问题，我无法去解决：
> 1. 集成应用 logo, 我没有找到合适的 logo，官方的svg放上去，显得格格不入。
> 2. i18n 支持，当前仅仅支持 English，中文简体，中文繁体。

## Translate
But there are some minor problems that I can't solve: 
1. Integrate the application logo. I didn't find a suitable logo. When I put the official svg on it, it seemed out of place. 
2. i18n support, currently only supports English, Simplified Chinese, and Traditional Chinese.


app/components/Analytics.tsx
```diff
@@ -106,6 +106,24 @@ const Analytics: React.FC = ({ children }: Props) => {
...
+  // Umami
+  React.useEffect(() => {
+    (env.analytics as PublicEnv["analytics"]).forEach((integration) => {
+      if (integration.service !== IntegrationService.Umami) {
+        return;
+      }
+
+      const script = document.createElement("script");
+      script.defer = true;
+      script.src = `${integration.settings?.instanceUrl}/${integration.settings?.scriptName}`;
```
tommoor, line 118: This is not acceptable as it would allow injecting any script at all, as far as I can tell it is always hosted at `/script.js` ?

QYG2297248353: Different, umami.js was specified in v1, and then the official default value given in v2 was script.js, and at the same time, self-built users were allowed to set the script name to avoid being blocked by AD advertising plug-ins.

image

You can refer to the official installation documentation of Umami. There is nothing to specify the script name. First of all, this is something that administrators have permission to do. 
This situation is only for self-built Umami, and the time value is fixed when using Umami Cloud.

[https://umami.is/docs/environment-variables](https://umami.is/docs/environment-variables)

tommoor: That's a shame, that does mean that this integration won't be possible to support in cloud then - you might as well change this so that `instanceUrl` is just the entire path to the script then? There isn't really a benefit to storing the parts separately and concatenating

server/env.ts
```diff
@@ -385,6 +385,31 @@ export class Environment {
     environment.GOOGLE_ANALYTICS_ID
   );
 
+  /**
+   * A Umami analytics domain to track user interactions.
+   */
+  @Public
+  @IsOptional()
+  public UMAMI_SCRIPT_DOMAIN = this.toOptionalString(
```
tommoor, line 393: There is no need for any of these additions in env, as it can all be configured through the integration settings page – the existing `GOOGLE_ANALYTICS_ID` here is just for backwards compatibility to much older versions of Outline

app/components/Analytics.tsx
QYG2297248353: According to the original development logic, I don't want to break the `instanceUrl` format and cause more trouble.

Secondly, the `instanceUrl ` should also be added to the SCP permission list, I think, to avoid the trouble of dealing with the `instanceUrl ` modification if it is standardized at a later stage.

server/env.ts
QYG2297248353: Ok, I can choose to remove it if I don't need it, I had hoped that, when the ops installed it, it would be done with a one-click deployment, eliminating the need for an administrator to log in and configure the integration.

pxlfrk:
Any update on this?

app.json
```diff
@@ -3,7 +3,13 @@
   "description": "Open source wiki and knowledge base for growing teams",
```
tommoor, line 3: Please revert all changes in this file.

server/services/web.ts
```diff
@@ -42,6 +43,14 @@ if (env.GOOGLE_ANALYTICS_ID) {
   scriptSrc.push("www.google-analytics.com");
 }
 
+if (environment.UMAMI_SCRIPT_DOMAIN) {
```
tommoor, line 46: Please remove changes in this file, csp is injected automatically for analytics integrations

shared/i18n/locales/zh_CN/translation.json
```diff
@@ -1017,8 +1017,12 @@
   "Configure a Matomo installation to send views and analytics from the workspace to your own Matomo instance.": "配置一个 Matomo 安装，将视图和分析从工作区发送到你的 Matomo 实例。",
   "Instance URL": "实例 URL",
   "The URL of your Matomo instance. If you are using Matomo Cloud it will end in matomo.cloud/": "你的 Matomo 实例 URL。如果你使用 Matomo 云，它将在 matomo.cloud/ 中结束。",
+  "The URL of your Umami instance. If you are using Umami Cloud it will end in https://cloud.umami.is/": "你的 Umami 实例 URL。如果你使用 Umami Cloud，请配置 https://cloud.umami.is/。",
```
tommoor, line 1020: Remove these changes as the file is auto-generated from [CrowdIn](https://translate.getoutline.com/)

shared/i18n/locales/zh_TW/translation.json
```diff
@@ -1016,9 +1016,13 @@
   "Create a \"Web\" stream in your Google Analytics admin dashboard and copy the measurement ID from the generated code snippet to install.": "在 Google Analytics 管理儀表板中創建 \"Web\" 流，並從生成的程式碼片段中複製 measurement ID 進行安裝。",
   "Configure a Matomo installation to send views and analytics from the workspace to your own Matomo instance.": "Configure a Matomo installation to send views and analytics from the workspace to your own Matomo instance.",
   "Instance URL": "Instance URL",
-  "The URL of your Matomo instance. If you are using Matomo Cloud it will end in matomo.cloud/": "The URL of your Matomo instance. If you are using Matomo Cloud it will end in matomo.cloud/",
-  "Site ID": "Site ID",
-  "An ID that uniquely identifies the website in your Matomo instance.": "An ID that uniquely identifies the website in your Matomo instance.",
+  "The URL of your Matomo instance. If you are using Matomo Cloud it will end in matomo.cloud/": "你的 Matomo 實例 URL。如果你使用 Matomo 雲，它將在 matomo.cloud/ 中結束。",
```
tommoor, line 1019: Remove these changes as the file is auto-generated from [CrowdIn](https://translate.getoutline.com/)

tommoor (commented):
If we can tidy up I can merge for next release

QYG2297248353:
> If we can tidy up and get lint passing I can merge for next release

Fix Finish

PR #7377, opened 2024-08-13
feat: Cache count of group members

tommoor:
This PR also adds a decorator to make it easy to cache any relationship count on any model going forward. 

This is useful foundational work towards https://github.com/outline/outline/pull/7275 as we really need to stop querying and returning all of the group memberships for performance (and reuse that for association for document memberships).

### DIFF
--- a/server/models/Group.ts
+++ b/server/models/Group.ts
@@ -1,4 +1,9 @@
-import { InferAttributes, InferCreationAttributes, Op } from "sequelize";
+import {
+  InferAttributes,
+  InferCreationAttributes,
+  Op,
+  type DestroyOptions,
+} from "sequelize";
 import {
   AfterDestroy,
   BelongsTo,
@@ -11,6 +16,7 @@ import {
   DataType,
   Scopes,
 } from "sequelize-typescript";
+import { CacheHelper } from "@server/utils/CacheHelper";
 import GroupMembership from "./GroupMembership";
 import GroupUser from "./GroupUser";
 import Team from "./Team";
@@ -81,7 +87,7 @@ class Group extends ParanoidModel<
   // hooks
 
   @AfterDestroy
-  static async deleteGroupUsers(model: Group) {
+  static async deleteGroupUsers(model: Group, options?: DestroyOptions<Group>) {
     if (!model.deletedAt) {
       return;
     }
@@ -89,11 +95,13 @@ class Group extends ParanoidModel<
       where: {
         groupId: model.id,
       },
+      transaction: options?.transaction,
     });
     await GroupMembership.destroy({
       where: {
         groupId: model.id,
       },
+      transaction: options?.transaction,
     });
   }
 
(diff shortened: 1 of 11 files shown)


server/models/Group.ts
```diff
@@ -78,25 +78,6 @@ class Group extends ParanoidModel<
   @Column
   name: string;
 
-  // hooks
-
-  @AfterDestroy
-  static async deleteGroupUsers(model: Group) {
```
tommoor, line 84: No longer needed, these relationships cascade deletes

server/models/Group.ts
```diff
@@ -128,6 +109,9 @@ class Group extends ParanoidModel<
 
   @BelongsToMany(() => User, () => GroupUser)
   users: User[];
+
+  @CounterCache(() => GroupUser, { as: "members", foreignKey: "groupId" })
+  memberCount: Promise<number>;
```
tommoor, line 114: All that's needed to create a cached count column 👌

server/routes/api/groups/groups.ts
```diff
@@ -284,12 +284,16 @@ router.post(
     });
 
     if (!groupUser) {
-      await group.$add("user", user, {
-        through: {
+      await GroupUser.create(
```
tommoor, line 287: Had to change these otherwise the `afterCreate` model hooks are not triggered

server/routes/api/groups/groups.ts
```diff
@@ -343,26 +347,37 @@ router.post(
...
-        },
       },
-      { transaction }
-    );
+      transaction,
+      lock: transaction.LOCK.UPDATE,
+    });
+
+    if (groupUser) {
+      await groupUser.destroy({ transaction });
```
tommoor, line 360: Had to change these otherwise the `afterDestroy` model hooks are not triggered

server/utils/CacheHelper.ts
```diff
@@ -1,32 +1,30 @@
...
 export class CacheHelper {
+  // Default expiry time for cache data in ms
   private static defaultDataExpiry = Day;
 
   /**
    * Given a key, gets the data from cache store
    *
    * @param key Key against which data will be accessed
    */
-  public static async getData(key: string): ReturnType<UnfurlSignature> {
```
tommoor, line 17: Weird that this helper had a bunch of Unfurl types @apoorv-mishra

server/utils/CacheHelper.ts
```diff
@@ -78,4 +62,16 @@ export class CacheHelper {
...
+
+  // keys
+
+  /**
+   * Gets key against which unfurl response for the given url is stored
+   *
+   * @param teamId The team ID to generate a key for
+   * @param url The url to generate a key for
+   */
+  public static getUnfurlKey(teamId: string, url = "") {
```
tommoor, line 74: Leaving this here for now

server/utils/CacheHelper.ts
apoorv-mishra: Yeah, doesn't quite make sense for a generic(hopefully) helper's methods. I don't recall a specific reason for this though. Perhaps, just an oversight.

server/models/Group.ts
apoorv-mishra: Sweet 👌

PR #7401, opened 2024-08-16
fix: Comment out IFRAMELY items in sample config

matthijskooijman:
The IFRAMELY_URL variable is marked as IsOptional and IsUrl in plugins/iframely/server/env.ts, but it seems that IsUrl refuses an empty value, but accepts a value that is absent.

In the .env.sample file the value is empty, causing the default config to fail with:

    [backend] [api] - IFRAMELY_URL must be a URL address

This commit comments out the IFRAMELY_URL variable (and for consistency also IFRAMELY_API_KEY) which makes the default config work without requiring changes to these variables (and anyone that wants to use IFRAMELY will figure out the uncomment them).


tommoor:
Thanks, I fixed the validation instead (a65d126cc)

matthijskooijman:
> Thanks, I fixed the validation instead (https://github.com/outline/outline/commit/a65d126ccf7755e491c79d62612fc3a99af99509)

Ah, that is indeed a better solution, but I did not know how (I thought that would need changes in class-validator maybe). Your solution looks good to me!

PR #7407, opened 2024-08-17
Client lifecycle hooks

tommoor:
We have over time created a bundle of imperative code handling side effects, this adds frontend model lifecycle hooks in the same format as the backend hooks provided by Sequelize to allow us to tidy these things up.


tommoor:
I'm going to merge as this is useful to complete https://github.com/outline/outline/pull/7460

Issue #4271, opened 2022-10-12
Mattermost Integration


tommoor:
We would like to build support for posting updates to Mattermost channels in the same way as Slack. It's likely that some of the Slack code can be refactored to be shared at the same time.

gnacoding:
Would love to see an mattermost integration!

antran22:
I think we can refactor: https://github.com/outline/outline/blob/3212d37ca5e6e061d454e10779a2ab03838cb8f3/plugins/slack/server/processors/SlackProcessor.ts#L17 into an abstract base class for Slack/Mattermost and other chat app.

tommoor:
I'm not sure exactly where the abstraction should be yet, but I'm sure it makes sense to have one for posting to chat providers. I think it makes sense to build out a Mattermost plugin first and then see where logic can be shared

almereyda:
That workflow has already proven suitable when support for generic OIDC providers had been shaped.

almereyda:
<details><summary>n8n workflow for converting Outline webhooks to Mattermost webhooks, incl. HMAC signature verification</summary>
<p>

This is an n8n node compiled from

* [How to "stringify" the raw binary data received by a webhook trigger? - Questions - n8n](https://community.n8n.io/t/how-to-stringify-the-raw-binary-data-received-by-a-webhook-trigger/18246/2)
* [Incoming webhooks](https://developers.mattermost.com/integrate/webhooks/incoming/)

```json
{
  "name": "Outline changes to Mattermost",
  "nodes": [
    {
      "parameters": {
        "conditions": {
          "string": [
            {
              "value1": "={{$node[\"Webhook\"].json[\"headers\"][\"outline-signature\"]}}",
              "value2": "={{`t=${$json[\"ts\"]},s=${$json[\"hmac\"]}`}}"
            }
          ]
        }
      },
      "id": "fdaf8d9c-48d3-4a28-9d00-a4db1d4be407",
      "name": "IF",
      "type": "n8n-nodes-base.if",
      "typeVersion": 1,
      "position": [
        620,
        580
      ]
    },
    {
      "parameters": {},
      "id": "d2d60417-bdc9-488e-9285-bab1f3226dce",
      "name": "Hmac Invalid",
      "type": "n8n-nodes-base.noOp",
      "typeVersion": 1,
      "position": [
        860,
        720
      ]
    },
    {
      "parameters": {},
      "id": "e27db27e-83a3-4784-9ed1-b6983ceadb5a",
[...]

PR #7408, opened 2024-08-18
feat: Link Mattermost account to Outline

hmacr:
Towards #4271 

The implementation is split into three parts.
1. Setup Mattermost connection.
2. Enable posting document updates to Mattermost channels.
3. Enable searching documents from Mattermost channels.

This PR adds the functionality for `1. Setup Mattermost connection`.

### Screen Recording

https://github.com/user-attachments/assets/92c10695-1376-41a6-aade-9f5279ed0e7c

## Test Setup
1. Setup a local Mattermost instance.
2. Enable CORS and restart the instance (ref: [doc](https://docs.mattermost.com/configure/integrations-configuration-settings.html#cors))


plugins/mattermost/server/api/mattermost.ts
```diff
@@ -0,0 +1,64 @@
...
+import validate from "@server/middlewares/validate";
+import { Integration, IntegrationAuthentication } from "@server/models";
+import { presentIntegration, presentPolicies } from "@server/presenters";
+import { APIContext } from "@server/types";
+import * as T from "./schema";
+
+const router = new Router();
+
+router.post(
+  "mattermost.connect",
```
hmacr, line 14: Added a new endpoint, instead of using `integrations.create`, since we need to store the api-key securely in `IntegrationAuthentication` model.

apoorv-mishra:
Slack uses OAuth to connect. We aren't doing the same here? 🤔

hmacr:
Agree on the difference with Slack.
Mattermost doesn't support granular OAuth scopes, so @tommoor and I figured there isn't much difference between an OAuth access and user API key.

plugins/mattermost/server/api/mattermost.ts
tommoor: I think this is a pretty common case actually, how about we update `integrations.create` to allow passing optional authentication token instead of an entire new endpoint. Less maintenance, thoughts?

plugins/mattermost/client/utils/useDictionary.ts
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+
+export default function useDictionary() {
```
tommoor, line 4: No need for a hook here, translations can be used directly in the components

plugins/mattermost/client/components/connection/TeamSelection.tsx
```diff
@@ -0,0 +1,134 @@
...
+      toast.error(err.message);
+    } finally {
+      setSaving(false);
+    }
+  }, [selectedTeam]);
+
+  return (
+    <Flex column gap={12}>
+      <div>
+        <Text as="h3">Account details</Text>
```
tommoor, line 57: Missing translation

plugins/mattermost/client/components/connection/TeamSelection.tsx
```diff
@@ -0,0 +1,134 @@
...
+        </Text>
+      </div>
+      <Flex justify="space-between">
+        <UserDataItem title={t("Username")} value={user.name} />
+        <UserDataItem title={t("Email")} value={user.email} />
+      </Flex>
+      <SelectTeam
+        teams={teams}
+        selectedTeam={selectedTeam}
+        onSelect={(team: TeamData) => setSelectedTeam(team)}
```
tommoor, line 79:
```suggestion
        onSelect={setSelectedTeam}
```

Perhaps

plugins/mattermost/client/components/connection/TeamSelection.tsx
```diff
@@ -0,0 +1,134 @@
...
+      />
+      <Flex justify="flex-end" gap={12} style={{ marginTop: "12px" }}>
+        <Button neutral onClick={() => onBack()}>
```
tommoor, line 82:
```suggestion
        <Button neutral onClick={onBack}>
```

plugins/mattermost/client/utils/mattermost.ts
```diff
@@ -0,0 +1,77 @@
+import { userDetailsUrl as userDataUrl, userTeamsUrl } from "../../shared/url";
+import { UserData, TeamData } from "./types";
+import { Dictionary } from "./useDictionary";
+
+export class MattermostClient {
+  #serverUrl: string;
```
tommoor, line 6:
```suggestion
  private serverUrl: string;
```

Prefer private and public keywords for these

shared/types.ts
```diff
@@ -148,6 +149,19 @@ export type IntegrationSettings<T> = T extends IntegrationType.Embed
       | { serviceTeamId: string }
       | { measurementId: string }
       | { slack: { serviceTeamId: string; serviceUserId: string } }
+      | {
+          url: string;
+          apiKey: string;
```
tommoor, line 154: `apiKey` shouldn't be here in the data

plugins/mattermost/client/utils/mattermost.ts
```diff
@@ -0,0 +1,77 @@
...
+    parseResponse,
+  }: {
+    url: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    parseResponse: (json: any) => T;
+  }) => {
+    let res: Response;
+
+    try {
+      res = await fetch(url, {
```
tommoor, line 59: Alternatively we could create an endpoint for this, wouldn't that avoid the need to enable CORS on the Mattermost server?

plugins/mattermost/server/api/mattermost.ts
hmacr: True - thought this might be a one-off case.

shared/types.ts
hmacr: Ah yes, oversight.

plugins/mattermost/client/utils/mattermost.ts
hmacr: It'd introduce an additional network hop to our backend, but I think it makes it cleaner to have all mattermost interactions from our backend; will make this change.

plugins/mattermost/client/index.tsx
```diff
@@ -11,6 +11,6 @@ PluginManager.add({
     group: "Integrations",
     icon: MattermostIcon,
     component: React.lazy(() => import("./Settings")),
-    enabled: (_, user) => [UserRole.Admin, UserRole.Member].includes(user.role),
+    enabled: (_, user) => [UserRole.Admin].includes(user.role),
```
hmacr, line 14: Modified the permission to match other integrations

server/routes/api/integrations/schema.ts
```diff
@@ -41,9 +42,16 @@ export const IntegrationsCreateSchema = BaseSchema.extend({
...
+    /** Integration access token (or) api key */
+    accessToken: z.string().optional(),
+
+    /** Integration refresh token */
+    refreshToken: z.string().optional(),
+
     /** Integration config/settings */
-    settings: z
-      .object({ url: z.string().url() })
+    settings: MattermostIntegrationSettingsSchema.or(
```
hmacr, line 52: Match mattermost settings first so that zod parsing doesn't short circuit to the base `{ url: string }` object.
Better way would be to introduce a type field in every object and use discrimination unions - could be done as a separate effort, imo.

tommoor: Yes we started doing this with newer objects. A hard lesson learned

app/models/Integration.ts
```diff
@@ -31,6 +31,10 @@ class Integration<T = unknown> extends Model {
 
   @observable
   settings: IntegrationSettings<T>;
+
+  accessToken?: string;
+
+  refreshToken?: string;
```
tommoor, line 37:
```suggestion
```

This change can be reverted, right?

tommoor (approved):
Okay, time to get some decisions on this old stuff!

This looks good apart from that one change on client model doesn't seem needed. I think I'll probably need to add some inline docs on how to get the API key. It seems every user would need to do so? 🤔

app/models/Integration.ts
hmacr: Let me take a look at this PR this week.. I think there are some unfinished things here from the feedback in #7461..

hmacr: Taking a while to get around to this.. I'll circle back here after finishing the current commitments.

tommoor: roger that

PR #7442, opened 2024-08-22
Delay loading of document policies until document menu is opened

### LINKED ISSUE #7441: Unnecessary `documents.info` call upon archiving a document
This unnecessary call is also broadcasted to multiple users over `wss`. Not a big problem now, but becomes dangerous after collection archival, where multiple(possibly, thousands) docs are archived at once.

Blocks #3141

apoorv-mishra:
Closes #7441


app/menus/DocumentMenu.tsx
```diff
@@ -225,8 +228,9 @@ function DocumentMenu({
       <ContextMenu
         {...menu}
         aria-label={t("Document options")}
-        onOpen={handleOpen}
+        onOpen={onOpen}
         onClose={onClose}
+        loading={loading || documents.isFetching}
```
tommoor, line 233: I don't think we want to show a loader here just because subscriptions is fetching, that's going to be quite annoying.

app/components/DocumentBreadcrumb.tsx
```diff
@@ -69,7 +69,7 @@ const DocumentBreadcrumb: React.FC<Props> = ({
     : undefined;
 
   React.useEffect(() => {
-    void document.loadRelations();
+    void document.loadRelations({ withoutPolicies: true });
```
tommoor, line 72: In what situation are we displaying a document breadcrumb but we don't need the document policies? 🤔

tommoor:
I think we'd be better off just delaying the rendering of the entire thing except the `...` button until you hover over/near the button. That way we can delay loading policies until needed _and_ delay all the other rendering here on lists of documents – two birds one stone. thoughts?

app/menus/DocumentMenu.tsx
apoorv-mishra: So, we don't show the loader at all?

tommoor: Right now sometimes the subscribe will switch to unsubscribe after a couple of ms, that seems a lot less jarring than the entire menu being in a loading state

app/components/DocumentBreadcrumb.tsx
apoorv-mishra: I would ask the opposite – in what situation do we need the document policies for displaying a document breadcrumb? 😄

apoorv-mishra:
>delaying the rendering of the entire thing except the ... button

Sorry, what do you mean by "entire thing" here? Document title, breadcrumb, viewership info – all of it? 🤔

tommoor:
No, just the entire content of the content menu – could avoid rendering everything except the button to open it

app/components/DocumentBreadcrumb.tsx
apoorv-mishra: Don't have a solution here yet but looks like this is an existing bug. I mean, even after loading policies, we don't really use those to _decide_ what the breadcrumb should look like.

app/menus/DocumentMenu.tsx
```diff
@@ -217,160 +212,164 @@ function DocumentMenu({
...
       )}
-      <ContextMenu
-        {...menu}
-        aria-label={t("Document options")}
-        onOpen={onOpen}
-        onClose={onClose}
-        loading={loading || documents.isFetching}
-      >
-        <Template
+      {data && !loading && !documents.isFetching ? (
```
tommoor, line 226: `documents.isFetching` this would prevent the menu from showing when _any_ fetch is happening…

apoorv-mishra: Ahh! Good catch, thanks 🙏

app/menus/DocumentMenu.tsx
```diff
@@ -214,159 +212,164 @@ function DocumentMenu({
...
         />
       )}
-      <ContextMenu
-        {...menu}
-        aria-label={t("Document options")}
-        onOpen={handleOpen}
-        onClose={onClose}
-      >
-        <Template
+      {data && !loading ? (
```
apoorv-mishra, line 226: A subtle race condition might occur here – subscriptions are loaded before policies are fetched 😬

tommoor: Subscriptions are not very important, I don't care to wait for them at all

app/menus/DocumentMenu.tsx
```diff
@@ -48,75 +50,120 @@ import {
...
 import useRequest from "~/hooks/useRequest";
 import useStores from "~/hooks/useStores";
 import { MenuItem } from "~/types";
 import { documentEditPath } from "~/utils/routeHelpers";
+import { MenuContext, useMenuContext } from "./MenuContext";
 
 type Props = {
+  /** Document for which the menu is to be shown */
   document: Document;
-  className?: string;
```
apoorv-mishra, line 61: I don't think we need this yet? Not used at any of the call sites of `DocumentMenu`.

app/menus/DocumentMenu.tsx
```diff
@@ -198,6 +417,10 @@ function DocumentMenu({
     [history, collection, documents, document.id]
   );
 
+  const handleMenuTrigger = React.useCallback(() => {
+    showMenu();
+  }, [showMenu]);
```
tommoor, line 422: Shouldn't be any need for this? Just use showMenu directly

app/menus/DocumentMenu.tsx
```diff
@@ -48,75 +50,120 @@ import {
...
+  }, [data, error, loading, request, document]);
 
-    if (onOpen) {
-      onOpen();
-    }
-  }, [data, loading, onOpen, request]);
+  return label ? (
+    <MenuButton
+      {...menuState}
+      onMouseEnter={handleMouseEnter}
```
tommoor, line 114: Lets use `onPointerEnter` as it's the modern equivalent

app/menus/DocumentMenu.tsx
```diff
@@ -48,75 +50,120 @@ import {
...
+      onClick={onTrigger}
+    >
+      {label}
+    </MenuButton>
+  ) : (
+    <OverflowMenuButton
+      aria-label={t("Show document menu")}
+      onMouseEnter={handleMouseEnter}
```
tommoor, line 122: Lets use `onPointerEnter` as it's the modern equivalent

app/menus/DocumentMenu.tsx
```diff
@@ -48,75 +50,120 @@ import {
...
+
+const MenuContent: React.FC<MenuContentProps> = ({
+  onOpen,
+  onClose,
+  onFindAndReplace,
+  onRename,
+  showDisplayOptions,
+  showToggleEmbeds,
+}) => {
+  const user = useCurrentUser();
```
tommoor, line 146: Can we remove the extra spaces between these

tommoor (approved):
Looks good after addressing quick comments 👍

Issue #6909, opened 2024-05-14
Logout not working with ODIC and Authentik (or Keycloak)


ksaadDE:
In this case ODIC + Authentik or  Keycloak is used. Which are both working according to the same principles.

I would expect after clicking on "logout" that I am not logged in anymore after page refresh. Sadly, Outline is unable to just delete the session. 

The discussion related to it can be found here https://github.com/outline/outline/discussions/3672

I consider it a bug, as there is a logout function that is not (properly) working (= not logging the user out and terminating the session). 

For security such functionalities are essential, yet alone for compliance.

tommoor:
Did you set the `OIDC_LOGOUT_URI` correctly? This has been available for a while

ksaadDE:
> Did you set the `OIDC_LOGOUT_URI` correctly? This has been available for a while

Hey, thank you very much! I checked the config and found out I haven't set the env var (docker). My bad!

Though I was not sure either whether that env var works at all, because the issue https://github.com/outline/outline/discussions/3672 does not give clear indications whether it got introduced or not. Maybe you could clarify that in the issue.

Furthermore there are issues with Authentik and Logout routes, but I found a way to make it happening. If anyone is interested, kindly refer to this discussion, where I documented that https://github.com/goauthentik/authentik/discussions/9737

Thank you very much! Great work!

blib:
I stumblled on this as well.

Here: 

```
const Logout = () => {
  const { auth } = useStores();

  void auth.logout().then(() => {
    if (env.OIDC_LOGOUT_URI) {
      window.location.replace(env.OIDC_LOGOUT_URI);
    }
  });

  return <Redirect to="/" />;
};
```

When you log out, `auth.logout` destroys the local session, and then the user is redirected to `/`. If the OIDC provider is the only configured authentication method, the user is immediately redirected to the OIDC provider’s authentication page. This triggers another `window.location.replace` call, which overwrites the previous destination. As a result, the user is not logged out of the OIDC provider and is instantly logged back in. You can observe this in the network requests, where the first `window.location.replace` is marked as canceled.

A potential workaround for this issue is to configure a second authentication provider. This way, the process will stop at the login page, giving the browser enough time to complete the initial `window.location.replace`.

PR #7446, opened 2024-08-22
logout for OIDC #6909

blib:
https://github.com/outline/outline/issues/6909

When you log out, auth.logout destroys the local session, and then the user is redirected to /. If the OIDC provider is the only configured authentication method, the user is immediately redirected to the OIDC provider’s authentication page. This triggers another window.location.replace call, which overwrites the previous destination. As a result, the user is not logged out of the OIDC provider and is instantly logged back in. You can observe this in the network requests, where the first window.location.replace is marked as canceled.

The fix simply prevents the redirect to the home page, allowing the browser enough time to complete the request.

### DIFF
--- a/app/actions/definitions/navigation.tsx
+++ b/app/actions/definitions/navigation.tsx
@@ -216,7 +216,9 @@ export const logout = createAction({
   perform: async () => {
     await stores.auth.logout();
     if (env.OIDC_LOGOUT_URI) {
-      window.location.replace(env.OIDC_LOGOUT_URI);
+      setTimeout(() => {
+        window.location.replace(env.OIDC_LOGOUT_URI);
+      });
     }
   },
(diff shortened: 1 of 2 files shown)


tommoor:
Hey, thanks for the PR – what is the timeout achieving?

tommoor:
Can requestAnimationFrame work? I would prefer to delay until specifically the next frame than some arbitrary point of time in the future if it will do the trick

blib:
> Can requestAnimationFrame work? I would prefer to delay until specifically the next frame than some arbitrary point of time in the future if it will do the trick

I tried that with requestAnimationFrame but it doesn't work. I don't like arbitrary timout either but I could not came up with anything better.

tommoor (approved):

PR #7456, opened 2024-08-23
fix: Set redis expiry in seconds

hmacr:
Came across this bug when working on Mattermost integration.
We've been using milliseconds for expiry when redis expects it in seconds.

redis: [doc](https://redis.io/docs/latest/commands/expire/)
ioredis: [doc](https://github.com/redis/ioredis?tab=readme-ov-file#expiration)

### DIFF
--- a/server/utils/CacheHelper.ts
+++ b/server/utils/CacheHelper.ts
@@ -1,4 +1,4 @@
-import { Day } from "@shared/utils/time";
+import { Day, Second } from "@shared/utils/time";
 import Logger from "@server/logging/Logger";
 import Redis from "@server/storage/redis";
 
@@ -6,8 +6,8 @@ import Redis from "@server/storage/redis";
  * A Helper class for server-side cache management
  */
 export class CacheHelper {
-  // Default expiry time for cache data in ms
-  private static defaultDataExpiry = Day;
+  // Default expiry time for cache data in seconds
+  private static defaultDataExpiry = Day / Second;
 
   /**
    * Given a key, gets the data from cache store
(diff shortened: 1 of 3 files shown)


hmacr:
Follow-up task would be to run a script to change the existing expiry value based on cache data type.
For ex, Iframely unfurls are supposed to be cached for a day but are cached for 2 years right now.

tommoor (commented):
Urgh, we also need to change the couple of places that call `setData` with cache expiry from the iframely integration 

cc @apoorv-mishra

server/utils/CacheHelper.ts
```diff
@@ -7,7 +7,7 @@ import Redis from "@server/storage/redis";
  */
 export class CacheHelper {
   // Default expiry time for cache data in seconds
-  private static defaultDataExpiry = Day / Second;
+  private static defaultDataExpiry = Day / 1000;
```
hmacr, line 10: Changed this since reading `Day / Second` conveys that the value is being converted to its minute equivalent.

apoorv-mishra:
Damn! A bit embarrassing I'll admit 🥲 

Also, relieved it didn't turn out the other way around 😥

server/utils/CacheHelper.ts
apoorv-mishra: Maybe we can extend these constants into classes with getters such as `milliseconds`, `seconds`, `hours`, so we have for example, `Day.milliseconds`, `Day.seconds`, `Day.hours`, etc. Might prevent such a  mistake going further. Thoughts?

tommoor (approved):

server/utils/CacheHelper.ts
tommoor: I'm down for this change

PR #7458, opened 2024-08-24
Specify time conversion unit

### LINKED ISSUE #7457: Specificity in time units
>Maybe we can extend these constants into classes with getters such as milliseconds, seconds, hours, so we have for example, Day.milliseconds, Day.seconds, Day.hours, etc. Might prevent such a mistake going further. Thoughts?

Ref: https://github.com/outline/outline/pull/7456#discussion_r1729944951

apoorv-mishra:
Closes #7457


tommoor:
Maybe just .ms is ok?

tommoor (approved):

PR #7459, opened 2024-08-25
chore: Script to invalidate unfurls cache

hmacr:
Follow-up task for #7456

### DIFF
--- a/server/scripts/20240825000000-invalidate-unfurls-cache.ts
+++ b/server/scripts/20240825000000-invalidate-unfurls-cache.ts
@@ -1,0 +1,75 @@
+import RedisAdapter from "@server/storage/redis";
+import "./bootstrap";
+
+const redis = RedisAdapter.defaultClient;
+
+const Keys = {
+  Pattern: "unfurl:*",
+  Total: 0,
+  Error: 0,
+};
+
+const Page = {
+  Count: 0,
+  Limit: 1000,
+};
+
+const deleteKeys = async (keys: string[]): Promise<void> => {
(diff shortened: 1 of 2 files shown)


server/scripts/20240825000000-invalidate-unfurls-cache.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  stream.on("data", (unfurlKeys) => {
+    if (!unfurlKeys.length) {
+      return;
+    }
+
+    Page.Count++;
+    console.log(`Starting page ${Page.Count}...`);
+
+    stream.pause();
```
hmacr, line 51: `ioredis` suggests us to pause and resume the stream when doing an async task inside the data handler.

server/scripts/20240825000000-invalidate-unfurls-cache.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+};
+
+const execute = async () => {
+  const stream = redis.scanStream({ match: Keys.Pattern, count: Page.Limit });
+
+  console.log("Invalidating unfurls cache...");
+
+  stream.on("data", (unfurlKeys) => {
+    if (!unfurlKeys.length) {
```
hmacr, line 44: The way redis stream works is,
Get the keys for the current page -> apply match filter -> return data to client.

It's sometimes possible that the returned data is empty but the stream iteration has not reached the end.

tommoor (approved):
🙏 works great

PR #7460, opened 2024-08-26
perf: Policies refactor

tommoor:
- Encodes membership ID's in policies, this allows improved performance by avoiding refetching policies with improved client logic in many scenarios.
- Refactor of websocket events

Follow up to https://github.com/outline/outline/pull/7448
Towards #6639
Towards #7275

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -28,6 +28,7 @@ import {
   WebsocketCollectionGroupEvent,
   WebsocketCollectionUpdateIndexEvent,
   WebsocketCollectionUserEvent,
+  WebsocketDocumentUserEvent,
   WebsocketEntitiesEvent,
   WebsocketEntityDeletedEvent,
 } from "~/types";
@@ -261,39 +262,21 @@ class WebsocketProvider extends React.Component<Props> {
       }
     );
 
-    // received when a user is given access to a document
-    this.socket.on(
-      "documents.add_user",
-      (event: PartialWithId<UserMembership>) => {
-        userMemberships.add(event);
-      }
-    );
+    this.socket.on("documents.add_user", async (event: UserMembership) => {
+      userMemberships.add(event);
+      await documents.fetch(event.documentId!, {
+        force: true,
+      });
+    });
 
     this.socket.on(
       "documents.remove_user",
-      (event: PartialWithId<UserMembership>) => {
-        if (event.userId) {
-          const userMembership = userMemberships.get(event.id);
+      (event: WebsocketDocumentUserEvent) => {
+        policies.removeForMembership(event.id);
+        userMemberships.remove(event.id);
 
-          // TODO: Possibly replace this with a one-to-many relation decorator.
-          if (userMembership) {
-            userMemberships
-              .filter({
-                userId: event.userId,
-                sourceId: userMembership.id,
-              })
-              .forEach((m) => {
-                m.documentId && documents.remove(m.documentId);
-              });
-          }
-
-          userMemberships.removeAll({
-            userId: event.userId,
-            documentId: event.documentId,
-          });
-        }
-
-        if (event.documentId && event.userId === auth.user?.id) {
+        const policy = policies.get(event.documentId);
+        if (policy && policy.abilities.read === false) {
           documents.remove(event.documentId);
         }
       }
@@ -417,12 +400,14 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on("collections.add_user", async (event: Membership) => {
       memberships.add(event);
-      await collections.fetch(event.collectionId);
+      await collections.fetch(event.collectionId, {
+        force: true,
+      });
     });
 
     this.socket.on(
       "collections.remove_user",
-      async (event: WebsocketCollectionUserEvent) => {
+      (event: WebsocketCollectionUserEvent) => {
         policies.removeForMembership(event.id);
         memberships.remove(event.id);
 
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -405,7 +405,7 @@ class WebsocketProvider extends React.Component<Props> {
       memberships.remove(event.id);
 
       const policy = policies.get(event.collectionId);
-      if (policy && policy.abilities.readDocument === false) {
+      if (policy && policy.abilities.read === false) {
         collections.remove(event.collectionId);
       }
     });
@@ -421,7 +421,7 @@ class WebsocketProvider extends React.Component<Props> {
         groupMemberships.remove(event.id);
 
         const policy = policies.get(event.collectionId!);
-        if (policy && policy.abilities.readDocument === false) {
+        if (policy && policy.abilities.read === false) {
           collections.remove(event.collectionId!);
         }
       }
@@ -431,10 +431,7 @@ class WebsocketProvider extends React.Component<Props> {
       "collections.update_index",
       action((event: WebsocketCollectionUpdateIndexEvent) => {
         const collection = collections.get(event.collectionId);
-
-        if (collection) {
-          collection.updateIndex(event.index);
-        }
+        collection?.updateIndex(event.index);
       })
     );
 
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -101,6 +101,8 @@ class WebsocketProvider extends React.Component<Props> {
       notifications,
     } = this.props;
 
+    const currentUserId = auth?.user?.id;
+
     // on reconnection, reset the transports option, as the Websocket
     // connection may have failed (caused by proxy, firewall, browser, ...)
     this.socket.io.on("reconnect_attempt", () => {
@@ -264,7 +266,7 @@ class WebsocketProvider extends React.Component<Props> {
     this.socket.on("documents.add_user", async (event: UserMembership) => {
       userMemberships.add(event);
       await documents.fetch(event.documentId!, {
-        force: true,
+        force: event.userId === currentUserId,
       });
     });
 
@@ -404,7 +406,7 @@ class WebsocketProvider extends React.Component<Props> {
     this.socket.on("collections.add_user", async (event: Membership) => {
       memberships.add(event);
       await collections.fetch(event.collectionId, {
-        force: true,
+        force: event.userId === currentUserId,
       });
     });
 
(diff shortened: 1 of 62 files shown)


server/policies/index.ts
```diff
@@ -25,31 +21,3 @@ import "./group";
...
 import "./userMembership";
-
-type Policy = Record<string, boolean>;
-
-/*
- * Given a user and a model – output an object which describes the actions the
- * user may take against the model. This serialized policy is used for testing
- * and sent in API responses to allow clients to adjust which UI is displayed.
- */
-export function serialize(model: User, target: Model | null): Policy {
```
tommoor, line 36: Moved to cancan class

server/policies/utils.ts
```diff
@@ -3,12 +3,16 @@ import { IncorrectEditionError } from "@server/errors";
 import { User, Team } from "@server/models";
 import Model from "@server/models/base/Model";
 
-export function and(...args: boolean[]) {
-  return args.every(Boolean);
+type Args = boolean | string | Args[];
```
tommoor, line 6: All of these changes are to allow policies to return `membershipIds` instead of just booleans

server/policies/cancan.ts
```diff
@@ -74,21 +78,58 @@ export class CanCan {
...
+      action,
+      target
     );
 
     // Check conditions only for matching abilities
-    return matchingAbilities.some(
-      (ability) =>
-        !ability.condition ||
-        ability.condition(performer, target, options || {})
+    const conditions = uniq(
```
tommoor, line 88: All of these changes are to allow policies to return `membershipIds` instead of just booleans

server/services/websockets.ts
```diff
@@ -161,14 +161,16 @@ async function authenticated(io: IO.Server, socket: SocketWithAuth) {
...
+  const [collectionIds, groupIds] = await Promise.all([
+    user.collectionIds(),
+    user.groupIds(),
+  ]);
 
-  collectionIds.forEach((collectionId) =>
-    rooms.push(`collection-${collectionId}`)
-  );
+  collectionIds.forEach((colId) => rooms.push(`collection-${colId}`));
+  groupIds.forEach((groupId) => rooms.push(`group-${groupId}`));
```
tommoor, line 173: Join a websocket channel for each group we're a member of

app/components/WebsocketProvider.tsx
```diff
@@ -413,56 +404,40 @@ class WebsocketProvider extends React.Component<Props> {
       }
     );
 
-    // received when a user is given access to a collection
-    // if the user is us then we go ahead and load the collection from API.
+    this.socket.on("collections.add_user", async (event: Membership) => {
+      memberships.add(event);
+      await collections.fetch(event.collectionId, {
```
apoorv-mishra, line 409: Earlier, collection was fetched _only_ for the user who was added, not it'll be fetched for all users in the collection room?

tommoor: Ah yep, thanks – I made this not a forced fetch for a bit in which case it didn't matter, but then moved back to a force fetfh

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -278,57 +282,63 @@ export default class WebsocketsProcessor {
...
+          .to(`group-${event.modelId}`)
+          .emit("collections.remove_group", membership);
 
         for (const groupUser of groupUsers) {
-          if (membershipUserIds.includes(groupUser.userId)) {
-            // the user still has access through some means...
-            // treat this like an add, so that the client re-syncs policies
-            socketio
-              .to(`user-${groupUser.userId}`)
-              .emit("collections.add_user", {
```
tommoor, line 317: Cleaned up this logic a lot, we won't emit `collections.add_user` events when a group is removed from the collection anymore.

app/components/WebsocketProvider.tsx
apoorv-mishra: Also, since `collections.add_user` endpoint is used not just for adding a user, but also for _updating_ an existing user's permission, then, we aren't updating the document policies for that case as we're doing earlier? I don't also see any hooks particularly for achieving that 👀

tommoor: It's true, thinking of using a lifecycle hook on membership update – thoughts?

apoorv-mishra: If we can achieve a _full cascade_, why not? I mean, since for this particular instance, a membership update should affect two other corresponding model instances:
1. Respective collection instance(w.r.t policies)
2. All documents within that collection(again w.r.t policies)

If we can achieve cascade from `membership` update -> `collection` policy update -> `documents'` policy update, we should be good?

Hope I was clear here 🤞

tommoor: image

tommoor: Ah yea, I have this on Membership but alternatively a similar hook could be on policy, perhaps that is better as it's more _direct_.

tommoor: Added here:
https://github.com/outline/outline/pull/7460/files#diff-66b70c39db74930a46b4a70e8c8c543bcc7881808c5b8ee1c1c7f8107ff9f587R33

tommoor: The problem is there is no relationship between policies, you'd need to go via the parent model. Definitely doable, but getting quite complex~ possible followup

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -224,49 +225,52 @@ export default class WebsocketsProcessor {
...
+          collectionId: event.collectionId,
+          id: event.modelId,
+        };
+
+        // let everyone with access to the collection know a user was removed
+        socketio
+          .to(`collection-${event.collectionId}`)
+          .emit("collections.remove_user", membership);
+
+        if (cannot(user, "read", collection)) {
```
apoorv-mishra, line 273: If user's an admin, they won't leave the channel.

app/stores/DocumentsStore.ts
```diff
@@ -680,17 +680,6 @@ export default class DocumentsStore extends Store<Document> {
     return this.add(res.data);
   };
 
-  @action
-  removeCollectionDocuments(collectionId: string) {
```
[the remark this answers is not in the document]
apoorv-mishra: I guess this will _remove_ drafts instead of _detaching_ them?

app/components/WebsocketProvider.tsx
```diff
@@ -413,56 +403,35 @@ class WebsocketProvider extends React.Component<Props> {
...
-          // Document policies might need updating as the permission changes
-          documents.inCollection(event.collectionId).forEach((document) => {
-            policies.remove(document.id);
-          });
-        }
+    this.socket.on("collections.remove_user", (event: Membership) => {
+      memberships.remove(event.id);
+
+      const policy = policies.get(event.collectionId);
+      if (policy && policy.abilities.readDocument === false) {
```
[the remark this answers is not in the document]
tommoor: I suppose it's not neccessary

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -466,153 +480,172 @@ export default class WebsocketsProcessor {
...
-              });
+          async (groupUsers) => {
+            for (const groupMembership of groupMemberships) {
+              if (!groupMembership.collectionId) {
+                continue;
+              }
+              const payload = presentGroupMembership(groupMembership);
+
+              for (const groupUser of groupUsers) {
+                socketio
```
tommoor, line 630: Unfortunately there is not enough information on the client to cleanup from a `groups.delete` event alone, so we send `collections.remove_group` for each individual collection (and soon document) membership instead.

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -466,153 +480,172 @@ export default class WebsocketsProcessor {
...
+        );
+
         return;
       }
 
       case "groups.delete": {
         socketio.to(`team-${event.teamId}`).emit(event.name, {
           modelId: event.modelId,
         });
+        socketio.to(`group-${event.modelId}`).emit("leave", {
```
tommoor, line 598: @apoorv-mishra I'm confident we can remove all these `leave` events in the near future with an abstraction on the client that maps a model removal to a channel disconnect

app/models/Collection.ts
```diff
@@ -207,14 +206,18 @@ export default class Collection extends ParanoidModel {
...
+  ) {
     if (!this.documents) {
       return;
     }
 
     const travelNodes = (nodes: NavigationNode[]) =>
       nodes.forEach((node) => {
         if (node.id === document.id) {
+          node.color = document.color ?? undefined;
+          node.icon = document.icon ?? undefined;
```
tommoor, line 220: Another small bug – icon and color not updated in collection structure here

tommoor:
I know this is basically unreviewable but it ended up being a lot of improvements, fixes, and perf improvements rolled into one. Will likely merge tomorrow to enable continued progress on #7275

server/queues/processors/WebsocketsProcessor.ts
apoorv-mishra: So I thought. Similar analogy holds for `join` events as well. I can take it up separately.

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -278,63 +282,73 @@ export default class WebsocketsProcessor {
...
-            });
+        // let everyone with access to the collection know a group was removed
+        socketio
+          .to(`collection-${event.collectionId}`)
+          .emit("collections.remove_group", membership);
+
+        // let everyone in the group know they were removed
+        socketio
+          .to(`group-${event.modelId}`)
+          .emit("collections.remove_group", membership);
```
apoorv-mishra, line 323: Everyone in `group-${event.modelId}` would still be members of the `collection-${event.collectionId}` room at this point. So this is perhaps redundant?

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -466,153 +480,172 @@ export default class WebsocketsProcessor {
...
-
-          if (!collection) {
-            continue;
-          }
+        await GroupMembership.findAllInBatches<GroupMembership>(
+          {
+            where: {
+              groupId: event.modelId,
+            },
+            limit: 100,
```
apoorv-mishra, line 507: Unrelated but I think `batchSize` here makes more sense than `limit`. Could be taken up separately if we intend to.

tommoor: Perhaps, or `batchLimit`

server/queues/processors/WebsocketsProcessor.ts
tommoor: Easy enough when the entire channel is disappearing, such as `groups.delete` but for kicking individual users out I still don't have a good method that doesn't involve a client roundtrip.

PR #7477, opened 2024-08-28
feat: Add email headers to enhance threading

### LINKED ISSUE #7444: Add email headers to enhance threading of notifications
If you have an example document like https://acme.com/doc/example-docs-hX4nZCk0ac then you can easily thread them like this:

```
In-Reply-To: <edit-hX4nZCk0ac@acme.com>
References: <edit-hX4nZCk0ac@acme.com>
Thread-Index: <edit-hX4nZCk0ac@acme.com>
```

I'm not sure if you'd want comment in a doc threaded with edits to a doc, they could be 2 sets of threads.

This way even if the document name changes they'll all get grouped together. The first two headers are the rfc headers, the second two add support for Outlook / Office365. Note that hX4nZCk0ac@acme.com is supposed to reference another message id but it doesn't need to reference a real one, all that is needed is for the related emails to reference the same id to group them. 

https://www.rfc-editor.org/rfc/rfc2822

_Originally posted by @brendanheywood in https://github.com/outline/outline/discussions/7424#discussioncomment-10415468_

hmacr:
Closes #7444


server/emails/mailer.tsx
```diff
@@ -180,6 +188,8 @@ export class Mailer {
           `Preview Url: ${nodemailer.getTestMessageUrl(info)}`
         );
       }
+
+      return info?.messageId;
```
tommoor, line 192: Ah, I didn't realize the `Message-ID` is something that comes back from the email server originally, I thought it was something we could provide.

server/emails/templates/BaseEmail.tsx
```diff
@@ -101,11 +104,43 @@ export default abstract class BaseEmail<
       return;
     }
 
+    let references: string[] | undefined;
+
+    if (notification) {
+      const prevNotifications = await Notification.unscoped().findAll({
```
tommoor, line 110: Could we move all this logic to a new static method on `Notification`, something like `Notification.getReferencesForNotification(notification)`, I think this will keep things tidy and create a more testable interface

server/emails/templates/BaseEmail.tsx
```diff
@@ -130,6 +165,7 @@ export default abstract class BaseEmail<
     if (notification) {
       try {
         notification.emailedAt = new Date();
+        notification.messageId = messageId ?? null;
```
tommoor, line 168: Return `messageId ?? null` from `sendMail` perhaps?

tommoor (commented):
Nice, clean and understandable

server/emails/templates/BaseEmail.tsx
hmacr: Yes, much cleaner.

server/emails/templates/BaseEmail.tsx
hmacr: Idea was to ensure the interfaces don't return `null` in the codebase. Thoughts?

server/emails/mailer.tsx
hmacr: We can provide it as well (just need to ensure it is unique) - thought we could let the the library handle it.
Perhaps `<notification-id@outline.com>`?

server/emails/templates/BaseEmail.tsx
```diff
@@ -101,11 +104,43 @@ export default abstract class BaseEmail<
       return;
     }
 
+    let references: string[] | undefined;
+
+    if (notification) {
+      const prevNotifications = await Notification.unscoped().findAll({
+        attributes: ["messageId"],
+        where: {
+          documentId: notification.documentId,
```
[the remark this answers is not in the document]
tommoor: I would suggest potentially an email thread for comments/mentions and a separate thread for document created/updated, we can include that logic inside the new method. If the event does not fall into those categories it can return an empty array of references.

server/migrations/20240828081032-add-notifications-index.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addIndex("notifications", ["documentId", "userId"]);
```
hmacr, line 6: Adding too many columns to this index could be detrimental to performance. 
We already have an index on `event` - postgres query planner should be smart enough to use multiple indexes when necessary.

server/emails/templates/CommentMentionedEmail.tsx
```diff
@@ -92,8 +92,8 @@ export default class CommentMentionedEmail extends BaseEmail<
     );
   }
 
-  protected subject({ actorName, document }: Props) {
-    return `${actorName} mentioned you in “${document.title}”`;
+  protected subject({ document }: Props) {
+    return `Mentioned you in “${document.title}”`;
```
hmacr, line 96: This matches the `DocumentMentionedEmail` subject now - helps with threading in Gmail.
Lmk if this needs to be reverted.

server/models/Notification.ts
```diff
@@ -232,41 +232,58 @@ class Notification extends Model<
...
+        limit,
+      });
+
+      if (!loadedNotifications.length) {
+        hasMore = false;
+      }
+
+      prevNotifications.push(...loadedNotifications);
+      offset += limit;
+    }
```
apoorv-mishra: We have `findAllInBatches` method if that's what you're looking for here.

server/models/Notification.ts
```diff
@@ -218,6 +225,63 @@ class Notification extends Model<
...
+
+    while (hasMore) {
+      const loadedNotifications = await this.unscoped().findAll({
+        attributes: ["id"],
+        where: {
+          id: {
+            [Op.ne]: notification.id,
+          },
+          event: notification.event,
+          documentId: notification.documentId,
```
tommoor, line 255: This isn't quite right as you'll start a separate thread for each notification type, we talked about having two separate threads – one for document edits and one for the conversation comment/mentions?

hmacr: This maintains separate threads for document edits, comments, and mentions.
Mention emails are in a separate thread since their `subject` is different from Comment emails. If we were to combine them, most email clients would dump the history in a single message instead of building the thread.

Essentially, Gmail and Outlook expect the `subject` to be consistent to present the thread tree in the UI (along with the necessary 'References' and `In-Reply-To` headers). So, I wouldn't suggest combining comment & mentions. Thoughts?

That said, this code has a bug - it doesn't combine the document mentions with the comment mentions (and creates 2 separate threads).

server/models/Notification.ts
hmacr: Ah great!! I wasn't aware of this, thanks 😄

server/models/Notification.ts
tommoor: > So, I wouldn't suggest combining comment & mentions. Thoughts?

Okay yes, it will be an improvement over today

server/utils/emails.ts
```diff
@@ -17,4 +17,27 @@ export const isEmailThreadSupportedNotification = (
...
+        NotificationEventType.MentionedInComment,
+      ];
+    case NotificationEventType.CreateComment:
+      return [NotificationEventType.CreateComment];
+    default:
+      return;
+  }
+};
+
+export const getEmailMessageId = (text: string) => `<${text}@${Domain}>`;
```
tommoor, line 43: Can we get the domain from `env.URL` – I don't want to use `outline.com` as we don't own it.

There is a `getBaseDomain` in `utils/domains`

server/models/Notification.ts
```diff
@@ -218,6 +226,65 @@ class Notification extends Model<
...
+          },
+          event: {
+            [Op.in]: events,
+          },
+          documentId: notification.documentId,
+          userId: notification.userId,
+        },
+        order: [["createdAt", "ASC"]],
+        offset: 0,
+        limit: 100,
```
tommoor, line 266: This changed to `batchLimit` on `main` here https://github.com/outline/outline/pull/7476

hmacr: Created #7511 for a bug related to this.
I'll push the review changes after #7511 is merged - Tests would fail otherwise.

tommoor (approved):

PR #7488, opened 2024-08-30
Show nested docs on Archive page

### LINKED ISSUE #7404: Children of archived document don't appear on the Archive page at once
Child doc appearing only after reload 👇 

https://github.com/user-attachments/assets/3ab9c8bf-dc61-4805-89a6-9a0870e3868c

apoorv-mishra:
Closes #7404

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -229,8 +229,7 @@ class WebsocketProvider extends React.Component<Props> {
     this.socket.on(
       "documents.archive",
       action((event: PartialWithId<Document>) => {
-        documents.add(event);
-        policies.remove(event.id);
+        documents.addToArchive(event);
 
         if (event.collectionId) {
           const collection = collections.get(event.collectionId);
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -25,7 +25,7 @@ import User from "~/models/User";
 import UserMembership from "~/models/UserMembership";
 import withStores from "~/components/withStores";
 import {
-  PartialWithId,
+  PartialExcept,
   WebsocketCollectionUpdateIndexEvent,
   WebsocketEntitiesEvent,
   WebsocketEntityDeletedEvent,
@@ -214,16 +214,14 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on(
       "documents.update",
-      action(
-        (event: PartialWithId<Document> & { title: string; url: string }) => {
-          documents.add(event);
+      action((event: PartialExcept<Document, "id" | "title" | "url">) => {
+        documents.add(event);
 
-          if (event.collectionId) {
-            const collection = collections.get(event.collectionId);
-            collection?.updateDocument(event);
-          }
+        if (event.collectionId) {
+          const collection = collections.get(event.collectionId);
+          collection?.updateDocument(event);
         }
-      )
+      })
     );
 
     this.socket.on(
@@ -240,7 +238,7 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on(
       "documents.delete",
-      action((event: PartialWithId<Document>) => {
+      action((event: PartialExcept<Document, "id">) => {
         documents.add(event);
         policies.remove(event.id);
 
@@ -264,7 +262,7 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on(
       "documents.add_user",
-      async (event: PartialWithId<UserMembership>) => {
+      async (event: PartialExcept<UserMembership, "id">) => {
         userMemberships.add(event);
 
         // Any existing child policies are now invalid
@@ -285,7 +283,7 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on(
       "documents.remove_user",
-      (event: PartialWithId<UserMembership>) => {
+      (event: PartialExcept<UserMembership, "id">) => {
         userMemberships.remove(event.id);
 
         // Any existing child policies are now invalid
@@ -307,7 +305,7 @@ class WebsocketProvider extends React.Component<Props> {
 
     this.socket.on(
       "documents.add_group",
-      (event: PartialWithId<GroupMembership>) => {
+      (event: PartialExcept<GroupMembership, "id">) => {
         groupMemberships.add(event);
 
         const group = groups.get(event.groupId!);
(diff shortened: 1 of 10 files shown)


app/models/base/Model.ts
```diff
@@ -55,7 +55,7 @@ export default abstract class Model {
       const store = this.store.rootStore.getStoreForModelName(
         properties.relationClassResolver().modelName
       );
-      if ("fetch" in store) {
+      if ("fetch" in store && this[properties.idKey]) {
```
apoorv-mishra, line 58: This prevents an existing issue where `*.info` calls were being rejected as `400` because the `id` was being passed as `undefined | null`.

apoorv-mishra: Looks like this has recently been fixed already.

app/stores/base/Store.ts
```diff
@@ -144,6 +149,42 @@ export default abstract class Store<T extends Model> {
...
+        });
+      }
+    });
+
+    // Remove associated policies automatically, not defined through Relation decorator.
+    if (this.modelName !== "Policy") {
+      this.rootStore.policies.remove(item.id);
+    }
+
+    item.archivedAt = new Date().toISOString();
```
tommoor, line 184: If item has `archivedAt` why are we resetting it here?

app/stores/base/Store.ts
```diff
@@ -144,6 +149,42 @@ export default abstract class Store<T extends Model> {
     LifecycleManager.executeHooks(model.constructor, "afterRemove", model);
   }
 
+  @action
+  addToArchive(item: PartialWithArchivedAt<T>): void {
```
tommoor, line 153:
```suggestion
  archive(item: PartialWithArchivedAt<T>): void {
```

I think just "archive" is good.

tommoor, line 153: I think there should be a check if the model has `archivedAt` property at all, and if not throw an error as a sanity check.

apoorv-mishra: Yeah, `archive` was already used, for example, in `DocumentsStore`, so I made it `addToArchive` for now.

tommoor: That's fine too

app/stores/base/Store.ts
apoorv-mishra: Not at this point. The items would have been archived on server but that info has not _yet_ propagated to all concerned clients yet.

tommoor: The type suggests it exists, I suppose this is just to make TS happy? Perhaps also solved by ArchivableModel then

app/stores/base/Store.ts
apoorv-mishra: Yeah, but the separation is also to avoid the `documents.archive` call.

app/stores/base/Store.ts
apoorv-mishra: Yeah, the type there is just a "hack" now to make it work. Had `ArchivableModel` in mind actually.

app/stores/base/Store.ts
tommoor: I have felt this tension that it's unclear when a store method is working just on local data vs mutating the server too. Need to come up with a consistent pattern here 🤔

app/stores/base/Store.ts
apoorv-mishra: Also, the type suggests that it has the `archivedAt` attribute, _not_ that it might be `null | undefined`.

app/stores/base/Store.ts
apoorv-mishra: Agreed! I'll give some thoughts...

app/stores/base/Store.ts
```diff
@@ -131,7 +128,11 @@ export default abstract class Store<T extends Model> {
           }
 
           if (deleteBehavior === "cascade") {
-            store.remove(item.id);
+            if (item instanceof ParanoidModel) {
+              item.deletedAt = new Date().toISOString();
+            } else {
+              store.remove(item.id);
+            }
```
apoorv-mishra, line 135: Regressed in https://github.com/outline/outline/pull/7460/files#diff-09465a4af35119447c885e5f289deca73e86112bb6574950d306f872ba4bb161L124-L128

app/stores/base/Store.ts
```diff
@@ -182,7 +183,7 @@ export default abstract class Store<T extends Model> {
     }
 
     item.archivedAt = new Date().toISOString();
-    this.add(item);
+    (this as unknown as Store<ArchivableModel>).add(item);
```
apoorv-mishra, line 186: This is kind of annoying but I don't know of a better way yet. I tried `item.updateData(...)` to circumvent this but that introduces an issue where the archived doc is not conveyed via `wss` to users(other than actor) if that's not in their stores.

server/models/base/ArchivableModel.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable @typescript-eslint/ban-types */
+import { AllowNull, Column, IsDate } from "sequelize-typescript";
+import ParanoidModel from "./ParanoidModel";
+
+class ArchivableModel<
+  TModelAttributes extends {} = any,
+  TCreationAttributes extends {} = TModelAttributes
+> extends ParanoidModel<TModelAttributes, TCreationAttributes> {
```
apoorv-mishra, line 8: Extending from `ParanoidModel` here. There's a limitation here that a non-paranoid model which may be archivable, can't really extend from `ArchivableModel`. `Notification` model is one such example.

It doesn't seem trivial because Typescript doesn't provide multiple inheritance out of the box. Haven't looked into this yet but seems like mixins can help achieve this. Could be scoped out separately if we want to explore that?

app/stores/base/Store.ts
tommoor: Yea it was removed because it caused a worse bug... I can't remember the exact scenario but basically it makes it impossible to actually remove these models from the store when you want to as they will always get marked as deleted instead

app/components/WebsocketProvider.tsx
```diff
@@ -214,34 +214,31 @@ class WebsocketProvider extends React.Component<Props> {
...
-      )
+      })
     );
 
     this.socket.on(
       "documents.archive",
-      action((event: PartialWithId<Document>) => {
-        documents.add(event);
-        policies.remove(event.id);
+      action((document: Document) => {
```
tommoor, line 229: We shouldn't change the type like this as it's incorrect - this isn't a `Document` model.

app/components/WebsocketProvider.tsx
```diff
@@ -214,34 +214,31 @@ class WebsocketProvider extends React.Component<Props> {
...
+        documents.addToArchive(document);
```
tommoor, line 230:
```suggestion
        documents.addToArchive(document as Document);
```

If neccessary, this is preferable

tommoor (approved):

PR #7495, opened 2024-08-31
fix: Refactor sidebar context management

tommoor:
This fixes several bugs where a document could be active highlighted in multiple locations and simplifies the logic for any number of sidebar "sections" going forward.

Each sub-tree of the sidebar must be wrapped in a `SidebarContext.Provider`.

### DIFF
--- a/app/components/Sidebar/components/CollectionLink.tsx
+++ b/app/components/Sidebar/components/CollectionLink.tsx
@@ -22,8 +22,8 @@ import CollectionMenu from "~/menus/CollectionMenu";
 import DropToImport from "./DropToImport";
 import EditableTitle, { RefHandle } from "./EditableTitle";
 import Relative from "./Relative";
+import { SidebarContextType, useSidebarContext } from "./SidebarContext";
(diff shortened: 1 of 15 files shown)


app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -131,17 +131,18 @@ const CollectionLink: React.FC<Props> = ({
...
           icon={<CollectionIcon collection={collection} expanded={expanded} />}
           showActions={menuOpen}
           isActiveDrop={isOver && canDrop}
-          isActive={(match, location: Location<{ starred?: boolean }>) =>
-            !!match && location.state?.starred === inStarredSection
-          }
+          isActive={(
+            match,
+            location: Location<{ sidebarContext?: SidebarContextType }>
+          ) => !!match && location.state?.sidebarContext === sidebarContext}
```
tommoor, line 145: TLDR: Instead of multiple hooks and things like `starred` just compare `sidebarContext` to see where in the sidebar the current component is being rendered

PR #7531, opened 2024-09-04
feat: Add background task for empty_trash action

hmacr:
This PR moves the heavy lifting to a background task.

### DIFF
--- a/server/routes/api/documents/documents.ts
+++ b/server/routes/api/documents/documents.ts
@@ -66,6 +66,7 @@ import {
 import DocumentImportTask, {
   DocumentImportTaskResponse,
 } from "@server/queues/tasks/DocumentImportTask";
+import EmptyTrashTask from "@server/queues/tasks/EmptyTrashTask";
 import FileStorage from "@server/storage/files";
 import { APIContext } from "@server/types";
 import { RateLimiterStrategy } from "@server/utils/RateLimiter";
@@ -1975,9 +1976,8 @@ router.post(
       paranoid: false,
     });
 
-    await documentPermanentDeleter(documents);
-    await Event.createFromContext(ctx, {
-      name: "documents.empty_trash",
+    await EmptyTrashTask.schedule({
+      documentIds: documents.map((doc) => doc.id),
     });
 
     ctx.body = {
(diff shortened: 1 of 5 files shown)


server/routes/api/documents/documents.ts
```diff
@@ -1975,9 +1976,8 @@ router.post(
       paranoid: false,
     });
 
-    await documentPermanentDeleter(documents);
-    await Event.createFromContext(ctx, {
```
hmacr, line 1979: Removed unnecessary event which was added in the initial implementation.

server/routes/api/documents/documents.ts
```diff
@@ -1975,9 +1976,8 @@ router.post(
...
-      name: "documents.empty_trash",
+    await EmptyTrashTask.schedule({
+      documentIds: documents.map((doc) => doc.id),
```
hmacr, line 1980: API determines the documents to delete at the time it was invoked by the user.

server/routes/api/documents/documents.ts
tommoor: I think the event is useful actually because this is a very destructive action… without it there is no record of who permanently deleted data 🙈

server/routes/api/documents/documents.ts
tommoor: Good call

tommoor: We could use `attributes` on the initial query to only retrieve the `id` and speed things up further

server/routes/api/documents/documents.ts
```diff
@@ -1975,7 +1977,9 @@ router.post(
       paranoid: false,
     });
 
-    await documentPermanentDeleter(documents);
+    await EmptyTrashTask.schedule({
```
tommoor, line 1980: Oh, lets add a guard here for `documents.length > 0`

hmacr: We have one in the task.. - but probably safer (and efficient) to have it at both the places 👍

tommoor: We can avoid creating the event too if no work is being done

tommoor (approved):

PR #7548, opened 2024-09-05
feat: group/groupuser commands

NextFire:
In an effort to add OIDC group mapping in our fork (https://github.com/Japan7/outline/commit/183b884911f227d58fc90c9aad65024b7611c3ac), I've created several commands for `Group` and `GroupUser` that I believe could also be integrated upstream.

### DIFF
--- a/server/commands/groupCreator.ts
+++ b/server/commands/groupCreator.ts
@@ -1,0 +1,39 @@
+import type { Transaction } from "sequelize";
+import { Event, Group, type User } from "@server/models";
+
+type Props = {
+  name: string;
+  actor: User;
+  ip: string;
+  transaction?: Transaction;
+};
+
+export default async function groupCreator({
+  name,
+  actor,
+  ip,
+  transaction,
+}: Props): Promise<Group> {
+  const group = await Group.create(
+    {
+      name,
+      teamId: actor.teamId,
+      createdById: actor.id,
+    },
+    { transaction }
+  );
+  await Event.create(
+    {
+      name: "groups.create",
+      modelId: group.id,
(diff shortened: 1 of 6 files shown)


server/commands/groupCreator.ts
```diff
@@ -0,0 +1,39 @@
...
+}: Props): Promise<Group> {
+  const group = await Group.create(
+    {
+      name,
+      teamId: actor.teamId,
+      createdById: actor.id,
+    },
+    { transaction }
+  );
+  await Event.create(
```
tommoor, line 25: We should retain all these as `Event.createFromContext` to automatically add the correct actor, team, and ip.

tommoor (commented):
Thanks, overall happy to merge this type of change but we should keep `createFromContext` rather than remove :)

server/commands/groupCreator.ts
NextFire: The other existing commands do not take the context in input and rely on `Event.create`.
I think there may be some cases where you want to call these commands but an `APIContext` is not available (that what I'm doing for OIDC group mapping).

tommoor: It is unfortunate that `accountProvisioner` does not already get passed `ctx` but it can/should and that would solve the issue

NextFire: This `ctx` would be a simple `Context`, not a complete `APIContext`, since the user hasn't been created yet and doesn't have an access token. I wouldn't be able to properly create an `APIContext` to pass to these commands. 🫤

tommoor (approved):

PR #7570, opened 2024-09-09
Update i18next-parser to fix cheerio bug

CKGrafico:
Just needed to fix this bug i18next/i18next-parser#1045  
  
If this is too small to request a PR feel free to cancel it


tommoor:
You hit an issue related to this? 🤔 The lockfile should prevent such things.

CKGrafico:
I'm super stupid sorry, your last version is updated enough I was in an older https://github.com/outline/outline/blob/209e5e20d5eada34bcd42aa4a8126be4b2042c72/package.json#L344

PR #7628, opened 2024-09-19
In `documents.archived`, allow `sort: "index"` in request

### LINKED ISSUE #7627: `documents.archived` fails with status 500 if the value of `sort` param in request is `"index"`
This has been the case since ages. It only got uncovered while working on #7266, particularly the part where "Documents" tab should show docs in an archived collection 👇

Screenshot 2024-09-19 at 9 54 45 PM

apoorv-mishra:
Also, enables filtering archived docs by `collectionId`.

closes #7627

### DIFF
--- a/server/routes/api/documents/documents.test.ts
+++ b/server/routes/api/documents/documents.test.ts
@@ -2066,6 +2066,101 @@ describe("#documents.templatize", () => {
 });
 
 describe("#documents.archived", () => {
+  it("should succeed with 200 ok if sort parameter in request is set to index", async () => {
+    const user = await buildUser();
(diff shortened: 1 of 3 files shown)


tommoor (approved):
Seems good, makes me think we should probably deprecate this endpoint in favor of a status filter on `documents.list` in the future

apoorv-mishra:
>we should probably deprecate this endpoint in favor of a status filter on documents.list

While I'm at it, thinking of doing this as part of #7266 – should be straightforward. Helps get rid of the unnecessary conditional flow on FE too there.

Issue #7635, opened 2024-09-22
0.80.0 - Docker - Version mismatch


gu1ll0me:
Using Docker image tag `0.80.0` displays version 0.79.1 in the Preferences menu.

- Install: self hosted
- Version: 0.80.0

b-d-e:
+1 to this, was just about to make an issue for the same thing
<img width="139" alt="image" src="https://github.com/user-attachments/assets/b5f15b63-a8ca-45a3-afef-3c5419f78852">
<img width="473" alt="image" src="https://github.com/user-attachments/assets/34929d3a-602a-466b-b38f-32d176e66ab5">

PR #7636, opened 2024-09-22
fix version mismatch to 0.80.0

b-d-e:
A few references to old version number 0.79.1 in the 0.80.0 release. Simple PR to fix. 

Fixes issue https://github.com/outline/outline/issues/7635


tommoor:
Thanks, but this doesn't really fix the issue as the docker image has already been baked.

Will put out a patch release later today with the correct version no.

PR #7658, opened 2024-09-26
perf: store document state in context

hmacr:
This PR enhances the document context to store state that's used across components.
This could potentially be used to handle all changes from the editor.


app/components/DocumentContext.tsx
```diff
@@ -0,0 +1,76 @@
+import { action, computed, observable } from "mobx";
+import React, { PropsWithChildren } from "react";
+import { Heading } from "@shared/utils/ProsemirrorHelper";
+import Document from "~/models/Document";
+import { Editor } from "~/editor";
+
+class DocumentContext {
```
hmacr, line 7: Having the document and editor instance here makes it easy to central the editor event updates.
One potential problem is, the context is provided high in the component tree (in AuthenticatedLayout). We can introduce another context for the state, but we'll end up duplicating the editor references across two contexts.

hmacr:
@tommoor The widget solution that we discussed may not work well since the `headings` state is used in many components. I figured context would be much clearer. Lmk your thoughts.

tommoor:
Yes, looking much cleaner… will have to measure the performance though as you're now calculating headings and tasks potentially more than previously?

tommoor:
I'd create a doc with lots of headings and tasks to get a feel for worst case scenario

hmacr:
Headings calculation count remains the same; tasks calculation has increased though.

re: perf, I tested with ~50 headings and ~100 tasks in a document. UI is smooth - I can see the update/render cycle takes ~5ms for each update.

app/components/DocumentContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+      "useDocumentContext must be used within DocumentContextProvider"
+    );
+  }
+  return ctx;
+};
+
+export const DocumentContextProvider = ({
+  children,
+}: PropsWithChildren<unknown>) => {
+  const context = new DocumentContext();
```
tommoor, line 74: Should this not be memoized? 🤔

hmacr: It's not affected by the page renders since it's at the layout level - gets created just once.

tommoor: ok 👍

tommoor (approved):

tommoor (changes requested):
I realised that because the DocumentContextProvider is only added within authenticated layout the logic errors on shared documents which are mounted outside that layout.

hmacr:
> the logic errors on shared documents

Thanks for finding the bug.

Just updated the PR to provide doc-context for shared documents.

app/components/DocumentContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+      "useDocumentContext must be used within DocumentContextProvider"
+    );
+  }
+  return ctx;
+};
+
+export const DocumentContextProvider = ({
+  children,
+}: PropsWithChildren<unknown>) => {
+  const context = React.useMemo(() => new DocumentContext(), []);
```
hmacr, line 74: Memoizing the context since the shared document page will re-render for every doc navigation.

tommoor (approved):

PR #7664, opened 2024-09-26
fix: check doc access before sending mention email

hmacr:
Follow-up to [this commit](https://github.com/outline/outline/commit/e032bb5ab8ace06329d131ce301af5d847d4f5ff)

### DIFF
--- a/server/queues/tasks/CommentUpdatedNotificationsTask.ts
+++ b/server/queues/tasks/CommentUpdatedNotificationsTask.ts
@@ -43,7 +43,7 @@ export default class CommentUpdatedNotificationsTask extends BaseTask<CommentEve
         recipient.subscribedToEventType(
           NotificationEventType.MentionedInComment
         ) &&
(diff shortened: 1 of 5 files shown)


server/queues/tasks/CommentUpdatedNotificationsTask.ts
```diff
@@ -41,14 +42,16 @@ export default class CommentUpdatedNotificationsTask extends BaseTask<CommentEve
...
+        ) &&
+        (await canUserAccessDocument({ user: recipient, docId: document.id }))
       ) {
         await Notification.create({
           event: NotificationEventType.MentionedInComment,
           userId: recipient.id,
           actorId: mention.actorId,
           teamId: document.teamId,
           documentId: document.id,
+          commentId: comment.id,
```
hmacr, line 54: Right now, we don't send comment-mentioned email in the update flow.

tommoor (approved):

PR #7671, opened 2024-09-28
fix: Correctly decorate urlId as Unique

### LINKED ISSUE #7589: Refreshing or Opening in new tab results in 404
Same as title, similar to #7188 

Started happening after updating from 0.74 to 0.79.1, and **only affects documents created/published after the update**.

Tried clearing cache, to no avail.

**To Reproduce**
1. Open Outline
2. Create a document and publish
4. Refresh / copy URL & open in new tab

Page shows "Not Found", in the Inspector the API call for `documents.info` (POST) returns 404.
The body of the request looks the same as for those that return correctly:

```
{id: "240911-sfg-meeting-QFXB0AEQKz"}
```

Response:
```
{ok: false, error: "not_found", status: 404, message: "Resource not found"}
```

I checked in the DB (PostgreSQL 14.2), the document exists with the correct `urlId` (`QFXB0AEQKz`), same as the "working" docs.

Everything works when navigating to the document manually by selecting it, because then `documents.info` is called with the UUID of the doc:
```
{id: "240603-spg-sync-meeting-Ws4kOcp6CB"}
```

**Expected behavior**
The document being reachable

- Install: self-hosted
- Version: v0.79.1 (https://github.com/outline/outline/commit/a6b96727791d625a7289367be6f0217cad3ee600)

 - OS: macOS 14.6
 - Browser: Chrome 128

tommoor:
closes #7589

### DIFF
--- a/server/models/Document.ts
+++ b/server/models/Document.ts
@@ -23,7 +23,6 @@ import {
   BelongsTo,
   Column,
   Default,
-  PrimaryKey,
   Table,
   BeforeValidate,
   BeforeCreate,
@@ -39,6 +38,7 @@ import {
   IsDate,
   AllowNull,
   BelongsToMany,
+  Unique,
 } from "sequelize-typescript";
 import isUUID from "validator/lib/isUUID";
 import type {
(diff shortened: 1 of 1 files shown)


kerberjg:
Interesting! How come was this the issue?

tommoor:
I don't know for sure as I could never reproduce your problem but the decorator does change how the query is constructed, `v0.80.3-0` is building right now and will be live on docker hub in 10mins for you to try.

PR #7685, opened 2024-09-28
Add recently viewed documents to top of command menu

### LINKED ISSUE #7677: Show recently viewed documents in the command menu
You get recently viewed on the home but elsewhere in the app there is no quick way to jump to these, quick win.

tommoor:
Also adds priority key to actions and sections and colocates all `CommandBar` components/hooks.

closes #7677


app/stores/CollectionsStore.ts
```diff
@@ -95,55 +72,6 @@ export default class CollectionsStore extends Store<Collection> {
     );
   }
 
-  /**
-   * List of paths to each of the documents, where paths are composed of id and title/name pairs
-   */
-  @computed
-  get pathsToDocuments(): DocumentPath[] {
```
tommoor, line 102: cleaned up a bunch of unused methods in the store here

PR #7697, opened 2024-09-29
perf: use findAll for querying document collaborators

### LINKED ISSUE #7687: Optimize `collaborators` query
Just noticed this is doing queries in a map loop – could easily be changed to a single query.

https://github.com/outline/outline/blob/a57a3749cd5bf2453c1a4ceca1d497e10a93c085/server/models/Document.ts#L813

hmacr:
Closes #7687

### DIFF
--- a/server/models/Document.ts
+++ b/server/models/Document.ts
@@ -807,17 +807,14 @@ class Document extends ParanoidModel<
    * @param options FindOptions
    * @returns A promise that resolve to a list of users
    */
-  collaborators = async (options?: FindOptions<User>): Promise<User[]> => {
-    const users = await User.findAll({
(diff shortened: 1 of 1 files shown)


server/models/Document.ts
```diff
@@ -808,11 +808,13 @@ class Document extends ParanoidModel<
...
-    );
+    const users = await User.findAll({
+      ...options,
+      where: {
+        ...options?.where,
+        id: this.collaboratorIds,
+      },
+    });
 
     return compact(users);
```
tommoor, line 819: We can remove the `compact` call too as `findAll` won't return empty items in the array

hmacr: Ah yes 🤦

tommoor (approved):

Issue #7675, opened 2024-09-28
Move API Keys to a workspace list


tommoor:
Currently API keys are managed on a per-user basis, this means that as an admin of a workspace you do not have any visibility into who has keys live, which are being used, when they expire etc.

The suggestion here is to make it possible for admins to view the existence of and manage all API keys in the workspace both via the API and the UI.

tommoor:
Ensure that admins can revoke other users API keys

PR #7699, opened 2024-09-30
Add hashed column for API keys

tommoor:
Towards #7675 – adds news column for storing the API key as a hashed value, and a preview of that value.

### DIFF
--- a/app/scenes/Settings/components/ApiKeyListItem.tsx
+++ b/app/scenes/Settings/components/ApiKeyListItem.tsx
@@ -61,11 +61,11 @@ const ApiKeyListItem = ({ apiKey, isCopied, onCopy }: Props) => {
           )}
           <Text
             type="tertiary"
-            size="small"
+            size="xsmall"
             style={{ marginRight: 8 }}
             monospace
           >
-            ol...{apiKey.last4}
+            {apiKey.obfuscatedValue}
           </Text>
           <ApiKeyMenu apiKey={apiKey} />
         </Flex>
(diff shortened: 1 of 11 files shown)


app/scenes/Settings/components/ApiKeyListItem.tsx
```diff
@@ -52,11 +52,21 @@ const ApiKeyListItem = ({ apiKey, isCopied, onCopy }: Props) => {
...
+              </Button>
+            </CopyToClipboard>
+          )}
+          <Text
+            type="tertiary"
+            size="small"
+            style={{ marginRight: 8 }}
+            monospace
+          >
+            ol...{apiKey.last4}
```
tommoor, line 68: As the API key is no longer available after creation, we show a preview of the value instead to allow identification.

server/migrations/20240929194201-add-hash-to-api-key.js
```diff
@@ -0,0 +1,35 @@
...
+
+  async down (queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async transaction => {
+      await queryInterface.removeColumn("apiKeys", "hash", { transaction });
+      await queryInterface.removeColumn("apiKeys", "last4", { transaction });
+      await queryInterface.changeColumn("apiKeys", "secret", {
+        type: Sequelize.STRING,
+        allowNull: false,
+      }, { transaction });
+    });
```
apoorv-mishra, line 33: Note that this would fail because `secret` would have `null` values!

tommoor: If any new API keys have been created it does fail yea, they need removing from the DB before rolling back.

Unfortunately by definition this change is lossy and non-revertable.

apoorv-mishra (approved):
LGTM 👍

PR #7731, opened 2024-10-04
[WIP] Run outline without a dedicated host

TimRepke:
In some scenarios, there is no easy way to set up a dedicated host/subdomain to serve outline. I spent the last few days trying to add configuration options that do not assume a dedicated host. Since this was ad-hoc and somewhat got out of hand and I haven't worked with React before, this might not be a clean and stable, but it works now.

Assume we have the server running and want to access it via https://domain.tld/wiki and serve static files at https://domain.tld/wiki/static. Then the config needs

```
URL=https://domain.tld/wiki
BASENAME=/wiki
CDN_URL=https://domain.tld/wiki/static/  # e.g. serving /srv/outline/build/app directly from nginx
FORCE_HTTPS=false  # prevent Location header, let nginx handle TLS
COLLABORATION_URL=https://domain.tld/wiki
```

In case you are interested in mainstreaming this I thought I'd share the changes. This still runs fine locally in development mode, but I haven't tested all integrations. There are many places where the backend returns a redirect URL that is missing the basename or simply strips the path, so I'm not sure that this is a complete transformation. I did notice several inconsistent /static links that might be worth double-checking in mainline either way.


tommoor:
Hi @TimRepke – we won't merge this I'm afraid due to the extra maintenance required to support this rare setup. I've seen clearly over the years how "small" changes like this result in many bugs over time.

I'm not sure in what situation it is difficult to give Outline a dedicated subdomain, but you're welcome to maintain this fork of course.

TimRepke:
Totally understandable, this PR most definitely has bugs I haven't found yet, it was really just meant as a pitch/prove of concept that I wanted to share as it seems to work for me.
Thanks for building this great platform, it looks beautiful and seems equipped with exactly what I was hoping to find.

danfergo:
Hi guys, I'm interested in this feature, too. I was just about to open a feature request for the same case. I'm trying to install Outline under **/wiki** with no success. In my case, I'm using TP-LINK Dynamic DNS and I don't have access to subdomains, so the **/wiki** path would be the only solution. I'm considering dropping Outline and sticking with nextcloud + nextcloud collectives.

TimRepke:
Hi @danfergo , in our deployment using my fork, outline has been working fine now with the changes I made. This is running for about two weeks now with about 20 active users. So it's probably stable enough for you to adopt this.
Unfortunately, I probably won't have any capacity to keep the fork updated with the mainline repository.

Issue #7700, opened 2024-09-30
Switch to `dockerhub` as source for "versions behind" notification


tommoor:
Currently we use the GitHub releases API to detect if the installed community edition is behind – this works well for community, but not for the enterprise image as there is no corresponding GitHub release. Instead we should switch to using the dockerhub API as the source for this information:

https://hub.docker.com/v2/repositories/outlinewiki/outline/tags?name=&ordering=last_updated&page=2&page_size=10

AlexandrZagorskiy:
Hi @tommoor ,

I’ve been exploring the issue you raised regarding version checking for the enterprise edition, and I’d like to share my approach. Here's what you can expect:

### Approach:
1. **New `DOCKERHUB_REPO_URL` environment variable**:
   - I’ll add this variable to the `.env` file to allow dynamic configuration for both the community and enterprise editions. The URL will be compiled based on the edition in use.

2. **Modification of the UI version check**:
   - I’ll update the current UI logic (found in `Version.tsx`) to fetch version information from DockerHub using `DOCKERHUB_REPO_URL`, instead of relying solely on the GitHub Releases API. This will make the UI capable of showing version status for the enterprise image hosted on DockerHub.

This change can serve as a foundational step towards addressing the request in [Discussion #7168](https://github.com/outline/outline/discussions/7168), which involves making the version accessible through the Outline Rest API. After the initial UI implementation, we can build on this to allow querying version data via the API.

### Expected Outcome:
Once this PR is merged, the UI will correctly display whether the installed enterprise version is up-to-date based on DockerHub data, just like it currently does for the community edition with GitHub.

Please let me know if this aligns with your expectations or if there are additional points you'd like me to consider!

tommoor:
Hi, I wasn't really expecting anyone else to take this on, but if you want to that's fine.

There is no need to make the URL an environment variable, just hardcode it the same way that the GitHub URL is today

AlexandrZagorskiy:
During my work on switching the version check from GitHub to DockerHub, I encountered an issue related to CORS when trying to fetch data from DockerHub directly on the client-side. Interestingly, GitHub's API does not trigger this problem, but DockerHub’s API does.

[...]

PR #7744, opened 2024-10-08
Feat/installation info endpoint

AlexandrZagorskiy:
This PR adds an endpoint for fetching version information, integrated with DockerHub. Discussed in #7700 

The stack and language used here are completely new for me, so I approached it intuitively. I would really appreciate any feedback or suggestions for improvement!

I have tested endpoint manually. I'm still working on the automatic tests for this endpoint, but wanted to get an initial review to ensure I'm on the right track.


server/presenters/installation.ts
```diff
@@ -0,0 +1,11 @@
+export default function presentInstallationInfo(data: {
```
tommoor, line 1: No need for the presenter, just return the object in the router directly 👍

server/routes/api/installation/installation.ts
```diff
@@ -0,0 +1,35 @@
...
+  getVersionsBehind,
+} from "@server/utils/getInstallationInfo";
+import * as T from "./schema";
+
+const router = new Router();
+
+router.post(
+  "installation.info",
+  auth(),
+  validate(T.InstallationInfoSchema),
```
tommoor, line 18: Can remove the validation and schema as there is no POST body required

server/utils/getInstallationInfo.ts
```diff
@@ -0,0 +1,50 @@
+import env from "@server/env";
+import { version } from "../../package.json";
+import fetch from "./fetch";
+
+const dockerhubLink =
+  "https://hub.docker.com/v2/repositories/outlinewiki/outline";
+
+export function getVersion(): string {
+  return env.VERSION ? env.VERSION : version;
```
tommoor, line 9: Don't use `env.VERSION` here, it might be a `sha` or some other non-semver value, the version in `package.json` is the right one to use

tommoor (commented):
Overall direction seems good, thanks!

AlexandrZagorskiy:
I mistakenly rebased my branch with the current version of the main branch, so I force-pushed the changes. However, I’ve kept all fixes and adjustments in separate commits for easier review.

Additionally, the documentation for this endpoint still needs to be written. Given my upcoming schedule and other commitments, I may not be able to complete it in the near future. It might be better to leave this as a separate issue for someone else to handle. I’m open to your thoughts on this!

That's how it looks on my local version: 
image
image

app/components/Sidebar/components/Version.tsx
```diff
@@ -2,41 +2,36 @@ import * as React from "react";
...
   React.useEffect(() => {
-    async function loadReleases() {
-      const res = await fetch(
-        "https://api.github.com/repos/outline/outline/releases"
-      );
-      const releases = await res.json();
+    async function loadVersionInfo() {
+      try {
+        // Fetch version info from the server-side proxy
+        const res = await fetch("/api/installation.info", {
```
tommoor, line 17: Oh, I just realised this should use `ApiClient` (see other usage) so you don't need to define headers, parse JSON, etc.

app/components/Sidebar/components/Version.tsx
```diff
@@ -2,41 +2,36 @@ import * as React from "react";
...
+          },
+        });
 
-      if (Array.isArray(releases)) {
-        const everyNewRelease = releases
-          .map((release) => release.tag_name)
-          .findIndex((tagName) => tagName === `v${version}`);
+        const { data } = await res.json();
 
-        const onlyFullNewRelease = releases
```
tommoor, line 24: One last thing is that I think we need to maintain some sense of this logic – in your example screenshot it says you're one version being because there is a pre-release that's newer but we definitely don't want to prompt people to upgrade to pre-releases, they should be ignored for the purposes of version comparison.

tommoor:
Thanks, no worries on the rebase

tommoor:
lmk if you're thinking of finishing this, or if you'd like someone to take it over. Thanks!

AlexandrZagorskiy:
Hey there,

First of all, I’d like to apologize for my absence over the past two weeks. I had some other priorities, but I’ve now returned to this issue and I’m ready to move it forward.

Regarding the full-release logic: I’ve assumed that a full release is any version that matches the following templates:

- "N.N.N" for [versions starting from 0.58.0 on the DockerHub](https://hub.docker.com/v2/repositories/outlinewiki/outline/tags?name=&ordering=last_updated&page=20&page_size=4) 
- "version-N.N.N" for versions before 0.58.0

I couldn’t find any field in DockerHub that explicitly identifies whether a version is a pre-release or full release, so this regex-based approach should cover most cases. If I’m mistaken and there’s a specific DockerHub field for this, the logic can be easily updated to accommodate that.

I’ve also included a few screenshots to demonstrate how the UI behaves when I manually change the version number in package.json:
image
image
image
image

This should give you a better sense of how the changes are reflected visually.

Let me know if you have any further suggestions or thoughts on improving this implementation!

app/components/Sidebar/components/Version.tsx
AlexandrZagorskiy: Done, check it [here](https://github.com/outline/outline/pull/7744/commits/de0a99e090996fe37d77efb06029d27b3504f6c1)

tommoor (approved):

tommoor:
> I couldn’t find any field in DockerHub that explicitly identifies whether a version is a pre-release or full release

That sounds right, It's just a matter of parsing the version number. I think one improvement here would be to use the `semver` library to deal with the version number. I'm going to merge now, but we could come back and fix that – it would make it more resilient and the dependency is already included.

AlexandrZagorskiy:
Thank you for the review and the approval. I’m looking forward to seeing this merged!

PR #7745, opened 2024-10-08
fix: show consistent structure for shared docs

### LINKED ISSUE #7502: Sub-documents within shared parent are not ordered correctly
**To Reproduce**
1. Share an individual document with another user
2. Add multiple sub-documents
3. Re-order the sub-documents
4. They will not always be the same order for the user they were shared with.

**Expected behavior**

Order matches on both clients

hmacr:
Closes #7502

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -27,6 +27,7 @@ import withStores from "~/components/withStores";
 import {
   PartialExcept,
   WebsocketCollectionUpdateIndexEvent,
+  WebsocketDocumentMovedEvent,
   WebsocketEntitiesEvent,
   WebsocketEntityDeletedEvent,
 } from "~/types";
@@ -147,9 +148,7 @@ class WebsocketProvider extends React.Component<Props> {
 
             // otherwise, grab the latest version of the document
             try {
-              document = await documents.fetch(documentId, {
-                force: true,
-              });
+              document = await documents.fetch(documentId, { force: true });
             } catch (err) {
               if (
                 err instanceof AuthorizationError ||
@@ -175,6 +174,16 @@ class WebsocketProvider extends React.Component<Props> {
                   id: document.collectionId,
                 });
               }
+            }
+
+            const parentDocuments = document?.parentDocuments ?? [];
+            for (const parentDoc of parentDocuments) {
+              await userMemberships
+                .find({ documentId: parentDoc.id })
+                ?.fetchDocuments({ force: true });
+              await groupMemberships
+                .find({ documentId: parentDoc.id })
+                ?.fetchDocuments({ force: true });
             }
           }
         }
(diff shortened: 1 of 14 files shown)


app/components/WebsocketProvider.tsx
```diff
@@ -260,6 +302,56 @@ class WebsocketProvider extends React.Component<Props> {
...
+            force: true,
+          });
+
+          // Refresh documents structure in case this document (or) any of its ancestors is shared to the user.
+          const documentIdsToRefresh = [
+            event.documentId,
+            ...document.parentDocuments.map((doc) => doc.id),
+          ];
+
+          for (const documentId of documentIdsToRefresh) {
```
tommoor, line 319: I'm very worried about the scalability of this, have been doing a lot of work in the last 6mo to move away from websocket events triggering fetches as it's inherently incompatible with large teams where a single person moving some documents can trigger thousands of online clients to hit the server.

app/models/UserMembership.ts
```diff
@@ -72,6 +80,85 @@ class UserMembership extends Model {
     return memberships[index + 1];
   }
 
+  /**
+   * Fetches the sub-documents structure, in case this membership provides access to a document.
+   */
+  fetchDocuments = async (options?: { force: boolean }) => {
```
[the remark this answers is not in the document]
hmacr: That makes sense - I wasn't sure if it was worth it to implement another class just for this.
Perhaps, a `NavigableModel`?

app/components/WebsocketProvider.tsx
hmacr: True - let me explore a more scalable solution that _might_ be used `collections.fetchDocuments` as well (or anything you've thought of already?)

app/models/UserMembership.ts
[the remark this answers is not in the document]
tommoor: At some point `Membership` can go, this could be used as a sensible name for a base model… until then `NavigableModel` seems ok

server/types.ts
```diff
@@ -200,6 +200,8 @@ export type DocumentEvent = BaseEvent<Document> &
         data: {
           collectionIds: string[];
           documentIds: string[];
+          prevParentDocumentId: string | null;
+          parentDocumentChanged: boolean;
```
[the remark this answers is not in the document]
tommoor: It's not a big change either way, just wanted to point it out

app/components/WebsocketProvider.tsx
hmacr: Whenever document structure is changed through various document events, e.g. create, delete, archive, move. etc., how about we send a `NavigationNode` in the websocket event? This will be used by the client to update its local doc structure data.
For this to work without race condition on the client, an `index` (fractional-indexing) field in `NavigationNode` can be introduced. This will be used only when the documents are manually sorted. Needs some tinkering to be backward compatible, I think.. Too complicated?

Another approach would be to send the whole doc structure; but this will consume a lot of bandwidth for very large collections / documents.

tommoor: I think sending the smallest subset of `NavigationNode` can work – if we always send the parent then no additional index would be needed at the cost of sending slightly more data

hmacr: Could be a problem when moving as a top-level document in a collection.
We'll have to send the whole collection structure, otherwise the insert location cannot be determined 🤔

tommoor: We could look at a packet inspired by https://jsonpatch.com 's move / remove operation

PR #7751, opened 2024-10-09
fix: show all document backlinks for a user

### LINKED ISSUE #7534: Backlinks section should show all documents that the current user has access to
**To Reproduce**
1. Link to a public doc from a private doc
2. Visit the public doc
3. Private doc does not appear in backlinks

**Expected behavior**

The private doc should be listed for the current user.

hmacr:
Closes #7534

### DIFF
--- a/server/models/Backlink.ts
+++ b/server/models/Backlink.ts
@@ -54,14 +54,7 @@ class Backlink extends IdModel<
       { userId: user.id }
     );
 
-    return documents
-      .filter(
-        (doc) =>
-          (doc.collection?.memberships.length || 0) > 0 ||
-          doc.memberships.length > 0 ||
-          doc.groupMemberships.length > 0
-      )
-      .map((doc) => doc.id);
+    return documents.map((doc) => doc.id);
   }
 }
 
--- a/server/models/Backlink.ts
+++ b/server/models/Backlink.ts
@@ -1,6 +1,3 @@
-import compact from "lodash/compact";
-import difference from "lodash/difference";
-import uniq from "lodash/uniq";
 import { InferAttributes, InferCreationAttributes } from "sequelize";
 import {
   DataType,
@@ -8,27 +5,12 @@ import {
   ForeignKey,
   Column,
   Table,
-  Scopes,
 } from "sequelize-typescript";
 import Document from "./Document";
-import Group from "./Group";
-import GroupMembership from "./GroupMembership";
-import GroupUser from "./GroupUser";
 import User from "./User";
-import UserMembership from "./UserMembership";
 import IdModel from "./base/IdModel";
 import Fix from "./decorators/Fix";
 
-@Scopes(() => ({
-  withSourceDocument: {
-    include: [
-      {
-        model: Document,
-        as: "reverseDocument",
-      },
-    ],
-  },
-}))
 @Table({ tableName: "backlinks", modelName: "backlink" })
 @Fix
 class Backlink extends IdModel<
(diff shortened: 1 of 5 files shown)


server/models/Backlink.ts
```diff
@@ -37,6 +55,67 @@ class Backlink extends IdModel<
   @ForeignKey(() => Document)
   @Column(DataType.UUID)
   reverseDocumentId: string;
+
+  public static async findSourceDocumentIdsForUser(
```
tommoor, line 59: This works but is quite complicated, I'm worried about bugs – given that `Document.findByPk` already has all the logic to scope to a user in just these lines:

```
https://github.com/outline/outline/blob/7a8d40b9e73454ee70cc5f4320d9e49ba0f48f3c/server/models/Document.ts#L687-L698
```

could we create something similar, perhaps `Document.findByIds` and just pass in the backlink document ID's?

hmacr: Agreed - We can filter for membership existence for the results from `Document.findByIds` (since the query uses left outer join).

server/models/Backlink.ts
```diff
@@ -37,6 +37,32 @@ class Backlink extends IdModel<
...
+      },
+    });
+
+    const documents = await Document.findByIds(
+      backlinks.map((backlink) => backlink.reverseDocumentId),
+      { userId: user.id }
+    );
+
+    return documents
+      .filter(
```
tommoor, line 58: Did you find this filter was required? I would not expect anything to get filtered here 🤔

server/models/Document.ts
```diff
@@ -733,6 +734,41 @@ class Document extends ArchivableModel<
...
+   * @param options FindOptions
+   * @returns A promise resolving to the list of documents
+   */
+  static async findByIds(
+    ids: string[],
+    options?: Omit<FindOptions<Document>, "where"> &
+      Omit<AdditionalFindOptions, "rejectOnEmpty">
+  ): Promise<Document[]> {
+    const userId = options?.userId;
+    const rest = omit(options, "userId");
```
tommoor, line 751:
```suggestion
    const { userId, ...rest } = options;
```

preferred? You would have to provide a default value of empty object for `options` is all

server/models/Backlink.ts
tommoor: looking much better

server/models/Backlink.ts
hmacr: I'm afraid, yes. The membership queries use left outer join (`required: false` [here](https://github.com/outline/outline/blob/main/server/models/Document.ts#L192) & [here](https://github.com/outline/outline/blob/main/server/models/Document.ts#L196) ) -  it'd return all the documents requested. We need to get only the source docs that the user has access to.

server/models/Document.ts
hmacr: `options` is an optional param here. Do we want to make it required?
I was thinking since it's a common `findByIds` method, this could potentially be used as `findByIds(list_of_ids)` in the future.

tommoor: `options = {}` solves it

server/models/Backlink.ts
hmacr: An example where the filter is necessary.

UserA: owns DocA and DocS.
UserB: owns DocB and has access to DocS.

Consider DocS has backlinks to DocA & DocB.
For UserA, we should list only DocA (likewise DocB for userB).

tommoor: Okay, makes sense – but shouldn't we move this filtering into `findByIds` otherwise it would be easy to misuse that method in the future

hmacr: True - missed it 🤦‍♂️
Filter when "userId" is present.

tommoor (approved):

PR #7756, opened 2024-10-10
feat: include content in document mentioned email

### LINKED ISSUE #7526: Comment email notifications should include a preview of the comment
Title says it all.

hmacr:
Closes #7526 

#### List / Paragraph content
Screenshot 2024-10-10 at 7 14 37 PM

#### Table content
Screenshot 2024-10-10 at 7 15 09 PM

### DIFF
--- a/server/emails/templates/DocumentMentionedEmail.tsx
+++ b/server/emails/templates/DocumentMentionedEmail.tsx
@@ -17,7 +17,7 @@ import Heading from "./components/Heading";
 
 type InputProps = EmailProps & {
   documentId: string;
-  revisionId: string;
+  revisionId: string | undefined;
   userId: string;
   actorName: string;
   teamUrl: string;
@@ -43,18 +43,24 @@ export default class DocumentMentionedEmail extends BaseEmail<
       return false;
     }
 
-    const currRevision = await Revision.findByPk(revisionId);
-    if (!currRevision) {
-      return false;
-    }
-    const prevRevision = await currRevision.before();
+    let currDoc: Document | Revision = document;
+    let prevDoc: Revision | undefined;
 
-    const currMentions = DocumentHelper.parseMentions(currRevision, {
+    if (revisionId) {
+      const revision = await Revision.findByPk(revisionId);
+      if (!revision) {
+        return false;
+      }
+      currDoc = revision;
+      prevDoc = (await revision.before()) ?? undefined;
+    }
+
+    const currMentions = DocumentHelper.parseMentions(currDoc, {
       type: "user",
       modelId: userId,
     });
-    const prevMentions = prevRevision
-      ? DocumentHelper.parseMentions(prevRevision, {
+    const prevMentions = prevDoc
+      ? DocumentHelper.parseMentions(prevDoc, {
           type: "user",
           modelId: userId,
         })
@@ -66,7 +72,7 @@ export default class DocumentMentionedEmail extends BaseEmail<
     }
 
     const node = ProsemirrorHelper.getNodeForMentionEmail(
-      DocumentHelper.toProsemirror(currRevision),
+      DocumentHelper.toProsemirror(currDoc),
       firstNewMention
     );
     if (!node) {
--- a/server/emails/templates/DocumentMentionedEmail.tsx
+++ b/server/emails/templates/DocumentMentionedEmail.tsx
@@ -1,20 +1,31 @@
+import differenceBy from "lodash/differenceBy";
 import * as React from "react";
-import { Document } from "@server/models";
+import { Day } from "@shared/utils/time";
+import { Document, Revision } from "@server/models";
+import { DocumentHelper } from "@server/models/helpers/DocumentHelper";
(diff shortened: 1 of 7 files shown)


server/emails/templates/DocumentMentionedEmail.tsx
```diff
@@ -26,13 +37,64 @@ export default class DocumentMentionedEmail extends BaseEmail<
...
+    const prevMentions = prevDoc
+      ? DocumentHelper.parseMentions(prevDoc, {
+          type: "user",
+          modelId: userId,
+        })
+      : [];
+
+    const firstNewMention = differenceBy(currMentions, prevMentions, "id")[0];
+    if (!firstNewMention) {
+      return false;
```
tommoor, line 71: I don't think this early return or the one below are ideal, as they will potentially result in an @mention not generating an email. It would be better to just fallback to the existing case of not having the preview if we cannot for some reason gather the correct document context

server/models/helpers/ProsemirrorHelper.tsx
```diff
@@ -155,6 +169,67 @@ export class ProsemirrorHelper {
     return mentions;
   }
 
+  static getNodeForMentionEmail(doc: Node, mention: MentionAttrs) {
+    let blockNode: Node | undefined;
+    const potentialBlockNodes = ["table", "checkbox_list", "paragraph"];
```
tommoor, line 174: We discussed a table row being the right point to cut, tables can be quite large so just wondering on the `table` inclusion here

server/models/helpers/ProsemirrorHelper.tsx
```diff
@@ -155,6 +169,67 @@ export class ProsemirrorHelper {
     return mentions;
   }
 
+  static getNodeForMentionEmail(doc: Node, mention: MentionAttrs) {
```
tommoor, line 172: Test would be great if possible

tommoor, line 172: JSDoc for new public methods appreciated

tommoor (commented):
niiiice

server/models/helpers/ProsemirrorHelper.tsx
hmacr: Oh yes, row is still the cut point.
We need the table as it has the necessary styles.

On L217, a new parent node gets created with a single child that has the mention.

server/models/helpers/ProsemirrorHelper.tsx
```diff
@@ -155,6 +169,74 @@ export class ProsemirrorHelper {
...
+  /**
+   * Find the nearest ancestor block node which contains the mention.
+   *
+   * @param doc The top-level doc node of a document / revision.
+   * @param mention The mention for which the ancestor node is needed.
+   * @returns A new top-level doc node with the ancestor node as the only child.
+   */
+  static getNodeForMentionEmail(doc: Node, mention: MentionAttrs) {
+    let blockNode: Node | undefined;
+    const potentialBlockNodes = ["table", "checkbox_list", "paragraph"];
```
tommoor, line 181: I think we should add `heading` in this list, currently a mention in a heading falls into the category of not getting a preview

server/emails/templates/DocumentMentionedEmail.tsx
```diff
@@ -74,6 +132,15 @@ Open Document: ${teamUrl}${document.url}
             {actorName} mentioned you in the document{" "}
             <a href={documentLink}>{document.title}</a>.
           </p>
+          {body && (
+            <>
+              <EmptySpace height={20} />
+              <Diff>
```
tommoor, line 138: Do you mind changing the style of `<Diff>` so that it uses a border instead of a background like this:

image

It will help to make mentions display correctly

server/models/helpers/ProsemirrorHelper.tsx
hmacr: Ugh, wasn't aware headings could have mentions 😓

server/emails/templates/DocumentMentionedEmail.tsx
hmacr: Just for the mention (or) the diff element in itself?

This styling of diff looks neat, imo.

tommoor: I was thinking for diff element itself, the grey background clashes with mentions and quotes so you can't see them properly

hmacr: Looks weird with too many lines for tables..
Screenshot 2024-10-11 at 10 34 23 PM

How does this look?
Screenshot 2024-10-11 at 10 39 15 PM

tommoor: I'm ok with the table case, it will be quite rare to just show up like this - changing the background color just has a few too many edgecases. 

I don't know if we could do something to even up the padding that would help a lot

tommoor (approved):

PR #7770, opened 2024-10-12
feat: allow sort by position for comments

### LINKED ISSUE #7725: Option to order comments based on document location, not chronologically
### Discussed in https://github.com/outline/outline/discussions/7399

<sup>Originally posted by **sjwant** August 16, 2024</sup>
Would love for comments to be arranged based on where they are in the document instead of chronologically</div>

hmacr:
Closes #7725


app/scenes/Document/components/CommentSortMenu.tsx
```diff
@@ -0,0 +1,61 @@
...
+      },
+    ],
+    [t, value, onSelect]
+  );
+
+  return (
+    <>
+      <MenuButton {...menu} aria-label={t("Sort comments")}>
+        {(props) => (
+          <Button {...props} neutral borderOnHover icon={<ManualSortIcon />} />
```
hmacr, line 51: Tippy doesn't play well with the menu button here. We'd have to maintain state and add some complex handlers (quite tricky, I think) to enable tooltip. Do we want to do it?

tommoor: No, it's okay – I think we need to change this somehow anyway – having three icons in a row is too much, especially as this sort option is something you'll change once and then likely never change again.

I'll have a think and get back to you

tommoor:
Gave this a test, apart from the icon comment above I noticed that if you have positional sorting enabled there is a flash of comments in chronological order when the document first loads

hmacr:
> if you have positional sorting enabled there is a flash of comments in chronological order when the document first loads

Thanks for catching this - It was quite subtle and sporadic. Just pushed a commit that fixes it.

app/scenes/Document/components/CommentSortMenu.tsx
hmacr: Probably better off moving this to preferences page?

tommoor:
Interesting, the problem makes sense but adding more callbacks and state is not ideal. I wonder if we could just have `getComments` return `undefined` until the editor is loaded and use that as the signal

hmacr:
> I wonder if we could just have `getComments` return `undefined` until the editor is loaded and use that as the signal.

I'm afraid we don't have a way to find if the editor is loaded. Do you know if prosemirror exposes it? I couldn't find it in their docs.
We can't rely on the `editor` ref as well &mdash; ref will be set by React, but then prosemirror wouldn't have finished building the nodes / view.
Without such "loaded" event from prosemirror, we need `isEditorInitialized` to track it.

app/scenes/Document/components/CommentSortMenu.tsx
tommoor: It's nice to directly see the effect it has, on preferences it's not as clear. I think the answer is to combine the resolved button and this into a "Display options" type of menu – I can probably make that change in the next few days, just need to iterate a bit on how it should be.

hmacr: Got it - I thought about this too, just wasn't sure if it's ideal to move the resolved comments toggle inside a menu. But I guess, we don't much option here..

Lmk in case you want me to implement it once the design is ready.

tommoor:
Unfortunately this last fix seems to have create a flash when opening the comment sidebar, it gets almost the entire way through the animation before the comments are rendered. Are you seeing that?

hmacr:
You mean this? 🤔
I see this on main too; think it has to do with framer unable to animate all the descendants properly &mdash; you can see the avatars are rendered without flash).

https://github.com/user-attachments/assets/67277434-becd-41ac-b131-fe731bd3e8b4

In my local, I solved it by wrapping [this comments container](https://github.com/outline/outline/blob/main/app/scenes/Document/components/Comments.tsx#L81) with a `<m.div>` 👇 Could you please add this to the menu redesign you're implementing? (it'd be easier that way I think)

https://github.com/user-attachments/assets/fbaaeb57-fa0d-43c8-ada4-29cd6807f615

tommoor:
Wrapping in another div prevents scrolling unfortunately, but you're right that it's sometimes visible in production – will consider separately.

app/stores/CommentsStore.ts
```diff
@@ -28,14 +32,29 @@ export default class CommentsStore extends Store<Comment> {
...
+    const comments = this.filter(
       (comment: Comment) =>
         comment.documentId === documentId &&
         !comment.parentCommentId &&
         (!comment.isNew ||
           comment.createdById === this.rootStore.auth.currentUserId)
     );
+
+    if (options.type === CommentSortType.Chrono) {
+      return comments;
```
hmacr, line 48: We should delete the overridden method here, as it is the `MostRecent` sort.
https://github.com/outline/outline/blob/e41519575f4ad1b4ee7e429b18feddc8f9ed4884/app/stores/CommentsStore.ts#L142-L144

app/types.ts
```diff
@@ -215,3 +215,12 @@ export type Properties<C> = {
     ? Property
     : never]?: C[Property];
 };
+
+export enum CommentSortType {
+  Chrono = "chronological",
```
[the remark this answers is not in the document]
hmacr: Updated the PR to address this and the other suggestion left below - please verify.

tommoor:
That last commit broke the sorting entirely I'm afraid

hmacr:
Could you try removing the `sortCommentsByPosition` user-pref field in the `users` table?
The rename in the latest commit causes deserialization errors.

tommoor:
I did see that and removed it, but the order still was no longer changed by switching the UI. Perhaps removing the computed getter is the culprit

hmacr:
Curious 🤔  Here's what I see in my local.

https://github.com/user-attachments/assets/38c5207a-7e39-4ffc-b3a0-2dd1809b3682


The computed getter should fallback to the parent method - I don't think it's an issue.
That said, I'm still scratching my head coz the commit has only renames except this.. 🤔

tommoor:
Okay, don't worry – I'll check it out and try again. Perhaps I messed up somehow

tommoor (approved):

PR #7790, opened 2024-10-17
feat: Comment reactions

### LINKED ISSUE #7137: Comment reactions
Badly missing out on comment reactions. We already have the picker ready so can reuse that.

hmacr:
Closes #7137

### Recording
https://github.com/user-attachments/assets/d4812b1f-73d1-4ca0-8e48-f641e256712d


app/menus/CommentMenu.tsx
```diff
@@ -57,6 +59,26 @@ function CommentMenu({
     }
   }, [t, document, comment]);
 
+  const fetchReactionData = React.useCallback(async () => {
```
tommoor, line 62: This seems unnecessary, we can do any such operations as part of the render loop in the component. We're also transforming this data twice – once here and then again inside the `ViewReactions` dialog.

Lets just call `comment.fetchReactions` inside the dialog

app/models/Comment.ts
```diff
@@ -84,6 +86,12 @@ class Comment extends Model {
    */
   resolvedById: string | null;
 
+  /**
+   * Active reactions for this comment.
+   */
+  @observable
+  reactions: Reaction[];
```
tommoor, line 93: Lets clarify that these are not full `Reaction` objects with a different type, `ReactionDetail` or something like that

app/scenes/Document/components/CommentThreadItem.tsx
```diff
@@ -107,6 +118,40 @@ function CommentThreadItem({
...
+  );
+
+  const handleRemoveReaction = React.useCallback(
+    async (emoji: string) => {
+      await comment.removeReaction({ emoji, userId: user.id });
+    },
+    [comment, user]
+  );
+
+  const fetchReactionData = React.useCallback(async () => {
```
tommoor, line 135: Again with the fetch + map, can we just `ReactionList` look after it's own data concerns?

app/components/Reactions/ViewReactionsDialog.tsx
```diff
@@ -0,0 +1,148 @@
...
+  () => true
+);
+PlaceHolder.displayName = "ViewReactionsPlaceholder";
+
+const TabActionsWrapper = styled(Flex)`
+  border-bottom: 1px solid ${s("inputBorder")};
+`;
+
+const StyledTab = styled(Tab)<{ $active: boolean }>`
+  position: relative;
```
tommoor, line 108: This is basically completely restyled, is it neccessary?

plugins/webhooks/server/tasks/DeliverWebhookTask.ts
```diff
@@ -178,6 +178,10 @@ export default class DeliverWebhookTask extends BaseTask<Props> {
       case "comments.delete":
         await this.handleCommentEvent(subscription, event);
         return;
+      case "comments.add_reaction":
+      case "comments.remove_reaction":
+        // Ignored
```
tommoor, line 183: We can create a follow up task for this

server/routes/api/comments/comments.ts
```diff
@@ -360,4 +360,119 @@ router.post(
   }
 );
 
+router.post(
+  "comments.add_reaction",
+  rateLimiter(RateLimiterStrategy.TwentyFivePerMinute),
```
hmacr, line 365: Do we want it to increase it? Something like 60/sec? (Likewise for `remove_reaction` API below).

app/components/Reactions/Reaction.tsx
```diff
@@ -0,0 +1,173 @@
...
+  padding: 8px;
+  border-radius: 12px;
+  border: 1px solid transparent;
+  transition: ${s("backgroundTransition")};
+
+  ${({ $active, theme }) =>
+    $active
+      ? theme.isDark
+        ? css`
+            background-color: ${darken(0.1, theme.accent)};
```
hmacr, line 144: Please feel free to update the styles if it can be improved 🙈

server/routes/api/reactions/reactions.ts
```diff
@@ -0,0 +1,45 @@
...
+  validate(T.ReactionsListSchema),
+  async (ctx: APIContext<T.ReactionsListReq>) => {
+    const { commentId } = ctx.input.body;
+    const { user } = ctx.state.auth;
+
+    const comment = await Comment.findByPk(commentId, {
+      rejectOnEmpty: true,
+    });
+
+    authorize(user, "readReaction", comment);
```
tommoor, line 24: Lets also check you have permission to read the associated document here, as here:

https://github.com/outline/outline/blob/d3ddb25c760aeb29120d8d1284e6f40b81e7e731/server/routes/api/comments/comments.ts#L73

server/routes/api/reactions/reactions.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const reactions: Reaction[] = [];
+
+    await Reaction.findAllInBatches<Reaction>(
```
tommoor, line 28: I think this should use the pagination pattern instead, you can use `fetchAll` on the client to make the usage there the same.

server/routes/api/comments/comments.ts
```diff
@@ -360,4 +360,119 @@ router.post(
...
+
+    await Reaction.findOrCreate({
+      where: {
+        emoji,
+        userId: user.id,
+        commentId: id,
+      },
+      transaction,
+    });
+    const added = await comment.updateReactions({
```
tommoor, line 393: Move and split this logic in `updateReactions` into `AfterCreate` and `AfterDestroy` hooks on the `Reaction` model, I promise everything will end up cleaner.

You can then also use the return property from `findOrCreate` to know if it was added or not.

server/routes/api/comments/comments.ts
```diff
@@ -360,4 +360,119 @@ router.post(
...
+
+    const comment = await Comment.findByPk(id, {
+      transaction,
+      rejectOnEmpty: true,
+      lock: {
+        level: transaction.LOCK.UPDATE,
+        of: Comment,
+      },
+    });
+    authorize(user, "addReaction", comment);
```
tommoor, line 383: This permission is currently just checking the user and comment are in the same workspace, it's not enough – lets also check they can comment on the document.

server/routes/api/comments/comments.ts
```diff
@@ -360,4 +360,119 @@ router.post(
...
+
+    const comment = await Comment.findByPk(id, {
+      transaction,
+      rejectOnEmpty: true,
+      lock: {
+        level: transaction.LOCK.UPDATE,
+        of: Comment,
+      },
+    });
+    authorize(user, "removeReaction", comment);
```
tommoor, line 441: This permission is currently just checking the user and comment are in the same workspace, it's not enough – lets also check they can comment on the document.

app/hooks/useHover.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Hook to detect if an element is hovered by the user.
+ *
+ * This signals the hover only on the first time an element is hovered.
+ *
+ * @returns {Object} object - The returned object.
+ * @returns {boolean} object.hovered - Signals when an element is hovered by the user.
+ * @returns {function(): void} object.onMouseEnter - Attach this function to the observed element's onMouseEnter event handler.
+ * @returns {function(): void} object.onMouseLeave - Attach this function to the observed element's onMouseLeave event handler.
+ */
+const useHover = ({ duration }: Props) => {
```
tommoor, line 19: Would it be a nicer API to pass a `ref` as a prop and handle the event binding internally?

app/components/Reactions/ReactionList.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { hovered, onMouseEnter, onMouseLeave } = useHover({
+    duration: 250,
+  });
+
+  React.useEffect(() => {
+    const loadReactionData = async () => {
+      try {
+        const reactionData = await fetchReactionData();
+
+        const transformedData = reactionData.reduce((acc, data) => {
```
tommoor, line 41: This exact logic is in two spots, can we make it a `@computed` getter on the comment model?

app/components/Reactions/Reaction.tsx
```diff
@@ -0,0 +1,173 @@
...
+    if (idx === 0) {
+      return name;
+    } else if (idx === usernames.length - 1) {
+      return `${content} ${t("and")} ${name}`;
+    } else {
+      return `${content}, ${name}`;
+    }
+  }, "");
+
+  return `${joinedUsernames} ${t("reacted with")} :${getEmojiId(emoji)}:`;
```
tommoor, line 81: Hard to provide exact feedback on how to change this, but we can't construct translated sentences from parts like this as every language has different grammatical structures. It needs to be done using variables in the translation as much as possible, e.g.

`t("{{ usernames }} reacted with {{ emoji }}")`

It's okay to have a couple of variants

server/models/Comment.ts
```diff
@@ -51,6 +56,9 @@ class Comment extends ParanoidModel<
   @Column(DataType.JSONB)
   data: ProsemirrorData;
 
+  @Column(DataType.JSONB)
+  reactions: Reaction[] | null;
```
hmacr, line 60: Please note the structure has changed. The array structure makes it simple to construct mobx observables in the frontend `Store` (follows `collection.documentStructure`).
**Proposed**
```
{
  "emoji_a": ["user_id_a", "user_id_b"],
  "emoji_b": ["user_id_b"]
}
```
**Current**
```
[
  {
    "emoji": "emoji_a",
    "userIds": ["user_id_a", "user_id_b"]
  },
  {
    "emoji": "emoji_b",
    "userIds": ["user_id_b"]
  }
]
```

app/menus/CommentMenu.tsx
tommoor: (there is another suggestion related to moving the transform to a computed getter further down)

server/routes/api/comments/comments.ts
tommoor: Feels ok

app/scenes/Document/components/CommentThreadItem.tsx
hmacr: The intention is to have `ReactionList` generalized to render on any context &mdash; comment for now, could be used for document later (same as a generalized `ViewReactionsDialog`).

app/components/Reactions/ViewReactionsDialog.tsx
hmacr: This unsets the `reakit` tab styles with `Outline` styles.

server/routes/api/comments/comments.ts
hmacr: It sure did, thank you!

app/hooks/useHover.ts
hmacr: Agree - makes it clean.

server/routes/api/reactions/reactions.ts
hmacr: I get what you mean here. However we don't have a store for reactions; can't use `fetchAll` :(
Let me implement pagination in the `comment.fetchReactionsData()` on the client..

app/components/Reactions/ReactionList.tsx
hmacr: Current code was doing unnecessary transformations. On refactoring, this data has become an `observable` (we need it to render the tooltip for reactions added through websockets).

app/components/Reactions/Reaction.tsx
hmacr: We could end up with a few variants like
- `t("You reacted with {{ emoji }}", { emoji })`
- `t("{{ username }} reacted with {{ emoji }}", { username, emoji })`
- `t("You and {{ username }} reacted with {{ emoji }}", { username, emoji })`
- `t("{{ first_username }} and {{ second_username }} reacted with {{ emoji }}", { first_username, second_username, emoji })`
- a couple more for three users
- and finally something like `t("You, {{ usernames }}, and {{ rem_count }} others reacted with {{ emoji }}", { usernames,  rem_count, emoji }`

I'm not really sure if this is the right way to handle the `and` conjunction. Should it be added to the translated string like above?

tommoor: "You" can be considered a username, so the `t("You reacted with {{ emoji }}", { emoji })` variant isn't needed. I think that leaves us with 4 variants:

- `t("{{ username }} reacted with {{ emoji }}", { username, emoji })`
- `t("{{ first_username }} and {{ second_username }} reacted with {{ emoji }}", { first_username, second_username, emoji })`
- `t("{{ username }} and {{ count }} others reacted with {{ emoji }}", { username,  count, emoji })`

Note it's important for the variable to be `count` on that last one for the i18n lib to generate the plural and singular variations

hmacr: Oh, I was thinking that we need to translate "You" as well since it's not a username and more of a text that can be localised. But, I suppose the same can be argued for the actual usernames too - can be localised but not advisable to expose PII.

tommoor: Localization happens at build time, so there's no exposing of PII. Take a look at this for an example of what I mean:

https://github.com/outline/outline/blob/b6fc8fb4b111cccd3db7e6d1f7f678281470cf3e/app/scenes/Settings/components/FileOperationListItem.tsx#L117-L122

app/components/Reactions/Reaction.tsx
```diff
@@ -46,44 +46,81 @@ const useTooltipContent = ({
...
+          secondUsername,
+          emoji: `:${getEmojiId(emoji)}:`,
+        }
+      );
+    }
 
-  const otherUsernames = reactedUsers
-    .filter((user) => user.id !== currUser.id)
-    .map((user) => user.name);
+    case 3: {
```
tommoor, line 73: Lets lose this case, find to just be "username and 2 others"

app/components/WebsocketProvider.tsx
```diff
@@ -352,7 +352,12 @@ class WebsocketProvider extends React.Component<Props> {
         comment?.updateReaction({
           type: "add",
           emoji: event.emoji,
-          userId: event.userId,
+          userId: event.user.id,
+        });
+        comment?.updateReactedUser({
```
tommoor, line 357: Why have this separate? Any reason we'd call `updateReaction` but not `updateReactedUser`?

hmacr: Not really.. can be consolidated into a single function 👍

tommoor:
Alrighty, finally got to a UX review – a few things, mostly small:

- [ ] When creating a new comment there is a JS error in the last commit
    image

- [ ] Lets add a 4px gap between the emoji react and "..." button, it's a bit tight:
    image

- [ ] It would be nice if we could add the emoji+ button to the right of any existing reactions once some exist on a comment, this has become a pretty standard pattern, similar to:
    image

- [ ] Can we align the popover with the right side of the button, for some reason it's awkwardly butted up against the edge of the page currently
    image

- [ ] When I open the skin tone picker it sometimes causes the entire sidebar to grow in width, might be solved by fixing the positioning of the popover.
    image

- [ ] The cursor on reactions should be `cursor: var(--pointer);` rather than hardcoded as the hand pointer

hmacr:
Addressed the UX feedback except for the below ones.

> When I open the skin tone picker it sometimes causes the entire sidebar to grow in width

I wasn't able to reproduce this - for safety, I've aligned this to the right edge of the button.

> The cursor on reactions should be `cursor: var(--pointer)` rather than hardcoded as the hand pointer

I don't think it's the case now 🤔 We reset the pointer to the default one for resolved comments (more like showing it's disabled) - [code](https://github.com/outline/outline/pull/7790/files#diff-62c6d7f1448a449dd7bfcce70021425c21fa23b53bec9b278b1818d971951855R150)

app/components/Reactions/ReactionPicker.tsx
```diff
@@ -0,0 +1,178 @@
...
+        <Popover
+          {...popover}
+          ref={contentRef}
+          width={popoverWidth}
+          shrink
+          aria-label={t("Reaction picker")}
+          onClick={(e) => e.stopPropagation()}
+          hideOnClickOutside={false}
+        >
+          <React.Suspense fallback={<PlaceHolder />}>
```
tommoor, line 122: Honestly just `null` as placeholder might be fine, I can't manage to trigger it – even with using 3G network speeds

hmacr: I don't think we can see the placeholder in local &mdash;`EmojiPanel` component gets loaded inside the doc title's icon-picker which is lazy loaded on render. I've added the suspense (with fallback) for the concurrent access cases (for safety). A bit too much you think?

That said, here's how the placeholder looks &mdash; added an ad-hoc delay for the lazy import.

https://github.com/user-attachments/assets/7cd6075e-263b-498a-b225-2b0474deb62a

hmacr:
Looks beautiful!! 👌

tommoor:
Discussed in DM – one more small refactor to remove `reactedUsers` and this should be good for deploy

hmacr:
`reactedUsers` has been removed in favour of using `users` store.
Please note that I fixed a nasty bug in remove_reaction API.. embarrassing 🤦‍♂️

server/routes/api/comments/comments.ts
```diff
@@ -447,7 +447,7 @@ router.post(
     authorize(user, "removeReaction", comment);
 
     const reaction = await Reaction.findOne({
-      where: { emoji, userId: user.id },
```
tommoor: Oof

Issue #7817, opened 2024-10-22
Comments in sidebar jump in during animation


tommoor:
**To Reproduce**
1. Add comments to a document
2. Toggle the sidebar open
3. Comments appear half way through animation (most of the time)

tommoor:
I can no longer reproduce this 🤷

hmacr:
Just had the chance to check the recent changes.. I still see this jank when there are quite a few comments 😞 

https://github.com/user-attachments/assets/65f2eedc-d41a-4d6e-b16a-ceeead7e30b8

PR #7825, opened 2024-10-24
Comments: scroll to most-recent during load and when switching sort setting

### LINKED ISSUE #7818: Ensure comment sidebar is scrolled to bottom on load
When ordered by "Most recent", follow up to https://github.com/outline/outline/pull/7770

infinite-persistence:
## Ticket
Closes #7818

## Notes
- Factored out `readyToDisplay` to use as the dependency since refs can't be one. The effect then covers both the mounting + switching cases.
    - It also ensures `document` can never be null, hence the typescript suppression.
- Due to the `50vh` bottom padding, the last comment appears in the middle of the screen. But I think this is the expected style given the padding.  If not, the scroll position can be adjusted further.
    - In addition, the lack of a scrollbar might make it look like there are less comments, but again, I think this is expected (I got the same feeling when I opened the `/` command popup for the first time).

_p/s: I skipped the pre-PR discussion part given the size of the PR. I'm totally ok with re-doing/scrapping/dropping, etc._

### DIFF
--- a/app/scenes/Document/components/Comments.tsx
+++ b/app/scenes/Document/components/Comments.tsx
@@ -72,11 +72,13 @@ function Comments() {
   };
 
   const handleScroll = () => {
+    const BUFFER_PX = 50;
+
     if (scrollableRef.current) {
       const sh = scrollableRef.current.scrollHeight;
       const st = scrollableRef.current.scrollTop;
       const ch = scrollableRef.current.clientHeight;
-      isAtBottom.current = Math.abs(sh - (st + ch)) <= 1;
+      isAtBottom.current = Math.abs(sh - (st + ch)) <= BUFFER_PX;
 
       if (isAtBottom.current) {
         setShowJumpToRecentBtn(false);
--- a/app/scenes/Document/components/Comments.tsx
+++ b/app/scenes/Document/components/Comments.tsx
@@ -5,6 +5,7 @@ import { useTranslation } from "react-i18next";
 import { useRouteMatch } from "react-router-dom";
 import styled from "styled-components";
 import { ProsemirrorData, UserPreference } from "@shared/types";
+import ButtonSmall from "~/components/ButtonSmall";
 import { useDocumentContext } from "~/components/DocumentContext";
 import Empty from "~/components/Empty";
 import Flex from "~/components/Flex";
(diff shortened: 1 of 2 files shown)


tommoor:
Thanks! We can lose the 50vh bottom padding btw, the aim here is to make it feel a bit more like a standard chat – so it should start at the bottom and stay pinned to the bottom if you add a new comment to the document.

The big problem with this as it stands is the smooth scrolling – the user should not be able to tell that any scrolling occurred as it's quite disorienting.

infinite-persistence:
Made some changes. Full details inside each commit message.


https://github.com/user-attachments/assets/0bc836a9-e2fb-4785-ab37-15843a6c9865

app/scenes/Document/components/Comments.tsx
```diff
@@ -50,19 +56,62 @@ function Comments() {
...
+          scrollToBottom(); // Remain pinned to bottom on new comments
+        } else {
+          setShowJumpToRecentBtn(true);
+        }
+      }
+    }
+    prevThreadCount.current = threads.length;
+  }, [threads.length]);
 
-  if (!document || !isEditorInitialized) {
```
tommoor, line 57: Why did you remove the `document` guard in favor of lots of non-null assertion? Seems objectively worse 🤔

app/scenes/Document/components/Comments.tsx
```diff
@@ -132,10 +186,32 @@ const NoComments = styled(Flex)`
 `;
 
 const Wrapper = styled.div<{ $hasComments: boolean }>`
-  padding-bottom: ${(props) => (props.$hasComments ? "50vh" : "0")};
   height: ${(props) => (props.$hasComments ? "auto" : "100%")};
 `;
 
+const ScrollToRecentSize = "32px";
+
+const ScrollToRecent = styled.div`
```
tommoor, line 194: Could you use the existing `ButtonSmall` component and just make it rounded? Could use the text `New comments` perhaps

tommoor (commented):
Seems to work well! Need to figure out what's causing https://github.com/outline/outline/issues/7817 for it to feel really solid when opening

app/scenes/Document/components/Comments.tsx
infinite-persistence: The null guard was DRY-ed for re-use as an effect dependency, but Typescript for some reason can't figure out that `document` will never be null, resulting in the need to suppress the check everywhere.

Agree that it's more messy, so un-DRY-ed

app/scenes/Document/components/Comments.tsx
infinite-persistence: Converted.

(_and wow, the translation script seems to be able to figure out where to inject the new string in the json :+1:_ )

tommoor (approved):

PR #7827, opened 2024-10-24
Move `editorToolbar` below `header`

### LINKED ISSUE #7826: Code block's copy button interference
Copy button will interfeere with UI when using seperate editing mode:

image

**To Reproduce**
1. Go to a large code block
2. Click into the Codeblok
3. Scroll down
4. See the Copy button "glitch" into the "action bar"

**Expected behavior**
The copy button should align itself underneath said bar.

**Screenshots**

image

- Install: self-hosted
- Version: v0.80.2

 - OS: Arch linux
 - Browser: Firefox
 - Version: 131.0.3 (64-bit)

**Aditional information:**
A coupple of collegues also have seen this happen, luckily it isnt a huge problem just a bit annoying.

infinite-persistence:
Closes #7826

On top of `FloatingToolbar`, this change also affects the `SuggestionMenu`, but the fix makes sense there as well.

### DIFF
--- a/shared/styles/depths.ts
+++ b/shared/styles/depths.ts
@@ -1,8 +1,8 @@
 const depths = {
   toc: 100,
+  editorToolbar: 780,
   header: 800,
   sidebar: 900,
-  editorToolbar: 925,
   mobileSidebar: 930,
   hoverPreview: 950,
   // Note: editor lightbox is z-index 999


tommoor:
I believe the better solution (suggested in the issue) is to make the toolbar position sticky below the height of the header rather than have it become hidden

PR #7830, opened 2024-10-25
fix: wait for shared document to load

hmacr:
This PR fixes a bug where the first load of a publicly shared doc(with sub docs) causes the main content to slide into the view.

By waiting until the first response is available, we ensure the app doesn't have a chance to render without a sidebar.

### DIFF
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -172,7 +172,7 @@ function SharedDocumentScene(props: Props) {
     }
   }
 
-  if (!response) {
+  if (!response?.sharedTree) {
     return <Loading location={props.location} />;
   }
 


app/scenes/Document/Shared.tsx
```diff
@@ -172,7 +172,7 @@ function SharedDocumentScene(props: Props) {
     }
   }
 
-  if (!response) {
+  if (!response?.sharedTree) {
```
tommoor, line 175: In what situation is there a response but not a sharedTree component 🤔

hmacr: It regressed here - on the initial app load, the sharedTree would be undefined until the response is received.
https://github.com/outline/outline/blob/f22bd1d7c8bf25923728f9be41be45bf73315072/app/scenes/Document/Shared.tsx#L132-L134

tommoor: Would it also work to check for `response.document` then? I think that would make more sense if so

tommoor (approved):

PR #7835, opened 2024-10-26
feat: create toast for websocket notification

### LINKED ISSUE #7723: Show toast for applicable realtime notifications
:bulb: For realtime notifications, a neat little addition is to show a toast based on current page context.
eg: A toast if someone comments on the page you're viewing with a link to navigate to it

hmacr:
Closes #7723 

**Follow-on tasks**
- Post #7790 merge, add the reaction events to `WebsocketToasts`.
- Use the `Notification.isRead` and `Notification.isUnread` methods in place of raw `Notification.viewedAt` comparisons.

### DIFF
--- a/shared/types.ts
+++ b/shared/types.ts
@@ -264,6 +264,11 @@ export type CollectionSort = {
   direction: "asc" | "desc";
 };
 
+export enum NotificationSource {
+  Api = "api",
+  Websocket = "websocket",
+}
+
 export enum NotificationEventType {
   PublishDocument = "documents.publish",
   UpdateDocument = "documents.update",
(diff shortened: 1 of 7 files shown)


shared/types.ts
```diff
@@ -264,6 +264,11 @@ export type CollectionSort = {
   direction: "asc" | "desc";
 };
 
+export enum NotificationSource {
```
hmacr, line 267: Co-located with notification related types.

tommoor:
It ends up that https://github.com/outline/outline/pull/7825 included an indicator when there are new comments out of view. 

After thinking some more on this, we should probably continue with that approach (maybe adapt it for the "order in doc" mode) and not add an additional toast.

PR #7838, opened 2024-10-26
fix: don't show comment marks for other users' drafts

### LINKED ISSUE #7470: Incomplete inline comments can appear as "missing" comments to other users
**To Reproduce**
1. Go to a document, select text, and choose "comment"
2. Close the tab
3. Visit the document as another user
4. "ghost" inline comment appears 

**Expected behavior**

The incomplete comment is not visible to other users.

hmacr:
Closes #7470 

Note that this PR provides a forward fix. I don't think it's simple/straight-forward to fix the existing cases - need to load the localstorage in the comment mark instance. I guess the drafts count should be fairly low to ignore them.

### DIFF
--- a/app/editor/index.tsx
+++ b/app/editor/index.tsx
@@ -690,7 +690,10 @@ export class Editor extends React.PureComponent<
    * @param commentId The id of the comment to remove
    * @param attrs The attributes to update
    */
-  public updateComment = (commentId: string, attrs: { resolved: boolean }) => {
+  public updateComment = (
+    commentId: string,
+    attrs: { resolved?: boolean; draft?: boolean }
+  ) => {
     const { state, dispatch } = this.view;
     const tr = state.tr;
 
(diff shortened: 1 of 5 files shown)


tommoor:
I'm ok with leaving the existing data, it should be minimal and easy to replace. This approach is sensible and probably what I'd do if I was starting from scratch today

hmacr:
Note that the toolbar will have an active comment item for inline draft comment - at least it's hidden :) Fixing this has an unintended consequence of the second user _stealing_ the existing comment mark.
Screenshot 2024-10-28 at 2 22 51 AM


I've created #7847 that explores a solution to allow multiple users to create inline draft comments - this will fix the above bug.

tommoor (approved):
I need to triple check before merging, but given this follows the patterns of resolved comments that work well I think this is good 👍 

Trying to remember why I didn't do it this way in the first place

PR #7840, opened 2024-10-26
Set reply-to email address to actor when permissions allow

tommoor:
We get a lot of folks replying to outgoing notifications, it would be better if these are directed to the originating user when the receiving user has access to their email address.

In the case of guests the standard reply-to address is kept, so as not to leak the email.


server/emails/templates/CommentCreatedEmail.tsx
```diff
@@ -119,6 +120,15 @@ export default class CommentCreatedEmail extends BaseEmail<
     return actorName;
   }
 
+  protected replyTo({ notification }: Props) {
+    if (notification?.user && notification.actor?.email) {
+      if (can(notification.user, "readEmail", notification.actor)) {
```
hmacr, line 125: "readEmail" permission does not allow viewers - is it intentional?
https://github.com/outline/outline/blob/7bdae0cbda21d01a431b42a3dc43344e662c1c8a/server/policies/user.ts#L41-L48

server/emails/templates/BaseEmail.tsx
```diff
@@ -10,7 +10,10 @@ import { NotificationMetadata } from "@server/types";
 import { getEmailMessageId } from "@server/utils/emails";
 
 export interface EmailProps {
+  /** The email address being sent to. */
   to: string | null;
+  /** The notification that triggered the email, if any. */
+  notification?: Notification;
 }
 
 export default abstract class BaseEmail<
```
hmacr, line 19: Could include Invite related emails?

server/emails/templates/CommentCreatedEmail.tsx
tommoor: It is… but also, it might not be neccessary to restrict to just members. A bit undecided

server/emails/templates/BaseEmail.tsx
tommoor: Wouldn't hurt!

PR #7844, opened 2024-10-27
Make from address for authentication related emails unguessable

tommoor:
This is a proactive change to prevent potential attacks in the vein of the recently discovered Zendesk vulnerability here:
https://gist.github.com/hackermondev/68ec8ed145fcee49d2f5e2b9d2cf2e52

By ensuring emails related to login/auth are sent from a unique email address each time we can protect against guessing the email in a third-party system

### DIFF
--- a/server/emails/templates/BaseEmail.tsx
+++ b/server/emails/templates/BaseEmail.tsx
@@ -1,6 +1,10 @@
+import addressparser from "addressparser";
 import Bull from "bull";
+import invariant from "invariant";
+import randomstring from "randomstring";
 import * as React from "react";
 import mailer from "@server/emails/mailer";
+import env from "@server/env";
 import Logger from "@server/logging/Logger";
 import Metrics from "@server/logging/Metrics";
 import Notification from "@server/models/Notification";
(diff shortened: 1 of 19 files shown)


server/emails/templates/BaseEmail.tsx
```diff
@@ -9,6 +13,13 @@ import { TaskPriority } from "@server/queues/tasks/BaseTask";
 import { NotificationMetadata } from "@server/types";
 import { getEmailMessageId } from "@server/utils/emails";
 
+export enum EmailMessageCategory {
+  Authentication = "authentication",
```
tommoor, line 17: Currently only `Authentication` is used for anything, but while we're here might as well categorize all emails for safety

server/emails/templates/BaseEmail.tsx
```diff
@@ -148,6 +162,31 @@ export default abstract class BaseEmail<
...
+      env.SMTP_FROM_EMAIL,
+      "SMTP_FROM_EMAIL is required to send emails"
+    );
+
+    const parsedFrom = addressparser(env.SMTP_FROM_EMAIL)[0];
+
+    if (this.category === EmailMessageCategory.Authentication) {
+      const domain = parsedFrom.address.split("@")[1];
+      return {
+        name: parsedFrom.name,
```
hmacr, line 176: Would it better to retain the behaviour of overriding `fromName` here?

tommoor: I guess so, although I don't think that combination will ever be needed it seems safer

hmacr: Ok 👍

PR #7865, opened 2024-10-29
Update mermaid to V11

Shuttleu:
Update Mermaid to V11 to add latest changes and graphs

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "markdown-it": "^13.0.2",
     "markdown-it-container": "^3.0.0",
     "markdown-it-emoji": "^2.0.0",
-    "mermaid": "9.3.0",
+    "mermaid": "11.3.0",
     "mime-types": "^2.1.35",
     "mobx": "^4.15.4",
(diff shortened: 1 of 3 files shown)


devinnovqube:
> The bundle size increase reveals why this upgrade has not been made - this single dependency upgrade increases the size of the entire software by 20%, which is unacceptable

Some of the new mermaid features are indeed useful. However, 20% of the size for a few features seems quite substantial. Perhaps we could find a way to allow users to choose between version 9 and version 11 when deploying the application. 

Mermaid is regularly improving, and this could certainly become an issue in the future.

tommoor:
I added additional fixes and merged here https://github.com/outline/outline/pull/7964

owanio1992:
@tommoor 
it's happy to see upgrade mermaid version  
I have a little question here 
does this cdn link don't need upgrade ?
https://github.com/outline/outline/blob/main/server/models/helpers/ProsemirrorHelper.tsx#L561

tommoor:
Good call

PR #7894, opened 2024-11-05
Fix header key for react-table v7

infinite-persistence:
Removes annoying warning + technically more correct.

(when opening `settings/shares`)

---

`header.id` does not exist in v7 (it does in v8). `@types/react-table` lied.

Stashed v7 documentation:
https://github.com/TanStack/table/blob/v7/docs/src/pages/docs/api/useTable.md#headergroup-properties

### DIFF
--- a/app/components/Table.tsx
+++ b/app/components/Table.tsx
@@ -120,30 +120,33 @@ function Table({
       <Anchor ref={topRef} />
       <InnerTable {...getTableProps()}>
         <thead>
-          {headerGroups.map((headerGroup) => (
-            <tr {...headerGroup.getHeaderGroupProps()} key={headerGroup.id}>
-              {headerGroup.headers.map((column) => (
-                <Head
-                  {...column.getHeaderProps(column.getSortByToggleProps())}
-                  key={column.id}
-                >
-                  <SortWrapper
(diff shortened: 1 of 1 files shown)


app/components/Table.tsx
```diff
@@ -121,7 +121,10 @@ function Table({
       <InnerTable {...getTableProps()}>
         <thead>
           {headerGroups.map((headerGroup) => (
-            <tr {...headerGroup.getHeaderGroupProps()} key={headerGroup.id}>
+            <tr
+              {...headerGroup.getHeaderGroupProps()}
+              key={headerGroup.getHeaderGroupProps().key}
```
tommoor, line 126: This should be fine just with the spread then, guessing without the explicit `key` there are eslint errors?

tommoor: Neatest is probably to assign to a local variable and use that.

```
const { key, ...groupProps } = headerGroup.getHeaderGroupProps();
```
?

infinite-persistence: Updated with local variable. Avoided double spread in both the destructing and in `tr`

tommoor (approved):

PR #7904, opened 2024-11-06
fix: heading disclosure transform

### LINKED ISSUE #7899: Heading disclosure no longer rotates when collapsed
The arrow should rotate 90deg and stay visible when the content beneath the header is collapsed, this seems to have regressed at some point – likely a small thing.

hmacr:
Closes #7899


tommoor:
I guarantee this was copilot 😠

tommoor (approved):

PR #7912, opened 2024-11-07
feat: Adds support for importing CSV files

tommoor:
I wasn't planning on adding this, but it's a quick win and people keep trying to import these files – so 🤷


hmacr:
Looks good 💯

PR #7930, opened 2024-11-12
Add Pinterest embed

fuekiin:
As discussed [here](https://github.com/outline/outline/discussions/7923).

Supports Pinterest Boards and Profiles. Works with dark mode. 

Looks like this:

Bildschirmfoto 2024-11-12 um 12 06 26

Bildschirmfoto 2024-11-12 um 12 06 38

Bildschirmfoto 2024-11-12 um 12 06 51

Bildschirmfoto 2024-11-12 um 12 06 59


tommoor:
This looks good, will test it out locally soon – I think we're at the point where these hosted embeds can probably be refactored to reduce some of the repetition

tommoor (approved):

Issue #7916, opened 2024-11-08
Templates shared at workspace level are not being displayed to users


notfounnd:
Issue as requested in https://github.com/outline/outline/discussions/7915 .

**To Reproduce**

1. As admin, create a template.
2. Share this template at workspace level.
3. As user, create new document.
4. Try use new template previously created.

- User not able to use the doc template shared at workspace level.

**Expected behavior**

- All users should able to use any doc template shared at workspace level.

**Screenshots**

User view:
![image](https://github.com/user-attachments/assets/11fb47ce-8d4f-4b74-8dea-86eb32ab8f3a)

Template config (OPD not able in user view):
![image](https://github.com/user-attachments/assets/9c203dc0-05f5-4be7-b7b8-970cdfa78e3c)

**Outline (please complete the following information):**
- Install: Self hosted
- Version: v0.80.2

**Desktop (please complete the following information):**
 - OS: Win10 / Win11
 - Browser: Chrome
 - Version: 130.0.6723.93

notfounnd:
@hmacr yes, I also checked it by performing logout/login to verify behaviors about these issue. It persist too.

Another thing I also noticed is that after sharing the template via group and remove the document from the group, it is in view for those who already had access.

I deleted the template that I used for testing it kept appearing to the user.

In this process of deletion, only updating with CTRL+SHIFT+F5 the user no longer has a view of the template that I used in the tests.

notfounnd:
I believe we have pros and cons in both use contexts.

But I agree that if the template is shared at the workspace level, the document would not need additional actions to be accessible by all users.

Today I kept shared so only for people to use. After resolution of the issue here, I intend to follow the Outline guide use.

tommoor:
closing as tenatively fixed, we can reopen if it still proves to be a problem after these fixes ☝️ . At that point we'll probably have to look in the database to see what's happening

PR #7937, opened 2024-11-12
hide share option for templates

hmacr:
Towards #7916.

Split the `manageUser` and `duplicate` policies for documents.
Right now, draft sharing is being allowed; so kept as-is. Not completely sure if this needs to be changed.


tommoor:
> Right now, draft sharing is being allowed; so kept as-is.

Yes, want to retain this it's very useful!

tommoor (approved):

PR #7941, opened 2024-11-13
Comment: collapse long replies

### LINKED ISSUE #5079: Collapse long threads (x more)
On long comment threads there should be a break in the middle with "X more comments…" that you can click to expand.

infinite-persistence:
## Ticket
Closes #5079

## Review
- For the case of RTL, I followed how "Reply" is implemented (assumed that is the desired way). If it needs to be re-aligned, it can be fixed together with "Reply" later.
- The threshold number can be moved to `constants.ts` if we don't want to pollute the props.

### DIFF
--- a/app/scenes/Document/components/CommentThread.tsx
+++ b/app/scenes/Document/components/CommentThread.tsx
@@ -10,8 +10,9 @@ import { s } from "@shared/styles";
 import { ProsemirrorData } from "@shared/types";
 import Comment from "~/models/Comment";
 import Document from "~/models/Document";
-import { Avatar } from "~/components/Avatar";
+import { Avatar, AvatarSize } from "~/components/Avatar";
 import { useDocumentContext } from "~/components/DocumentContext";
+import Facepile from "~/components/Facepile";
 import Fade from "~/components/Fade";
 import Flex from "~/components/Flex";
 import { ResizingHeightContainer } from "~/components/ResizingHeightContainer";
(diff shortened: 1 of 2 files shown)


infinite-persistence:
Ah wait, the avatar-chaining is incorrect. Be right back with a fix...

infinite-persistence:
Fixed, ready for review

app/scenes/Document/components/CommentThread.tsx
```diff
@@ -192,6 +214,19 @@ function CommentThread({
       onClick={handleClickThread}
     >
       {commentsInThread.map((comment, index) => {
+        if (collapse !== null) {
+          if (index === collapse.begin) {
+            const count = collapse.final - collapse.begin + 1;
+            return (
+              <ShowMore onClick={handleClickExpand} key="show-more">
+                {t("Show {{ count }} reply", { count })}
```
infinite-persistence, line 222: Hm ... not show why it isn't pluralized. Doc says just use `count` as the variable, right?

tommoor: That's right, but the english translations need adjusting manually in this case.

tommoor:
All seems good, I think the styling will need tweaking though – I want to keep the aesthetic that each thread is a "card" and the way this looks right now it feels like the card has a big gap in it with the background showing through. I can probably provide some designs in a few datys

infinite-persistence:
Updated design

image

app/scenes/Document/components/CommentThread.tsx
infinite-persistence: Mybad. I just discovered `build:i18n`. It works now :smile:

tommoor:
IMG_1692

How about like this? Having the `<Facepile />` on there is quite useful as you get a preview of who is hidden. It is similar to the first one but the difference in background would be much more subtle.

tommoor:
Nice, thanks! I might want to tweak the background color a bit but probably easiest to do that post-merge

tommoor (approved):

PR #7943, opened 2024-11-13
`documents.viewed`: include direct shares

### LINKED ISSUE #7849: Shared docs do not show up in the "Recently viewed" home tab.
**To Reproduce**
1. Open a document shared directly to you (or) a group you're part of (wait for more than 3 seconds).
2. Go back to the home page - Notice that the document is listed in the "Recently viewed" tab.
3. Refresh the page.
4. Notice that the document does not show up in the "Recently viewed" tab (_bug_).

**Expected behavior**
Shared document shows up in the "Recently viewed" home tab.

infinite-persistence:
## Ticket
Closes #7849

## Root-cause
The endpoint is filtered to active collections only (perhaps intentional?)

## Walkthrough
- Originally attempted a single-call solution (see https://github.com/infinite-persistence/outline/commit/6703c95c0edc07575933b782063d96527338161f), but I couldn't get it to work for all cases. It's either a Sequealize or skills issue.
- Moved to a `.filter` solution instead.
    - Given that we've reduced 1 loop from `user.collectionIds()`, the amount of work remain roughly the same as before?

## Question
- Do we need to retain the original behavior of the endpoint for existing clients? If so, we can add a param to include direct and group shares instead.

### DIFF
--- a/server/routes/api/documents/documents.ts
+++ b/server/routes/api/documents/documents.ts
@@ -11,12 +11,7 @@ import uniq from "lodash/uniq";
 import mime from "mime-types";
 import { Op, ScopeOptions, Sequelize, WhereOptions } from "sequelize";
 import { v4 as uuidv4 } from "uuid";
-import {
-  CollectionPermission,
-  StatusFilter,
-  TeamPreference,
-  UserRole,
-} from "@shared/types";
+import { StatusFilter, TeamPreference, UserRole } from "@shared/types";
 import { subtractDate } from "@shared/utils/date";
 import slugify from "@shared/utils/slugify";
 import documentCreator from "@server/commands/documentCreator";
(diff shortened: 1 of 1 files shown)


tommoor:
> Do we need to retain the original behavior of the endpoint for existing clients? If so, we can add a param to include direct and group shares instead

No no, it's fine to change.

> Moved to a .filter solution instead.

Thanks for taking this on. I do think this solution is liable to be problematic in the future though with all these manual permission checks. Because losing access to things is quite rare maybe one way to achieve would be to just query the views, get the documentIds and then query the documents with those ID's using the same methodology as eg `documents.list`. 

The downside is that sometimes the request will return less than the defined limit 🤔

infinite-persistence:
Second attempt

Given that we are ok with grabbing everything on the js side (previously, I felt this defeated the purpose of pagination), we can use that call to circumvent the `offset/limit` with Sequelize -- by doing revocation-check on one call, and pagination on the other. This should also address the "return less than requested" problem since it is "pre-filtered".

In the second call, I changed to "all memberships", assuming that it's faster without having to check for userId again.

tommoor:
> Given that we are ok with grabbing everything on the js side 

I'm actually not really following here what this is referring to

infinite-persistence:
Oh, it's just referencing...
> maybe one way to achieve would be to just query the views, get the documentIds` and then query the documents with those ID's

...where I interpreter that as "_1 `findAll` to get the documentIds, and another to get the docs_". The first one kind of already grabbed the full set instead of paginated set (?).

tommoor:
Hmm no, I mean to get the views paginated, map over the documentIds and then query those with the same type of query in `documents.list` that accounts for all the current permissions. In this case sometimes there will be less results as the viewed document may no longer be accessible

infinite-persistence:
> All of this to happen on the server

Right, this feels more normal. I over-assumed :sweat_smile: 

In that case, I'm closing it as I'm not sure how to solve the potential reduced results without pre-filtering.
Thought of just showing the history as-is and let the link lead to a dead end (it is a history list after all), but I don't think this will be what you want.

PR #7951, opened 2024-11-14
fix: restore workspace templates

hmacr:
This PR fixes a bug wherein workspace templates could not be restored from trash.


tommoor (approved):
At this point im convinced having templates mixed with documents was a huge mistake 😔

hmacr:
I think we've crossed almost all the complex use-cases around them.
We can always refactor later 🙂

PR #7952, opened 2024-11-14
Improvements around templates

hmacr:
This PR provides the following improvements around templates.

- Hide "New from template" menu item for the following cases.
  - User does not have access to create document in a collection (An user with view permission to a collection).
  - Template is deleted.
  - Template is a draft.
- Conditionally show "New from template" action in the command bar.
  - When user does not have create document permission in a collection, the action is hidden.
  - When a collection is active, the templates in this collection + workspace templates are listed.
  - When no collection is active, only the workspace templates are listed.
- Viewing a deleted template will now stay in the "Trash" page itself. Earlier, it was navigating to the templates page.

### DIFF
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -131,11 +131,30 @@ export const createDocumentFromTemplate = createAction({
   section: DocumentSection,
   icon: <NewDocumentIcon />,
   keywords: "create",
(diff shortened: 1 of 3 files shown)


hmacr:
Just noticed that the command bar needs some changes to hide the "New from template" action - will convert it to draft for now.

tommoor (approved):

PR #7956, opened 2024-11-15
fix: auto-scroll sidebar to show active document

### LINKED ISSUE #7939: Sidebar does not always scroll automatically to show focused document
For me, the sidebar does not scroll automatically at all

_Originally posted by @matthew-ant in https://github.com/outline/outline/discussions/7931#discussioncomment-11227334_

hmacr:
Closes #7939 

Currently, starred documents and collections are not focused when they are opened.
As part of this PR, stars will now be prioritised for focus over the actual document / collection.


app/components/Sidebar/components/SidebarContext.ts
```diff
@@ -1,9 +1,20 @@
 import * as React from "react";
 
-export type SidebarContextType = "collections" | "starred" | string | undefined;
+export type SidebarContextType =
```
hmacr, line 3: Refactored for type-safety

app/models/Document.ts
```diff
@@ -429,6 +429,20 @@ export default class Document extends ArchivableModel {
     return this.title || i18n.t("Untitled");
   }
 
+  @computed
+  get membershipType(): "collection" | "direct" | "group" {
```
hmacr, line 433: Rename `direct` to `document`? Or something different maybe..?

app/components/Sidebar/components/SidebarContext.ts
```diff
@@ -1,9 +1,20 @@
...
+  | "collections"
+  | "shared"
+  | `group-${string}`
```
tommoor, line 6: `group-${string}`

I had no idea you can do this… are you sure?

app/models/Document.ts
tommoor: `document` is good

app/components/Sidebar/components/SidebarContext.ts
hmacr: Yes, it works!

tommoor: mind blown

app/components/DocumentListItem.tsx
```diff
@@ -78,6 +83,23 @@ function DocumentListItem(
     !!document.title.toLowerCase().includes(highlight.toLowerCase());
   const canStar = !document.isArchived && !document.isTemplate;
 
+  const membershipType = document.membershipType;
```
tommoor, line 86: This logic leads to some off behavior, for example if you are viewing a collection which contains a starred document and then click on that document it now jumps the sidebar to the top to focus it in the starred section.

We can probably lose this and `membershipType` but retain the rest of the fix?

hmacr: I'm afraid we need the `membershipType` so that shared docs to an user and within a group can be focused correctly in the sidebar.

re: stars: it was intentional to prefer stars over the actual document - It looks nice from "Home" page but it's unpleasant from "Collection" page 😞 I can think of a couple solutions.
1. Lose the whole "stars" priority? This will mean the stars won't be focused when opened from the document list in home / collections.
2. Focus the stars only if it's opened from Home?

tommoor: The second option seems ideal from a UX point of view, you're right... how to prevent this becoming overly complex

hmacr: We can listen to `history` events and conditionally update the location state based on the context in which the navigation happens. I think this helps centralize the logic, although it could become a bit complex to account for various navigations, e.g shared -> starred, home -> starred, shared -> shared, etc.

tommoor: I think `DocumentListItem` can be aware of where it's being rendered, either through a prop or a provider/context – we only need to know home for now so lets not over complicate.

hmacr: That was my initial thought, but there are quite a few cases which we can cover to provide a better UX. e.g. Navigate within the starred context when there's a link from a starred document to another starred document - This happens inside editor event handler, let me see if it's possible to cover them with the current solution 🤔 .

hmacr: Just updated the PR by introducing a function `determineSidebarContext` that uses the document and the current sidebar context to determine the next one.
Please note that most of the changes in this PR is forwarding the location state in case of any in-doc history updates.

app/scenes/Document/index.tsx
```diff
@@ -32,6 +35,16 @@ export default function DocumentScene(props: Props) {
 
   React.useEffect(() => () => ui.clearActiveDocument(), [ui]);
 
+  React.useEffect(() => {
+    // When opening a document directly on app load, sidebarContext will not be set.
```
tommoor, line 39: Oof, this is going to result in another render on every fresh document load :/

hmacr: When we navigate between documents (both fresh and cached), this won't re-render since the location state will have sidebar context..
I think we'll have an additional render only when we navigate from a non-collection and non-document space 🤔

app/actions/definitions/comments.tsx
```diff
@@ -55,7 +55,10 @@ export const resolveCommentFactory = ({
 
       history.replace({
         ...history.location,
-        state: null,
+        state: {
+          ...(history.location.state as Record<string, unknown>),
```
tommoor, line 59: I'm a bit worried about spreading the entire state like this, perhaps we should be picking just the `sidebarContext`? I don't want to have other random state following between navigations 🤔

app/components/Sidebar/components/StarredLink.tsx
```diff
@@ -40,25 +40,35 @@ function StarredLink({ star }: Props) {
...
-      star.documentId === ui.activeDocumentId &&
-      sidebarContext === locationSidebarContext
-    ) {
-      setExpanded(true);
-    } else if (
-      star.collectionId === ui.activeCollectionId &&
-      sidebarContext === locationSidebarContext
-    ) {
-      setExpanded(true);
+    if (starType === "document") {
```
tommoor, line 55: I don't really know that this addition is neccessary?

app/components/Sidebar/components/StarredLink.tsx
```diff
@@ -40,25 +40,35 @@ function StarredLink({ star }: Props) {
   const { documentId, collectionId } = star;
   const collection = collections.get(collectionId);
   const locationSidebarContext = useLocationState();
-  const sidebarContext = useSidebarContext();
+  const starType = star.documentId ? "document" : "collection";
+  const sidebarContext = starredSidebarContext(
+    starType === "document" ? star.documentId! : star.collectionId
```
tommoor, line 45:
```suggestion
    star.documentId ?? star.collectionId
```

?

app/scenes/Document/components/CommentThread.tsx
```diff
@@ -101,7 +101,10 @@ function CommentThread({
       history.replace({
         search: location.search,
         pathname: location.pathname,
-        state: { commentId: undefined },
+        state: {
+          ...(location.state as Record<string, unknown>),
```
tommoor, line 105: Similar re spreading everything

app/components/Sidebar/components/StarredLink.tsx
hmacr: Yes, that works. Added the `starType` for readability - will remove it.

app/components/Sidebar/components/StarredLink.tsx
hmacr: Not really, added for readability - will remove it.

app/actions/definitions/comments.tsx
hmacr: Thought of being extra safe and send everythin, but you're right - there could be unnecessary state we may not be interested in that lead to bugs.

app/scenes/Document/index.tsx
tommoor: Feels okay in use... maybe we'll get away with it

tommoor (approved):

tommoor (approved):

PR #7959, opened 2024-11-15
show resolved view when a resolved comment is opened from notif email

### LINKED ISSUE #7948: Links to resolved threads
If a notification email contains a link to a thread that is closed, opening the link will not show the thread, only the still open threads, greyed out.

**Expected behavior**
The comments panel should show the resolved threads in this case, highlighting the one that is mentioned in the link

hmacr:
Closes #7948

### DIFF
--- a/app/scenes/Document/components/Editor.tsx
+++ b/app/scenes/Document/components/Editor.tsx
@@ -27,6 +27,7 @@ import useCurrentTeam from "~/hooks/useCurrentTeam";
 import useCurrentUser from "~/hooks/useCurrentUser";
 import useFocusedComment from "~/hooks/useFocusedComment";
 import usePolicy from "~/hooks/usePolicy";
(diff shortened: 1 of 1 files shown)


app/scenes/Document/components/Editor.tsx
```diff
@@ -103,9 +103,14 @@ function DocumentEditor(props: Props, ref: React.RefObject<any>) {
 
   React.useEffect(() => {
     if (focusedComment) {
+      history.replace({
```
tommoor, line 106: I think this will result in a double render when focusing a comment always. We should check if we're already on the correct route before calling `replace`?

hmacr: Ah yes.. It affects all types of comment focus (I was just thinking of the email flow)..

tommoor (approved):

PR #7968, opened 2024-11-17
Fix spacing on invite box

ghiculescu:
Fix the spacing just before the "As an admin...":

image

I didn't test my changes, just made them through the github UI, so not sure if they will work (:


tommoor:
Hi, thanks – this particular fix doesn't work as it introduces a n extra space _inside_ the translated string. I could take over the fix given it's small or you could address, lmk

ghiculescu:
Thanks @tommoor

PR #7985, opened 2024-11-19
feat: Audio component support

l1n:
Similar to the Video component, but without resizing.


tommoor:
👍 down to merge something like this, perhaps both Audio and Video can inherit from a `Media` node so that there isn't so much copy and paste?

We'll need to add handling so that the correct node is created when the mimetype begins with `audio/` and add an item to the block menu to easily insert these too

tommoor:
Closing as abandoned, feel free to reopen if you decide to take this on.

PR #7996, opened 2024-11-20
fix: Rare loop of storage events between tabs causing flickering UI

tommoor:
It is possible to rapidly toggle some UI elements faster than these changes can be communicated between tabs via the `"storage"` event. When this happens it can cause an infinite loop that can only be stopped by closing a tab.

### DIFF
--- a/app/stores/UiStore.ts
+++ b/app/stores/UiStore.ts
@@ -1,4 +1,4 @@
-import { action, autorun, computed, observable } from "mobx";
+import { action, computed, observable } from "mobx";
 import { flushSync } from "react-dom";
 import { light as defaultTheme } from "@shared/styles/theme";
 import Storage from "@shared/utils/Storage";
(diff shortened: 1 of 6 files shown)


app/stores/UiStore.ts
```diff
@@ -324,6 +307,10 @@ class UiStore {
       theme: this.theme,
     };
   }
+
+  private persist = () => {
+    Storage.set(UI_STORE, this.asJson);
```
tommoor, line 312: Overall this solution is pretty dated, but we'll keep it for now for backwards compat

app/stores/UiStore.ts
```diff
@@ -134,10 +135,6 @@ class UiStore {
         this.tocVisible = newData.tocVisible;
       }
     });
-
-    autorun(() => {
```
tommoor, line 138: Removing the autorun means we won't write to storage unless the user _in this tab_ interacts.

PR #8009, opened 2024-11-23
Check flag emoji is supported

hmacr:
Related #8006 

The logic is ported from [emoji-mart](https://github.com/missive/emoji-mart/blob/16978d04a766eec6455e2e8bb21cd8dc0b3c7436/packages/emoji-mart/src/helpers/native-support.ts#L45).

This has been tested in mac, ubuntu and windows platform.

### DIFF
--- a/shared/utils/emoji.ts
+++ b/shared/utils/emoji.ts
@@ -5,29 +5,98 @@ import FuzzySearch from "fuzzy-search";
 import capitalize from "lodash/capitalize";
 import sortBy from "lodash/sortBy";
 import { Emoji, EmojiCategory, EmojiSkinTone, EmojiVariants } from "../types";
-import { isMac } from "./browser";
-
-const isMacEnv = isMac();
 
 init({ data: RawData });
 
 // Data has the pre-processed "search" terms.
 const TypedData = Data as EmojiMartData;
 
+// Slightly modified version of https://github.com/koala-interactive/is-emoji-supported/blob/master/src/is-emoji-supported.ts
+const isFlagEmojiSupported = (): boolean => {
+  const emoji = "🇺🇸";
+  let ctx = null;
+  try {
+    ctx = document
+      .createElement("canvas")
+      .getContext("2d", { willReadFrequently: true });
+    // eslint-disable-next-line no-empty
+  } catch {}
+
+  if (!ctx) {
+    return false;
+  }
+
+  const CANVAS_HEIGHT = 25;
+  const CANVAS_WIDTH = 20;
+  const textSize = Math.floor(CANVAS_HEIGHT / 2);
+
+  // Initialize convas context
+  ctx.font = textSize + "px Arial, Sans-Serif";
+  ctx.textBaseline = "top";
+  ctx.canvas.width = CANVAS_WIDTH * 2;
+  ctx.canvas.height = CANVAS_HEIGHT;
+
+  // Draw in red on the left
+  ctx.fillStyle = "#FF0000";
+  ctx.fillText(emoji, 0, 22);
+
+  // Draw in blue on right
+  ctx.fillStyle = "#0000FF";
+  ctx.fillText(emoji, CANVAS_WIDTH, 22);
+
+  const a = ctx.getImageData(0, 0, CANVAS_WIDTH, CANVAS_HEIGHT).data;
+  const count = a.length;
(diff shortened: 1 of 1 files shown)


wouteroosterveld:
I did some testing of PR #8006 and PR #8009 in comparison to the current main branch (5102cfe8eb7f5edd21bd4104e5bfd73f00f0be1d):

Platform | Browser | With Filter (5102cfe8eb7f5edd21bd4104e5bfd73f00f0be1d) | (No filter) PR #8006 | (With check flag) PR #8009
-- | -- | -- | -- | --
Debian Linux 12.8 | Firefox 132.0.2 | Cannot add flags - Can view flags | OK | OK
Debian Linux 12.8 | Chromium 130 | Cannot add flags - Can view flags | OK | OK
Mac Sonoma 14.4.1 | Firefox 131.0.3 | OK | OK | OK
Mac Sonoma 14.4.1 | Chrome 131 | OK | OK | OK
Windows 11 Pro 24H2 | Firefox 132 | Cannot add flags - Can view flags | OK | OK
Windows 11 Pro 24H2 | Chromium 133 | Cannot add flags - Existing flags substituted with country code | Cannot add flags - Existing flags substituted with country code | Cannot add flags - Existing flags substituted with country code

I did not find a functional difference between #8006 and #8009 in the tests above.

tommoor:
Thanks for testing across platforms 🙏 

That last Windows case where they are substituted with a country code is the one we're trying to avoid and the code here should be covering – so this would suggest that it isn't working correctly.

tommoor:
> Is the expected behaviour that the underlying emoji code is in the document, but nothing is shown in Chromium on Windows?

No, the expected behavior is as you are seeing – it will display as best as the platform can if the emoji is in a document or comment.

This code should be purely to filter Flags out of the emoji _picker_ which you can access through a comment reaction or choosing an icon for collections/documents.

wouteroosterveld:
> > Is the expected behaviour that the underlying emoji code is in the document, but nothing is shown in Chromium on Windows?
> 
> No, the expected behavior is as you are seeing – it will display as best as the platform can if the emoji is in a document or comment.
> 
> This code should be purely to filter Flags out of the emoji _picker_ which you can access through a comment reaction or choosing an icon for collections/documents.

On Chromium on Windows I _cannot_ pick a flag using #8009, but I _can_ pick a flag using Firefox on Windows.

wouteroosterveld:
> > Is the expected behaviour that the underlying emoji code is in the document, but nothing is shown in Chromium on Windows?
> 
> No, the expected behavior is as you are seeing – it will display as best as the platform can if the emoji is in a document or comment.
> 
> This code should be purely to filter Flags out of the emoji _picker_ which you can access through a comment reaction or choosing an icon for collections/documents.

Even with the original code, just with the filter removed (#8006), Chromium on Windows does not show the country flags in the emoji picker. Adding an emoji flag in Firefox results in a country code in Chrome/Windows, same as #8009.

tommoor:
> Even with the original code, just with the filter removed (https://github.com/outline/outline/pull/8006), Chromium on Windows does not show the country flags in the emoji picker.

Well they would show up as country codes instead, which is the problem. Screenshots of the flags section might be usefiul

hmacr:
Just to pitch in - As mentioned by @tommoor, the filter is to remove the flags section from the _emoji picker_ so that the user cannot add any flag emojis. The existing flag emojis, if any, would be displayed as such in case the platform/browser supports it, otherwise it'd fallback to the country code.

The existing behaviour and this PR addresses the emoji picker side of things _only_. Currently we show flag emojis in Mac because we're sure that Mac as a platform supports them (Picker won't have them in non-mac platforms).
But it looks like Linux (all variants?) supports it, and firefox has its own built-in support to cover all platforms. So, it may not be ideal to just filter them based on platform (or) browser.

I came across the solution implemented in this PR in emoji-mart which was replaced as part of #7038.

wouteroosterveld:
This is on #8009.

Screenshot 1: typing `:france` in Firefox on Windows. The emoji picker shows the country flag on top of the selection list:
Screenshot_20241124_110055


Screenshot 2: typing :france in Chromium on Windows. The emoji picker does not show the country flag on top of the selection list. The baguette is an interesting suggestion though.
Screenshot_20241124_105959

tommoor (approved):

PR #8010, opened 2024-11-23
fix: Flash of styles when printing dark mode

### LINKED ISSUE #8005: Theme toggles in a weird way when printing
- [ ] If printing from context menu, the theme changes to light and burns retinas.
- [ ] If printing from browser (Ctrl + P), theme stays correct but there is flashing, which can be repeated by toggling print settings.

https://github.com/user-attachments/assets/82c1dd8a-8525-4a5d-8294-81156b470a12

tommoor:
The background transitions are no longer needed since we're now using the `startTransition` API to switch between themes. Note: This does not fix the page becoming white when using `window.print` from the contenxt menu, that seems like a browser issue as we don't have any control beyond calling print.

closes #8005

### DIFF
--- a/app/components/Actions.ts
+++ b/app/components/Actions.ts
@@ -31,7 +31,6 @@ const Actions = styled(Flex)`
   left: 0;
   border-radius: 3px;
   background: ${s("background")};
-  transition: ${s("backgroundTransition")};
   padding: 12px;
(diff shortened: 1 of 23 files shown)


tommoor:
cc @infinite-persistence

PR #8030, opened 2024-11-27
Allow user to select doc-copy destination

### LINKED ISSUE #7193: Allow choosing location when duplicating document
Currently when duplicating it just places the document next to the existing automatically, this is often not desired – it would be better to show the file explorer and allow choosing where to duplicate to.

infinite-persistence:
Closes #7193

See commit messages for concerns.


tommoor:
Functionally this works well, but the dialog is not quite there – the title below the explorer looks weird and the explorer itself is not aligned, having a bunch of extra padding on the left and the right for some reason.

image

I wonder if it might be best to just drop the existing duplicate dialog altogether and start fresh off the "Move" dialog, it includes some niceties like the explainer text in the footer that is reassuring and the layout is very neat. If we were to do that and just change "Move" to "Copy" we could drop the ability to change the title here, it's easy enough to change the title in the resulting doc? 

lmk your thoughts

infinite-persistence:
> having a bunch of extra padding on the left and the right for some reason
Not intentional, just missed out few more overrides. But agree with the other suggestions, so this doesn't matter anymore :D

> drop the existing duplicate dialog altogether and start fresh off the "Move" dialog
To confirm, this is duplicating `DocumentMove` as `DocumentCopy`, and not augmenting `DocumentMove` to support both, right?

> The Publish toggle could go next to the copy button
As in below the `Copy to xxx    Copy` line?

ok

tommoor:
> To confirm, this is duplicating DocumentMove as DocumentCopy, and not augmenting DocumentMove to support both, right?

Probably the place to start, and then maybe there will be some sub-components that could be shared

> As in below the Copy to xxx    Copy line?

To the right perhaps? Between that line and the primary "Copy" button? If it doesn't fit well then it would probably have to be above in the main content area.

infinite-persistence:
Ok. There is potential 2nd toggle buttons (another for "recursive"), so I'll put it in the main content area then.

infinite-persistence:
Updated. 




Still not sure if the toggle-option's position is best, especially when the tree is expanded. Is it ok?

tommoor:
I think this is pretty good, probably about as good as we can make it while retaining these component 👍

tommoor (approved):

PR #8031, opened 2024-11-27
Tracking of total uploaded attachments / team

tommoor:
Some busy work that's been needed for a while for cloud hosting – track on team the amount of storage size being used.

### DIFF
--- a/server/queues/tasks/UpdateTeamsAttachmentsSizeTask.ts
+++ b/server/queues/tasks/UpdateTeamsAttachmentsSizeTask.ts
@@ -1,0 +1,49 @@
+import { subDays } from "date-fns";
+import { Op } from "sequelize";
+import Logger from "@server/logging/Logger";
+import { Attachment } from "@server/models";
+import { sequelize } from "@server/storage/database";
+import BaseTask, { TaskPriority, TaskSchedule } from "./BaseTask";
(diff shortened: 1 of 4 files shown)


server/queues/tasks/UpdateTeamsAttachmentsSizeTask.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  public async perform({ limit }: Props) {
+    Logger.info(
+      "task",
+      `Recalculating attachment sizes for upto ${limit} teams…`
+    );
+
+    // Find unique attachment teamIds created in the last day, update only
+    // those teams' approximate attachment sizes
+    const rows = await Attachment.findAll({
```
hmacr, line 24: `findAllInBatches` perhaps?

server/queues/tasks/UpdateTeamsAttachmentsSizeTask.ts
```diff
@@ -0,0 +1,49 @@
...
+      attributes: [
+        [sequelize.fn("DISTINCT", sequelize.col("teamId")), "teamId"],
+      ],
+      where: {
+        createdAt: {
```
hmacr, line 29: I think an index on `createdAt` would be beneficial here.

PR #8042, opened 2024-11-29
Hmacr/7907 detach archived doc

### LINKED ISSUE #7907: Archived documents cannot be viewed when the collection is deleted
**To Reproduce**
1. Archive a document in a collection.
2. Delete the collection.
3. Notice that the document is not displayed in the "Archive" page.

**Expected behavior**
Archived document should be displayed in the "Archive" page.

---

Not completely sure if this is the intended behaviour (or) a bug.
In case it's a bug, archived documents should probably be detached just like drafts.

hmacr:
Closes #7907 

I've identified a couple of related bugs, stated below, for which separate PRs will be created.
1. Deleting an archived collection does not delete its documents.
2. Archive a document -> Then archive the containing collection -> Notice that the archived collection's structure does not have the previously archived document.

### DIFF
--- a/server/queues/tasks/DetachDocumentsFromCollectionTask.test.ts
+++ b/server/queues/tasks/DetachDocumentsFromCollectionTask.test.ts
@@ -51,4 +51,30 @@ describe("DetachDocumentsFromCollectionTask", () => {
     expect(draft?.deletedAt).toBe(null);
     expect(draft?.collectionId).toBe(null);
   });
+
+  it("should detach archived documents from deleted collection", async () => {
+    const collection = await buildCollection();
+    const document = await buildDocument({
+      title: "test",
+      collectionId: collection.id,
+      archivedAt: new Date(),
+      createdById: collection.createdById,
+      teamId: collection.teamId,
+    });
+    await collection.destroy({ hooks: false });
+
+    const task = new DetachDocumentsFromCollectionTask();
+    await task.perform({
+      collectionId: collection.id,
+      ip,
+      actorId: collection.createdById,
+    });
+
+    const detachedDocument = await Document.findByPk(document.id);
+    expect(detachedDocument).not.toBe(null);
+    expect(detachedDocument?.archivedAt).toBeTruthy();
+    expect(detachedDocument?.publishedAt).toBeTruthy();
+    expect(detachedDocument?.deletedAt).toBe(null);
+    expect(detachedDocument?.collectionId).toBe(null);
+  });
 });
--- a/server/queues/tasks/DetachDocumentsFromCollectionTask.test.ts
+++ b/server/queues/tasks/DetachDocumentsFromCollectionTask.test.ts
@@ -1,8 +1,8 @@
 import { Document } from "@server/models";
 import { buildCollection, buildDocument } from "@server/test/factories";
-import DetachDraftsFromCollectionTask from "./DetachDraftsFromCollectionTask";
+import DetachDocumentsFromCollectionTask from "./DetachDocumentsFromCollectionTask";
 
-describe("DetachDraftsFromCollectionTask", () => {
+describe("DetachDocumentsFromCollectionTask", () => {
   const ip = "127.0.0.1";
   it("should detach drafts from deleted collection", async () => {
     const collection = await buildCollection();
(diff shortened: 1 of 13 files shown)


server/queues/tasks/DetachDocumentsFromCollectionTask.test.ts
```diff
@@ -51,4 +51,30 @@ describe("DetachDraftsFromCollectionTask", () => {
...
+      ip,
+      actorId: collection.createdById,
+    });
+
+    const detachedDocument = await Document.findByPk(document.id);
+    expect(detachedDocument).not.toBe(null);
+    expect(detachedDocument?.archivedAt).toBeTruthy();
+    expect(detachedDocument?.publishedAt).toBeTruthy();
+    expect(detachedDocument?.deletedAt).toBe(null);
+    expect(detachedDocument?.collectionId).toBe(null);
```
apoorv-mishra, line 78: This is interesting. We have a kinda business rule where if a document is published, it's implied that it has a parent collection(since we always publish a doc _to_ a collection). We'd have to be mindful of this case since it violates the rule. Hopefully, doesn't regress since we already filter out archived/deleted docs in most of the scenarios.

server/routes/api/documents/documents.ts
```diff
@@ -343,7 +343,11 @@ router.post(
       const collectionIds = await user.collectionIds();
       where = {
         ...where,
-        collectionId: collectionIds,
+        collectionId: can(user, "readDocument", user.team)
+          ? {
+              [Op.or]: [{ [Op.in]: collectionIds }, { [Op.is]: null }],
+            }
+          : collectionIds,
```
apoorv-mishra, line 350: Haven't tested it but _theoretically_ it seems this won't include the archived doc which was _individually shared_(the ones that appear under "Shared with me" section) with a user formerly, after the collection is deleted?

hmacr: Great catch - I think this is the behaviour on prod. Probably separate the fix out to another PR so as to not over-complicate this one?

On a related note - I've come across this particular bug in other flows as well (e.g. backlinks, viewed docs). I think it'd be worthwhile to invest a separate effort to implement a solution that handles them all.

apoorv-mishra: Yeah, separate PR would be good.

server/queues/tasks/DetachDocumentsFromCollectionTask.ts
```diff
@@ -27,13 +27,24 @@ export default class DetachDraftsFromCollectionTask extends BaseTask<Props> {
       return;
     }
 
+    // Detach drafts and archived documents when an unarchived collection is deleted; otherwise detach drafts only.
+    const whereOpts: WhereOptions<Document> =
```
tommoor, line 31: Looking at it with fresh eyes – this all seems like a bit of an accident waiting to happen :( – these archived documents will suddenly become available to a broader audience when the collection is deleted?

hmacr: It's controlled with this policy [here](https://github.com/outline/outline/pull/8042/files#diff-aa7db55bb651c45df93af7d14ab2a0489bb846337b4aa80e1bbb7b31cb346d5dR53). This is broader than we want it to be - I think, in this case, the document has to be restricted to the owner and admin. Thoughts?

tommoor: Yea, admin and owner would work ok in this case

server/routes/api/documents/documents.ts
```diff
@@ -343,7 +343,25 @@ router.post(
...
       where = {
         ...where,
-        collectionId: collectionIds,
+        [Op.or]: [
+          {
+            collectionId: {
+              [Op.in]: collectionIds,
+            },
+          },
+          can(user, "readDocument", user.team)
```
apoorv-mishra, line 352: Looking at this now, I'm not sure if an indirection through policy is really required here? Also the policy `readDocument` seems very restrictive to allow only admins. Instead, should we just get rid of the policy and directly use `isTeamAdmin` here? Makes it more explicit IMO.

cc: @tommoor

hmacr: We want to retrieve the workspace level archived documents the user has access to. For workspace level docs (regardless of their archived status), admins should be able to view everything, while other users should be able to access the docs owned by them. The condition statement here and `readDocument` policy allow us to achieve this.

re: policy, I think it gives us the flexibility to change who can access workspace level docs later.

apoorv-mishra: Fine with this, no big deal.

apoorv-mishra (approved):

PR #8045, opened 2024-11-30
Notification for resolved comment

### LINKED ISSUE #7180: Add notifications for resolved comments
If you're involved in a comment thread there should be a notification when it is marked as resolved.

Follow up to #5078

infinite-persistence:
Closes #7180 

_Skipped discussions -- as always, ready to iterate or even close if there is a different plan_

## Spec
- [x]  Send notification + email to:
    - [x]  **Commenter** if the comment was resolved by someone else.
    - [x]  **Repliers**
    - [x]  **Mentioned users** in both the comment and replies.
- [ ]  Notify on “un-resolve” (probably no need?)

## Decision
The biggest one is between:
1. Overloading `NotificationEventType.MentionedInDocument` to handle both "mentioned" and "resolved", vs
    - con: potentially confusing code and setting
2. Creating a new `NotificationEventType.ResolveComment`
    - con: creates another setting (too fine-grained?), migration

Went with "1" for this iteration.


tommoor:
I think overloading `MentionedInComment` to this extent is a bit much, it creates all sorts of logical forks and potential bugs if you get mentioned in a resolved comment thread (I don't know that this is possible today, but I wouldn't want to count it out).

Creating another event type _is_ quite fine-grained, but we can default it to enabled (no db migration required) and I think thread resolution is different enough that it makes sense. I'd rather keep the events in the backend quite specific and the UI can be changed to toggle multiple at once if necessary.

infinite-persistence:
Updated.

Behavioral changes in this iteration:
- The `comments.update` webhook used to include resolve, but that is now split into `comments.resolve`.

tommoor:
I don't really follow why it was neccessary to add the new event type? Generally trying to avoid adding new specific events as they're not backwards compatible and require all that extra plumbing that could just use `comments.update`

infinite-persistence:
_Thanks for patience._

Updated:
- Only split `NotificationEventType`, not `CommentUpdateEvent` (you're right).
- Fixed logic in previous "New comment ↓" hint.

app/scenes/Settings/Notifications.tsx
```diff
@@ -65,6 +66,14 @@ function Notifications() {
         "Receive a notification when someone mentions you in a document or comment"
       ),
     },
+    {
+      event: NotificationEventType.ResolveComment,
+      icon: <DoneIcon />,
+      title: t("Resolved"),
+      description: t(
+        "Receive a notification when someone resolves a comment you created or were involved in"
```
tommoor, line 74:
```suggestion
        "Receive a notification when a comment thread you were involved in is resolved"
```

server/emails/templates/CommentResolvedEmail.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+    return;
+  }
+
+  protected subject({ document }: Props) {
+    return `Resolved a comment in “${document.title}”`;
+  }
+
+  protected preview({ actorName }: Props): string {
+    return `${actorName} resolved a comment you were involved in`;
```
tommoor, line 99:
```suggestion
    return `${actorName} resolved a comment thread`;
```

server/emails/templates/CommentResolvedEmail.tsx
```diff
@@ -0,0 +1,163 @@
...
+    if (notification?.user && notification.actor?.email) {
+      if (can(notification.user, "readEmail", notification.actor)) {
+        return notification.actor.email;
+      }
+    }
+    return;
+  }
+
+  protected subject({ document }: Props) {
+    return `Resolved a comment in “${document.title}”`;
```
tommoor, line 95:
```suggestion
    return `Resolved a comment thread in “${document.title}”`;
```

server/emails/templates/CommentResolvedEmail.tsx
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  protected renderAsText({
+    actorName,
+    teamUrl,
+    document,
+    commentId,
+    collection,
+  }: Props): string {
+    const t1 = `${actorName} resolved a comment on "${document.title}"`;
```
tommoor, line 113:
```suggestion
    const t1 = `${actorName} resolved a comment thread on "${document.title}"`;
```

server/queues/tasks/CommentUpdatedNotificationsTask.ts
```diff
@@ -53,6 +65,86 @@ export default class CommentUpdatedNotificationsTask extends BaseTask<CommentEve
...
+          user.id !== event.actorId &&
+          user.subscribedToEventType(NotificationEventType.ResolveComment) &&
+          (await canUserAccessDocument(user, document.id))
+        ) {
+          await send(user.id);
+          userIdsNotified.push(user.id);
+        }
+      }
+
+      // -- mentions
```
tommoor, line 127: Can we refactor this loop so that mentions is first. We can parse the mentioned user ID's, append the `createdById` to that array and then avoid having the rest of the logic twice here and the indirection to the `send` method.

tommoor (commented):
so close 👌

server/queues/tasks/CommentUpdatedNotificationsTask.ts
infinite-persistence: :flushed:

tommoor: I think it's an improvement? Less bug oppos

tommoor (approved):

server/queues/tasks/CommentUpdatedNotificationsTask.ts
infinite-persistence: Ah, I meant to say "I'm embarrassed for not doing that sooner" through that emoji 😄 .  So much cleaner.
I was fixated at retaining the copied logic from the mentions for some reason, and only factored out `send` :(

Issue #8026, opened 2024-11-27
Multiple context menus can be opened at once


tommoor:
Opening a context menu should close any others that are already open.

PR #8069, opened 2024-12-05
fix: Use singleton for tooltips, ensures that only one is visible at a time.

tommoor:
This feels about 7000% better

### Before

https://github.com/user-attachments/assets/d34ff6ac-a936-4068-84c3-761cee3d1777

### After

https://github.com/user-attachments/assets/a283a349-f2f6-456c-af3e-8073427dfc3e

Towards #8026

### DIFF
--- a/app/components/Tooltip.tsx
+++ b/app/components/Tooltip.tsx
@@ -4,6 +4,7 @@ import styled, { createGlobalStyle } from "styled-components";
 import { roundArrow } from "tippy.js";
 import { s } from "@shared/styles";
 import useMobile from "~/hooks/useMobile";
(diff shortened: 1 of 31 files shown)


app/components/Tooltip.tsx
```diff
@@ -35,6 +37,7 @@ function Tooltip({ shortcut, content: tooltip, delay = 50, ...rest }: Props) {
       animation="shift-away"
```
hmacr, line 37: Can these props be removed since the source tooltip in `TooltipContext` has them?
Maybe have just the `delay` supported here to allow override?

PR #8070, opened 2024-12-05
Adds count of occurences and index to find and replace

### LINKED ISSUE #7819: Show count of occurrences in find and replace
It would be nice to include the equivalent of this `1/x` language from chromium. Customer requested

image

This UI from Pitch is a lot bigger but could inspire a little redesign here soon 

image

tommoor:
image

closes #7819

### DIFF
--- a/app/editor/components/FindAndReplace.tsx
+++ b/app/editor/components/FindAndReplace.tsx
@@ -280,6 +280,7 @@ export default function FindAndReplace({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [popover.visible]);
 
+  const disabled = totalResults === 0;
   const navigation = (
     <>
       <Tooltip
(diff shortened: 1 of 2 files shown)


app/editor/components/FindAndReplace.tsx
```diff
@@ -278,7 +288,10 @@ export default function FindAndReplace({
         delay={500}
         placement="bottom"
       >
-        <ButtonLarge onClick={() => editor.commands.prevSearchMatch()}>
+        <ButtonLarge
+          disabled={totalResults === 0}
```
tommoor, line 292: Disable the navigation buttons if there are no results

hmacr: Disable the replace / replaceAll buttons as well?

While we're here, do you think adding a close button would be beneficial?

tommoor: Not going to add a close button right now as there is already an abundance of controls in this small space and there  are two ways to close with `cmd+f` and `esc`

PR #8082, opened 2024-12-06
fix(Grist): Improve integration of Grist links

kosssi:
When I publicly share a Grist spreadsheet the link is:
```
https://{{ grist_url }}/o/docs/{{ doc_id }}/{{ doc_name }}?utm_id=share-doc
```

The `utm_id` is create [here](https://github.com/gristlabs/grist-core/blob/efefc961e0c98d54bf22f0bafab629b6f534c0f6/app/client/ui/ShareMenu.ts#L331)

If I paste the link on Outline the link is transform by 
```
https://{{ grist_url }}/o/docs/{{ doc_id }}/{{ doc_name }}?utm_id=share-doc?embed=true
```

The code which add this is [here](https://github.com/outline/outline/blob/6c1e4a5b4041caa0c9ee1b429bc7a2beaea19b1d/shared/editor/embeds/index.tsx#L411).

The rendering is therefore not done correctly.

### DIFF
--- a/shared/editor/embeds/index.tsx
+++ b/shared/editor/embeds/index.tsx
@@ -404,11 +404,11 @@ const embeds: EmbedDescriptor[] = [
     transformMatch: (matches: RegExpMatchArray) => {
       const input = matches.input ?? matches[0];
 
-      if (input.includes("style=singlePage")) {
+      if (input.includes("style=singlePage") || input.includes("embed=true")) {
         return input;
       }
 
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks, this works but it's gotten a bit messy, in hindsight regex here was a bad idea. We should use `new URL`  to parse the url and `query-string` which is already imported into the client to add the embed query.

kosssi:
@tommoor Don't hesitate to change my request to add or create one on your side. I won't have much time this week to contribute.

PR #8085, opened 2024-12-07
feat: Improve inline rule matching

### LINKED ISSUE #8083: Allow inline code marks to trigger when prededed by brackets and braces
A quick piece of feedback regarding backticks:
&gt; There seems to be a formatting issue that comes up if you put a backtick after an open paren. For example, if I write out this string in outline:
&gt; ```(`/something`)```
&gt; then what's inside the parens doesn't get properly formatted as code. It should look like this:
&gt;&gt; (`/something`)
&gt; 
&gt; But instead it looks like
&gt;&gt; (`/something`)
In Outline, I noticed that backticks don't work if they are preceded by any character. It seems Slack makes special cases for backticks after (, {, and [. I think that is more in line with what folks expect elsewhere and is hopefully an easy improvement.

tommoor:
Refactors inline mark matching to use a single consistent rule, adds ability to match when preceded by "(" "[" "{" characters.

closes #8083

### DIFF
--- a/shared/editor/marks/Italic.ts
+++ b/shared/editor/marks/Italic.ts
@@ -2,7 +2,7 @@ import { toggleMark } from "prosemirror-commands";
 import { InputRule } from "prosemirror-inputrules";
 import { MarkSpec, MarkType } from "prosemirror-model";
 import { Command } from "prosemirror-state";
-import markInputRule from "../lib/markInputRule";
+import { markInputRuleForPattern } from "../lib/markInputRule";
 import Mark from "./Mark";
 
 export default class Italic extends Mark {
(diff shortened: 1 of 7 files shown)


shared/editor/marks/Italic.ts
```diff
@@ -25,29 +25,9 @@ export default class Italic extends Mark {
   }
 
   inputRules({ type }: { type: MarkType }): InputRule[] {
-    /**
-     * Due to use of snake_case strings common in docs the matching conditions
```
tommoor, line 29: Tested all these cases and they still function correctly with `markInputRuleForPattern` 💪

PR #8119, opened 2024-12-17
feat: Allow users to change email in-app

### LINKED ISSUE #7387: Allow updating a user's email from member management
This would need to send an email to the new address, which when clicked would perform the email change, rather than immediately modifying the database.

towards https://github.com/outline/outline/issues/6159

tommoor:
Allows users to change their own email address, with confirmation.

towards #7387


aeonoea:
Hi @tommoor, do you think it's possible that after changing the e-mail address, to "lock" it?

I use SSO to log in, and changing the e-mail address works, but after each login it gets reset to the old e-mail address that is used in the SSO database.

tommoor:
@aeonoea is there a reason you wouldn't update in the SSO provider in this case?

aeonoea:
Sure, it's because our SSO (Synology) only allows setting one email address per user. Using the work email instead of the private would break the password recovery process.

tommoor:
It's doable but we'd need a new checkbox/storage bit for whether to sync the email with SSO.

aeonoea:
That would be amazing, thank you.

Issue #7652, opened 2024-09-24
Add ability to resize embeds (iframes)


Sotneo:
Description:
Currently, the height of embeds (iframes) in Outline is fixed at 400 pixels. This can cause issues with content display if it has a larger or smaller height.

Proposal:
Add the ability to resize embeds (iframes) in Outline. 
My current workaround, I rebuild docker image with 600 px height, but ability to set custom height would be great.

Sotneo:
Sorrry, can't change label to enhancement

PR #8154, opened 2024-12-27
feat: Allow resizing of embed height

tommoor:
Towards #7652


tommoor:
I'm swaying if to do an entire rebuild of resizing or just get this improvement merged and come back to it

Ali-ovo:
I have an idea, we get the aspect ratio, by jumping to the width of the automatic modification of the height can be equal to the proportion resize

tommoor:
There are no "natural" proportions for many of the embeds, so in that sense they're not like images.

Ali-ovo:
> There are no "natural" proportions for many of the embeds, so in that sense they're not like images.

I have used notion before, and its embed has a similar resize feature, and I think the experience is OK
（You can try embedding a youtube video in it to experience this feature）

tommoor:
Videos are kind of the one exception where they do have a natural size and act like images

PR #8157, opened 2024-12-28
Upgrade and virtualize table component

hmacr:
This PR virtualizes the `Table` component to fetch and show more data on scroll instead of prev/back buttons. Along the way, `react-table` library is upgraded to the latest version (`tanstack/react-table`).

For virtualization, `tanstack/react-virtual` is used instead of the existing `react-window` library.
`tanstack/react-virtual` APIs are more modern and more importantly, it uses `transform` to show/hide virtualized items (`react-window` uses `top` - we should replace its existing usages).


app/scenes/Settings/components/PeopleTable.tsx
```diff
@@ -1,99 +1,113 @@
...
 
-function PeopleTable({ canManage, ...rest }: Props) {
+export function PeopleTable({ canManage, ...rest }: Props) {
   const { t } = useTranslation();
   const currentUser = useCurrentUser();
-  const columns = React.useMemo(
+
+  const gridColumns = React.useMemo(() => {
+    if (canManage) {
+      return "4fr 4fr 2fr 2fr 0.5fr"; // all columns will be displayed.
```
hmacr, line 28: This grid becomes necessary since absolutely positioned table rows are not stylable to take up the full width. Instead, the rows are rendered as grid, with the cells being the grid items.

app/scenes/Settings/components/SharesTable.tsx
```diff
@@ -1,108 +1,137 @@
...
-function SharesTable({ canManage, data, ...rest }: Props) {
+export function SharesTable({ data, canManage, ...rest }: Props) {
   const { t } = useTranslation();
   const language = useUserLocale();
   const hasDomain = data.some((share) => share.domain);
 
-  const columns = React.useMemo(
+  const gridColumns = React.useMemo(() => {
+    if (canManage && hasDomain) {
+      return "4fr 2fr 2fr 2fr 1.5fr 1fr 0.5fr"; // all columns will be displayed.
```
hmacr, line 32: Same as `PeopleTable`.. I'm not completely sure how to improve this.
One iteration of the implementation had this width config as part of the individual column defs itself - but I figured having the grid styles together (like how it's in CSS) provided better understanding.

server/routes/api/shares/shares.ts
```diff
@@ -152,7 +152,28 @@ router.post(
         offset: ctx.state.pagination.offset,
         limit: ctx.state.pagination.limit,
       }),
-      Share.count({ where }),
+      Share.count({
```
hmacr, line 155: Tiny bug with shares count which was not in parity with the actual available data.
For admins, we did not check if they actually had access to the shared doc (ex: doc in another user's private collection).

app/scenes/Settings/components/PeopleTable.tsx
tommoor: Hmmm yea I really don't like this defined here – the css should be derived from the columns within the Table component somehow. We can define some sort of `width` property on the column definition below…?

app/scenes/Settings/components/SharesTable.tsx
tommoor: As above, definitely prefer figuring out how to get the width as a config

tommoor (commented):
I noticed that there's some extra double scrolling going on now… _Ideally_  it would appear like a normal settings page that just scrolls forever. If that's too hard with the virtualization then the main page scrolling could be fixed, but either way there shouldn't be multiple scroll areas :)

hmacr:
> I noticed that there's some extra double scrolling going on now

Oh, I think the table height needs adjusting. I _intentionally_ fixed the page height so that only the table area is scrollable. That way, the filters at the top are always accessible.

> Ideally it would appear like a normal settings page that just scrolls forever. If that's too hard with the virtualization then the main page scrolling could be fixed

Both should be doable.
I prefer a fixed main page scroll - do you prefer to keep the existing behaviour of the whole page scroll?

app/scenes/Settings/components/PeopleTable.tsx
hmacr: Do you mean to skip the whole grid styles? It becomes tricky to accept percentage as column width since we need to ensure all the columns add up to 100% for a full-width table.
The simpler option would be to move the individual parts of `4fr 4fr 2fr 2fr 0.5fr` to its column def. Would that be ok?

tommoor: Grid style is fine, with grid you can an use mixed widths so one column could be a fixed px and another could be `fr` unit? I think it will work okay if we can move to config

tommoor:
> I prefer a fixed main page scroll - do you prefer to keep the existing behaviour of the whole page scroll?

In a perfect world it would be a full page scroll as that then triggers the sticky header instead of having the top 30% of the screen be used up by the header.

Otherwise this is great though, thanks

hmacr:
> In a perfect world it would be a full page scroll as that then triggers the sticky header instead of having the top 30% of the screen be used up by the header

Couldn't agree more! I tried that initially but it was very tricky to get it right for table and absolute positioned elements (quite a few quirks for table in CSS). An alternative we can try is to replace tables with divs and aria roles. That might help a lot in navigating these quirks. Is that something we can explore?

tommoor:
Happy to explore that if the table is the limiting factor

hmacr:
I just pushed a commit that implements the full page scroll, please test and let me know if something needs to be changed 😄

app/components/FilterOptions.tsx
```diff
@@ -41,12 +42,13 @@ const FilterOptions = ({
...
   fetchQuery,
   fetchQueryOptions,
+  modalMenu,
 }: Props) => {
   const { t } = useTranslation();
   const searchInputRef = React.useRef<HTMLInputElement>(null);
   const listRef = React.useRef<HTMLDivElement | null>(null);
   const menu = useMenuState({
-    modal: true,
+    modal: modalMenu ?? true,
```
hmacr, line 51: Quirks of reakit.. scroll position changes unexpectedly when opening the menu in `modal` mode
https://github.com/ariakit/ariakit/issues/856 - Seems this issue is fixed in v2; all the more reason to move to Radix soon.

app/components/Scene.tsx
```diff
@@ -21,6 +21,8 @@ type Props = {
   wide?: boolean;
   /** The content of the scene */
   children?: React.ReactNode;
+  /** Whether to shrink padding */
+  shrink?: boolean;
```
tommoor, line 25: I'm not really clear on how this differs from the `wide` option – aka when would we use one without the other?

app/components/FilterOptions.tsx
tommoor: Ok, I guess my only thought is can we just default this to `false` then and avoid the prop drilling? It's not used in many spots

app/scenes/Settings/Members.tsx
```diff
@@ -120,43 +88,31 @@ function Members() {
...
-  );
+  React.useEffect(() => {
+    if (error) {
+      toast.error(t("Could not load members"));
+    }
+  }, [t, error]);
 
-  const appName = env.APP_NAME;
+  React.useEffect(() => {
+    const timeout = setTimeout(() => updateParams("query", query), 250);
```
tommoor, line 113: Why the timeout? Probably fine but worth a comment

tommoor (commented):
v good

tommoor:
image

Oop, just spotted the sticky filters are infront of dialogs

app/components/Scene.tsx
hmacr: Yeah, this is not needed now..
Earlier, with the fixed height table, this was needed to reduce the bottom padding (and hence the body scroll).

app/scenes/Settings/Members.tsx
hmacr: It's more of a "hacky" debounce for the filter typed in the search box (earlier, we were invoking the API for each keypress).

app/components/FilterOptions.tsx
hmacr: With `modal` true, the tab focus stays within the rendered menu - it may not be as important I think.
Will change it 👍

tommoor (approved):

Issue #7920, opened 2024-11-11
Move event writing to model layer


tommoor:
`events` should be written automatically based on mutations in the model layer. To achieve this the request context will be passed down to model mutations. Translation must happen on a model by model basis.

### TODO
- [x] ApiKey
- [x] Attachment
- [ ] AuthenticationProvider
- [ ] Backlink
- [ ] Collection
- [ ] Comment
- [ ] Document
- [ ] Event
- [x] FileOperation
- [ ] Group
- [ ] GroupMembership
- [ ] GroupUser
- [ ] IntegrationAuthentication
- [ ] Notification
- [ ] Pin
- [ ] Reaction
- [ ] Revision
- [ ] SearchQuery
- [ ] Share
- [ ] Star
- [ ] Subscription
- [ ] Team
- [ ] TeamDomain
- [ ] User
- [ ] UserAuthentication
- [ ] UserMembership
- [ ] View
- [ ] WebhookDelivery
- [ ] WebhookSubscription

PR #8161, opened 2024-12-29
Convert `WebhookSubscription` mutations

hmacr:
Towards #7920

### DIFF
--- a/plugins/webhooks/server/api/webhookSubscriptions.ts
+++ b/plugins/webhooks/server/api/webhookSubscriptions.ts
@@ -5,7 +5,7 @@ import { UserRole } from "@shared/types";
 import auth from "@server/middlewares/authentication";
 import { transaction } from "@server/middlewares/transaction";
 import validate from "@server/middlewares/validate";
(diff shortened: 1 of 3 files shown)


plugins/webhooks/server/api/webhookSubscriptions.ts
```diff
@@ -120,37 +98,24 @@ router.post(
...
-
-    await Event.createFromContext(ctx, {
-      name: "webhookSubscriptions.update",
-      modelId: webhookSubscription.id,
-      data: {
-        name: webhookSubscription.name,
-        url: webhookSubscription.url,
-        events: webhookSubscription.events,
-      },
+    webhookSubscription = await webhookSubscription.updateWithCtx(ctx, {
```
tommoor, line 113: Is the reassignment needed? If so might need to change `updateWithCtx` to match `update` behavior

hmacr: It ends up being an unnecessary assignment - changed it now.

tommoor (approved):

PR #8166, opened 2024-12-30
Convert `Subscription` mutations

hmacr:
Towards #7920


server/routes/api/subscriptions/subscriptions.ts
```diff
@@ -146,7 +143,13 @@ router.get(
 
     authorize(user, "delete", subscription);
 
-    await subscription.destroy({ transaction });
+    ctx.context = {
```
hmacr, line 146: For email unsubscribe flow, this enables the background tasks. Should this be removed?

server/models/base/Model.ts
```diff
@@ -63,6 +65,13 @@ class Model<
     return this.destroy(ctx.context as InstanceDestroyOptions);
   }
 
+  /**
+   * Restore the row corresponding to this instance. Only available for paranoid models.
+   */
+  public restoreWithCtx(ctx: APIContext) {
```
hmacr, line 71: `changeset` is empty for restore flow ([sequelize open issue](https://github.com/sequelize/sequelize/issues/12912)). This means the `changes` will not be populated in the published event. I think this should be fine in this case - we can always manually build it if the need arises.

server/routes/api/subscriptions/subscriptions.ts
tommoor: I don't love having the auth type as `API`, will make it indistinguishable from a true API request. Can `auth.type` be left empty?

Otherwise we can also use `createContext` helper here

hmacr: I wasn't sure of using `APP` vs `API` (email action being an one-off case for now) - will use `createContext` here.

server/routes/api/subscriptions/subscriptions.ts
```diff
@@ -143,11 +144,11 @@ router.get(
 
     authorize(user, "delete", subscription);
 
-    ctx.context = {
-      auth: { user, type: AuthenticationType.API },
-      transaction,
+    ctx.context = createContext({
```
tommoor, line 147: You can pass `createContext` straight through to `destroyWithCtx`, no need to override

tommoor (approved):

PR #8177, opened 2024-12-31
Convert `Share` mutations

hmacr:
Towards #7920 

Marking it as draft since it needs to be rebased after #8166 (change in `createContext` method signature).


server/types.ts
```diff
@@ -424,9 +424,6 @@ export type ShareEvent = BaseEvent<Share> & {
   modelId: string;
   documentId: string;
   collectionId?: string;
-  data: {
-    name: string;
```
hmacr, line 428: Removed `name` data as it wasn't used anywhere in the background jobs.

tommoor: sg, as long as it's in the changeset we should be removing usage of `data` whereever possible.

server/models/Share.ts
```diff
@@ -188,10 +192,11 @@ class Share extends IdModel<
   @Column
   allowIndexing: boolean;
 
-  revoke(userId: string) {
+  revoke(ctx: APIContext, eventOverride?: EventOverride) {
```
tommoor, line 195: No need to allow passing `eventOverride` here, we can just hardcode it on L199?

hmacr: Yeah 🤦‍♂️

PR #8179, opened 2025-01-01
Move Group to model event writing

tommoor:
Towards #7920


hmacr (approved):

hmacr:
Just noticed the `GroupEvent` type can be cleaned up to remove the `data` field.

https://github.com/outline/outline/blob/b8c07eb2984a95bbc8a341f982c437225c1bfe15/server/types.ts#L357-L359

PR #8180, opened 2025-01-01
chore: Cleanup unused pinDestroyer

tommoor:
Follow on from #7993


hmacr (approved):
I can't recall a good reason why this wasn't done in #7993

PR #8181, opened 2025-01-02
Convert `Comment` and `Reaction` mutations

hmacr:
Towards #7920

Marking as draft since this builds on #8177 `eventOverride` field.


server/models/Reaction.ts
```diff
@@ -71,7 +77,7 @@ class Reaction extends IdModel<
       return;
     }
 
-    const reactions = comment.reactions ?? [];
+    const reactions = cloneDeep(comment.reactions) ?? [];
```
hmacr, line 80: `cloneDeep` is used instead of setting `changed` field in the model. This allows us to generate the right changeset  in the event.

server/routes/api/comments/comments.ts
```diff
@@ -245,20 +252,33 @@ router.post(
...
-    authorize(user, "comment", document);
     authorize(user, "delete", comment);
+    authorize(user, "comment", document);
 
-    await commentDestroyer({
-      user,
-      comment,
-      ip: ctx.request.ip,
+    // Delete child comments.
+    const childComments = await Comment.findAll({
```
tommoor, line 267: Thinking we should move this into `AfterDestroy` hook?

server/models/Reaction.ts
```diff
@@ -81,16 +87,28 @@ class Reaction extends IdModel<
...
-    comment.changed("reactions", true);
-    await comment.save({ fields: ["reactions"], transaction, silent: true });
+
+    // Pass only the fields needed in APIContext; otherwise sequelize props will be overwritten.
+    const context = {
+      context: { auth: ctx.auth, ip: ctx.ip, transaction: ctx.transaction },
+    } as APIContext;
+
+    await comment.saveWithCtx(
+      context,
```
tommoor, line 97: use `createContext`?

server/routes/api/comments/comments.ts
hmacr: Yes, much better.

I just realised that we might be better off deleting the associated reactions as well - thinking of a background task in a separate PR.

tommoor: _probably_ unnecessary, they will get cleaned up through db cascade if the comment is every fully removed

hmacr: That's true

tommoor (approved):

hmacr:
I had to update the PR to resolve merge conflicts - sorry about the re-review / re-approval 😓

tommoor:
All good, I expected this which is why it hasn't been merged ;)

tommoor (approved):

PR #8188, opened 2025-01-04
Publish event in `withCtx` flow only

hmacr:
Towards #7920 

Ensure that event is published in the `withCtx` flows only. This avoids accidental event publishing from other model persistence touchpoints.

### DIFF
--- a/server/models/base/Model.ts
+++ b/server/models/base/Model.ts
@@ -31,7 +31,9 @@ type EventOverrideOptions = {
 };
 
 type EventOptions = EventOverrideOptions & {
-  /** Whether to publish event to the job queue. */
+  /**
+   * Whether to publish event to the job queue. Defaults to true when using any `withCtx` methods.
+   */
   publish: boolean;
 };
 
@@ -50,14 +52,11 @@ class Model<
   /**
    * Validates this instance, and if the validation passes, persists it to the database.
    */
-  public saveWithCtx(
-    ctx: APIContext,
-    eventOverrideOpts?: EventOverrideOptions
-  ) {
+  public saveWithCtx(ctx: APIContext, eventOpts?: EventOverrideOptions) {
     const hookContext: HookContext = {
       ...ctx.context,
       event: {
-        ...eventOverrideOpts,
+        ...eventOpts,
         publish: true,
       },
     };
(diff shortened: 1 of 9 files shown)


tommoor:
As discussed in DM, we can introduce a separate option for this to avoid accidentally removing critical hooks

tommoor:
I had to quickly patch this one too (https://github.com/outline/outline/commit/e93ef8b3927cbac623adfbedd6eedb078ec67f82) as it was creating an anonymous event on every view of a share

hmacr:
I think the best way to move forward is to set `events: true` from the various `withCtx` persistence methods. This feels right since we want the auto-events in the `withCtx` flows only (also avoids these accidental events throughout the app).

hmacr:
With the recent commit, I think it looks a lot clearer.
Also, cleaned up `hooks: false` for GroupUser. We still have FileOperation and Share cleanups which will be a follow-up PR to not increase the scope further.

server/models/base/Model.ts
```diff
@@ -30,10 +25,20 @@ import Logger from "@server/logging/Logger";
...
+type EventOverrideOptions = {
+  /** Override the default event name. */
   name?: string;
 };
 
+type EventOptions = EventOverrideOptions & {
+  /**
+   * Whether to publish event to the job queue. Defaults to true when using any `withCtx` methods.
+   */
+  publish: boolean;
```
tommoor, line 37: I think "publish" is just a little overloaded in Outline already in regards to publishing documents. It will also _write_ events to the db – maybe just "create" is ok?

server/models/base/Model.ts
```diff
@@ -47,34 +52,65 @@ class Model<
   /**
    * Validates this instance, and if the validation passes, persists it to the database.
    */
-  public saveWithCtx(ctx: APIContext, eventOverride?: EventOverride) {
-    this.eventOverride = eventOverride;
+  public saveWithCtx(ctx: APIContext, eventOpts?: EventOverrideOptions) {
+    const hookContext: HookContext = {
+      ...ctx.context,
+      event: {
+        ...eventOpts,
```
tommoor, line 59: We should move the spreads below so that it can be overridden

hmacr: The idea is, `withCtx` methods will always publish an event - so those methods accept `EventOverrideOptions` (without `publish` method) so as to not allow override. We can use the vanilla persistence methods from Sequelize in case we don't want to trigger any event.

`insertEvent` method, which is invoked by hooks, receives the `EventOptions` with `publish` field that'll be
- true in the `withCtx` flows.
- false in the vanilla persistence flows (can be overridden in a non-typesafe way).

PR #8189, opened 2025-01-04
Simplify email references determination

### LINKED ISSUE #8048: Threading of emails from document edits is too deeply nested
**To Reproduce**
1) Have multuple edits to a document over time which sends many emails (eg over days)
2) look at the nesting of emails in thunderbird

image

Not all email clients do this, eg gmail ignores the email headers and always does single level threading.

Conceptually a series of 10 edits in serial against a document is 10 comments against the same item (ie the document). Deeply nested threads like above are appropriate for when a user is commenting on a comment, or commenting on a comment on a comment etc which isn't the case here.

**Expected behavior**

The emails should have the threading headers in place to group them, but they should only be nested to a single level and not deeply recursively nested.

Current state of headers:
```
In-Reply-To: <774d50e7-1610-4134-917f-429f16b452da@docs.catalyst-it.io>
References: <bc5cecb4-8945-4733-b1b0-4fb4101596cd@docs.catalyst-it.io>
 <15d071ae-1b59-4d1a-bc4f-c59957215c63@docs.catalyst-it.io>
 <1bbef69c-f01c-4346-9932-41e575b8c79f@docs.catalyst-it.io>
 <7656aaf9-9ff1-44fc-8e51-c80913ea2b41@docs.catalyst-it.io>
 <cd56ba46-728e-4864-b8ba-344a9cece832@docs.catalyst-it.io>
 <774d50e7-1610-4134-917f-429f16b452da@docs.catalyst-it.io>
```

Desired state of headers:

It isn't clear to me exactly what the id's used in these headers are, but the `References` header should only have a single value which would be some ID tied to the document id. the `In-Reply-To` header can be identical.

```
In-Reply-To: <774d50e7-1610-4134-917f-429f16b452da@docs.catalyst-it.io>
References: <774d50e7-1610-4134-917f-429f16b452da@docs.catalyst-it.io>
```
for instance this doc has a url stub of /doc/weekly-tasks-7MPst8lrrR so it would seem simpler to me to have these headers:

```
In-Reply-To: <7MPst8lrrR@docs.catalyst-it.io>
References: <7MPst8lrrR@docs.catalyst-it.io>
```

The values of these do not need to point to valid Message ID's, they just need to be the same between emails to group them.

Commit https://github.com/outline/outline/commit/86c3ea8e

Ubuntu thunderbird

hmacr:
Closes #8048 

I don't see a difference in the threading behavior in Thunderbird. That is, the previous implementation also did not create nested threads shown in #8048 (probably an older version Thunderbird behavior).
However, the current implementation vastly simplifies determining the reference.

The references are listed below.
| Event | Reference |
|--------|--------|
| Document Publish, Document Update | {documentId}-updates | 
| Document Mention, Comment Mention | {documentId}-mentions | 
| Comment Creation | {commentId}-comments |


tommoor:
Based on the issue it seems like this simplification should work, one thing that immediately jumps out though is that each _reply_ in an Outline comment thread should probably be it's own thread in the email inbox – otherwise you'll have unrelated conversations interleaved.

hmacr:
By reply, do you mean individual email thread for each comment thread? It makes sense to have each comment thread separately in the email. One little detail is that, when the user is mentioned in one of the thread comments, that email will be in a separate thread (document mentions). Rest all will be threaded like a chat from oldest to newest.

tommoor:
> By reply, do you mean individual email thread for each comment thread? I

Yep, that's right. If mentions are separate still that seems reasonable

hmacr:
I just pushed a commit that creates a new email thread for each Outline comment thread - please lmk if something needs to be changed.

server/emails/templates/CommentCreatedEmail.tsx
```diff
@@ -91,8 +93,20 @@ export default class CommentCreatedEmail extends BaseEmail<
...
+      )
+    );
+    const trimmedText =
+      commentText.length <= MAX_SUBJECT_CONTENT
+        ? commentText
+        : `${commentText.slice(0, MAX_SUBJECT_CONTENT)}...`;
+
+    return `${parentComment ? "Re: " : ""}New comment on “${
+      document.title
+    }” - ${trimmedText}`;
```
tommoor, line 109: Do we need to the beginning of the comment as well here? I'm not really sure it adds much… but if we do there is a `toPlainText` helper already that can be used

hmacr: Yeah, it's necessary to separate the individual comment threads within the document - Gmail, and the likes, uses subject to auto-thread the emails (in addition to the references).
Each email client has its own behavior of threading - we want to ensure we cover the base as much as possible by providing unique subject and references.

tommoor (approved):

PR #8194, opened 2025-01-05
fix: Notify previously mentioned users when new comment is added to a thread

### LINKED ISSUE #8192: Mentioned users in a comment thread don't receive email notifications when new replies are added.
Just as the title says - unless any of those users have created a comment in the thread earlier.

hmacr:
Closes #8192


tommoor (approved):
great find, thanks

PR #8196, opened 2025-01-06
feat: Option to return anchor text for comments

### LINKED ISSUE #8071: Add parameter to return selected text with comments via API
When a comment is on selected text it would be useful to be able to return that via the API.

We can get this by parsing the document state and finding the matching mark with comment ID – this will be relatively expensive and should be optional.

hmacr:
Closes #8071

### DIFF
--- a/server/models/helpers/ProsemirrorHelper.tsx
+++ b/server/models/helpers/ProsemirrorHelper.tsx
@@ -549,4 +549,30 @@ export class ProsemirrorHelper {
 
     return dom.serialize();
   }
+
+  static getAnchorTextForComment(doc: Node, commentId: string) {
+    let textNode: Node | undefined;
+
+    doc.descendants((node) => {
+      if (textNode) {
+        return false;
+      }
+
+      if (node.type.name !== "text") {
+        return true;
+      }
+
+      const commentMark = node.marks.find(
+        (mark) => mark.type.name === "comment" && mark.attrs.id === commentId
+      );
+
+      if (commentMark) {
+        textNode = node;
+      }
+
+      return true;
+    });
+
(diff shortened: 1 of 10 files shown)


server/routes/api/comments/comments.ts
```diff
@@ -72,8 +74,17 @@ router.post(
     authorize(user, "read", comment);
     authorize(user, "read", document);
 
+    let anchorText: string | undefined;
+
+    if (includeAnchorText) {
+      anchorText = ProsemirrorHelper.getAnchorTextForComment(
```
tommoor, line 80: Move this inside `presentComment(comment, { includeAnchorText })` and I think it'll cleanup nicely. Would just need to ensure `comment.document` is available

server/models/helpers/ProsemirrorHelper.tsx
```diff
@@ -549,4 +549,30 @@ export class ProsemirrorHelper {
...
+    doc.descendants((node) => {
+      if (textNode) {
+        return false;
+      }
+
+      if (node.type.name !== "text") {
+        return true;
+      }
+
+      const commentMark = node.marks.find(
```
tommoor, line 565: There can be multiple comment marks with the same ID which represent a single "anchor", it's calculated on the client here:

https://github.com/outline/outline/blob/main/app/scenes/Document/components/CommentThread.tsx#L77-L80

Ideally we can keep the logic identical to make sure they don't get out of sync?

hmacr: Oh, I wasn't aware the comment marks can be split across multiple text nodes.

tommoor (approved):

PR #8212, opened 2025-01-09
Move group management to table

### LINKED ISSUE #7484: Move group management to table
Currently a list, and one that is badly ordered at that. Lets use the same setup as the Members settings

hmacr:
Closes #7484 

This PR moves group management page to use `Table` component. Along the way, group modals that were scattered in many files are consolidated into a single `GroupDialogs` file.


hmacr:
There's a subtle regression in nested group modals wherein closing a nested modal does not bring the focus back automatically to the parent modal - we render each modal as a separate top level dialog; reakit suggests the child modal to be rendered inside the parent to ensure proper focus on close (https://reakit.io/docs/dialog/#nested-dialogs).

Fixing this would require changing [Dialogs](https://github.com/outline/outline/blob/main/app/components/Dialogs.tsx#L21) and [Modal](https://github.com/outline/outline/blob/main/app/components/Modal.tsx) component. I was wondering if we can live with this slight regression for a couple weeks until the modal component is migrated to Radix (my next big task), which will anyway need some re-implementation.

app/scenes/Settings/components/GroupsTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+            return (
+              <Flex>
+                <Facepile users={users} overflow={overflow} />
+              </Flex>
+            );
+          },
+          width: "1fr",
+          sortable: false,
+        },
+        !currentUser.isGuest
```
tommoor, line 91: We shouldn't check roles directly like this, instead check if the policy allows `can.read` or `can.update` for the group

app/scenes/Settings/components/GroupsTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+            return (
+              <Flex>
+                <Facepile users={users} overflow={overflow} />
+              </Flex>
+            );
+          },
+          width: "1fr",
+          sortable: false,
+        },
```
tommoor, line 90: Lets add a sortable column for `createdAt` while we're improving this

app/scenes/Settings/Groups.tsx
```diff
@@ -46,43 +135,48 @@ function Groups() {
...
-        isOpen={newGroupModalOpen}
-      >
-        <GroupNew onSubmit={handleNewGroupModalClose} />
-      </Modal>
+      </DelayedMount>
     </Scene>
   );
 }
 
+const StickyFilter = styled(Flex)`
```
tommoor, line 174: Should we split this into `/settings/components`, used in multiple spots

app/scenes/Settings/Groups.tsx
```diff
@@ -46,43 +135,48 @@ function Groups() {
...
-            renderActions={({ openMembersModal }) => (
-              <GroupMenu group={item} onMembers={openMembersModal} />
-            )}
-            showFacepile
-          />
+      <DelayedMount>
+        {isEmpty ? (
+          <Empty>{t("No groups have been created yet")}</Empty>
+        ) : (
+          <Fade>
```
tommoor, line 150: Is the fade neccessary? Especially around the filter UI as well It would be better to appear instantly

app/scenes/Settings/components/GroupsTable.tsx
hmacr: That's true - we need a way to skip rendering the whole actions column; might have been better off with a private `useGroupAction` hook that allows us to use `usePolicy` hook.

That said, I think the check is _redundant_ since the groups page is hidden is hidden for non-guests - will remove this.

app/scenes/Settings/Groups.tsx
hmacr: I added it to avoid UI flicker - I'm not able to reproduce it now 🤔 
Will check and remove it 👍

tommoor (approved):

PR #8225, opened 2025-01-11
Document mentions

### LINKED ISSUE #7680: Document @mentions
We want to introduce the concept of mentioning documents alongside users, this would have a "chip" style in the same way that would show the document icon and title.

tommoor:
This PR adds the ability to mention documents using `@` and in doing so removes the document search from the link editor, which becomes a plain link input. As a nice side effect `cmd+k` is now also useable while the editor is focused.

Document mentions are synced with the title and icon of the mentioned document. Any existing links will stay the same for now and would need converting to the mention format if desired.

In order to use the `Icon` inside editor it moved to `/shared/` hence a lot of changes here.

Setup for more mention types down the line.

closes #7680


tommoor:
Okay, I _think_ this is good enough to go.

app/editor/components/MentionMenu.tsx
```diff
@@ -60,28 +67,92 @@ function MentionMenu({ search, isActive, ...rest }: Props) {
...
-          actorId: auth.currentUserId ?? undefined,
-          label: user.name,
-        },
-      }));
+    if (data && actorId && !loading) {
+      const items = data.users
+        .map(
+          (user) =>
+            ({
+              name: "mention",
```
hmacr, line 75: Can we use priority to order the items instead of sorting by section name in `SuggestionsMenu`?

shared/editor/components/Mentions.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { users } = useStores();
+  const user = users.get(node.attrs.modelId);
+
+  return (
+    <span
+      className={cn({
+        "ProseMirror-selectednode": isSelected,
+        "use-hover-preview": true,
+        mention: true,
+      })}
```
hmacr, line 23: We need to set the data attributes here for the hover preview and current user badge styling to work.

hmacr (changes requested):
Works great! Just some tiny things I found.
On a side note, copying an existing mention (user/document) and pasting it re-uses the `id` attribute. Is that something we can fix now?
This type of user mention pasting is causing errors in the backend events - the actorId column is not set properly.

app/editor/components/MentionMenu.tsx
tommoor: I should probably just change it to order by section priority

hmacr: Yea, that's even better.

tommoor: https://x.com/tommoor/status/1879530146287489339

hmacr:
Sorry, I just noticed something else when testing the latest updates - mentions are not displayed in the comments until they are clicked (something to do with node views being lazy rendered?).

https://github.com/user-attachments/assets/cf380209-363b-41de-9448-e22e65062604

tommoor:
Good find!

hmacr (approved):
Works great!!

apoorv-mishra:
Not a pressing issue, but the title of a doc from a private collection is revealed to another user if mentioned in a mutually accessible doc(though, the user won't be able to access the doc, through hover or click, which is good)


https://github.com/user-attachments/assets/66b86534-8baa-4c84-b846-c1bc21bb2639

tommoor:
Definitely by design, and no different from the way linking a document works today – although I'll make sure this behavior is mentioned in the docs.

Issue #8150, opened 2024-12-27
Behavior of pasting links cannot be selected


Ali-ovo:
When I paste a link in it automatically shows the card preview, I want to be able to choose whether to paste it as a embed preview or just the link address

and then I couldn't resize the embed preview site either

Expected behavior
can control the presentation of the pasted link： either as a embed preview or just the link address.
can change the size of the embed preview.

Screenshots
paste a link like this
<img width="788" alt="image" src="https://github.com/user-attachments/assets/faa9a0ab-df2c-4d74-8ba3-b61aca720d24" />

embed  resize like this:
<img width="785" alt="image" src="https://github.com/user-attachments/assets/5cf22e71-9598-486e-aa32-0dc00eb05706" />

Browser: Google Chrome
Device: MacBook Pro
OS: macOS 14.6.1
Installation: Self-hosted server
Outline Version: 0.81.1

PR #8229, opened 2025-01-13
feat: add paste menu

Ali-ovo:
close: https://github.com/outline/outline/issues/8150

add paste menu:

image


tommoor:
Hey, thanks for the contribution – will thoroughly review soon

app/editor/components/PasteMenu.tsx
```diff
@@ -0,0 +1,43 @@
...
+type Props = Omit<
+  SuggestionsMenuProps,
+  "renderMenuItem" | "items" | "embeds" | "trigger"
+>;
+
+const PasteMenu = (props: Props) => {
+  const items = [
+    {
+      name: "link",
+      title: "URL",
```
tommoor, line 16: These strings need to be wrapping in `t` for translation (see elsewhere)

app/editor/components/PasteMenu.tsx
```diff
@@ -0,0 +1,43 @@
...
+    {
+      name: "embed",
+      title: "Embed",
+    },
+  ];
+
+  return (
+    <SuggestionsMenu
+      {...props}
+      trigger=""
```
tommoor, line 27: We should make `trigger` prop optional

app/editor/extensions/PasteHandler.tsx
```diff
@@ -165,9 +162,8 @@ export default class PasteHandler extends Extension {
                     }
                     const matches = embed.matcher(text);
                     if (matches) {
-                      this.editor.commands.embed({
-                        href: text,
-                      });
+                      this.insertLink(text);
+                      this.showPasteMenu(text);
```
tommoor, line 166: This code path can be removed entirely if the behavior is now the same for all pasted links

app/editor/extensions/PasteHandler.tsx
```diff
@@ -193,18 +189,19 @@ export default class PasteHandler extends Extension {
                           const title = `${
                             hasEmoji ? document.icon + " " : ""
                           }${document.titleWithDefault}`;
-                          insertLink(`${document.path}${hash}`, title);
+                          this.insertLink(`${document.path}${hash}`, title);
```
tommoor, line 192: There will be a small merge conflict here with #7680 in the next few days

app/editor/extensions/PasteHandler.tsx
```diff
@@ -310,6 +307,89 @@ export default class PasteHandler extends Extension {
...
+  }
+
+  private replaceOldLink(href: string) {
+    const { view } = this.editor;
+    const { state } = view;
+    const preLen = href.length;
+
+    const transaction = view.state.tr.delete(
+      state.selection.from - preLen,
+      state.selection.to
```
tommoor, line 347: This assumes that the selection has not moved, but unfortunately it can move either by typing extra space characters or through another user editing the document – in both cases the embed will be placed in the wrong position.

There are two options for a more reliable approach:

- Insert a decoration around the link with an ID upon paste, the decoration will be moved with document edits so it can be used for the final insert position, similar to what happens with image uploads
- Search the doc for the _current_ position of the link mark matching the paste data, do not rely on the selection

app/editor/components/PasteMenu.tsx
Ali-ovo: I found that `trigger` was mandatory in `SuggestionsMenuProps`, so I couldn't make her optional

app/editor/extensions/PasteHandler.tsx
Ali-ovo: i got it

app/editor/components/PasteMenu.tsx
Ali-ovo: done

app/editor/extensions/PasteHandler.tsx
Ali-ovo: I don't think it's going to be a big problem

app/editor/extensions/PasteHandler.tsx
Ali-ovo: I actually don't quite understand what to do with this , I haven't touched this feature before, can you fix this for me if possible 🙏

tommoor: It is a bit complicated, I will probably merge https://github.com/outline/outline/pull/8225 first and then look into this

Ali-ovo: Okay, just in time for me to deal with the above conflict when you're done merging.

tommoor: I'm working on the refactor to allow for multiplayer, should have something in the next day or so

app/editor/extensions/PasteHandler.tsx
```diff
@@ -363,20 +422,39 @@ export default class PasteHandler extends Extension {
...
+          }
+          case "embed": {
+            const result = this.findPlaceholder(state, this.state.pastedText);
+
+            if (result) {
+              view.dispatch(
+                state.tr.replaceWith(
+                  result[0],
+                  result[1],
+                  this.editor.schema.text("")
```
Ali-ovo, line 447: image

When I select embed, the page gives me an error and doesn't take effect, so I think it's a problem here.

tommoor: Yep, you're right – pull the latest commit :)

PR #8237, opened 2025-01-15
fix: NodeView does not re-render when editable prop changes

### LINKED ISSUE #8220: Can't download attachments after editing them or adding them to a document
After editing a document and adding or editing an existing attachment, you cannot download the attachment until you first navigate to a different document and then back again.

1. Click "Edit" on a document
2. add an attachment to the document or edit/select an existing attachment
3. Click "Done Editing"
4. Hover the mouse over the attachment. The download icon is missing, and clicking on the attachment does not download the document.
5. Navigate away from the document and then back to it, and the download button will now work.

The behavior is the same in the Windows desktop version and the online version

See attached video

 - OS: Windows 11
 - Browser Chrome Version 131.0.6778.265 (Official Build) (64-bit)

https://github.com/user-attachments/assets/3249bbe0-1dc2-4597-b665-9c8b2a94d7ae

tommoor:
closes #8220

### DIFF
--- a/app/editor/index.tsx
+++ b/app/editor/index.tsx
@@ -1,5 +1,6 @@
 /* global File Promise */
 import { PluginSimple } from "markdown-it";
+import { Observer } from "mobx-react";
 import { darken, transparentize } from "polished";
 import { baseKeymap } from "prosemirror-commands";
 import { dropCursor } from "prosemirror-dropcursor";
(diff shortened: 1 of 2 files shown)


app/editor/index.tsx
```diff
@@ -849,7 +856,11 @@ export class Editor extends React.PureComponent<
               Object.values(this.widgets).map((Widget, index) => (
                 <Widget key={String(index)} rtl={isRTL} readOnly={readOnly} />
               ))}
-            {Array.from(this.renderers).map((view) => view.content)}
+            <Observer>
```
tommoor, line 859: Shame to add this, but it's one of the only remaining class components

hmacr: https://github.com/outline/outline/blob/8fec6758b899b867c1479f342a8ecd5d3d8922d2/app/editor/components/ComponentView.tsx#L70

Would something like this work? 👇 I didn't test it - technically it should work fine..
```JS
this.renderer = new NodeViewRenderer(this.dom, observer(this.component), this.props);
```

tommoor: Maybe, I will check!

tommoor: No luck, it needs to be around the `computed` content here:

https://github.com/outline/outline/blob/70f103742379aa27e7fae6dc5b822138b34c0271/app/editor/components/NodeViewRenderer.tsx#L17

hmacr: Ok 😞

PR #8252, opened 2025-01-17
Remove namespace filter for publishing events

hmacr:
Towards #7920 

This PR removes the model namespace as a filter for publishing events. As such, it'll be used to customise the event's namespace when it's different from the model's table name.

The affected models are listed below.

| Model | Table name | Event namespace | Namespace customisation |
| ApiKey | apiKeys | api_keys | Retained |
| Attachment | attachments | attachments | Removed |
| Collection | collections | collections | Removed |
| Comment | comments | comments | Removed |
| Document | documents | documents | Removed |
| FileOperation | file_operations | fileOperations | Retained |
| Group | groups | groups | Removed |
| GroupMembership | group_permissions | N/A | N/A |
| GroupUser | group_users | groups | Added |
| Pin | pins | pins | Removed |
| Reaction | reactions | N/A | N/A |
| Share | shares | shares | Removed |
| Star | stars | stars | Removed |
| Subscription | subscriptions | subscriptions | Removed |
| WebhookSubscription | webhook_subscriptions | webhookSubscriptions | Retained |

### DIFF
--- a/server/routes/api/collections/collections.ts
+++ b/server/routes/api/collections/collections.ts
@@ -211,24 +211,22 @@ router.post(
     authorize(user, "update", collection);
     authorize(user, "read", group);
 
-    const [membership, created] = await GroupMembership.findOrCreateWithCtx(
-      ctx,
-      {
-        where: {
-          collectionId: id,
-          groupId,
-        },
-        defaults: {
-          permission,
-          createdById: user.id,
-        },
-        lock: transaction.LOCK.UPDATE,
-      }
-    );
+    const [membership, created] = await GroupMembership.findOrCreate({
+      where: {
+        collectionId: id,
+        groupId,
+      },
+      defaults: {
+        permission,
+        createdById: user.id,
+      },
+      lock: transaction.LOCK.UPDATE,
+      ...ctx.context,
+    });
 
     if (!created) {
       membership.permission = permission;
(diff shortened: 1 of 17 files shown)


server/routes/api/collections/collections.ts
```diff
@@ -211,24 +211,22 @@ router.post(
...
-        },
-        defaults: {
-          permission,
-          createdById: user.id,
-        },
-        lock: transaction.LOCK.UPDATE,
-      }
-    );
+    const [membership, created] = await GroupMembership.findOrCreate({
+      where: {
```
tommoor, line 215: Not following why these were reverted?

hmacr: We don't want to trigger the auto events by using the `withCtx` methods. GroupMembership is kind of an exception - it can be in `collections` or `documents` namespace. This is handled by the hooks in the GroupMembership model.

tommoor: True, we had the `create: false` option for this though I thought

hmacr: Ah right - `withCtx` methods were designed to always publish events - I thought it's simpler to reason about using `withCtx` vs vanilla persistence methods.
Should we change it?

tommoor (approved):

Issue #8239, opened 2025-01-15
Notify the group members when a group is added to a collection


hmacr:
Same as adding groups to a document, we should notify the members when the group is provided access to a collection.

PR #8263, opened 2025-01-21
Prevent duplicate emails when user has existing access to a document.

hmacr:
Related #8239


server/queues/tasks/DocumentAddGroupNotificationsTask.ts
```diff
@@ -1,12 +1,31 @@
...
+
+    if (!permission) {
+      Logger.debug(
+        "task",
+        `Suppressing notification for group ${event.modelId} as permission not available`,
+        event.data
+      );
+    }
+
+    const addUserTaskData: DocumentUserEvent["data"] = {
```
hmacr, line 23: This will be unnecessary when `UserMembership` is migrated to auto events - draft PR until then to avoid rework?

server/models/User.ts
```diff
@@ -616,6 +635,127 @@ class User extends ParanoidModel<
...
+  }
+
+  /**
+   * Returns the user's highest access level to a document.
+   *
+   * @param {string} documentId The document to check.
+   * @param {string} skipMembershipId The membership to skip when comparing the existing permissions.
+   * @returns {DocumentPermission | undefined} Highest access level, if it exists.
+   */
+  public async getDocumentPermission(
```
tommoor, line 678:
```suggestion
  private async getDocumentPermission(
```

I don't love this being on `user`, but if it must we can make it private

server/queues/tasks/DocumentAddGroupNotificationsTask.ts
```diff
@@ -1,12 +1,28 @@
...
 export default class DocumentAddGroupNotificationsTask extends BaseTask<DocumentGroupEvent> {
   public async perform(event: DocumentGroupEvent) {
+    const permission = event.changes?.attributes.permission as
+      | DocumentPermission
+      | undefined;
+
+    if (!permission) {
+      Logger.debug(
+        "task",
+        `Suppressing notification for group ${event.modelId} as permission not available`,
```
tommoor, line 18: Do you remember what is the situation where this is the case?

server/queues/tasks/DocumentAddUserNotificationsTask.ts
```diff
@@ -1,27 +1,62 @@
...
+      return;
     }
+
+    const hasHigherPermission = await recipient.hasHigherDocumentPermission({
+      documentId: event.documentId,
+      permission,
+      skipMembershipId: event.modelId,
+    });
+
+    if (hasHigherPermission) {
```
tommoor, line 40: So this suppresses if they already have a higher permission, what if they have the same permission already – shouldn't we suppress in that situation too? 🤔

server/queues/tasks/DocumentAddGroupNotificationsTask.ts
hmacr: It was added during Group model auto event migration to account for older events which won't have `changes` field set.
I suppose it's not needed now.

server/models/User.ts
hmacr: yes, I tried moving just this method to the existing "policies" file (renamed to permissions), but it led to circular imports with `User`, so I had to move both methods out (figured a separate `getDocumentPermission` would be beneficial in future).

tommoor (approved):

PR #8264, opened 2025-01-21
fix: Render TOC only when the shared document has headings

### LINKED ISSUE #7944: Static-share page's content not centered when window width changes
- In editor mode, the content's left edge is adjusted accordingly when the browser width is reduced:
    - 
- In it's corresponding static share, the left-edge is stuck and often leaves a lot of blank space on the left:
    -

hmacr:
Closes #7944

### DIFF
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -225,7 +225,9 @@ const SharedDocument = observer(
       return null;
     }
 
-    const tocPosition = response.team?.tocPosition ?? TOCPosition.Left;
+    const tocPosition = hasHeadings
+      ? response.team?.tocPosition ?? TOCPosition.Left
+      : false;
     setDocument(response.document);
 
     return (
@@ -234,7 +236,6 @@ const SharedDocument = observer(
         document={response.document}
         sharedTree={response.sharedTree}
         shareId={shareId}
-        shareHasHeadings={hasHeadings}
         tocPosition={tocPosition}
         readOnly
       />
--- a/app/scenes/Document/Shared.tsx
+++ b/app/scenes/Document/Shared.tsx
@@ -137,13 +137,6 @@ function SharedDocumentScene(props: Props) {
     }
   }, [ui, documentId]);
 
-  // On initial load, share's TOC should be visible.
-  React.useEffect(() => {
-    if (ui.tocVisible === undefined) {
-      ui.set({ tocVisible: true });
-    }
-  }, [ui]);
-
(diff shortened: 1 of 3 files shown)


app/scenes/Document/Shared.tsx
```diff
@@ -137,6 +137,13 @@ function SharedDocumentScene(props: Props) {
     }
   }, [ui, documentId]);
 
+  // On initial load, share's TOC should be visible.
+  React.useEffect(() => {
+    if (ui.tocVisible === undefined) {
+      ui.set({ tocVisible: true });
```
tommoor, line 143: I'm not sure about this, one – we'll get a frame of `ui.tocVisible === undefined` and two if you're logged in when visiting the share link it will enable TOC in your settings every time.

hmacr: Ah, did not account for the logged in user.
This was added to fix TOC button's inconsistent state transition (undefined -> true, when user wishes to hide the TOC shown by default) - just pushed a commit that handles it for shares only.

app/scenes/Document/Shared.tsx
```diff
@@ -217,33 +217,30 @@ function SharedDocumentScene(props: Props) {
...
+    const tocPosition = response.team?.tocPosition ?? TOCPosition.Left;
+    setDocument(response.document);
+
+    return (
+      <Document
+        abilities={EMPTY_OBJECT}
+        document={response.document}
+        sharedTree={response.sharedTree}
+        shareId={shareId}
+        shareHasHeadings={hasHeadings}
```
tommoor, line 237: It is a bit weird to pass this in from the outside, no? Is it just because it's kind of hard to get the information in the class component

hmacr: That yea - although it can be overcome by using `contextType` I think.
But primarily, to avoid rendering the document when the first heading is added (or) the last heading is removed. With shares, we're guaranteed that edits can't happen.

tommoor: > With shares, we're guaranteed that edits can't happen.

That's fair, just kind of hate having yet-another-prop here on document, and especially one that's only applicable to shares, the point of `Document/Shared` was ideally to isolate that logic. 

Random thought – what about an option for `tocPosition={false}` or "disabled"?

hmacr: `tocPosition={false}` - Yea, this is a nice alternative 👍

tommoor (approved):

PR #8274, opened 2025-01-22
Use explicitly passed `title` and `text` in favor of template title and text

### LINKED ISSUE #8272: `text` and `title` should take precedence when passing `templateId` to `documents.create`
> if I add a templateId at create time, it works too, but the title is the one of the template, not the one in title data.

This is a bug, the passed data should take precedence. In the meantime if you want to workaround it then you would have to read the template separately and apply in your code before POSTing.

_Originally posted by @tommoor in https://github.com/outline/outline/discussions/8268#discussioncomment-11909751_

apoorv-mishra:
Closes #8272

### DIFF
--- a/server/commands/documentCreator.ts
+++ b/server/commands/documentCreator.ts
@@ -35,8 +35,8 @@ type Props = Optional<
 };
 
 export default async function documentCreator({
-  title = "",
-  text = "",
+  title,
+  text,
   icon,
   color,
   state,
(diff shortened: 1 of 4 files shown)


server/routes/api/documents/schema.ts
```diff
@@ -331,10 +331,10 @@ export const DocumentsCreateSchema = BaseSchema.extend({
     id: z.string().uuid().optional(),
 
     /** Document title */
-    title: z.string().default(""),
+    title: z.string().optional(),
```
apoorv-mishra, line 334: > Using ?? to handle the case where if a user intends to use empty title for the doc as opposed to reusing template title. If the user passes no title at all, then we fallback to template title.

☝️ required this particular change in schema.

server/commands/documentCreator.ts
```diff
@@ -101,14 +101,20 @@ export default async function documentCreator({
...
-      ),
-      text: TextHelper.replaceTemplateVariables(
-        templateDocument ? templateDocument.text : text,
-        user
-      ),
+      title:
+        title ??
+        (templateDocument
+          ? template
+            ? templateDocument.title
```
apoorv-mishra, line 108: IMO, it makes sense to retain template variables if we're creating a template from another template.

tommoor: clever, makes sense

tommoor (approved):

PR #8284, opened 2025-01-23
fix: Redirect previous subdomains #8266

### LINKED ISSUE #8266: Redirect previous subdomains
If a workspace changes subdomain it would be nice to redirect requests to the new one if here are no other workspaces that are using the subdomain as their _primary_. This can be achieved by:

- Add a new column: `previousSubdomains` with is an array of previous subdomains capped to a length (lets say 5)
- Fallback to searching this if no valid `subdomain` is found
- Respond with a redirect

SachinC2003:
## Changes
- Added a new field `previousSubdomains` to store up to 5 previous subdomains.
- Implemented fallback logic for subdomains when the current subdomain is not valid.

## Motivation
This update allows the system to redirect users to the new subdomain if there are no other workspaces using the old subdomain, improving the user experience.

## Related Issue
Fixes #8266

## Testing
- Tested by manually changing subdomains and verifying the redirection logic works as expected.

### DIFF
--- a/app/models/Team.ts
+++ b/app/models/Team.ts
@@ -54,6 +54,10 @@ class Team extends Model {
 
   @Field
   @observable
+  previousSubdomains: string[] | null | undefined;
+  
+  @Field
+  @observable
   defaultUserRole: UserRole;
 
   @Field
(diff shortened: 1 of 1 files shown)


tommoor:
This isn't close to being the full solution I'm afraid, previous subdomains need storing in the database and the redirect would happen on the server. In fact, I don't think _any_ clientside changes in `/app/` are neccessary

PR #8286, opened 2025-01-23
fix: Skip events for sourced group memberships

hmacr:
Not a functional impact, however dangling events are created for sourced memberships which fail in the background task.

### DIFF
--- a/server/models/GroupMembership.ts
+++ b/server/models/GroupMembership.ts
@@ -344,6 +344,7 @@ class GroupMembership extends ParanoidModel<
         },
         {
           transaction,
+          hooks: false,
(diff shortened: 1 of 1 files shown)


server/models/GroupMembership.ts
```diff
@@ -344,6 +344,7 @@ class GroupMembership extends ParanoidModel<
         },
         {
           transaction,
+          hooks: false,
```
hmacr, line 347: An alternative would have been to check `model.sourceId` in the event publishing hook, this looks cleaner imo.

tommoor (approved):

PR #8293, opened 2025-01-23
chore: Bump rollup-plugin-webpack-stats to v2.0.1

vio:
[`rollup-plugin-webpack-stats@v2`](https://github.com/relative-ci/rollup-plugin-webpack-stats/releases) introduces a new way of detecting the correct chunk "initial" flag based on the dependency tree.

According to the report generated on the fork, multiple chunks are affected by the change, but from the tests the initial flag is set correctly (including the index entry): https://bit.ly/40IHFNz.

---

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -353,7 +353,7 @@
     "prettier": "^2.8.8",
     "react-refresh": "^0.14.2",
     "rimraf": "^2.5.4",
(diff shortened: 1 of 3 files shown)


tommoor:
Thanks 🙏, looks like CI didn't build so I'll test locally first

tommoor (approved):

PR #8297, opened 2025-01-25
feat: API key resource scoping

### LINKED ISSUE #8186: API key scopes
Add `scope` column to `ApiKey`. Scopes is a space separated list of endpoints (or prefixes) which are accessible to the key. For example, valid values:

`documents`
`documents.info`
`documents.list documents.info`
`collections.info documents`

- Because the scope is a list of endpoint definitions it can be easily and cheaply checked as part of the authentication middleware.
- An `ApiKey` with empty `scope` has no limitations, the same access as the user.

tommoor:
closes #8186


server/models/ApiKey.test.ts
```diff
@@ -44,4 +44,49 @@ describe("#ApiKey", () => {
...
     });
   });
+
+  describe("canAccess", () => {
+    it("should return true for all resources if no scope", async () => {
+      const apiKey = await buildApiKey({
+        name: "Dev",
+      });
+
+      expect(apiKey.canAccess("documents.info")).toBe(true);
```
tommoor, line 54: With another 12 hours of ruminating I think these should probably be URLs, like:

`/api/documents.info` 

This allows for easier inclusion of other scope types such as `read` later, and we can differentiate with the includes of `/`

hmacr: I'm thinking if we're better off with a scope object, something like
```
{
  "read": ["documents.info", "users.info"],
  "read_write": ["documents.create"],
  ...some more access types
}
```
With this, a scope middleware could be like `(accessType: "read" | "read_write", resource: string)` which can be integrated with the APIs incrementally.

tommoor: Read write is already encoded in the endpoint type so it's duplicative. I'm also trying to make this compatible with OAuth flow

hmacr: Fair :)

tommoor:
I will update to store the url paths rather than just method names and use native string array in Postgres

app/scenes/ApiKeyNew/index.tsx
```diff
@@ -105,6 +107,20 @@ function ApiKeyNew({ onSubmit }: Props) {
           autoFocus
           flex
         />
+        <Input
+          type="text"
+          label={t("Scopes")}
```
hmacr, line 112: I'm not sure if this should be a text input - very easy for typos, missing space between two scopes, user needing to consult doc / code?
Maybe a follow-on PR to convert this to a list of checkbox? (like GitHub's token page)

app/scenes/Settings/components/ApiKeyListItem.tsx
```diff
@@ -35,16 +36,29 @@ const ApiKeyListItem = ({ apiKey }: Props) => {
         &middot;{" "}
       </Text>
       {apiKey.lastActiveAt && (
-        <Text type={"tertiary"}>
-          {t("Last used")} <Time dateTime={apiKey.lastActiveAt} addSuffix />{" "}
-          &middot;{" "}
+        <Text type="tertiary">
+          {t("Last used")} <Time dateTime={apiKey.lastActiveAt} addSuffix />
+          {apiKey.expiresAt && <> &middot; </>}
```
hmacr, line 41: Conditional rendering is not needed here - `middot` is missing when the key has no expiry.

Screenshot 2025-01-28 at 7 48 32 PM

app/scenes/ApiKeyNew/index.tsx
tommoor: I think ideally this would be a token input, perhaps with some good auto suggest? Not sure – I'm okay leaving it loosey for now and will add documentation – but agreed overall.

PR #8301, opened 2025-01-27
feat: Upload remote-hosted images on paste

### LINKED ISSUE #5713: Upload remote images to storage when pasting content into editor
- During paste detect remote image urls
- Kick off upload async
- Replace urls in attributes when upload completes asyncronously

### Discussed in https://github.com/outline/outline/discussions/5069

<sup>Originally posted by **WarlockFish** March 19, 2023</sup>
 Sometimes, the articles are very good, and I want to import them into the outline. I can use the "[Web Clipper](https://github.com/webclipper/web-clipper)"  to quickly save text    to markdown file  and import the outline, but images cannot be quickly imported. Is there any way to quickly import text and images </div>

tommoor:
closes #5713


server/routes/api/attachments/attachments.ts
```diff
@@ -105,6 +111,74 @@ router.post(
...
+      id: modelId,
+      name,
+      userId: user.id,
+    });
+
+    // Does not use transaction middleware, as attachment must be persisted
+    // before the job is scheduled.
+    const attachment = await sequelize.transaction(async (transaction) =>
+      Attachment.createWithCtx(
+        { ...ctx, transaction },
```
hmacr, line 151: This needs to be `{...ctx, context: { ...ctx.context, transaction }` - right now, event insertion happens outside the transaction (also missing the auth-type and ip)

server/routes/api/attachments/attachments.ts
```diff
@@ -105,6 +111,74 @@ router.post(
...
 );
 
+router.post(
+  "attachments.createFromUrl",
+  rateLimiter(RateLimiterStrategy.TwentyFivePerMinute),
+  auth(),
+  validate(T.AttachmentsCreateFromUrlSchema),
+  async (ctx: APIContext<T.AttachmentCreateFromUrlReq>) => {
+    const { url, documentId, preset } = ctx.input.body;
+    const { auth, transaction } = ctx.state;
```
hmacr, line 121: We don't use the transaction middleware currently - can remove here and its touchpoints?

While we're here, a _nit_ 🙈  - comment in L124 does not align with the scope; can be modified or removed?

server/queues/tasks/UploadAttachmentFromUrlTask.ts
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      const res = await FileStorage.storeFromUrl(
+        props.url,
+        attachment.key,
+        attachment.acl
+      );
+
+      if (res?.url) {
+        const ctx = createContext({ user: attachment.user });
+        await attachment.updateWithCtx(ctx, {
```
hmacr, line 32: I think we need to add an `attachments.update` event type [here](https://github.com/outline/outline/blob/23a89c4d7bc66a9db542a37a7e583c5ed5bc351a/server/types.ts#L110) and handle it in `DeliverWebhookTask`.

hmacr (commented):
Works great!

server/routes/api/attachments/attachments.ts
tommoor: > We don't use the transaction middleware currently - can remove here and its touchpoints?

Shame we don't have type safety here :(

server/queues/tasks/UploadAttachmentFromUrlTask.ts
tommoor: All of the `attachments.` events are actually ignored in `DeliverWebhookTask` as they're kind of considered an internal concern but will add for completeness

hmacr: Should be fine, I think - `DeliverWebhookTask` just adds a warning log for unknown events.

hmacr (approved):

PR #8317, opened 2025-01-28
Toggle block

### LINKED ISSUE #7676: Add a toggle element in the editor
The problem today is that while headings are toggleable they are syncronized via multiplayer collaboration – so one user collapsing the header does so for everyone, this is very confusing.

Ideally we can make headers toggling a per-user setting, the best way to achieve this would be to first introduce a dedicated toggleable element that hides it's child nodes when collapsed. This logic will actually be much easier than that for headings, as only _children_ will be collapsed.

apoorv-mishra:
Closes #7676

### DIFF
--- a/shared/editor/nodes/Heading.ts
+++ b/shared/editor/nodes/Heading.ts
@@ -1,5 +1,5 @@
 import copy from "copy-to-clipboard";
-import { textblockTypeInputRule } from "prosemirror-inputrules";
+import { InputRule } from "prosemirror-inputrules";
 import {
   Node as ProsemirrorNode,
   NodeSpec,
@@ -7,6 +7,7 @@ import {
   Schema,
 } from "prosemirror-model";
 import { Command, Plugin, Selection } from "prosemirror-state";
+import { findWrapping } from "prosemirror-transform";
 import { Decoration, DecorationSet } from "prosemirror-view";
 import { toast } from "sonner";
 import { Primitive } from "utility-types";
@@ -277,11 +278,35 @@ export default class Heading extends Node {
     return [new FoldingHeadersPlugin(this.editor.props.id), plugin];
   }
 
-  inputRules({ type }: { type: NodeType }) {
-    return this.options.levels.map((level: number) =>
-      textblockTypeInputRule(new RegExp(`^(#{1,${level}})\\s$`), type, () => ({
-        level,
-      }))
+  inputRules({ type, schema }: { type: NodeType; schema: Schema }) {
+    return this.options.levels.map(
+      (level: number) =>
+        new InputRule(
+          new RegExp(`^(#{1,${level}})\\s$`),
+          (state, _match, start, end) => {
+            const $start = state.doc.resolve(start);
+            if (
+              !$start
+                .node(-1)
+                .canReplaceWith($start.index(-1), $start.indexAfter(-1), type)
+            ) {
+              return null;
+            }
+
+            const tr = state.tr
+              .delete(start, end)
+              .setBlockType(start, start, type, { level });
+
+            const range = tr.doc.resolve(start).blockRange();
+            const wrappers =
+              range && findWrapping(range, schema.nodes["toggle_block"]);
+            if (!wrappers) {
+              return null;
+            }
+
+            return tr.wrap(range!, wrappers);
+          }
+        )
     );
   }
 }
--- a/shared/editor/nodes/ToggleBlock.ts
+++ b/shared/editor/nodes/ToggleBlock.ts
@@ -1,0 +1,16 @@
+import { NodeSpec } from "prosemirror-model";
+import Node from "./Node";
+
+export default class ToggleBlock extends Node {
+  get name() {
+    return "toggle_block";
+  }
+
+  get schema(): NodeSpec {
+    return {
+      content: "block+",
+      group: "block",
+      toDOM: () => ["div", { class: "toggle-block" }, 0],
+    };
+  }
+}
--- a/shared/editor/nodes/index.ts
+++ b/shared/editor/nodes/index.ts
@@ -42,6 +42,7 @@ import TableCell from "./TableCell";
 import TableHeader from "./TableHeader";
 import TableRow from "./TableRow";
 import Text from "./Text";
+import ToggleBlock from "./ToggleBlock";
 import Video from "./Video";
 
 type Nodes = (typeof Node | typeof Mark | typeof Extension)[];
@@ -107,6 +108,7 @@ export const richExtensions: Nodes = [
   TemplatePlaceholder,
   Math,
   MathBlock,
+  ToggleBlock,
   // Container type nodes should be last so that key handlers are registered for content inside
   // the container nodes first.
   ...listExtensions,


shared/editor/nodes/Heading.ts
```diff
@@ -277,11 +278,35 @@ export default class Heading extends Node {
...
+              .setBlockType(start, start, type, { level });
+
+            const range = tr.doc.resolve(start).blockRange();
+            const wrappers =
+              range && findWrapping(range, schema.nodes["toggle_block"]);
+            if (!wrappers) {
+              return null;
+            }
+
+            return tr.wrap(range!, wrappers);
```
apoorv-mishra, line 307: Attempted this through `inputRules` on the `ToggleBlock` node and using `wrappingInputRule` but, couldn't make it work because the required regex was similar to that of `Heading` node and observed that only one of the input rules executes, depending on the order of nodes in `shared/editor/nodes/index.ts`. So, had to put the wrapping login inside of `Heading` node itself.

_Note to self_: A cleaner way could be to _detect_ when a particular node is created/inserted, then, capture that transaction and `tr.wrap` that particular node inside a toggle block. That way, the logic could be moved out of the node and can be shared across multiple nodes. Although, couldn't find a direct way to detect node creation/insertion.

tommoor: Can we leave Headings how are they are and implement toggle block entirely separately? The toggle mechanism is quite different

apoorv-mishra: Hmm, so heading won't be wrapped in a toggle block by default? Is the toggle block supposed to be triggered via command menu, for example?

tommoor: Toggle is a new element inserted through the block menu that lets you show and hide it's children. We can't change the structure of existing headings as it would modify existing docs. Also, the logic is quite different as heading toggle iterates through siblings until it finds a heading of the _same level or above_ rather than just children

apoorv-mishra: > We can't change the structure of existing headings as it would modify existing docs.

Noted 👍

> the logic is quite different as heading toggle iterates through siblings until it finds a heading of the same level or above rather than just children

Yeah, I mistakenly presumed this needed changing altogether with toggle block, which would make it easy to toggle _just_ its children. Thanks for the clarity 🙏

app/editor/menus/block.tsx
```diff
@@ -214,5 +214,11 @@ export default function blockMenuItems(dictionary: Dictionary): MenuItem[] {
       keywords: "diagram flowchart",
       attrs: { language: "mermaidjs" },
     },
+    {
+      name: "toggle_block",
+      title: dictionary.toggleBlock,
+      icon: <Img src="/images/mermaidjs.png" alt="Mermaid Diagram" />,
```
apoorv-mishra, line 220: - [ ]  Replace this with the toggle icon

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -0,0 +1,28 @@
...
+export default class ToggleBlock extends Node {
+  folded = false;
+
+  get name() {
+    return "toggle_block";
+  }
+
+  get schema(): NodeSpec {
+    return {
+      content: "toggle_head toggle_body?",
```
apoorv-mishra, line 14: `toggle_block` is composed of `toggle_head` & `toggle_body` now. This choice was made so that it's convenient to apply decoration to hide _only_ `toggle_body`. Also, the `toggle_body` can be easily _indented_ to distinguish it from whatever is outside of toggle block.
Screenshot 2025-02-19 at 9 04 32 PM

tommoor: Okay, will have to test! Not totally against this solution – though I feel that toggle_body isn't _strictly_ necessary as we can always hide all the children except the first (the heading)

apoorv-mishra: Also, simplifies the problem indenting toggle block content. Not strictly necessary though, agreed!

shared/editor/components/Styles.ts
```diff
@@ -1903,6 +1903,54 @@ del[data-operation-index] {
...
+  &.folded {
+    &:dir(rtl) {
+      --rotate-by: 90deg;
+    }
+    &:dir(ltr) {
+      --rotate-by: -90deg;
+    }
+    > .toggle-block-content > :not(:first-child) {
+        display: none;
+      }
```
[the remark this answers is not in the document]
apoorv-mishra: Looks like we don't yet need `toggle_head` and `toggle_body`. The problem presented here has been dealt with in [5dc5e53](https://github.com/outline/outline/pull/8317/commits/5dc5e53c736694a5725dbe8b0221756f1a72f617).
The key is to apply a separate decoration to the first child node of `toggle_block` by setting [`nodeName`](https://prosemirror.net/docs/ref/#view.DecorationAttrs.nodeName), which wraps the first child node's DOM node in a separate DOM node. Then, we can simply exclude this separate DOM node while hiding all children.

https://github.com/user-attachments/assets/d2e7ddf7-1af7-4bb7-a3a3-de67e587c43b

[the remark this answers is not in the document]
Enissay: I didnt follow the full discussion here but I hope the toggle block doesnt require the first line to have any specific formatting as I see in you last vid. It could be just a normal text to fold some large bloc (typically code in my case).

shared/styles/globals.ts
```diff
@@ -14,6 +14,11 @@ export default createGlobalStyle<Props>`
     box-sizing: border-box;
   }
 
+  html {
```
tommoor, line 17: I'm not really following why all these additions were needed on this branch?

shared/editor/rules/toggleBlocks.ts
```diff
@@ -0,0 +1,9 @@
+import MarkdownIt from "markdown-it";
+import customFence from "markdown-it-container";
+
+export default function toggleBlocks(md: MarkdownIt): void {
+  return customFence(md, "toggle_block", {
```
tommoor, line 5:
```suggestion
  return customFence(md, "toggle", {
```

Lets just call it "toggle" so it ends up `container_toggle`

shared/editor/components/Styles.ts
```diff
@@ -1935,6 +1943,98 @@ del[data-operation-index] {
...
+    }
+  }
+
+  > .toggle-block-button {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+
+    &:has(+ .toggle-block-content > .toggle-block-head > p) {
```
tommoor, line 1976: Similar, not following why these overrides are required lmk!

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -0,0 +1,605 @@
...
+  CHANGE,
+  FOLD,
+  UNFOLD,
+}
+
+export default class ToggleBlock extends Node {
+  static pluginKey = new PluginKey<DecorationSet>("toggleBlockPlugin");
+
+  get name() {
+    return "container_toggle_block";
```
tommoor, line 59:
```suggestion
    return "container_toggle";
```

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -0,0 +1,605 @@
...
+  }
+
+  get schema(): NodeSpec {
+    return {
+      content: "(paragraph | heading) block*",
```
tommoor, line 64: 👍

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -0,0 +1,605 @@
...
+
+  toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
+    state.write(state.repeat("+", 3 + height(node)) + "\n");
+    state.renderContent(node);
+    state.write(state.repeat("+", 3 + height(node)) + "\n");
+  }
+
+  parseMarkdown(): ParseSpec | void {
+    return {
+      block: "container_toggle_block",
```
tommoor, line 476:
```suggestion
      block: "container_toggle",
```

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -0,0 +1,605 @@
...
+      "Shift-Tab": liftLastBlockOutOf(type),
+    };
+  }
+
+  commands({ type }: { type: NodeType }) {
+    return () => wrapIn(type);
+  }
+
+  toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
+    state.write(state.repeat("+", 3 + height(node)) + "\n");
```
tommoor, line 469: Is this right? I'd expect the fence to always be `+++`

tommoor (commented):
There's _a lot_ of code here, I can't really dive into all the specifics yet - it might be best to get the UX right first and then iterate on the code that's left? Amazing how involved this has ended up, so nice work!

A couple of key things that don't quite feel right from a UX point of view just yet:

- When you create a new toggle block it should start opened so that you can hit `enter` after typing the title and go straight into adding the content.
- An empty content area should always show at least one paragraph with a placeholder so that you can see where you should be typing. We might be able to tweak/reuse the logic on the existing placeholder?
- Cmd+Enter could toggle the current toggle block open/closed
- We can set up a `+++` input rule as a shortcut to add a new toggle block
-

shared/editor/nodes/ToggleBlock.ts
apoorv-mishra: _Think [nesting](https://github.com/markdown-it/markdown-it-container/issues/33#issuecomment-551111227)..._

shared/editor/components/Styles.ts
apoorv-mishra: These are here to set the `height` of toggle button _conditionally_ so that it's aligned as expected, based on the height of adjacent `<p>` or `<h1>-<h4>` DOM nodes. If the adjacent DOM node is `<p>`, then height of button = height of `<p>`, which in turn is a product of `line-height` and `font-size` of `<p>`. It's derived for `<h1>-<h4>` in similar manner.

apoorv-mishra:
> We can set up a +++ input rule as a shortcut to add a new toggle block

What about `> `. I know it's already used for `blockquote`. We can change that to `" ` and use `> ` for toggle block? Fewer keystrokes compared to `+++`.

tommoor:
We can't change blockquote as it's based on markdown, also been that way for 6 years so there's a lot of muscle memory!

tommoor:
Yep, but the input rule is based on people writing markdown naturally :)

apoorv-mishra:
So, the muscle memory is much older than 6 years then... ;)

shared/editor/nodes/ToggleBlock.ts
apoorv-mishra: Just pushed this change. Make sure to _erase_ all toggle blocks in your existing test docs before pulling latest changes @tommoor

tommoor: roger that

Uriziel01 (approved):
Just had a quick look. Also, can't we implement missing/undefined behaviour based on https://linear.app/ as it already has a working `Collapsible section` there? Also the comments from @tommoor suggest this is the direction he still wants it to go here.

image

Uriziel01 (changes requested):
Just had a quick look. Also, can't we implement missing/undefined behaviour based on https://linear.app/ as it already has a working `Collapsible section` there? Also the comments from @tommoor suggest this is the direction he still wants it to go here.

image

app/editor/menus/formatting.tsx
```diff
@@ -163,6 +165,15 @@ export default function formattingMenuItems(
       attrs: { level: 2 },
       visible: !isCodeBlock && (!isMobile || isEmpty),
     },
+    {
+      name: "container_toggle",
+      // tooltip: dictionary.quote,
+      // shortcut: `${metaDisplay}+]`,
```
apoorv-mishra: - [ ] Decide upon tooltip & shortcut.

tommoor:
Unfortunately this is so big it's really difficult to review from a code perspective. I'm inclined to merge it, but to begin with behind a flag on the "Features" page to enable – this also gives us a path to testing it with the collapsible headings still in place as it becomes opt-in for a period

tommoor:
Did a functional test, largely seems to work really well at this point – feedback:

- [ ] Toggles should start in the collapsed state – so if you go to a document that you've never been to before that has toggles they should all be collapsed. This is the key difference to regular content.
- [ ] `Tab` behavior is a bit confusing, hitting tab inside a toggle causes the entire thing to get nested – is this the designed/expected behavior?

apoorv-mishra:
- [x] Replace [`uuid`](https://github.com/outline/outline/pull/10318)

michabbb:
**please:** can someone please fix that ? 🙏 it´s terribly annoying to get notification for thing that haven´t changed 😢

alexhillnz:
This is missing a tooltip in the formatting menu. Wouldn't know what the button does if I hadn't seen the feature in the changelog

ados8:
Been waiting for auto collapsing blocks for ages, but this doesn't seem to be it? Unless I'm mistaken, it requires the keyboard shortcut or mouse click to close. Was really hoping for the behaviour of every time you leave/navigate away the blocks auto close.

tommoor:
> This is missing a tooltip in the formatting menu. Wouldn't know what the button does if I hadn't seen the feature in the changelog

Yea, spotted and fixed already.


> Was really hoping for the behaviour of every time you leave/navigate away the blocks auto close.

I don't know of any similar implementation that does that, but it could be an interesting user preference

ados8:
Another app I used had that by default.
The general reason to use collapsible sections is density and neatness.
Having to manual collapse them is inefficient and in my opinion against its core purpose.

tommoor:
Cmd+Enter allows you to toggle quickly. Which app?

ados8:
Problem is the keyboard shortcut only work when editing. If you navigate to the page and expand a bunch of blocks you need to compress them manually or they remain expanded.
The app was Bookstack from memory.
There's many features I miss from that app, some unique functions.

notfounnd:
@tommoor @apoorv-mishra  great feature!

I'm checking this enhanced in Outline docs guide and I get this behaviour.

When collapse an area that have toggles block, the component not fold with the collapsed area. Is it expected?

- https://docs.getoutline.com/s/guide/doc/toggles-LNZgXe0o2b

image

---

image

tommoor:
Those toggleable headers will be going away very soon as they are too similar to the new toggles but work differently (and worse).

This does seem like a bug, but probably not worth fixing given that

Issue #2722, opened 2021-11-03
Make collection descriptions collaborative 


tommoor:
Now that document collaborative editing is complete we can add this functionality to collection descriptions too.

matbrgz:
Please change the view from collection to the same of document, there is no reason to be different. I need to use embed services in collection home. I will appreciate your review on that. Thank you

matbrgz:
Up

tommoor:
It's going to happen

almereyda:
Preceded by the work merged in #5553

PR #8326, opened 2025-02-03
Move collection description into dedicated tab

tommoor:
Moves collection description to an "Overview" tab so it has much more space, fixes some animation bugs on the tabs, simplifies everything.

- Users with view-only permissions will only see an "Overview" tab if one has been written
- All users will default back to the last tab they were looking at, per-collection

Related #2722


app/scenes/Collection/index.tsx
```diff
@@ -198,135 +212,117 @@ function CollectionScene() {
...
+                        parentDocumentId: null,
+                        sort: collection.sort.field,
+                        direction: collection.sort.direction,
+                      }}
+                      showParentDocuments
+                    />
+                  </Route>
+                </>
+              ) : (
                 <Route path={collectionPath(collection.path)} exact>
```
hmacr, line 322: This needs to be `collectionPath(collection.path, "recent")` to load/show documents in an archived collection.

There are quite a few path variants to keep track of, I'm thinking if we can DRY it up a bit - store the path types in an enum `CollectionPath` with a function `getCollectionPath(CollectionPath)`. That way, `tabProps` function would also accept `CollectionPath`.

app/scenes/Collection/index.tsx
```diff
@@ -198,135 +212,117 @@ function CollectionScene() {
...
-                    fetch={documents.fetchLeastRecentlyUpdated}
-                    options={{
-                      collectionId: collection.id,
+                />
+              </Route>
+              <Route path={collectionPath(collection.path, "overview")}>
+                {hasOverview ? (
+                  <CollectionDescription collection={collection} />
+                ) : (
+                  <Redirect
```
hmacr, line 246: Can update the local storage to "recent"? One less redirect :)

PR #8333, opened 2025-02-04
fix: Skip permission checks on the app when moving document using DnD

### LINKED ISSUE #8313: Manual sorting does not work until document policies are loaded
>  What I noticed is that sometimes when I
try to drag a document to be under another document or between two
documents it does not allow me to do it (I see the crossed red circle
icon). However, if I click on the document I am trying to move and the
document(s) where I try to move, then it works. I

Seems to be you just can't drag to reorder when the document policies have not been loaded. Pretty bad bug, @hmacr not sure if you have time to take a look at a solution – probably have to allow all drop targets and then check permission post-drop?

hmacr:
Closes #8313

Removed permission checks for document in the app - server will now perform the necessary validation.

I also noted a couple of improvs and refactors around the DnD which I'll raise as a follow-on PR (specifically, better drop indicator to differentiate between moving inside a collection vs another parent document).


app/components/WebsocketProvider.tsx
```diff
@@ -190,7 +190,7 @@ class WebsocketProvider extends React.Component<Props> {
             if (collection?.updatedAt === collectionDescriptor.updatedAt) {
               continue;
             }
-            if (!collection?.documents?.length && !event.fetchIfMissing) {
+            if (!collection?.documents && !event.fetchIfMissing) {
```
tommoor, line 193: This seems mistaken, we should check the length as it could be an empty array

app/components/Sidebar/hooks/useDragAndDrop.tsx
```diff
@@ -357,7 +451,22 @@ export function useDropToReorderDocument(
...
           });
         } else {
-          void documents.move(params);
+          try {
+            await documents.move(params);
+          } catch (err) {
+            if (err instanceof AuthorizationError) {
+              toast.error(
+                t(
+                  "You do not have permission to move {{ documentName }} document",
```
tommoor, line 460:
```suggestion
                  "The document {{ documentName }} cannot be moved here",
```

app/components/Sidebar/hooks/useDragAndDrop.tsx
```diff
@@ -233,22 +330,40 @@ export function useDropToReparentDocument(
...
+          await documents.move({
+            documentId: item.id,
+            parentDocumentId: node.id,
+          });
+          setExpanded();
+        } catch (err) {
+          if (err instanceof AuthorizationError) {
+            toast.error(
+              t(
+                "You do not have permission to move {{ documentName }} document to {{ parentDocumentName }} document",
```
tommoor, line 343:
```suggestion
                "{{ documentName }} cannot be moved within {{ parentDocumentName }}",
```

app/components/Sidebar/hooks/useDragAndDrop.tsx
```diff
@@ -162,6 +201,80 @@ export function useDragDocument(
...
+        });
+      } else {
+        try {
+          await documents.move({ documentId: id, collectionId: collection.id });
+          expandNode();
+        } catch (err) {
+          if (err instanceof AuthorizationError) {
+            toast.error(
+              t(
+                "You do not have permission to move {{ documentName }} document to {{ collectionName }} collection",
```
tommoor, line 248:
```suggestion
                "You do not have permission to move {{ documentName }} to the {{ collectionName }} collection",
```

app/components/ConfirmMoveDialog.tsx
```diff
@@ -30,12 +32,29 @@ function ConfirmMoveDialog({ collection, item, ...rest }: Props) {
...
+      await documents.move({
+        documentId: item.id,
+        collectionId: collection.id,
+        ...rest,
+      });
+    } catch (err) {
+      if (err instanceof AuthorizationError) {
+        toast.error(
+          t(
+            "You do not have permission to move {{ documentName }} document to {{ collectionName }} collection",
```
tommoor, line 45:
```suggestion
            "You do not have permission to move {{ documentName }} to the {{ collectionName }} collection",
```

app/components/WebsocketProvider.tsx
hmacr: This was needed to fix an edge case when a document is moved into an empty collection - doc structure is not being updated until a refresh in these cases.

tommoor:
Just did a round of testing and it looks like we lost the ability to drag a document to the first position in a collection

app/components/Sidebar/components/CollectionLinkChildren.tsx
```diff
@@ -36,55 +34,17 @@ function CollectionLinkChildren({
...
-  const manualSort = collection.sort.field === "index";
-  const { documents, dialogs, collections } = useStores();
+  const { documents } = useStores();
   const { t } = useTranslation();
   const childDocuments = useCollectionDocuments(collection, documents.active);
   const [showing, setShowing] = React.useState(pageSize);
+  const dummyRef = React.useRef<HTMLDivElement>(null);
 
-  // Drop to reorder document
-  const [{ isOverReorder, isDraggingAnyDocument }, dropToReorder] = useDrop({
```
tommoor, line 47: This change seems to be the culprit

tommoor:
One more – I can no longer drag a document into a manually sorted collection, it's disabled

hmacr:
> I can no longer drag a document into a manually sorted collection by dragging over the collection name, it's disabled

I'm not able to repro this 🤔  Collection is disabled when the user does not have permission to create docs. Another scenario is related to dragging membership docs onto a collection (this is left as-is to not increase the scope). Any of these cases maybe?

tommoor:
Oh I had it the wrong way around – it's for collections that are _not_ manually sorted.

hmacr:
> it's for collections that are not manually sorted

Ah ok - good catch!

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -117,59 +82,63 @@ const CollectionLink: React.FC<Props> = ({
...
-                  onOpen={handleMenuOpen}
-                  onClose={handleMenuClose}
-                />
-              </Fade>
-            )
-          }
-        />
-      </DropToImport>
+    <Relative ref={parentRef}>
+      <div ref={dropRef}>
```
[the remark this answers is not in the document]
hmacr: Thanks! I incorrectly assumed merge-refs would not work here..

tommoor (approved):

PR #8349, opened 2025-02-07
Unsubscribe from document updates when a user is removed from a document

hmacr:
This PR fixes a bug where document update notifications were sent to a subscribed user even after their access was revoked from a document.

I noticed that we don't manage documents subscriptions for groups - will create a follow-on PR once this is merged.


server/queues/processors/DocumentSubscriptionProcessor.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      default:
+    }
+  }
+
+  private async addUser(event: DocumentUserEvent, user: User) {
+    await sequelize.transaction(async (transaction) => {
+      await subscriptionCreator({
+        ctx: createContext({
```
tommoor, line 37: Thinking that we should do a little refactor soon to make a prebuilt `ctx` a parameter of `perform`

tommoor (approved):

Issue #7340, opened 2024-08-02
Subscribe to receive notifications for a whole collection


tommoor:
### Discussed in https://github.com/outline/outline/discussions/7339

<div type='discussions-op-text'>

<sup>Originally posted by **phoegon** August  2, 2024</sup>
I would like to subscribe to all pages in a collection, so I can receive notifications about all changes in a collection. This subscription should also extend to all pages created after I subscribe to the collection.</div>

PR #8368, opened 2025-02-10
Fetch subscription data using 'subscriptions.info' API

hmacr:
Allow reusing subscription data from the store.

Towards #7340

### DIFF
--- a/server/routes/api/subscriptions/subscriptions.ts
+++ b/server/routes/api/subscriptions/subscriptions.ts
@@ -4,7 +4,6 @@ import { QueryNotices } from "@shared/types";
 import subscriptionCreator from "@server/commands/subscriptionCreator";
 import { createContext } from "@server/context";
 import env from "@server/env";
-import { NotFoundError } from "@server/errors";
 import auth from "@server/middlewares/authentication";
 import { rateLimiter } from "@server/middlewares/rateLimiter";
 import { transaction } from "@server/middlewares/transaction";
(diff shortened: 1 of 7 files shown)


server/routes/api/subscriptions/subscriptions.ts
```diff
@@ -69,11 +69,11 @@ router.post(
         documentId: document.id,
         event,
       },
-      rejectOnEmpty: true,
```
tommoor, line 72: I'm not really following why change this, we should have consistency here across the API when a resource isn't found `404` is the correct response code

hmacr: I added this to ensure the store adds the subscription only when the user has an active subscription.

Agree on the consistency part - will handle the _not found_ error in the store.

hmacr: Just updated the PR to throw 404.

`rejectOnEmpty: true` by default throws a 500, so it had to be changed.

tommoor: It doesn't, but there _is_ a problem in the logging that makes it appear as though it does – if you check the network response from the server it's correct

hmacr: Ahh! UT does pass with just `rejectOnEmpty: true` 🤦‍♂️

tommoor (approved):

PR #8369, opened 2025-02-10
fix: `InviteAcceptedEmail` not sent if user joins via SSO

tommoor:
Resolves long standing TODO

### DIFF
--- a/plugins/email/server/auth/email.ts
+++ b/plugins/email/server/auth/email.ts
@@ -1,9 +1,6 @@
 import Router from "koa-router";
-import { NotificationEventType } from "@shared/types";
 import { parseDomain } from "@shared/utils/domains";
-import InviteAcceptedEmail from "@server/emails/templates/InviteAcceptedEmail";
 import SigninEmail from "@server/emails/templates/SigninEmail";
-import WelcomeEmail from "@server/emails/templates/WelcomeEmail";
 import env from "@server/env";
 import { AuthorizationError } from "@server/errors";
 import { rateLimiter } from "@server/middlewares/rateLimiter";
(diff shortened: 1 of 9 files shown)


plugins/email/server/auth/email.ts
```diff
@@ -117,26 +114,6 @@ router.get(
       return ctx.redirect("/?notice=user-suspended");
     }
 
-    if (user.isInvited) {
```
tommoor, line 120: Moved this to a processor, as here we'll only send the `InviteAcceptedEmail` if our invited user joins with email. But, it's actually possible and likely they will sign-in with SSO – in which case this email is never sent.

server/queues/processors/UserSigninProcessor.ts
```diff
@@ -0,0 +1,40 @@
...
+import WelcomeEmail from "@server/emails/templates/WelcomeEmail";
+import { User } from "@server/models";
+import { Event as TEvent, UserEvent } from "@server/types";
+import BaseProcessor from "./BaseProcessor";
+
+export default class UserSigninProcessor extends BaseProcessor {
+  static applicableEvents: TEvent["name"][] = ["users.signin"];
+
+  async perform(event: UserEvent) {
+    const inviteAccepted =
```
hmacr, line 12: I haven't tested this yet - it looks like the welcome email _might_ be skipped in oauth invite flow..? 🤔 

https://github.com/outline/outline/blob/main/server/commands/userProvisioner.ts#L133

hmacr: Nvm - UserCreatedProcessor takes care of it.

PR #8373, opened 2025-02-10
If cursor ends up under a collapsed heading, unfold the heading

### LINKED ISSUE #8327: Weird editing behaviour past a collapsed heading

### Current Behavior

https://github.com/user-attachments/assets/fb00fcb0-0b7a-4a50-a681-df1fe468c7ca

### Expected Behavior

_No response_

### Steps To Reproduce

1. Type out a heading
2. Type out some content under it
3. Collapse the heading
4. Press `Enter` multiple times
5. Notice cursor moving one line down and staying there
6. Now press `Backspace` multiple times
7. Toggle back the heading and notice the erased content without any visual cue

### Environment

```markdown
- Outline: latest `main`
```

### Anything else?

_No response_

apoorv-mishra:
Closes #8327


shared/editor/plugins/HeadingTracker.ts
```diff
@@ -0,0 +1,224 @@
...
+  constructor() {
+    super({
+      appendTransaction: (transactions, _oldState, newState) => {
+        if (!transactions.some((tr) => tr.docChanged)) {
+          return;
+        }
+        this.doc = newState.doc;
+        this.updateHeadingPositions();
+        const cursorPos = newState.selection.from;
+        this.collectPrevClosestHeadingsFrom(cursorPos);
```
apoorv-mishra, line 81: Some after-thoughts on this...
Just pointing it out that this runs upon _every_ transaction that triggers a doc update. And, as a result of that it build up the entire sorted list containing heading positions in the doc, every time! Kind of inefficient perf-wise but hopefully works _fine_ out in the wild for most cases. That being said, it'd be great to figure out a way to just update the list(and not rebuild everytime), the way decorations are mapped through changes.

shared/editor/nodes/Node.ts
```diff
@@ -19,6 +19,10 @@ export default abstract class Node extends Extension {
     return {};
   }
 
+  get attrs(): NodeSpec["attrs"] {
+    return {};
+  }
```
apoorv-mishra, line 24: This is done to please `tsc` when accessing `attrs` on _our_ nodes, for example `headingNode.attrs.collapsed`.

shared/editor/plugins/HeadingTracker.ts
tommoor: Yea, this is a lot for a very small fix – I think if we're going to do this we should at least combine the other logic which tracks headings for the table of contents and have it use this same underlying data structure?

apoorv-mishra: If the same data structure _can_ be used for another purpose, why not! I'm in favor of tackling this problem separately though.

apoorv-mishra: Okay, so there are two issues as I read it:
1. Mimicking `DecorationSet.map` to _our_ data structure so that it updates, rather than re-builds
2. Use the same data structure to _update_ headings in ToC

I'll try to tackle (2) above to re-use the data structure barring the perf-issue.

PR #8392, opened 2025-02-12
feat: Collection subscription

### LINKED ISSUE #7340: Subscribe to receive notifications for a whole collection

hmacr:
Closes #7340 

Potential merge conflict with #8503.


app/actions/definitions/documents.tsx
```diff
@@ -319,6 +319,7 @@ export const subscribeDocument = createAction({
     const document = stores.documents.get(activeDocumentId);
 
     return (
+      !document?.collection?.isSubscribed &&
```
hmacr, line 322: Hides document subscription menu item when collection subscription is active.

server/routes/api/subscriptions/schema.ts
```diff
@@ -1,36 +1,35 @@
 import { z } from "zod";
+import { SubscriptionType } from "@shared/types";
 import { ValidateDocumentId } from "@server/validation";
 import { BaseSchema } from "../schema";
 
-export const SubscriptionsListSchema = BaseSchema.extend({
-  body: z.object({
+const SubscriptionBody = z.discriminatedUnion("event", [
```
hmacr, line 6: Common body that's used across list, info, and create endpoints.

app/stores/CollectionsStore.ts
```diff
@@ -213,6 +214,20 @@ export default class CollectionsStore extends Store<Collection> {
     await star?.delete();
   };
 
+  subscribe = (collection: Collection) =>
+    this.rootStore.subscriptions.create({
+      collectionId: collection.id,
+      event: SubscriptionType.Collection,
```
tommoor, line 220: This part to me doesn't make sense, we're still subscribing to document update events – just within the _context_ of the collection, right? 

We can use the same event type here as before and I think it makes sense to

hmacr: Yep - I was skeptical of a "Collection" subscription type for this use case because it doesn't align with what it enables. Your suggestion makes the most sense 👍

app/actions/definitions/documents.tsx
tommoor: I think this is perhaps more confusing in the long run as it will be unclear why sometimes the option appears and sometimes it doesnt. We kind of of have this problem today with sharing where it can be disabled at the collection level and then folks are very unclear as to why they can't share some documents (sidenote: Should fix this too!)

So, I think it's better to leave the option and show it as subscribed – _ideally_ it would be disabled with a tooltip 😬. thoughts?

app/actions/definitions/collections.tsx
```diff
@@ -205,6 +207,66 @@ export const unstarCollection = createAction({
...
+  perform: async ({ activeCollectionId, stores, t }) => {
+    if (!activeCollectionId) {
+      return;
+    }
+
+    const collection = stores.collections.get(activeCollectionId);
+
+    await collection?.subscribe();
+
+    toast.success(t("Subscribed to collection notifications"));
```
tommoor, line 236:
```suggestion
    toast.success(t("Subscribed to document notifications"));
```

I would keep the same language

app/actions/definitions/collections.tsx
```diff
@@ -205,6 +207,66 @@ export const unstarCollection = createAction({
...
+  perform: async ({ activeCollectionId, currentUserId, stores, t }) => {
+    if (!activeCollectionId || !currentUserId) {
+      return;
+    }
+
+    const collection = stores.collections.get(activeCollectionId);
+
+    await collection?.unsubscribe();
+
+    toast.success(t("Unsubscribed from collection notifications"));
```
tommoor, line 266:
```suggestion
    toast.success(t("Unsubscribed from document notifications"));
```

server/models/helpers/NotificationHelper.ts
```diff
@@ -160,8 +168,13 @@ export default class NotificationHelper {
...
           userId: recipients.map((recipient) => recipient.id),
-          documentId: document.id,
-          event: eventType,
+          [Op.or]: [
+            {
+              collectionId: document.collectionId,
+              event: SubscriptionType.Document,
+            },
+            { documentId: document.id, event: SubscriptionType.Document },
+          ],
```
tommoor, line 177:
```suggestion
          event: SubscriptionType.Document,
          [Op.or]: [
            { collectionId: document.collectionId },
            { documentId: document.id },
          ],
```

tommoor (commented):
Looking good

app/actions/definitions/documents.tsx
hmacr: Agreed - disabled with a tooltip will be more user-friendly.

hmacr: Tooltip could be "Subscribed to all documents in collection"?

tommoor:
Small merge conflict here when you get to it, no rush

hmacr:
> Small merge conflict here

yea, I expected it after #8503 - just resolved :)

tommoor (approved):

PR #8397, opened 2025-02-13
fix: Notifications sent for insignificant changes

tommoor:
Prevents notification emails being sent for small and invisible changes

### DIFF
--- a/server/models/helpers/DocumentHelper.tsx
+++ b/server/models/helpers/DocumentHelper.tsx
@@ -1,5 +1,6 @@
 import { JSDOM } from "jsdom";
 import { Node } from "prosemirror-model";
+import ukkonen from "ukkonen";
 import { updateYFragment, yDocToProsemirrorJSON } from "y-prosemirror";
 import * as Y from "yjs";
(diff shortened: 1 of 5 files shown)


server/models/helpers/DocumentHelper.tsx
```diff
@@ -478,25 +479,26 @@ export class DocumentHelper {
...
       return false;
     }
 
-    return (
-      before.title === after.title &&
-      this.toMarkdown(before) === this.toMarkdown(after)
-    );
+    const first = before.title + this.toPlainText(before);
+    const second = after.title + this.toPlainText(after);
+    const distance = ukkonen(first, second);
```
hmacr, line 500: `ukkonen` accepts a third parameter `threshold` that's used to short-circuit/limit the distance computation.
Since we are interested in only knowing whether a certain threshold is crossed, do you think it's better to change the impl to `ukkonen(first, second, inputThreshold + 1)` and check if the returned distance is equal to `inputThreshold + 1`? This will be more performant for large documents - not very pressing since this is in the async flow. Otherwise, PR works great 👍

PR #8400, opened 2025-02-13
fix: Hide doc creation button for archived collection

hmacr:
Hides doc creation button for archived collection.


tommoor:
Fixed in edd413fba with a policy instead, thanks for spotting

Issue #5694, opened 2023-08-16
Option to step through changes in revision history


tommoor:
### Discussed in https://github.com/outline/outline/discussions/5693

<div type='discussions-op-text'>

<sup>Originally posted by **thumDer** August 16, 2023</sup>
It would be great to have a control on the revision views which allows the user to step through each and every changes that are part of the current revision. Currently the user has to go through the whole document and look for the highlighted parts, which is quite cumbersome on long documents. The way it is implemented in vs cvode is quite convinient:
![kép](https://github.com/outline/outline/assets/54556708/09393bcc-35fd-45e4-a5ca-efe84c4049ac)
</div>

includesec-erik:
This is a great idea thanks for suggesting it @thumder!

It does feel like it's a bit of heavy lift for the UI to code this. 

Suggestion (if doing this all in one go is too much work), break this into two parts:
1) Adding the "list of changes as arrows" in the right nav so that one can see from the micro-nav bar on the right where the changes are and scroll until the change is in the view port of the text box
2) Finish the feature with"up/down nav" and keeping a quantified total of number of changes "x of y changes" as the up/down buttons are clicked

almereyda:
Still relevant, if not as a reminder for related UX around versioning.

almereyda:
In GitHub and GitLab PRs, this is the review mode.

Might depend on

- #1084

and more, or less, on

- #5310

includesec-erik:
Still a good feature to add, but if you don't think it makes it on to the next year's roadmap perhaps you can throw it to a back burner queue @tommoor ?

includesec-erik:
@hmacr what do you think, still remove stale flag? As a person who uses the diff UI often, I'd love to see this feature

hmacr:
There was ongoing work in the recent past, we'll see if that can be taken forward.

includesec-erik:
@hmacr @tommoor it sounds like earlier this year there was some movement towards supporting this, any chance we can get it back on the roadmap for 2026?

It seems like a really cool feature that advanced users of Outline (such as all of the folks at my company) would love!

PR #8404, opened 2025-02-13
Revision diff navigator

infinite-persistence:
Closes [#5694 Option to step through changes in revision history ](https://github.com/outline/outline/issues/5694)


tommoor:
Cool, nice first pass! I would probably consider moving this up until the header area next to the restore button – did you try that?

infinite-persistence:
Ah, putting the new component in `Header` was my first choice too, but:
- Would need to put the `editorRef` and `selectedOpIndex` into `uiStore` to communicate with `RevisionViewer`. Is storing ref in store normal?
- Next idea is to just `position: fixed` it beside the Restore button, but that would require listening to `resize` to handle re-adjustments.  Ok to add listeners for this case?
    - If the extra listener is ok, we can also change the highlight-animation to a small overlay marker instead (ala Google Docs). I think it's a better visual indicator, but I dropped it to avoid the extra listener and extra handling for weird zoom ratios.
        - Also considered letting `RevisionViewer` handle a local copy of `revision.html` and change the css (to avoid positioning issues with the overlay method), but I that felt like too much processing.

What do you think?

tommoor:
Would not put these in the store, a reference to the editor can be gotten through `useEditor` context. For `selectedOpIndex` you would probably need to move that state in `Document.tsx` and pass through to `RevisionViewer` that way.

infinite-persistence:
> a reference to the editor can be gotten through useEditor context

Hm ... for this case, it's not an actual editor, but just a div to display the server-rendered diff?

> For selectedOpIndex you would probably need to move that state in Document.tsx and pass through to RevisionViewer that way.

Ah, cool. I'll take a look.

infinite-persistence:
**Update:** Relocate to header, refactor as standalone component.

_See commit messages for details. Deviated from suggestion a bit._

PR #8410, opened 2025-02-14
fix: Guard templates dropdown menu

hmacr:
Hides templates dropdown menu for viewers and guests

### DIFF
--- a/app/scenes/Document/components/Header.tsx
+++ b/app/scenes/Document/components/Header.tsx
@@ -281,15 +281,18 @@ function DocumentHeader({
                 limit={isCompact ? 3 : undefined}
               />
             )}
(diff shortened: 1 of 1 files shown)


tommoor:
Seems fine, but doesn't `isEditing` cover this? 🤔

hmacr:
I'm afraid `isEditing` just denotes if the document is in editable state - not a public share and user does not have a separate edit mode. This was being used in multiple places, hence didn't want to touch it for now.

As a follow-on, I think it'd be worthwhile to refactor the `Header` rendering a bit - too many conditionals involved here..

tommoor:
> As a follow-on, I think it'd be worthwhile to refactor the Header rendering a bit - too many conditionals involved here..

For sure, it's gotten a little messy

tommoor (approved):

PR #8411, opened 2025-02-14
Enable dragging a document into drafts

### LINKED ISSUE #8346: Enable dragging a document into drafts to remove from collection
This makes sense, I just tried to do it – if the document is published it should unpublish too.

hmacr:
Closes #8346

### DIFF
--- a/app/components/Sidebar/components/DraftsLink.tsx
+++ b/app/components/Sidebar/components/DraftsLink.tsx
@@ -1,0 +1,41 @@
+import { observer } from "mobx-react";
+import { DraftsIcon } from "outline-icons";
+import React from "react";
+import { useTranslation } from "react-i18next";
+import styled from "styled-components";
+import Flex from "~/components/Flex";
+import Text from "~/components/Text";
+import useStores from "~/hooks/useStores";
+import { draftsPath } from "~/utils/routeHelpers";
+import { useDropToUnpublish } from "../hooks/useDragAndDrop";
+import SidebarLink from "./SidebarLink";
+
+export const DraftsLink = observer(() => {
+  const { t } = useTranslation();
+  const { documents } = useStores();
+  const [{ isOver, canDrop }, dropRef] = useDropToUnpublish();
+
+  return (
+    <div ref={dropRef}>
+      <SidebarLink
+        to={draftsPath()}
+        icon={<DraftsIcon />}
+        label={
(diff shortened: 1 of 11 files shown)


app/components/Sidebar/components/DraftsLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTranslation } from "react-i18next";
+import styled from "styled-components";
+import Flex from "~/components/Flex";
+import Text from "~/components/Text";
+import useStores from "~/hooks/useStores";
+import { draftsPath } from "~/utils/routeHelpers";
+import { useDropToUnpublish } from "../hooks/useDragAndDrop";
+import SidebarLink from "./SidebarLink";
+
+export const DraftsLink = observer(() => {
```
tommoor, line 13: Nice 🆗

app/components/Sidebar/hooks/useDragAndDrop.tsx
```diff
@@ -586,3 +586,42 @@ export function useDropToArchive() {
...
+    { isOver: boolean; canDrop: boolean }
+  >({
+    accept: "document",
+    drop: async (item) => {
+      const document = documents.get(item.id);
+      if (!document) {
+        return;
+      }
+
+      await document.unpublish();
```
tommoor, line 606: Is it possible to also clear the `collectionId`? I think it would feel much more natural like you're moving it into the drafts section

hmacr: Ahh yep - that would be the expected outcome for the user, thanks for the suggestion.

server/queues/processors/WebsocketsProcessor.ts
```diff
@@ -73,6 +72,28 @@ export default class WebsocketsProcessor {
...
+        if (!document) {
+          return;
+        }
+
+        const documentToPresent = await presentDocument(undefined, document);
+
+        const channels = await this.getDocumentEventChannels(event, document);
+
+        // We need to add the collection channel to let the members update the doc structure.
+        channels.push(`collection-${event.collectionId}`);
```
hmacr, line 89: Fixed a bug along the way - unpublish events weren't sent to the interested users before.

This could've been added in the `getDocumentEventChannels` method itself - but we generally don't send draft document updates except this special case.

tommoor (approved):

PR #8458, opened 2025-02-17
Move document history to `revisions.list` API

### LINKED ISSUE #8403: Revision entry sometimes show wrong diff

### Current Behavior

Sometimes, clicking on a fresh revision in the History page shows diffs from a prior change.

### Expected Behavior

_No response_

### Steps To Reproduce

1. Make a simple word insertion to a doc.
2. Wait for idle, until status says "You updated less than a minute ago"
3. `...` > History
4. Click on the latest (topmost) entry. The diff is wrong ❌ 
5. Click on the second latest entry, and then back to the latest entry. It's correct now ✔ 

### Environment

```markdown
`ac487671`
```

### Anything else?

_No response_

### LINKED ISSUE #8424: Move document history sidebar to be driven by `revisions.list`
As discussed, `events.list` will no longer be a reliable source of information.

hmacr:
Closes #8424 

This doesn't fix #8403 - it needs a bit more work to consolidate local document state.

I also noticed that document restore doesn't create a `documents.restore` event which sometimes messes up the history - will create a separate PR.

### DIFF
--- a/app/components/PaginatedList.tsx
+++ b/app/components/PaginatedList.tsx
@@ -60,7 +60,7 @@ class PaginatedList<T extends PaginatedItem> extends React.PureComponent<
   fetchCounter = 0;
 
   @observable
-  renderCount = 15;
+  renderCount = Pagination.defaultLimit;
 
   @observable
   offset = 0;
@@ -108,13 +108,16 @@ class PaginatedList<T extends PaginatedItem> extends React.PureComponent<
         ...this.props.options,
       });
 
+      if (this.offset !== 0) {
+        this.renderCount += limit;
+      }
+
       if (results && (results.length === 0 || results.length < limit)) {
         this.allowLoadMore = false;
       } else {
         this.offset += limit;
       }
 
-      this.renderCount += limit;
       this.isFetchingInitial = false;
     } catch (err) {
       this.error = err;
@@ -248,7 +251,9 @@ class PaginatedList<T extends PaginatedItem> extends React.PureComponent<
           }}
         </ArrowKeyNavigation>
         {this.allowLoadMore && (
-          <Waypoint key={this.renderCount} onEnter={this.loadMoreResults} />
+          <div style={{ height: "1px" }}>
+            <Waypoint key={this.renderCount} onEnter={this.loadMoreResults} />
+          </div>
         )}
       </>
     );
--- a/app/scenes/Document/components/History.tsx
+++ b/app/scenes/Document/components/History.tsx
@@ -106,18 +106,26 @@ function History() {
     return pageEvents;
   }, [document, revisions, events, toEvent, offset]);
 
-  const revisionEvents = document
-    ? revisions
-        .filter({ documentId: document.id })
-        .slice(0, offset.revisions + 1) // take one extra to account for realtime edits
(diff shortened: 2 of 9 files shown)


app/components/PaginatedList.tsx
```diff
@@ -248,7 +251,9 @@ class PaginatedList<T extends PaginatedItem> extends React.PureComponent<
           }}
         </ArrowKeyNavigation>
         {this.allowLoadMore && (
-          <Waypoint key={this.renderCount} onEnter={this.loadMoreResults} />
+          <div style={{ height: "1px" }}>
```
hmacr, line 254: I noticed that waypoint doesn't get triggered without a parent div. This happens only in the right sidebar though.. 🤔

server/routes/api/events/schema.ts
```diff
@@ -15,7 +26,9 @@ export const EventsListSchema = BaseSchema.extend({
     /** Whether to include audit events */
     auditLog: z.boolean().default(false),
 
-    /** Name of the event to retrieve */
+    /** @deprecated, use 'events' instead
+     * Name of the event to retrieve
+     */
```
tommoor, line 31:
```suggestion
    /** 
     * @deprecated, use `events` parameter instead
     * Name of the event to retrieve
     */
```

server/routes/api/events/events.ts
```diff
@@ -20,37 +20,41 @@ router.post(
...
-      where = { ...where, actorId };
-    }
-
-    if (documentId) {
-      where = { ...where, documentId };
-    }
-
-    if (auditLog) {
+    if (events?.length) {
+      where.name = events;
```
tommoor, line 38: This will allow retrieving audit events without the `audit` policy, might be easiest to just filter?

app/scenes/Document/components/History.tsx
```diff
@@ -1,34 +1,147 @@
...
+        createdAt: document.updatedAt,
+        actor: document.updatedBy!,
+        latest: true,
+      });
+    } else {
+      (latestRevision as RevisionEvent).latest = true;
+    }
+  }
+
+  const mergedEvents = orderBy(
```
tommoor, line 138: Should memo this?

hmacr:
> If the history is entirely driven from revisions then this should become a non-issue?

Yes, the whole impl becomes a lot simpler too - no need to keep track of individual offsets between revisions and events. Should be fine to just list the `revisions.list`?

server/routes/api/events/events.ts
hmacr: Thanks for catching this bug.

tommoor:
Well not sure I want to _remove_ them, I guess I'm just not sure how not having `documents.restore` is an issue here as we'll have an associated `revision`

hmacr:
I hear you - I mean it's not a functional issue, just that the history might not have it which should be fine because data from events are _nice to haves_ anyway with the cleanup task going in later.

tommoor:
Perhaps ideally there would be a separate model here like `DocumentHistory` that would combine revisions and events, leave events as the audit trail it was supposed to be.

Something for the future

tommoor (approved):

PR #8484, opened 2025-02-18
fix: Don't remove data from store when fetching archived docs

### LINKED ISSUE #8473: Archive Nest Documents

### Current Behavior

I have a Collection named `Test`, and created a sub-documents (named `subTest`)which also contained many sub-sub-documents, if I archive
`subTest,` I could not find the `subTest` at the sidebar (under the Archive). They all showed on the right side and lost their structure.

And if the `subTest` contains many sub-sub-documents, I can't scroll up when I scroll down at the `Archive page`

1.  At beginning 
Image

2. After I scroll down

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- Outline:
- Browser:
```

### Anything else?

_No response_

hmacr:
Closes #8473 

Feels right, but I couldn't think of a reason why the filtering was needed before.

### DIFF
--- a/app/stores/DocumentsStore.ts
+++ b/app/stores/DocumentsStore.ts
@@ -343,18 +343,8 @@ export default class DocumentsStore extends Store<Document> {
   };
 
   @action
-  fetchArchived = async (options?: PaginationParams): Promise<Document[]> => {
-    const archivedInResponse = await this.fetchNamedPage("archived", options);
-    const archivedInMemory = this.archived;
-
-    archivedInMemory.forEach((docInMemory) => {
-      !archivedInResponse.find(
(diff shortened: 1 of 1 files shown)


tommoor:
@apoorv-mishra you only added this a couple of months ago, anything still in your memory?

apoorv-mishra:
Dug [this](https://github.com/outline/outline/pull/7266/commits/e03b627e2c54cd11fcb6a9f77686d5e439760c70) up from commit history, but can't recall the corresponding bug 🙄

tommoor:
Okay, I guess we'll remove for now – i suspect the issue here would be better solved another way if there still is one. Perhaps weren't receiving the right events through websocket for example

PR #8497, opened 2025-02-19
Move document history to `revisions.list` API

### LINKED ISSUE #8487: History: minor glitch in avatar for drafts
### Current Behavior

- Seems to only impact documents that became unpublished.
- F5 sometimes fixes it, sometimes not.

Image

### Environment

c6ae6e0

tommoor:
Reverts outline/outline#8495 – @hmacr this reverts the revert, easiest would be to push your fixes to this branch

Also, closes #8487

### DIFF
--- a/app/components/EventListItem.tsx
+++ b/app/components/EventListItem.tsx
@@ -14,8 +14,9 @@ import { useLocation } from "react-router-dom";
 import styled, { css } from "styled-components";
 import EventBoundary from "@shared/components/EventBoundary";
 import { s, hover } from "@shared/styles";
+import { RevisionHelper } from "@shared/utils/RevisionHelper";
 import Document from "~/models/Document";
-import Event from "~/models/Event";
+import User from "~/models/User";
 import { Avatar, AvatarSize } from "~/components/Avatar";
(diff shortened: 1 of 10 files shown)


app/components/EventListItem.tsx
```diff
@@ -50,23 +77,32 @@ const EventListItem = ({ event, latest, document, ...rest }: Props) => {
   };
 
   const prefetchRevision = async () => {
-    if (event.name === "revisions.create" && event.modelId) {
-      await revisions.fetch(event.modelId);
+    if (
+      !document.isDeleted &&
```
hmacr, line 81: Viewing the diff is not allowed for deleted docs - might as well suppress the prefetch.

Should this behavior be extended to archived docs? (as a follow-on)

hmacr:
@tommoor I've tested the fix for various document flows - please let me know if you find something.

app/components/EventListItem.tsx
tommoor: It's ok / useful to see history of archived stuff

tommoor:
Ok, merging for now – although it's like this already in production, it might be better as a follow up to move all this event-munging logic into a computed getter in the store.

PR #8499, opened 2025-02-19
fix: Prevent last user/group with collection manage permission being removed

### LINKED ISSUE #8243: Last user with collection manage permission can be removed
Imagine this scenario: You are an administrator setting up permissions for different collections. While doing so, you accidentally remove your own Manage permissions access for one of the collections. No big deal, right? As an administrator, you can usually restore your permissions.

But what if that collection has its access set to No access for All members? In that case, the collection becomes invisible—not just to you but to everyone. It effectively disappears from view, leaving it inaccessible to all users and lost within the database.

**To Reproduce**
1. Create a new collection
2. Set All members to No access
3. Remove yourself from the permission list (so nobody will have Manage permissions and nobody can see the collection)
4. The collection disappears from the list of collections, and nobody can find it from that point on.

**Expected behavior**
There must be at least 1 permission rule having Manage level. You cannot remove all.
Or ... administrators can see hidden collections from them. At least in some administrator special GUI.

**Screenshots**
giph2

- Install: self-hosted
- Version: v0.81.1

tommoor:
closes #8243

### DIFF
--- a/server/models/GroupMembership.ts
+++ b/server/models/GroupMembership.ts
@@ -20,6 +20,7 @@ import {
   AfterUpdate,
   AfterDestroy,
   BeforeDestroy,
+  BeforeUpdate,
 } from "sequelize-typescript";
 import { CollectionPermission, DocumentPermission } from "@shared/types";
 import { ValidationError } from "@server/errors";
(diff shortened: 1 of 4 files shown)


server/models/GroupMembership.ts
```diff
@@ -247,6 +250,47 @@ class GroupMembership extends ParanoidModel<
     }
   }
 
+  @BeforeDestroy
+  static async checkLastAdminPermission(
```
hmacr, line 254: Think we need to account for Admin -> Viewer/Editor change in a BeforeUpdate hook as well. I was able to remove myself by performing this action which led to a dangling collection.

PR #8504, opened 2025-02-19
fix: Update local storage when creating/deleting pins

hmacr:
Fixes a bug where the documents list sometimes slides up in the collection scene.

This was because the local cache wasn't updated when a pin is created / deleted. So, when there are no more pins to show, they still take up space until the latest data is retrieved from the server.

### DIFF
--- a/app/models/Document.ts
+++ b/app/models/Document.ts
@@ -27,6 +27,7 @@ import { client } from "~/utils/ApiClient";
 import { settingsPath } from "~/utils/routeHelpers";
 import Collection from "./Collection";
 import Notification from "./Notification";
+import Pin from "./Pin";
 import View from "./View";
 import ArchivableModel from "./base/ArchivableModel";
 import Field from "./decorators/Field";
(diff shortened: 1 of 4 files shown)


app/models/Document.ts
```diff
@@ -465,12 +466,17 @@ export default class Document extends ArchivableModel implements Searchable {
   };
 
   @action
-  pin = (collectionId?: string | null) =>
-    this.store.rootStore.pins.create({
+  pin = async (collectionId?: string | null) => {
+    const pin = new Pin({}, this.store.rootStore.pins);
+
+    await pin.save({
```
hmacr, line 472: Helps run `AfterCreate` hook.

app/models/Pin.ts
```diff
@@ -26,6 +30,30 @@ class Pin extends Model {
...
+      setPersistedState(pinsCacheKey("home"), pins.home.length);
+      return;
+    }
+
+    const collection = pins.rootStore.collections.get(model.collectionId);
+    if (!collection) {
+      return;
+    }
+
+    const urlId = collection.path.replace("/collection/", "");
```
tommoor, line 51: I think `collection.path` is what you want here

tommoor, line 51: It sucks that `urlId` is misused in the existing keys – I'd be up for just changing the key and ignoring the existing cache to be honest

tommoor (approved):

app/models/Pin.ts
hmacr: Let me push a commit quickly to use collection id.

tommoor: Well, I think collection.id is not usable because it means waiting for the network request that resolves `urlId` -> `id`, however we can use `urlId` the part without the title slug as it's stable.

tommoor (approved):

Issue #8521, opened 2025-02-21
Image failed to load after Upgrade


huntx:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

After upgrading to 0.82.0, the users unable to view image of certain article.

![Image](https://github.com/user-attachments/assets/dc4fecbc-35c5-4df3-a4a5-032542ee8140)

A little backstory, this self-hosted Outline had migrated from self-hosted Minio bucket to local storage since 0.80.2, and everything was working fine.

Some observation on this issue is,
-  the author of the article (who uploaded the image) can view the image. But, other users not able to view.
-  only the image that was uploaded prior the local storage migration is affected

### Environment

```markdown
- Outline: 0.82.0
- Browser: Tested with Chrome, Edge, Firefox
```

tommoor:
- What is the url of the image that's failing? You can find by looking in the network inspector in developer tools.
- What is the error in the network inspector?
- To confirm, you're using the `v0.82.0` docker image?

huntx:
I not able to reproduce the error now because I have fixed the issue by manually reupload all the affected images.

Before I implement the fix, I did check through the logs.

This is URL of images when the user unable to view from Nginx logs

[21/Feb/2025:13:33:11 +0000] "GET /api/files.get?key=uploads/d10c1877-ccb5-43a7-8f8a-d12774f9ceee/a23f9b2a-87b0-4dad-b89f-51ab365154aa/image.png HTTP/2.0" 403 87 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36" "-"

But, if the user manually paste the <kb.domain>/api/files.get?key=uploads/d10c1877-ccb5-43a7-8f8a-d12774f9ceee/a23f9b2a-87b0-4dad-b89f-51ab365154aa/image.png in browser address bar, they able to view the image.

I further notice that only one of the document path "d10c1877-ccb5-43a7-8f8a-d12774f9ceee" is affected, while others are working normally.

ls -la data/outline/uploads/
total 28
drwxr-xr-x  5 1001 1001  4096 Nov 26 09:34 .
drwxr-xr-x  5 1001 1001  4096 Oct 22 03:51 ..
[...]

PR #8524, opened 2025-02-21
fix: Allow downloading exported file

### LINKED ISSUE #8508: Can not download exported file.

### Current Behavior

After documents being exported, I cannot download the zip file to my machine. 
```
NotFoundError: Resource not found
    ...
```

### Expected Behavior

_No response_

### Steps To Reproduce

1. Export any documents by clicking `Preferences >  Export > Export data...`
2. Click the exported file to download it.

### Environment

```markdown
- Outline: 0.80.0
- Browser: Chrome
```

### Anything else?

The error started occurring from [#f469212](https://github.com/outline/outline/commit/f46921275dca713d25e953c2c7610dec5d030d28). The test results indicate that the error is caused by modifications to [this file](https://github.com/outline/outline/commit/f46921275dca713d25e953c2c7610dec5d030d28#diff-26a1619ef7484a9d9fcd5eee18e291e33994bd335a9f7da57ede9ce674efb825), making the exported file untraceable through `Attachment.findOne`.

hmacr:
Fixes #8508 

In case of exports, an associated attachment will not be available.


tommoor:
The original report said the issue was in `v0.80.0` which is why I dismissed it 🤷 , not on `main`.

tommoor (approved):

hmacr:
There were three issues filed in succession, all of which were related to images - other two being #8521 and #8509. So, I did a deep dive to see if there was any regression recently.

tommoor:
Kind of coincidence then given this code change is only a couple of days old and not in `v0.82.0`

PR #8531, opened 2025-02-22
Wrap headings on mobile ToC

### LINKED ISSUE #8526: The "no headings" TOC advisory cut off in mobile
### Expected Behavior

Wrapped text, maybe?  Would assume it would impact localized text even if shortened.

hmacr:
Closes #8526

### DIFF
--- a/app/menus/TableOfContentsMenu.tsx
+++ b/app/menus/TableOfContentsMenu.tsx
@@ -1,8 +1,10 @@
 import { observer } from "mobx-react";
 import { TableOfContentsIcon } from "outline-icons";
+import { ellipsis } from "polished";
 import * as React from "react";
 import { useTranslation } from "react-i18next";
 import { MenuButton, useMenuState } from "reakit/Menu";
+import styled from "styled-components";
 import Button from "~/components/Button";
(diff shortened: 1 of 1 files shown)


tommoor:
Not sure here, I don't think there is anything specific about TOC that warrants special treatment? Like just apply truncation to all the menus on mobile if we want to go that route? 🤔

hmacr:
It was one of the options I considered - there's an existing [ellipsis](https://github.com/outline/outline/blob/main/app/components/ContextMenu/MenuItem.tsx#L109) style in the menu item, but it doesn't work due to flexbox compatibility.
There's a few places where the `Title`'s children is not a simple text node, for e.g. https://github.com/outline/outline/blob/87fcf35956d171ab00777842b04d38b9a2fd1313/app/editor/components/SuggestionsMenuItem.tsx#L52-L62
This will require a bit of refactor throughout the app to truncate just the title text - worth doing it now? 🤔

tommoor:
Fair enough, not worth it now. Might be better if it wrapped here as you can't really access the elided text on mobile?

tommoor (approved):

PR #8548, opened 2025-02-24
fix: Horizontal scroll in workspace details page

hmacr:
Fixes the horizontal scroll caused by action row in workspace details page.


tommoor:
Damn, being on macOS with scrollbars turned off I never noticed this

PR #8549, opened 2025-02-24
chore: Use stores from context in collection actions

hmacr:
Fixes the lint warnings shown in IDE / PRs.
Verified all collection action touchpoints pass stores in the context.

---

Note: Other actions also use the imported stores, but don't cause lint errors - might be better to modify them to use from action context for consistency (can go incrementally here).

### DIFF
--- a/app/actions/definitions/collections.tsx
+++ b/app/actions/definitions/collections.tsx
@@ -15,7 +15,6 @@ import {
 } from "outline-icons";
 import * as React from "react";
 import { toast } from "sonner";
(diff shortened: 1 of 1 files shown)


tommoor:
I think what happened is that I moved CI from CircleCI -> GitHub at the weekend and now linting warnings are appearing on PRs, however there are so many warnings that it's only showing the first few in the list.

I'll see if there is a way to suppress if they're not affecting the files in the PR itself

PR #8567, opened 2025-02-26
fix: Show diff from document when revision has not been created yet

### LINKED ISSUE #8403: Revision entry sometimes show wrong diff

hmacr:
Closes #8403 

Also fixes the spurious revision shown as the latest when any non-revision events like delete, unpublish happens.

### DIFF
--- a/app/scenes/Document/components/History.tsx
+++ b/app/scenes/Document/components/History.tsx
@@ -136,14 +136,19 @@ function History() {
       "desc"
     );
 
-    const latestEvent = merged[0];
+    const latestRevisionEvent = merged.find(
+      (event) => event.name === "revisions.create"
+    );
 
-    if (latestEvent && document) {
-      const latestRevisionEvent = merged.find(
-        (event) => event.name === "revisions.create"
-      );
+    if (latestRevisionEvent && document) {
+      const latestRevision = revisions.get(latestRevisionEvent.id);
 
-      if (latestEvent.createdAt !== document.updatedAt) {
+      const isDocUpdated =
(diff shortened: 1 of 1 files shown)


app/scenes/Document/components/History.tsx
```diff
@@ -136,14 +136,19 @@ function History() {
...
+    if (latestRevisionEvent && document) {
+      const latestRevision = revisions.get(latestRevisionEvent.id);
 
-    if (latestEvent && document) {
-      const latestRevisionEvent = merged.find(
-        (event) => event.name === "revisions.create"
-      );
+      const isDocUpdated =
+        latestRevision?.title !== document.title ||
+        JSON.stringify(latestRevision.data) !== JSON.stringify(document.data);
```
tommoor, line 148: Do we need to stringify? Might it be faster to use `isEqual` is it can bail as soon as an inequality is found

hmacr: yes, `fast-deep-equal` totally slipped my mind.

tommoor (approved):

PR #8630, opened 2025-03-05
fix: Skip auto creating subscriptions when user/group is added to a document

hmacr:
Changing the auto subscription behaviour as it may lead to noisy notifications.


tommoor:
Reverts most of https://github.com/outline/outline/pull/8354

tommoor (approved):

PR #8632, opened 2025-03-05
fix: First item in list must be a paragraph

### LINKED ISSUE #8216: Cannot enter code block inside list
Markdown, and also outline, supports having a code block inside a list. However, I cannot seem to enter this formatting in a clean way.

In Markdown, this is possible by indenting the code block to the same level as the list. E.g. this markdown:

````
1. This is a list
   ```
   with a code block
   of two lines
   ```
2. and a second item
````

Produces this output in github markdown:

1. This is a list
   ```
   with a code block
   of two lines
   ```
2. and a second item

If I copy the above markdown source into outline with control-V, it is also rendered as expected, so outline does support this structure:

image

But when I type it manually (using shift-enter to continue the first list item), it does not recognize this at all:

image

Similarly, using `/code` inside the list also does not work. When typing / after the shift-enter, nothing happens, no popup. When typing /code directly at the start of a list item, I get a popup, but then selecting /code and pressing enter does not do anything (the popup just disappears).

- Install: self-hosted (docker version)
- Version: v0.81.1

 - OS: Linux, Ubuntu 24.10
 - Browser Firefox
 - Version 133.0

### LINKED ISSUE #8611: Callouts not working inside ordered/bulleted lists

### Current Behavior

When creating a callout (info, warning, tip, etc.) if you are at the root of the page, or inside a bulleted/ordered list but not on a tabbed section you can create a callout.

However, if you ARE inside a tabbed section and type /info then nothing happens.

### Expected Behavior

Callouts should work inside bulleted and/or ordered lists.

### Steps To Reproduce

Watch behavior

Image

### Environment

```markdown
- Outline: Hosted
- Browser: Edge or Chrome
```

### Anything else?

_No response_

tommoor:
A simple change in the spec

closes #8611
closes #8216

### DIFF
--- a/shared/editor/nodes/CheckboxItem.ts
+++ b/shared/editor/nodes/CheckboxItem.ts
@@ -22,7 +22,7 @@ export default class CheckboxItem extends Node {
           default: false,
         },
       },
-      content: "paragraph block*",
+      content: "block+",
       defining: true,
       draggable: true,
       parseDOM: [
--- a/shared/editor/nodes/ListItem.ts
+++ b/shared/editor/nodes/ListItem.ts
@@ -26,7 +26,7 @@ export default class ListItem extends Node {
 
   get schema(): NodeSpec {
     return {
-      content: "paragraph block*",
+      content: "block+",
       defining: true,
       draggable: true,
       parseDOM: [{ tag: "li" }],


apoorv-mishra (approved):
Tested 👍

tommoor:
lil related discussion here: https://discuss.prosemirror.net/t/recommended-spec-for-list-item-content/8247/3

apoorv-mishra:
Yeah, knew about that, but we can expect users to not nest like that, purposely? Also that `!` thingy should work, considering the `content` takes regex-like strings?

apoorv-mishra:
Hmm, looks like content expressions borrowed only simpler regex features, like quanitfiers(`+`, `*`, etc.). I think the other solution here might be use `appendTransaction` and issue a `tr` to delete the unexpectedly entered list item.

tommoor:
I tried creating a group that only contains non-list blocks and enforcing this as the first node in a list – unfortunately the behavior is much worse with this setup. I think what's here is tehe best solution

apoorv-mishra:
> I think what's here is the best solution

"here" meaning? keeping `content: block+` is the best solution?

tommoor:
I mean what's in this PR :)

apoorv-mishra:
On another note, we can also deliver an alternate UX with the following automatic transform, from 👇 
Screenshot 2025-03-12 at 10 10 10 AM
to 👇 
Screenshot 2025-03-12 at 10 11 31 AM

apoorv-mishra:
Sure, can pick after toggle block.

PR #8673, opened 2025-03-12
fix: Potential unsafe `content-type` check

tommoor:
Using `startsWith` to parse content type is not safe, better to check the entire string.

### DIFF
--- a/shared/editor/lib/FileHelper.ts
+++ b/shared/editor/lib/FileHelper.ts
@@ -8,7 +8,7 @@ export default class FileHelper {
    * @returns True if the file is an image
    */
   static isImage(contentType: string) {
-    return /^image\/[a-z0-9_\-\+]+$/i.test(contentType);
+    return /^image\/[!#$%&'*+.^\w`|~-]+$/i.test(contentType);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


shared/editor/lib/FileHelper.ts
```diff
@@ -4,21 +4,31 @@ export default class FileHelper {
...
    * Checks if a file is an image.
    *
-   * @param file The file to check
+   * @param contentType The content type of the file
    * @returns True if the file is an image
    */
-  static isImage(file: File) {
-    return file.type.startsWith("image/");
+  static isImage(contentType: string) {
+    return /^image\/[a-z0-9_\-\+]+$/i.test(contentType);
```
hmacr, line 11: Some additional characters like ".", "!" can be part of the subtype, for e.g. `image/vnd.adobe.photoshop`.
Can we use a library like [fast-content-type-parse](https://www.npmjs.com/package/fast-content-type-parse) instead which takes care of these nuances?

Possible Content-type: https://www.iana.org/assignments/media-types/media-types.xhtml
Related RFC - [7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1), [7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) (this has info allowed characters)

tommoor: I don't want to pull in another library for something so simple, we need to reduce deps if anything 😉 

There is no `image/vnd.adobe.photoshop` from what I can tell, though there is `application/vnd.adobe.photoshop`. I'll add the additional characters, although the usage here does not really necessitate it

hmacr: Ah ok - I think we should be fine with just the dot character then :)

PR #8697, opened 2025-03-15
Require collection admin permission to update template

### LINKED ISSUE #8692: Change templates to only editable with collection manage permission
This change will be for everyone going forward

tommoor:
Changes behavior so that:

- Admin permission is required to create and update workspace templates
- Collection admin permission is required to create and update collection templates

closes #8692

### DIFF
--- a/app/components/TeamLogo.ts
+++ b/app/components/TeamLogo.ts
@@ -4,8 +4,20 @@ import { Avatar } from "./Avatar";
 
 const TeamLogo = styled(Avatar)`
   border-radius: 4px;
-  box-shadow: inset 0 0 0 1px ${s("divider")};
   border: 0;
+  position: relative;
+
+  &::before {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    border-radius: 4px;
+    box-shadow: inset 0 0 0 1px ${s("divider")};
+    z-index: -1;
+  }
 `;
 
 export default TeamLogo;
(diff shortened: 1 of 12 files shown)


hmacr:
I think we need to modify the template creation flow as well. A collection member can create a blank template, but they won't be able to add content to it.
Cleaner and predictable behavior would be to restrict the create/edit flows for only collection admins as long as it's a collection template.

tommoor:
We desperately need a refactor of templates, separated into their own model – this logic is way more complex than should be needed and is going to lead to permissions bugs

app/components/TeamLogo.ts
```diff
@@ -4,8 +4,20 @@ import { Avatar } from "./Avatar";
 
 const TeamLogo = styled(Avatar)`
   border-radius: 4px;
-  box-shadow: inset 0 0 0 1px ${s("divider")};
   border: 0;
+  position: relative;
+
+  &::before {
```
tommoor, line 10: Just a random styling fix I noticed in the menus here – avoiding a grey outline around all team logos

hmacr:
I'll test this tomorrow, if it can wait.
Otherwise, code looks right 👍

hmacr:
Sorry for not noticing earlier - "unpublish/publish" and "move" flows need handling as well (other cases of permission bugs 😞).

tommoor:
publish and move both are supersets of `update` permission, so it should be taken into account 🤔

hmacr:
`update` / `updateDocuments` permissions allow an "editor" to perform these actions which we want to restrict in these PR 🤔 
Move needs a bit more control I think - we want to show the target collections only where the current user is an Admin.

tommoor:
closing in favor of https://github.com/outline/outline/pull/8705

PR #8710, opened 2025-03-17
API importer for Notion

hmacr:
This PR enables importing Notion documents using an API integration.

### Pending tasks
- [x] JSDoc.
- [x] Restrict imports to "one" active per team.
- [x] Cleanup task for stuck imports.
- [x] Cleanup task for completed import_tasks.
- [x] Add a cancel option in UI.
- [x] Tests for APIs.

Note: bigger PR, but not as big as it seems to be.. ~9k lines are related to tests.


server/constants.ts
```diff
@@ -0,0 +1 @@
+export const PagePerImportTask = 3;
```
tommoor, line 1: Not worth a new file here

server/models/Import.ts
```diff
@@ -0,0 +1,87 @@
...
+  @Column(DataType.STRING)
+  state: ImportState;
+
+  @Column(DataType.JSONB)
+  input: ImportInput<T>;
+
+  @IsNumeric
+  @Default(0)
+  @Column(DataType.INTEGER)
+  pageCount: number;
```
tommoor, line 61: Is there a more generic term here thinking this will be used for more than just Notion.

`progress`? 
`completedCount`?

server/policies/import.ts
```diff
@@ -0,0 +1,16 @@
+import { ImportState } from "@shared/types";
+import { User, Team, Import } from "@server/models";
+import { allow } from "./cancan";
+import { and, isTeamAdmin, isTeamMutable } from "./utils";
+
+allow(User, ["createImport", "readImport"], Team, (actor, team) =>
```
tommoor, line 6:
```suggestion
allow(User, ["createImport", "listImports"], Team, (actor, team) =>
```

This would be more typical

server/queues/processors/ImportsProcessor.ts
```diff
@@ -0,0 +1,434 @@
...
+              model: Collection,
+              as: "collection",
+            },
+          ],
+          order: [["createdAt", "DESC"]],
+          transaction,
+        },
+        async (documents) => {
+          for (const document of documents) {
+            // Without reload, sequelize overwrites the updates to collection.
```
tommoor, line 140: This is unfortunate, in what way does it overwrite?

server/queues/processors/ImportsProcessor.ts
```diff
@@ -0,0 +1,434 @@
...
+  static applicableEvents: Event["name"][] = [
+    "imports.create",
+    "imports.processed",
+    "imports.delete",
+  ];
+
+  public async perform(event: ImportEvent) {
+    const importModel = await Import.findByPk<Import<T>>(event.modelId, {
+      rejectOnEmpty: true,
+      paranoid: false,
```
tommoor, line 48: Should probably query with an update lock

server/routes/api/imports/imports.ts
```diff
@@ -0,0 +1,154 @@
...
+  pagination(),
+  validate(T.ImportsListSchema),
+  async (ctx: APIContext<T.ImportsListReq>) => {
+    const { service, sort, direction } = ctx.input.body;
+    const { user } = ctx.state.auth;
+
+    authorize(user, "readImport", user.team);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const where: WhereOptions<Import<any>> = service
```
tommoor, line 77: Need to include `teamId` in query here

server/routes/api/imports/imports.ts
```diff
@@ -0,0 +1,154 @@
...
+  transaction(),
+  async (ctx: APIContext<T.ImportsInfoReq>) => {
+    const { id } = ctx.input.body;
+    const { user } = ctx.state.auth;
+    const { transaction } = ctx.state;
+
+    const importModel = await Import.findByPk(id, {
+      rejectOnEmpty: true,
+      transaction,
+      lock: transaction.LOCK.UPDATE,
```
tommoor, line 118: No need for update lock or transaction here as we're only querying

server/routes/api/imports/imports.ts
```diff
@@ -0,0 +1,154 @@
...
+  auth({ role: UserRole.Admin }),
+  validate(T.ImportsCreateSchema),
+  transaction(),
+  async (ctx: APIContext<T.ImportsCreateReq>) => {
+    const { integrationId, service, input } = ctx.input.body;
+    const { user } = ctx.state.auth;
+
+    authorize(user, "createImport", user.team);
+
+    const pendingImport = await Import.findOne({
```
tommoor, line 30: nbd but could just use `Import.count` here

server/queues/tasks/UploadAttachmentsForImportTask.ts
```diff
@@ -0,0 +1,69 @@
...
+        items.length > ConcurrentUploads ? items.length : ConcurrentUploads,
+    });
+
+    const uploadPromises = items.map(async (item) => {
+      try {
+        await sema.acquire();
+
+        const attachment = await Attachment.findByPk(item.attachmentId, {
+          rejectOnEmpty: true,
+          include: [{ association: "user" }],
```
tommoor, line 33: Not clear why `user` is needed here?

app/hooks/usePolicy.ts
```diff
@@ -8,9 +8,14 @@ import useStores from "./useStores";
...
  *
  * @param entity The model or model id
+ * @param {Object} options - Options for loading relations.
+ * @param {string} options.force - Whether to force load the relations, default: false.
  * @returns The policy for the model
  */
-export default function usePolicy(entity?: string | Model | null) {
+export default function usePolicy(
+  entity?: string | Model | null,
+  { force }: { force: boolean } = { force: false }
```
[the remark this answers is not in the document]
hmacr: Hmm we need a way to reload policy when a property in `entity` changes 🤔
In almost all cases, `usePolicy` is used in a fresh page / render - so `entity` remains fresh for the `useEffect` dep reference checks. But for `ImportListItem`, the ref remains the same as the realtime progress count keeps getting updated.
Perhaps, force-loading the import model in `ImportListItem` is better as an one-off case?

plugins/notion/server/oauth.ts
```diff
@@ -0,0 +1,40 @@
+import { z } from "zod";
+import fetch from "@server/utils/fetch";
+import { NotionUtils } from "../shared/NotionUtils";
+import env from "./env";
+
+export class NotionOAuth {
```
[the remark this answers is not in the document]
hmacr: yea, can be included in `NotionClient`.

server/queues/tasks/UploadAttachmentsForImportTask.ts
hmacr: my bad, not needed here.

server/models/Import.ts
hmacr: I'm thinking `pageCount` is generic enough? In the end, a _document_ or a  _page_ from an external source will be imported? 🤔

`documentCount` maybe?

server/queues/processors/ImportsProcessor.ts
hmacr: It's just the `documentStructure` because of the way `findAllInBatches` work - document from a subsequent batch could overwrite the associated collection.
But, looking at it with fresh eyes, I think this document refetch might not be needed - we can return the collections and documents from `createCollectionsAndDocuments` method -> update collection structure in place for all root documents -> save all collections. Let me see if this works.

hmacr: It works great - very efficient!

server/queues/processors/ImportsProcessor.ts
```diff
@@ -0,0 +1,438 @@
...
+              );
+
+              continue;
+            }
+
+            // Document at the root of a collection when there's no parent (or) the parent is the collection.
+            const isRootDocument =
+              !parentExternalId || !!importInput[parentExternalId];
+
+            await Document.createWithCtx(
```
hmacr, line 319: On par with what's happening in existing file-based import flow, i.e. publish `documents.create` event for each document.

I noticed that this leads to flood of requests from the app when the import is complete. I _really_ think we should suppress this event, and retain just the `collections.create` events.

server/models/Import.ts
tommoor: `documentCount` is better - I would rather use our language than notion's

server/queues/processors/ImportsProcessor.ts
tommoor: One thing you'll see is that a lot of processors suppress handling when the event is from an import, e.g.

https://github.com/outline/outline/blob/2b44a61f0226309386a33f56459c728c13e81461/server/queues/processors/NotificationsProcessor.ts#L60

hmacr:
All comments are addressed, pushed a couple of improvements - lmk if something needs changing.

Meanwhile, I'll work on adding a "cancel" option and some tests for APIs & tasks.

PR #8541, opened 2025-02-23
Radix input select component

hmacr:
Towards #6456 

This PR implements an input select component based on Radix primitives. The affected components are listed below - input selects that are inside another "reakit" component is best migrated later as the compatibility with Radix is not great (especially w.r.t keeping the component mounted).

### Affected components
| Component | Migrated? |
|--------|--------|
| DefaultCollectionInputSelect |  :white_check_mark: First component that was migrated. |
| Details (Settings)<br>(ToC preference) | :white_check_mark:  Migrated since it renders right next to `DefaultCollectionInputSelect`. |
| CommentSortMenu | :white_check_mark: Migrated to test "separator" variant in select options. |
| Preferences (Settings)<br>(Language and theme preferences) | :soon: In a follow-up to keep a smaller testing scope. |
|Security (Settings)<br>(Default role for new users) | :soon: In a follow-up to keep a smaller testing scope. |
| InputMemberPermissionSelect | :x: Renders in reakit popover - best to migrate post popover component migration. |
| InputSelectPermission | :x: Renders in reakit popover & dialog - best to migrate post popover and dialog components migration. |
| SelectLocation (TemplatizeDialog) | :x: Renders in reakit dialog - best to migrate post dialog component migration. |
| ApiKeyNew | :x: Renders in reakit dialog - best to migrate post dialog component migration. |
| Invite | :x: Renders in reakit dialog - best to migrate post dialog component migration. |

#### Desktop
- Checkmark is now on the right side - nicely aligns the options' icon with the button's.

| Before | After |
|--------|--------|
| Screenshot 2025-03-03 at 2 55 39 PM | Screenshot 2025-03-03 at 2 53 45 PM |

#### Mobile
- Mandatory label at the top - I think it gives context.
- Removed margin with the viewport edges - works like a native drawer.
- Checkmark is now on the right side - keeps it consistent with desktop variant.

| Before | After |
|--------|--------|
| Screenshot 2025-03-03 at 2 56 01 PM | Screenshot 2025-03-03 at 2 54 04 PM |

PR #8711, opened 2025-03-17
Migrate language, theme and user-role input-select to Radix

hmacr:
Towards #6456 
Additional components from #8541

### DIFF
--- a/app/scenes/Settings/Preferences.tsx
+++ b/app/scenes/Settings/Preferences.tsx
@@ -3,12 +3,12 @@ import { SettingsIcon } from "outline-icons";
 import * as React from "react";
 import { Trans, useTranslation } from "react-i18next";
 import { toast } from "sonner";
-import { languageOptions } from "@shared/i18n";
+import { languageOptions as availableLanguages } from "@shared/i18n";
 import { TeamPreference, UserPreference } from "@shared/types";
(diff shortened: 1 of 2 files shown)


app/scenes/Settings/Preferences.tsx
```diff
@@ -77,31 +111,27 @@ function Preferences() {
...
-        <InputSelect
-          ariaLabel={t("Appearance")}
-          options={[
-            { label: t("Light"), value: Theme.Light },
-            { label: t("Dark"), value: Theme.Dark },
-            { label: t("System"), value: Theme.System },
-          ]}
+        <InputSelectNew
+          options={themeOptions}
           value={ui.resolvedTheme}
```
hmacr, line 130: I noticed that `resolvedTheme` is shown as the select value which is different from the actual setting the user chose (Also not in sync with that's shown in account menu in sidebar)
For e.g., user selects "System" -> page shows "System" and theme pref is persisted -> on refresh, theme is resolved to "light" / "dark".
Is this intentional? 🤔 

### Preferences screen
Screenshot 2025-03-17 at 8 27 10 PM

### Sidebar
Screenshot 2025-03-17 at 8 26 57 PM

tommoor: Not intentional, should always show System

PR #8726, opened 2025-03-19
fix: "unknown" users visible in users table

tommoor:
Okay, this happens when a user has been deleted and there is only the headstone record left. If you then open a document that they interacted with this will be in the UsersStore and liable to show up in other views around the app.

The fix is to make sure we send the `deletedAt` timestamp in responses and filter out of displayed data.


app/stores/UsersStore.ts
```diff
@@ -27,46 +27,46 @@ export default class UsersStore extends Store<User> {
 
   @computed
   get active(): User[] {
-    return this.orderedData.filter(
-      (user) => !user.isSuspended && user.lastActiveAt
-    );
+    return this.all.filter((user) => !user.isSuspended);
```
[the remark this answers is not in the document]
tommoor: The data is filtered in `all` now

app/scenes/Settings/Members.tsx
```diff
@@ -194,7 +194,7 @@ function getFilteredUsers({
 
   switch (filter) {
     case "all":
-      filteredUsers = users.orderedData;
+      filteredUsers = users.all;
```
hmacr, line 197: Maybe filter just the deleted users? "All" is kinda misleading by filtering out the invited users as well..

hmacr:
Very tricky to find the bug!
Oh, we need to handle the same in mention menu.

app/scenes/Settings/Members.tsx
tommoor: true

PR #8734, opened 2025-03-21
fix: Unsubscribe link for collection subscriptions 

tommoor:
Hopefully all the existing tests make this easier to review – TLDR: We include a link to unsubscribe in document notification emails, when #8392 was added this wasn't updated to account for collection subscriptions.

### DIFF
--- a/app/hooks/useQueryNotices.ts
+++ b/app/hooks/useQueryNotices.ts
@@ -24,6 +24,14 @@ export default function useQueryNotices() {
         );
         break;
       }
+      case QueryNotices.UnsubscribeCollection: {
+        toast.success(
+          t("Unsubscribed from collection", {
+            type: "success",
+          })
+        );
+        break;
+      }
       default:
     }
   }, [t, notice]);
--- a/server/models/helpers/SubscriptionHelper.ts
+++ b/server/models/helpers/SubscriptionHelper.ts
@@ -1,4 +1,5 @@
 import crypto from "crypto";
+import queryString from "query-string";
 import env from "@server/env";
 
 /**
@@ -15,12 +16,22 @@ export default class SubscriptionHelper {
    * @returns The unsubscribe URL
    */
   public static unsubscribeUrl(userId: string, documentId: string) {
-    return `${env.URL}/api/subscriptions.delete?token=${this.unsubscribeToken(
+    const token = this.unsubscribeToken(userId, documentId);
+
+    return `${env.URL}/api/subscriptions.delete?${queryString.stringify({
+      token,
       userId,
-      documentId
-    )}&userId=${userId}&documentId=${documentId}`;
+      documentId,
+    })}`;
   }
 
+  /**
+   * Generate a token for unsubscribing a user from a document or collection.
+   *
+   * @param userId The user ID to unsubscribe
+   * @param documentId The document ID to unsubscribe from
+   * @returns The unsubscribe token
+   */
   public static unsubscribeToken(userId: string, documentId: string) {
     const hash = crypto.createHash("sha256");
     hash.update(`${userId}-${env.SECRET_KEY}-${documentId}`);
--- a/server/routes/api/subscriptions/subscriptions.ts
+++ b/server/routes/api/subscriptions/subscriptions.ts
@@ -185,7 +185,7 @@ router.get(
 
     let collectionSubscription;
     if (!documentSubscription) {
-      const document = await Document.findOne({
+      const document = await Document.unscoped().findOne({
         attributes: ["collectionId"],
         where: {
           id: documentId,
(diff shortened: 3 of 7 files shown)


server/models/helpers/NotificationHelper.ts
```diff
@@ -17,6 +17,11 @@ import {
 import { can } from "@server/policies";
 import { ProsemirrorHelper } from "./ProsemirrorHelper";
 
+type Recipient = {
```
tommoor, line 20: Pretty much all the changes relate to changing this interface from a simple `User[]` return to one that also includes the subscription, if present

server/models/helpers/NotificationHelper.ts
```diff
@@ -146,64 +153,75 @@ export default class NotificationHelper {
     notificationType: NotificationEventType;
     onlySubscribers: boolean;
     actorId: string;
-  }): Promise<User[]> => {
-    // First find all the users that have notifications enabled for this event
-    // type at all and aren't the one that performed the action.
-    let recipients = await User.findAll({
```
tommoor, line 152: I also changed this to avoid having to query all of the users in a workspace when `onlySubscribers: true` (the most common case)

server/models/helpers/SubscriptionHelper.ts
```diff
@@ -12,18 +12,39 @@ export default class SubscriptionHelper {
    *
    * @param userId The user ID to unsubscribe
    * @param documentId The document ID to unsubscribe from
+   * @param subscriptionId The subscription ID to unsubscribe from
    * @returns The unsubscribe URL
    */
-  public static unsubscribeUrl(userId: string, documentId: string) {
+  public static unsubscribeUrl(
```
tommoor, line 18: Changes here are optional and backwards compatible

server/routes/api/subscriptions/subscriptions.ts
```diff
@@ -164,12 +165,13 @@ router.get(
...
-          documentId,
-        },
+        where: subscriptionId
+          ? { id: subscriptionId }
+          : {
+              userId,
+              documentId,
+            },
         lock: Transaction.LOCK.UPDATE,
-        rejectOnEmpty: true,
```
tommoor, line 172: Removed on purpose, previously this would just show a 404 error if you were already unsubscribed – not a great UX.

hmacr:
oof terrible miss! thanks for fixing it 🙏 will test now.

hmacr:
Passing `subscriptionId` could inadvertently unsubscribe from a different collection (`collectionA`) when a document is moved to another collection (`collectionB`) before the user clicks (could be after X days) on unsubscribe link. However, the end state feels correct as the user originally was subscribed to `collectionA` which allowed the email notif and they intend to unsubscribe from it.

Should be fine right? 🤔

tommoor:
I believe the intent is OK, should be very rare. Interesting though

tommoor:
Alternatively I could grab the `collectionId` from the subscription and use that in the unsubscribe url. The majority of changes here remain the same

hmacr:
> collectionId from the subscription and use that in the unsubscribe url

I think you meant using the latest `collectionId` in the deletion handler 🤔

> honestly not sure which I prefer

Same, this edge-case must be veryyy rare - think we can leave it as-is? Just wanted to bring it up to cover the bases.

tommoor:
> I think you meant using the latest collectionId in the deletion handler 🤔

well to fix the issue you raised it would need to be passed into the email -> unsubscribe url -> deletion handler yep

app/hooks/useQueryNotices.ts
```diff
@@ -24,6 +24,14 @@ export default function useQueryNotices() {
         );
         break;
       }
+      case QueryNotices.UnsubscribeCollection: {
+        toast.success(
+          t("Unsubscribed from collection", {
```
hmacr, line 29: Should this be changed to `unsubscribed from document`? We use the same language in subscription flow and in the email.

hmacr: Ok 👍 as long as it wasn't an oversight :)

hmacr (approved):
Works great!

server/routes/api/subscriptions/subscriptions.ts
```diff
@@ -150,7 +150,8 @@ router.get(
   transaction(),
   async (ctx: APIContext<T.SubscriptionsDeleteTokenReq>) => {
     const { transaction } = ctx.state;
-    const { follow, userId, documentId, subscriptionId, token } = ctx.input.query;
+    const { follow, userId, documentId, subscriptionId, token } =
```
hmacr, line 153: 👆commit 🤣

PR #8754, opened 2025-03-23
chore: Move notification event writing to model layer

tommoor:
- Move notification event writing to model layer
- fix: Bulk notification action does not reflect on other clients
- Remove `notificationUpdater`

towards #7920


server/routes/api/notifications/notifications.ts
```diff
@@ -216,7 +233,19 @@ router.post(
       };
     }
 
-    const [total] = await Notification.update(values, { where });
+    let total = 0;
+    if (!isEmpty(values)) {
+      total = await Notification.findAllInBatches(
```
hmacr, line 238: It looks like `notifications.update_all` is missing the transaction middleware.
Also, I think the db query can be unscoped here - we aren't using the joined models in this flow.

hmacr:
Tested the changes - works as expected :)

server/routes/api/notifications/notifications.ts
tommoor: Yea… decided to not add it in for now given the amount of writes might be massive and this is not super important data

PR #8765, opened 2025-03-24
fix: Release redis lock only when it hasn't expired

hmacr:
This PR suppresses any exceptions that could arise during redlock flow. It's fine to suppress as sometimes the operation could fail when the lock has already expired (or) some other process has active lock on it.
More on this here - https://github.com/mike-marcacci/node-redlock/issues/168

This exception ends up failing the document update emails - noticed it in #8763

### DIFF
--- a/server/utils/CacheHelper.ts
+++ b/server/utils/CacheHelper.ts
@@ -54,7 +54,12 @@ export class CacheHelper {
       }
       return value;
     } finally {
-      await lock?.release();
+      try {
+        await lock?.release();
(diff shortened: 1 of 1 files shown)


tommoor:
What do you think about this suggestion?

https://github.com/mike-marcacci/node-redlock/issues/168#issuecomment-2100800912

Catching all errors in any situation always makes me wary

hmacr:
yep - that's better.

tommoor (approved):

PR #8803, opened 2025-03-26
fix: Handle index collision when creating a collection

### LINKED ISSUE #8795: Error:  % >=  %
Sentry Issue: [OUTLINE-CLOUD-6K3](https://outline-jl.sentry.io/issues/5242321680/?referrer=github_integration)

```
Error:  % >=  %
  at <object>.drop (../../../app/components/Sidebar/components/DraggableCollectionLink.tsx:50:9)
  at Array.forEach (<anonymous>)
...
(4 additional frame(s) were not displayed)
```

hmacr:
Closes #8795 

~Draft since we might need a script to correct the existing indexes - could be a separate PR?~
Correction script will be a separate PR

### DIFF
--- a/server/models/Collection.ts
+++ b/server/models/Collection.ts
@@ -1,4 +1,5 @@
 /* eslint-disable lines-between-class-members */
+import fractionalIndex from "fractional-index";
 import find from "lodash/find";
 import findIndex from "lodash/findIndex";
 import remove from "lodash/remove";
@@ -11,6 +12,8 @@ import {
   InferAttributes,
   InferCreationAttributes,
   EmptyResultError,
+  type CreateOptions,
+  type UpdateOptions,
 } from "sequelize";
 import {
   Sequelize,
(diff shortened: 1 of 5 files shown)


server/models/Collection.ts
```diff
@@ -217,8 +222,8 @@ class Collection extends ParanoidModel<
   color: string | null;
 
   @Length({
-    max: 256,
-    msg: `index must be 256 characters or less`,
+    max: ValidateIndex.maxLength,
```
hmacr, line 225: Unrelated, noticed that API and model validation were not in sync.

server/models/Collection.ts
```diff
@@ -324,6 +329,30 @@ class Collection extends ParanoidModel<
     }
   }
 
+  @BeforeCreate
```
hmacr, line 332: Added just the `BeforeCreate` hook.

`BeforeUpdate` hook is not ideal as we will be guarding the collision logic with `model.changed("index")` so as to not run it for all updates to a collection. This ends up not removing the collision when the new index for `collections.move` is the same as the existing index.

tommoor: > This ends up not removing the collision when the new index for collections.move is the same as the existing index.

Being dumb – Isn't it a no-op if the index stays the same?

hmacr: Yep 🤦‍♂️ I guess I was unnecessarily covering for some failing tests.

tommoor (approved):
Nice, this is the refactor I wanted to do – would be good to include a test case for the bug in question

PR #8809, opened 2025-03-27
fix: Improve performance of notification queries

### LINKED ISSUE #8807: Comment notifications are sent to all users with read access
Maybe introduced here https://github.com/outline/outline/issues/4984

tommoor:
So far I've been unable to reproduce the reported issue of users receiving emails for comments they didn't participate in, but this is a nice refactor that simplifies, improves performance, and removes the dangerous `onlySubscribers` prop

closes #8807

### DIFF
--- a/server/models/helpers/NotificationHelper.test.ts
+++ b/server/models/helpers/NotificationHelper.test.ts
@@ -9,7 +9,7 @@ import NotificationHelper from "./NotificationHelper";
 
 describe("NotificationHelper", () => {
   describe("getCommentNotificationRecipients", () => {
-    it("should return users who have notification enabled for comment creation and are subscribed to the document in case of parent comment", async () => {
+    it("should only return users who have notification enabled for comment creation and are subscribed to the document in case of new thread", async () => {
       const documentAuthor = await buildUser();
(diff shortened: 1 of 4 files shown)


hmacr:
> So far I've been unable to reproduce the reported issue of users receiving emails for comments they didn't participate in

https://github.com/outline/outline/pull/8349 - seems a bit related to this fix 🤔

hmacr:
I also noticed an edge case where we could be sending emails after a user is removed from a collection - https://github.com/outline/outline/pull/8821 😶 

On a related note, do you think maybe an active collection subscription is throwing them off when they mean "didn't participate in a document"? Perhaps, the users would have expected to receive new document / revision updates but not comments?

tommoor:
> On a related note, do you think maybe an active collection subscription is throwing them off when they mean "didn't participate in a document"? Perhaps, the users would have expected to receive new document / revision updates but not comments?

They don't have a collection subscription in this case, the menu shows the "Subscribe" option not disabled. I'm out of ideas, the only thing I have left is that they are receiving notifications for a comment thread they are involved in. Will find out a bit more later today hopefully 🤞

PR #8821, opened 2025-03-28
Manage collection subscriptions when user (or) group is removed from a collection

hmacr:
Missing piece from #8392 🤦‍♂️

### DIFF
--- a/server/queues/processors/DocumentSubscriptionProcessor.ts
+++ b/server/queues/processors/DocumentSubscriptionProcessor.ts
@@ -1,30 +1,78 @@
 import { Op } from "sequelize";
 import { GroupUser } from "@server/models";
-import { DocumentGroupEvent, DocumentUserEvent, Event } from "@server/types";
+import {
+  CollectionGroupEvent,
+  CollectionUserEvent,
+  DocumentGroupEvent,
+  DocumentUserEvent,
+  Event,
+} from "@server/types";
+import CollectionSubscriptionTask from "../tasks/CollectionSubscriptionTask";
 import DocumentSubscriptionTask from "../tasks/DocumentSubscriptionTask";
 import BaseProcessor from "./BaseProcessor";
 
+type ReceivedEvent =
+  | CollectionUserEvent
+  | CollectionGroupEvent
+  | DocumentUserEvent
+  | DocumentGroupEvent;
+
 export default class DocumentSubscriptionProcessor extends BaseProcessor {
   static applicableEvents: Event["name"][] = [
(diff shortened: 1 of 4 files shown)


server/queues/tasks/CollectionSubscriptionTask.ts
```diff
@@ -0,0 +1,52 @@
...
+import { SubscriptionType } from "@shared/types";
+import { createContext } from "@server/context";
+import Logger from "@server/logging/Logger";
+import { Collection, Subscription, User } from "@server/models";
+import { can } from "@server/policies";
+import { sequelize } from "@server/storage/database";
+import { CollectionUserEvent } from "@server/types";
+import BaseTask from "./BaseTask";
+
+export default class CollectionSubscriptionTask extends BaseTask<CollectionUserEvent> {
```
tommoor, line 11: Lets rename these tasks to better reflect their function – e.g. `CollectionSubscriptionRemoveUserTask`

We can also remove the `event.name` check as we're in charge of when this runs.

server/queues/tasks/DocumentSubscriptionRemoveUserTask.ts
```diff
@@ -8,11 +8,11 @@ import { sequelize } from "@server/storage/database";
 import { DocumentUserEvent } from "@server/types";
 import BaseTask from "./BaseTask";
 
-export default class DocumentSubscriptionTask extends BaseTask<DocumentUserEvent> {
+export default class DocumentSubscriptionRemoveUserTask extends BaseTask<DocumentUserEvent> {
```
hmacr, line 11: While we're here, renamed the task for completeness.

server/queues/processors/DocumentSubscriptionProcessor.ts
```diff
@@ -1,30 +1,78 @@
...
   }
 
-  private async handleGroup(event: DocumentGroupEvent) {
+  private async handleRemoveGroupFromCollection(event: CollectionGroupEvent) {
+    await GroupUser.findAllInBatches<GroupUser>(
+      {
+        where: {
+          groupId: event.modelId,
+          userId: {
+            [Op.ne]: event.actorId,
```
tommoor, line 56: Why do we skip the actor? Feels like the check for remaining access removes the need for this

hmacr: agreed, thanks for finding it :)
I believe this wasn't removed when the document access check was added later in the downstream task.

PR #8828, opened 2025-03-28
fix: Handle switching between regular lists and todo lists

### LINKED ISSUE #8826: Converting a list to checkbox is broken

### Current Behavior

it somehow doubles up the bullets, and so now the grabbers are in the way of checking it off, and it results in some crazy states when you drag those lists

https://github.com/user-attachments/assets/28c72c32-520a-4822-819c-03d19677fc30

### Expected Behavior

it should conver to checkboxes in place without added grabbers/bullets

### Steps To Reproduce

1. make list
2. highlight list
3. change to checkboxes

### Environment

```markdown
- Outline: prod
- Browser: chrome
```

### Anything else?

hi @tommoor

adityash1:
Closes #8826

### DIFF
--- a/shared/editor/commands/toggleList.ts
+++ b/shared/editor/commands/toggleList.ts
@@ -1,5 +1,5 @@
 import { NodeType } from "prosemirror-model";
-import { wrapInList, liftListItem } from "prosemirror-schema-list";
+import { liftListItem, wrapInList } from "prosemirror-schema-list";
 import { Command } from "prosemirror-state";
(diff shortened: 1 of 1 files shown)


tommoor (approved):
This actually works quite well! It does cause the structure to get lost if you have nested lists but it's still a lot better than today and the behavior seems to not have changed when switching between ordered/unordered (that's a good thing)

adityash1:
yes. nested lists scenario is still not properly handled but this is better for now

PR #8838, opened 2025-03-30
fix: New comments are measured incorrectly

### LINKED ISSUE #8834: Overlapping comments when adding new comment in sidebar
Image

tommoor:
closes #8834
regressed in https://github.com/outline/outline/pull/8338

### DIFF
--- a/shared/hooks/useComponentSize.ts
+++ b/shared/hooks/useComponentSize.ts
@@ -21,23 +21,14 @@ export function useComponentSize(
   input: HTMLElement | null | React.RefObject<HTMLElement | null>
 ): DOMRect | typeof defaultRect {
   const element = input instanceof HTMLElement ? input : input?.current;
-  const [size, setSize] = useState(() => element?.getBoundingClientRect());
-
-  useLayoutEffect(() => {
-    const sizeObserver = new ResizeObserver(() => {
-      element?.dispatchEvent(new CustomEvent("resize"));
-    });
-    if (element) {
-      sizeObserver.observe(element);
-    }
-    return () => sizeObserver.disconnect();
-  }, [element]);
+  const [size, setSize] = useState<DOMRect | undefined>(
+    () => element?.getBoundingClientRect() || new DOMRect()
+  );
 
   useLayoutEffect(() => {
     const handleResize = () => {
       setSize((state) => {
         const rect = element?.getBoundingClientRect();
-
         if (
           rect &&
           state &&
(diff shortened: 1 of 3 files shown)


shared/hooks/useComponentSize.ts
```diff
@@ -21,23 +21,14 @@ export function useComponentSize(
...
-    const sizeObserver = new ResizeObserver(() => {
-      element?.dispatchEvent(new CustomEvent("resize"));
-    });
-    if (element) {
-      sizeObserver.observe(element);
-    }
-    return () => sizeObserver.disconnect();
-  }, [element]);
+  const [size, setSize] = useState<DOMRect | undefined>(
+    () => element?.getBoundingClientRect() || new DOMRect()
```
tommoor, line 25: This is the actual fix, ensuring this is set avoids the size being skipped on L34

shared/hooks/useComponentSize.ts
```diff
@@ -63,5 +55,15 @@ export function useComponentSize(
     };
   });
 
+  useLayoutEffect(() => {
```
tommoor, line 58: Moved this afterwards as it makes sense to setup the listeners _after_ the handlers.

shared/hooks/useComponentSize.ts
hmacr: I might be wrong, this seems unnecessary with the call to `handleResize` on uLE setup 🤔 Let me test and see if this works well without it.

tommoor: It's very finicky

shared/hooks/useComponentSize.ts
```diff
@@ -55,6 +46,7 @@ export function useComponentSize(
     window.addEventListener("click", handleResize);
     window.addEventListener("resize", handleResize);
     element?.addEventListener("resize", handleResize);
+    handleResize();
 
     return () => {
```
hmacr, line 41: Not a feedback on this PR.. totally unrelated TIL for me - in these cases of cleaning up multiple event listeners, using an AbortController's signal makes it very neat :)

https://kettanaito.com/blog/dont-sleep-on-abort-controller

tommoor: TIL indeed!

shared/hooks/useComponentSize.ts
hmacr: Doesn't work without it - not worth debugging I suppose.

hmacr (approved):
Tested 👍

Issue #8506, opened 2025-02-20
Improvements to GitHub integration


hmacr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Environment

```markdown

```

### Anything else?

Improve the base GitHub integration workflow so it's easier to extend to other issue tracker services, eg. Linear, Jira.

- [ ] Convert widget to "mention" model, that displays title & status.
- [ ] Show additional info like description, assignee on hover.
- [ ] Add a "Convert to issue" popup when selecting a text in document. This should create a new issue and replace the text with the mention.

PR #8870, opened 2025-04-02
Show GitHub issues and pull requests as mentions

hmacr:
Towards #8506 

https://github.com/user-attachments/assets/6c5eea1c-abad-4c14-86da-1b189851aa11


app/editor/components/PasteMenu.tsx
```diff
@@ -41,9 +54,19 @@ const PasteMenu = ({ embeds, ...props }: Props) => {
...
         keywords: embed?.keywords,
       },
-    ],
-    [embed, t]
-  );
+    ];
+
+    if (mentionType) {
+      menuItems.push({
+        name: "mention",
```
hmacr, line 61: Having a mention icon would be a nice addition here.

app/editor/extensions/PasteHandler.tsx
```diff
@@ -415,6 +415,27 @@ export default class PasteHandler extends Extension {
...
+      const tr = state.tr.deleteRange(result[0], result[1]);
+      view.dispatch(
+        tr.setSelection(TextSelection.near(tr.doc.resolve(result[0])))
+      );
+    }
+
+    this.editor.commands.mention({
+      id: v4(),
+      type: attrs.type as MentionType,
+      label: this.state.pastedText,
```
hmacr, line 433: Contemplated between adding a `href` attribute vs re-using `label`.
For these mentions, we don't really have any use for a separate `label`, so ended up storing the url in label.

app/stores/UnfurlsStore.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default class UnfurlsStore {
+  private rootStore: RootStore;
+
+  constructor(rootStore: RootStore) {
+    this.rootStore = rootStore;
+  }
+
+  fetch = async <UnfurlType extends UnfurlResourceType>(url: string) => {
+    const data = await client.post("/urls.unfurl", {
```
hmacr, line 13: ESLint prevents importing `client` in shared module - so, created this store for the backend api call.
In future, this can hold an `Unfurl` model which will allow realtime updates in the webhooks -> websockets flow.

shared/components/Text.tsx
```diff
@@ -0,0 +1,73 @@
...
+  ellipsis?: boolean;
+  /** Whether the text should be monospaced */
+  monospace?: boolean;
+};
+
+/**
+ * Use this component for all interface text that should not be selectable
+ * by the user, this is the majority of UI text explainers, notes, headings.
+ */
+const Text = styled.span<Props>`
```
hmacr, line 27: Essentially a copy of `Text` component from app to use in `shared` module.
The one is `app` was being used in many components - moving it would have led to a lot of import changes.

app/editor/extensions/PasteHandler.tsx
hmacr: Added a `href` attribute which holds the issue / pr url. `label` stores the title.
Additionally, `unfurl` attribute has the dump of full unfurl data (except description).

app/editor/components/PasteMenu.tsx
```diff
@@ -52,9 +86,7 @@ const PasteMenu = ({ embeds, ...props }: Props) => {
       filterable={false}
       renderMenuItem={(item, _index, options) => (
         <SuggestionsMenuItem
-          onClick={() => {
-            props.onSelect?.(item);
-          }}
+          onClick={options.onClick}
```
hmacr, line 89: This doesn't make a difference when pasting embeds 🤷‍♂️

Seems it was added to handle "link" item here, but it was superseded by a "noop" action.
https://github.com/outline/outline/pull/8229/commits/c3ba07cee467ce5420f40f3e536a104c437a8b5f#diff-ec86617a54f34afbbbb2d9fd55d800d9a481e9325674aadd39ff4cc3f8326f43R15

shared/editor/nodes/Mention.tsx
```diff
@@ -149,29 +178,42 @@ export default class Mention extends Node {
...
-            selection.node.attrs.type === MentionType.Document
-              ? "doc"
-              : selection.node.attrs.type === MentionType.Collection
-              ? "collection"
-              : undefined;
+          if (
+            mentionType === MentionType.Issue ||
+            mentionType === MentionType.PullRequest
+          ) {
+            link = selection.node.attrs.label;
```
tommoor, line 204:
```suggestion
            link = selection.node.attrs.href;
```

right?

app/stores/UnfurlsStore.ts
tommoor: Even without the websocket this could be an asset – we can use it to cache the unfurl response and avoid fetching every time you open the document

shared/components/Text.tsx
tommoor: In the meantime we can switch `components/Text` to simply pass through to this shared module to avoid the duplication

server/models/helpers/ProsemirrorHelper.tsx
```diff
@@ -228,14 +232,15 @@ export class ProsemirrorHelper {
...
+    const mentionWithoutEmptyValues = omitBy(mention, isNil);
 
     const isNodeContainingMention = (node: Node) => {
       let foundMention = false;
 
       node.descendants((childNode: Node) => {
         if (
           childNode.type.name === "mention" &&
-          isEqual(childNode.attrs, mention)
+          isEqual(omitBy(childNode.attrs, isNil), mentionWithoutEmptyValues) // ignore null and undefined values for comparison
```
tommoor, line 243: We might be able to use [isMatch](https://lodash.com/docs/4.17.15#isMatch) here

app/models/Integration.ts
```diff
@@ -29,6 +30,45 @@ class Integration<T = unknown> extends Model {
 
   @observable
   settings: IntegrationSettings<T>;
+
+  isMentionable(url: URL) {
```
tommoor, line 34: Not blocking – I'm not certain these helpers belong here, could be some sort of mention lib / `MentionHelper` would be better.

tommoor (commented):
🙏 feels close to mergeable

shared/editor/nodes/Mention.tsx
```diff
@@ -218,4 +260,27 @@ export default class Mention extends Node {
...
+        unfurl.type === UnfurlResourceType.Issue ||
+        unfurl.type === UnfurlResourceType.PR
+          ? unfurl.title
+          : undefined;
+
+      const overrides: Record<string, unknown> = label ? { label } : {};
+      overrides.unfurl = unfurl;
+
+      const pos = getPos();
+      const transaction = tr.setNodeMarkup(pos, undefined, {
```
tommoor, line 280: This seems to be erroring currently, tried on a fresh paragraph and in a list 

image

shared/editor/nodes/Mention.tsx
hmacr: yep

app/stores/UnfurlsStore.ts
hmacr: With this PR, PR / issue mention misses out on latest updates from external source when the document is open. Only hover previews can fetch those updates which might lead to data mismatch between them.
As you suggested, I suppose the best thing is to introduce the `Unfurl` model. I'm leaning towards including it in this PR itself to prevent the data mismatch issue.

app/editor/components/PasteMenu.tsx
```diff
@@ -41,9 +57,27 @@ const PasteMenu = ({ embeds, ...props }: Props) => {
...
+      menuItems.push({
+        name: "mention",
+        title: t("Mention"),
+        icon: <EmailIcon />,
+        attrs: {
+          id: v4(),
+          type: mentionType,
+          label: pastedText,
+          href: pastedText,
+          modelId: v4(),
```
tommoor, line 72: Re https://github.com/outline/outline/pull/8920#discussion_r2035778943 – we can remove this

hmacr: Ok, just need to make sure that making `modelId` an optional attribute for the node does not break anything in server.

tommoor: mmm, it is finicky. Ok nevermind sorry

shared/editor/nodes/Mention.tsx
hmacr: Turns out, node attributes were updated after it was removed from doc.
This is fixed by checking whether the corresponding node view is still mounted by React.

tommoor (approved):
Fixed a couple of things, should be good

shared/editor/nodes/Mention.tsx
```diff
@@ -82,6 +82,10 @@ export default class Mention extends Node {
               actorId: dom.dataset.actorid,
               label: dom.innerText,
               id: dom.id,
+              href: dom.getAttribute("href"),
```
hmacr, line 85: 🤦

PR #8877, opened 2025-04-04
fix: Handle duplicate urlId in import persistence

### LINKED ISSUE #8876: SequelizeUniqueConstraintError: Validation error
Sentry Issue: [OUTLINE-CLOUD-8HZ](https://outline-jl.sentry.io/issues/6505757479/?referrer=github_integration)

```
SequelizeUniqueConstraintError: Validation error
  File "/app/build/server/queues/processors/ImportsProcessor.js", line 340, in <anonymous>
    await document.saveWithCtx(ctx, {
  File "/app/build/server/models/base/Model.js", line 197, in import_task.findAllInBatches
    await callback(results, mappedQuery);
  File "/app/build/server/queues/processors/ImportsProcessor.js", line 230, in NotionImportsProcessor.createCollectionsAndDocuments
    await _ImportTask.default.findAllInBatches({
  File "/app/build/server/queues/processors/ImportsProcessor.js", line 127, in NotionImportsProcessor.onProcessed
    } = await this.createCollectionsAndDocuments({
  File "/app/build/server/queues/processors/ImportsProcessor.js", line 40, in NotionImportsProcessor.perform
    await _database.sequelize.transaction(async transaction => {
...
(7 additional frame(s) were not displayed)
```

hmacr:
Closes #8876

### DIFF
--- a/server/queues/processors/ImportsProcessor.ts
+++ b/server/queues/processors/ImportsProcessor.ts
@@ -26,10 +26,12 @@ import ImportTask, {
   ImportTaskAttributes,
   ImportTaskCreationAttributes,
 } from "@server/models/ImportTask";
+import { EventOverrideOptions } from "@server/models/base/Model";
 import { DocumentHelper } from "@server/models/helpers/DocumentHelper";
 import { ProsemirrorHelper } from "@server/models/helpers/ProsemirrorHelper";
 import { sequelize } from "@server/storage/database";
-import { Event, ImportEvent } from "@server/types";
+import { APIContext, Event, ImportEvent } from "@server/types";
+import { generateUrlId } from "@server/utils/url";
 import BaseProcessor from "./BaseProcessor";
 
 export const PagePerImportTask = 3;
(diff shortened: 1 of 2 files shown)


server/queues/processors/ImportsProcessor.ts
```diff
@@ -500,6 +496,42 @@ export default abstract class ImportsProcessor<
...
+    eventOpts: EventOverrideOptions
+  ): Promise<Collection> {
+    const existingUrlId = await Collection.count({
+      where: { urlId: collection.urlId },
+      paranoid: false,
+      transaction: ctx.context.transaction,
+    });
+
+    if (existingUrlId) {
+      collection.urlId = generateUrlId();
```
hmacr, line 511: This is in line with existing file-based imports flow.

Ideally, we'd catch `UniqueConstraintError` -> verify `urlId` is the cause before generating another one. But, with restrictions on sequelize [retrying a failed transaction](https://github.com/sequelize/sequelize/issues/12608), I'm not completely confident to use it.

hmacr: I couldn't repro this organically as well.
There's only two causes for this error in document persistence - `id` and `urlId`. `urlId` is a random generated string, so more likelihood of collision compared to `id` (backed by uuid) - I'm confident `id` isn't the cause.

For testing, I forcibly set a duplicate `urlId` for the persistence to ensure it's being replaced.

tommoor: Okay, in that case I would like to investigate further and avoid merging this complexity if it might not truly be the issue – this error has occurred 20 times so it's very unlikely to be random string collision

hmacr:
Will re-open once the root cause is narrowed down.

PR #8884, opened 2025-04-06
OAuth provider

tommoor:
This PR contains the necessary work to make Outline an OAuth provider including:

- OAuth app registration
- OAuth app management
- Private / public apps (Public in cloud only)
- Full OAuth 2.0 spec compatible authentication flow
- Granular scopes
- User token management screen in settings
- Associated API endpoints for programatic access

---

image

---

image

---

image

---

image


app/scenes/Login/OAuthScopeHelper.ts
```diff
@@ -0,0 +1,56 @@
...
+    const methodToReadable = {
+      list: t("read"),
+      info: t("read"),
+      read: t("read"),
+      write: t("write"),
+      create: t("write"),
+      update: t("write"),
+      delete: t("write"),
+      "*": t("manage"),
+    };
```
R167, line 17: It seems like the naming here differs slightly from what the backend calls it in methodToScope? Specifically the `"*": "manage"` entry

https://github.com/outline/outline/pull/8884/files#diff-f59462f1f54f9b78195ce4dc8b29cc24a94e74043d8959f8f69764d08590b543R9-R23

tommoor: "Manage" is kind of a catch all, maybe too ambiguous and should be changed to "Read & Write"?

server/routes/oauth/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    ctx.body = { code: authorizationCode };
+  }
+);
+
+router.post("/token", async (ctx) => {
+  // Note: These objects are mutated by the OAuth2Server library
+  const request = new OAuth2Server.Request(ctx.request);
+  const response = new OAuth2Server.Response(ctx.response);
+  const token = await oauth.token(request, response);
```
R167, line 71: Does this automatically pull in the existing defaults we generated for 30 day tokens during creation, or does it need the options specified a second time so it doesn't default to a 2w refresh token?

Looking at oauth-server code I think the answer is no.

server/routes/oauth/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const client = await OAuthClient.findByClientId(clientId);
+    authorize(user, "read", client);
+
+    // Note: These objects are mutated by the OAuth2Server library
+    const request = new OAuth2Server.Request(ctx.request);
+    const response = new OAuth2Server.Response(ctx.response);
+
+    const authorizationCode = await oauth.authorize(request, response, {
+      allowEmptyState: true,
```
R167, line 45: What's the rationale for allowing empty state here? it seems like this reduces csrf protection?

tommoor: Generally trying to keep the implementation as flexible as possible – even Google does not _require_ the state parameter, it is left as a recommended option for the client to implement

PR #8902, opened 2025-04-08
fix: Increase lock timeout for calculating document diff

tommoor:


server/utils/CacheHelper.ts
```diff
@@ -18,11 +18,14 @@ export class CacheHelper {
...
    * @param expiry Cache data expiry in seconds
+   * @param lockTimeout Lock timeout in milliseconds
+   * @returns The data from cache or the result of the callback
    */
   public static async getDataOrSet<T>(
     key: string,
     callback: () => Promise<T | undefined>,
-    expiry?: number
+    expiry: number,
+    lockTimeout: number = MutexLock.defaultLockTimeout
```
tommoor, line 28: Obviously I don't love that this is a mixture of seconds and milliseconds – if we change `expiry` here to be milliseconds then it's the only method in this helper like that which is also weird.

hmacr (approved):

PR #8918, opened 2025-04-09
fix: `collectionIndexing` results in `teamId` undefined error

tommoor:
This seems to be due to a bug in Sequelize when selecting fields to query. The bug can be reproduced by removing the changes in `indexing.ts` and running the included tests

### DIFF
--- a/server/utils/indexing.ts
+++ b/server/utils/indexing.ts
@@ -10,10 +10,8 @@ export async function collectionIndexing(
   const collections = await Collection.findAll({
     where: {
       teamId,
-      // no point in maintaining index of deleted collections.
-      deletedAt: null,
     },
-    attributes: ["id", "index", "name"],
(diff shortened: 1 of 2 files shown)


server/utils/indexing.ts
```diff
@@ -10,10 +10,8 @@ export async function collectionIndexing(
   const collections = await Collection.findAll({
     where: {
       teamId,
-      // no point in maintaining index of deleted collections.
-      deletedAt: null,
```
tommoor, line 14: This is pointless, default scope already removes deleted collections

hmacr:
Ugh.. looks like a side effect of https://github.com/outline/outline/pull/8803/files#diff-3d318c0a8ba5b1a20a9dd32bdabfbb86191f72f17e4bcee8b6850d349062405eR383

tommoor:
That'll be it 😢

PR #8920, opened 2025-04-09
Implement UnfurlsStore

hmacr:
Adds an `UnfurlsStore` and `Unfurl` model.

Foundational for #8870

### DIFF
--- a/app/components/HoverPreview/HoverPreview.tsx
+++ b/app/components/HoverPreview/HoverPreview.tsx
@@ -1,4 +1,5 @@
 import { m } from "framer-motion";
+import { observer } from "mobx-react";
 import * as React from "react";
 import { Portal } from "react-portal";
 import styled from "styled-components";
@@ -8,6 +9,7 @@ import useEventListener from "~/hooks/useEventListener";
 import useKeyDown from "~/hooks/useKeyDown";
 import useMobile from "~/hooks/useMobile";
 import useOnClickOutside from "~/hooks/useOnClickOutside";
+import useStores from "~/hooks/useStores";
 import LoadingIndicator from "../LoadingIndicator";
 import { CARD_MARGIN } from "./Components";
 import HoverPreviewDocument from "./HoverPreviewDocument";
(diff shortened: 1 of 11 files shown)


app/stores/UnfurlsStore.ts
```diff
@@ -0,0 +1,60 @@
...
+        url,
+        documentId,
+      });
+
+      // unfurls can succeed with no data.
+      if (!data) {
+        return;
+      }
+
+      return this.add({ id: url, type: data.type, data } as Unfurl<T>);
```
hmacr, line 44: Making `url` as the unique identified will allow duplicate requests for the same mentioned user, but I think this provides better code readability.
Alternative would have been to parse url -> parse `modelId` if it's a mention -> use `modelId` as unique identifier.

tommoor: Lets use `modelId` if it's available and otherwise use the url?

hmacr: Ok 👍

tommoor: Changed my mind sorry, we can simply and focus on url – occassional dupe is ok. Still you can just use `get(url)` to find in store

hmacr: yep - so PR is reviewable.

tommoor (approved):
image

app/components/HoverPreview/HoverPreview.tsx
```diff
@@ -36,151 +38,155 @@ enum Direction {
...
+    const [isVisible, setVisible] = React.useState(false);
+    const timerClose = React.useRef<ReturnType<typeof setTimeout>>();
+    const cardRef = React.useRef<HTMLDivElement | null>(null);
+    const { cardLeft, cardTop, pointerLeft, pointerTop, pointerDir } =
+      useHoverPosition({
+        cardRef,
+        element,
+        isVisible,
+      });
+    const data = unfurlId ? unfurls.get(unfurlId)?.data : undefined;
```
hmacr, line 53: Diff looks messy, but the changes are
- converted this component to an observer for fetching latest unfurl.
- replaced `data` with `unfurlId` in Props.

hmacr:
As discussed in DM, added refetch logic for unfurls that are older than 5 mins.

PR #8937, opened 2025-04-12
feat: Add PDF embedding functionality

hunmac9:
ezgif-456dc1c6878b32
This PR would add the ability to embed PDFs (addressing #1788, #7742, #7903, and #3542) directly into documents using the native browser renderer in an iframe, making it super lightweight and easy to maintain. I played with react-pdf and pdf.js implementations and I just wasn't happy with the speed/resource usage. This felt like a good middle ground between implementing a heavy handed solution and not having any pdf preview. It uses already present icon sets and acts nearly identically to current file attachments.

Theoretically this was possible with a PDF URL starting from PR #7319 but I encountered so many issues with this; for one, out of the box it doesn't work unless you remove sandbox attributes due to PDF scripting concerns, which I think is unreasonable for privileged environments like this. Even with that relatively minor modification, it doesn't upload files to the native database which makes it useless for multiplayer with local files. 

Here's a quick rundown of what I changed:

*   Added a new `pdf_document` node type (`shared/editor/nodes/Pdf.tsx`) to handle the PDF logic.
*   Created a React component (`shared/editor/components/PdfEmbed.tsx`) to actually render the PDF embed using an iframe. It includes lazy loading and resizing.
*   Updated the file drop logic in `app/components/Editor.tsx` and the `insertFiles` command (`shared/editor/commands/insertFiles.ts`) to recognize PDFs and create the right node.
*   Hooked it up to the UI: added an "Upload PDF" option in the slash menu (`app/editor/menus/block.tsx`, `app/editor/components/SuggestionsMenu.tsx`) and a context menu for selected PDFs (`app/editor/menus/pdfMenu.tsx`, `app/editor/components/SelectionToolbar.tsx`).
*   Tweaked the `Widget` component (`shared/editor/components/Widget.tsx`) a bit to play nicer with embeds like this.
*   Added the necessary translations entry and some basic styling.

I'm not a programmer by most definitions, and primarily added this for my own personal usage, so if there any issues I won't take criticism harshly.


tommoor:
Thanks for your contribution – I'm getting the impression this was "vibe coded"?

We don't want to introduce an entire new node type for this - it should still be a file attachment but with an additional attribute to denote that the preview is embedded in an acceptable implementation.

Issue #8930, opened 2025-04-11
Unable to recognize documents with the same title located in the same collection in the @ and link operation


YKDZ:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

My document collection contains multiple documents with the same title (located in different document tree levels). When using @ to link them, I am unable to identify the correct document I want to link from the information currently displayed, as they all look the same.

![Image](https://github.com/user-attachments/assets/33b23165-8acf-4758-ab2f-d7efb1bd1861)

### Expected Behavior

This pop-up window should display more detailed document path and location information, or display a document summary when hovering a line.

### Steps To Reproduce

1. Create document with same name in a collection
2. Use @ to link them in another document and find the issue 

### Environment

```markdown
- Outline: v0.82.0
- Browser: Chrome
```

YKDZ:
I just conducted a similar test in the link toolbar restored in v0.83.0 and found the same issue. Completely unable to recognize documents with the same name and located in the same collection.

YKDZ:
I still believe that it is reasonable to have documents with the same name and colloection. In my example, I have the following structure of documents:

- Plugin
  - Product
    - Event Hook
  - Shop
    - Event Hook
  - Menu
    - Icon
      - Event Hook  
    - Event Hook
 
Therefore, I believe the solution to this problem is to replace the current collection title with the titles of at most two parent objects (document or collection) as the subtitle for menu items:

- Event Hook · Plugin / Product
- Event Hook · Plugin / Shop
- Event Hook · Plugin / Menu
- Event Hook · Menu / Icon

And at the same time, add a document summary hover window similar to @ and link to the menu item:

![Image](https://github.com/user-attachments/assets/9b9e84cf-7e0f-429c-8068-7382477ee4d6)

tommoor:
I'm going to close this as a duplicate of https://github.com/outline/outline/issues/8676 but will prioritize for next release

YKDZ:
Sorry to mention the issue that has already been closed, but I have done something with DocumentBreadcrumb:

![Image](https://github.com/user-attachments/assets/c7442bf4-6191-4dfe-9d41-de0fe2a72ed7)

![Image](https://github.com/user-attachments/assets/dfec4203-5208-48aa-b880-639852fce24c)

Can I submit it as a pull request?

tommoor:
[...]

PR #8938, opened 2025-04-12
Display breadcrumb instead of collection name when link and mention document

YKDZ:
Attempt to implement the optimization mentioned in #8930 for the display content of the document search box item in the context and link.
image
image

### DIFF
--- a/app/components/DocumentBreadcrumb.tsx
+++ b/app/components/DocumentBreadcrumb.tsx
@@ -98,7 +98,7 @@ function DocumentBreadcrumb(
   const items = React.useMemo(() => {
     const output = [];
 
-    if (category && output.length) {
+    if (category) {
       output.push(category);
     }
     if (collectionNode) {
--- a/app/components/DocumentBreadcrumb.tsx
+++ b/app/components/DocumentBreadcrumb.tsx
@@ -19,7 +19,7 @@ type Props = {
   document: Document;
   onlyText?: boolean;
   reverse?: boolean;
-  reversedLength?: number;
+  maxDepth?: number;
 };
 
 function useCategory(document: Document): MenuInternalLink | null {
@@ -56,7 +56,7 @@ function useCategory(document: Document): MenuInternalLink | null {
 }
 
 function DocumentBreadcrumb(
-  { document, children, onlyText, reverse = false, reversedLength = 2 }: Props,
+  { document, children, onlyText, reverse = false, maxDepth }: Props,
   ref: React.RefObject<HTMLDivElement> | null
 ) {
   const { collections } = useStores();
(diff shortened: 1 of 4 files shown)


app/components/DocumentBreadcrumb.tsx
```diff
@@ -96,15 +98,20 @@ function DocumentBreadcrumb(
   const items = React.useMemo(() => {
     const output = [];
 
-    if (category) {
-      output.push(category);
+    if (!reverse) {
```
tommoor, line 101: Repeating this logic twice is not good, it would be simpler to reverse the breadcrumb at the end after all other logic is complete

app/components/DocumentBreadcrumb.tsx
```diff
@@ -18,6 +18,8 @@ type Props = {
   children?: React.ReactNode;
   document: Document;
   onlyText?: boolean;
+  reverse?: boolean;
+  reversedLength?: number;
```
tommoor, line 22: It doesn't make sense for this to apply only to reversing, a prop is fine but it should be called `maxLength` or `maxDepth` and apply to both cases

app/editor/components/LinkEditor.tsx
```diff
@@ -277,7 +280,7 @@ const Wrapper = styled(Flex)`
   gap: 8px;
 `;
 
-const SearchResults = styled(Scrollable)<{ $hasResults: boolean }>`
+const SearchResults = styled(Scrollable) <{ $hasResults: boolean }>`
```
tommoor, line 283:
```suggestion
const SearchResults = styled(Scrollable)<{ $hasResults: boolean }>`
```

app/editor/components/MentionMenu.tsx
```diff
@@ -57,7 +58,7 @@ function MentionMenu({ search, isActive, ...rest }: Props) {
       res.data.documents.map(documents.add);
       res.data.users.map(users.add);
       res.data.collections.map(collections.add);
-    }, [search, documents, users])
+    }, [search, documents, users, collections.add])
```
tommoor, line 61:
```suggestion
    }, [search, documents, users, collections])
```

app/editor/components/MentionMenu.tsx
```diff
@@ -176,7 +179,7 @@ function MentionMenu({ search, isActive, ...rest }: Props) {
       setItems(items);
       setLoaded(true);
     }
-  }, [t, actorId, loading, search, users, documents, maxResultsInSection]);
+  }, [t, actorId, loading, search, users, documents, maxResultsInSection, React.useEffect]);
```
tommoor, line 182:
```suggestion
  }, [t, actorId, loading, search, users, documents, maxResultsInSection]);
```

tommoor (changes requested):

app/components/DocumentBreadcrumb.tsx
```diff
@@ -91,20 +93,19 @@ function DocumentBreadcrumb(
...
   }
 
-  const path = document.pathTo;
+  const path = document.pathTo.slice(0, -1);
 
   const items = React.useMemo(() => {
     const output = [];
 
-    if (category) {
+    if (category && output.length) {
```
tommoor, line 101: This will never be true, the `output` array is initialized just above?

tommoor (approved):
Thanks for the improvement

PR #8949, opened 2025-04-13
fix: Reset avatar zoom, set missing text in upload button

hmacr:
Started out as a couple of bug fixes (stated in title) and ended up refactoring a bit to use existing patterns.


app/scenes/Settings/components/ImageUpload.tsx
```diff
@@ -121,31 +177,13 @@ const ImageUpload: React.FC<Props> = ({
...
-    </Modal>
-  );
-
-  if (isCropping && file) {
-    return renderCropping();
+    );
   }
-
-  return (
-    <Dropzone
```
hmacr, line 136: Turns out there is a `useDropZone` hook available for this HOC.

tommoor (approved):

PR #8952, opened 2025-04-13
Cache `issueSources` for embed integrations

hmacr:
Extracted this piece from #8951.

This will ensure new github integrations have their `issueSources` populated.
We also need a backfill task for the existing integrations which will be in a follow-on PR.


server/utils/BaseIssueProvider.ts
```diff
@@ -0,0 +1,15 @@
+import { IssueSource } from "@shared/schema";
+import { IntegrationType, IssueTrackerIntegrationService } from "@shared/types";
+import { Integration } from "@server/models";
+
+export abstract class BaseIssueProvider {
```
hmacr, line 5: Partial abstraction from https://github.com/outline/outline/pull/8951/files#diff-6550778890c7e92193c333f7e50e59ae578465f149fca710533a2b7b13b72ed1R10

server/queues/tasks/CacheIssueSourcesTask.ts
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+  integrationId: string;
+};
+
+export default class CacheIssueSourcesTask extends BaseTask<Props> {
+  async perform({ integrationId }: Props) {
+    await sequelize.transaction(async (transaction) => {
+      const integration = await Integration.findByPk(integrationId, {
+        transaction,
+        lock: transaction.LOCK.UPDATE,
```
tommoor, line 17: Lets not take a lock the entire time `fetchSources` is running as it may take a very long time

server/queues/tasks/CacheIssueSourcesTask.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const plugins = PluginManager.getHooks(Hook.IssueProvider);
+
+type Props = {
+  integrationId: string;
+};
+
+export default class CacheIssueSourcesTask extends BaseTask<Props> {
+  async perform({ integrationId }: Props) {
+    const integration = await Integration.findByPk(integrationId);
```
tommoor, line 14: Maybe we should grab a MutexLock to prevent multiple of these running at once, thoughts? defer until later?

tommoor (approved):

server/queues/tasks/CacheIssueSourcesTask.ts
hmacr: Feels very rare that queue would consider it a _stalled_ job and retry it, given that the task is I/O bound rather than CPU.

tommoor:
Small conflict now, can merge once resolved

PR #8978, opened 2025-04-15
Improve invite form UI with delete functionality

MehdiBouzouaya:
The invite modal becomes unusable with multiple users - action buttons disappear off-screen with no scroll functionality, forcing users to zoom out as a workaround.
i Added delete (×) buttons to invite rows to:
• Remove unnecessary entries and reduce modal height
• Keep action buttons accessible without zooming
• Improve overall form usability
image


tommoor:
Thanks, but this PR includes a lot of additional unneccessary changes – 

The core issue of the modal not being scrollable is fixed in https://github.com/outline/outline/pull/8981

PR #8981, opened 2025-04-15
fix: Small modal overflow scrolling behavior

### LINKED ISSUE #8966: "Invite to Workspace" modal overflows with multiple inputs.

### Current Behavior

When inviting multiple users via the "Invite to workspace" modal, causes the "Add another" and "Send Invites" buttons to move out of the visible area. There is no scroll bar, and the buttons become inaccessible unless the user zooms out significantly.

https://github.com/user-attachments/assets/13a3e0f6-86a2-45d2-9f8d-729b1d5798ed

### Expected Behavior

- Scroll vertically when it exceeds the viewport height,

- Provide an option to delete some of the input rows (email/name fields) to reduce the total height,

### Steps To Reproduce

1. Go to the "Invite to workspace" screen in Outline.
2. Add multiple invite rows by clicking "Add another" repeatedly (approx. 5–6 times).
3. Observe that the "Send Invites" and "Add another" buttons get pushed out of view.
4. Attempting to scroll does not bring them into view.
5. Only workaround is to zoom out in the browser, which is not user-friendly.

### Environment

```markdown
- Outline:Outline 0.83.0
- Browser: Chrome
```

### Anything else?

I’m currently working on a solution that provides an option to delete some of the input rows to help reduce the total height of the modal.

https://github.com/user-attachments/assets/34ed3a44-39c3-441b-a225-ea7767cba385

tommoor:
closes #8966

### DIFF
--- a/app/scenes/Invite.tsx
+++ b/app/scenes/Invite.tsx
@@ -209,7 +209,9 @@ function Invite({ onSubmit }: Props) {
                   placeholder={`name@${predictedDomain}`}
                   value={invite.email}
                   required={index === 0}
+                  autoComplete="off"
                   autoFocus
+                  data-1p-ignore
                   flex
                 />
                 <StyledInput
--- a/app/components/Modal.tsx
+++ b/app/components/Modal.tsx
@@ -114,6 +114,8 @@ const Modal: React.FC<Props> = ({
                 <Small {...props}>
                   <Centered
                     onClick={(ev) => ev.stopPropagation()}
+                    // maxHeight needed for proper overflow behavior in Safari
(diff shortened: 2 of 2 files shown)


app/scenes/Invite.tsx
```diff
@@ -209,7 +209,9 @@ function Invite({ onSubmit }: Props) {
                   placeholder={`name@${predictedDomain}`}
                   value={invite.email}
                   required={index === 0}
+                  autoComplete="off"
```
tommoor, line 212: While I'm here – fix password managers suggesting your own email address here

PR #8994, opened 2025-04-16
Fix: Prevent password managers from suggesting user's own email 

MehdiBouzouaya:
This PR prevents password managers from suggesting the current user's email address in the invite form.
Replaced autoComplete="off" with autoComplete="new-email" to better handle browser autofill behavior
Removed autoFocus.
I've removed unnecessary modifications to keep the implementation clean and focused. Thank you for the opportunity to make my first contribution.

### DIFF
--- a/app/scenes/Invite.tsx
+++ b/app/scenes/Invite.tsx
@@ -209,11 +209,11 @@ function Invite({ onSubmit }: Props) {
                   placeholder={`name@${predictedDomain}`}
                   value={invite.email}
                   required={index === 0}
+                  autoFocus
                   autoComplete="new-email"
                   data-1p-ignore
                   flex
                 />
-
                 <StyledInput
                   type="text"
                   name="name"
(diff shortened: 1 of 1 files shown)


tommoor:
It looks like you still removed `autoFocus`? Why – this is desired behavior so the next input is focused after it's added to the form

MehdiBouzouaya:
I will fix it right away

app/scenes/Invite.tsx
```diff
@@ -209,8 +209,8 @@ function Invite({ onSubmit }: Props) {
                   placeholder={`name@${predictedDomain}`}
                   value={invite.email}
                   required={index === 0}
-                  autoComplete="off"
                   autoFocus
+                  autoComplete="new-email"
```
tommoor, line 213: Where are you seeing that this is a valid value? It's not listed here for example:

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

tommoor: Got it, we don't want to include a non-standard value and undocumented in that case.

PR #9006, opened 2025-04-18
Transform issue and pull_request to unfurl shape in plugin

hmacr:
This PR moves the presenter logic tailored for GitHub into the corresponding plugin.

Base for adding other issue / pull_request unfurl providers like Linear, GitLab.


shared/i18n/locales/en_US/translation.json
```diff
@@ -520,7 +520,6 @@
   "Groups": "Groups",
   "Shared Links": "Shared Links",
   "Import": "Import",
-  "Self Hosted": "Self Hosted",
```
hmacr, line 523: Looks like missed translations from https://github.com/outline/outline/pull/9001

server/presenters/unfurl.ts
```diff
@@ -71,47 +71,61 @@ const presentDocument = (
...
-    avatarUrl: data.user.avatar_url,
-  },
-  state: {
-    name: data.merged ? "merged" : data.state,
-    color: GitHubUtils.getColorForStatus(data.merged ? "merged" : data.state),
-  },
-  createdAt: data.created_at,
-});
+): UnfurlResponse[UnfurlResourceType.PR] => {
+  if (data.transformed_unfurl) {
```
hmacr, line 75: This is needed for backward compatibility with existing cache entries - can be removed in a subsequent release.

plugins/github/server/github.ts
```diff
@@ -208,10 +212,59 @@ export class GitHub {
...
+
+      return GitHub.transformData(data, resource.type);
     } catch (err) {
       Logger.warn("Failed to fetch resource from GitHub", err);
       return { error: err.message || "Unknown error" };
     }
   };
+
+  private static transformData(
+    data: Record<string, any>,
```
tommoor, line 224: Can't we type this properly? Defintion should be inside Octokit

hmacr: Should be possible, let me see.

tommoor (approved):

PR #9009, opened 2025-04-18
fix: Parse emoji and url only as workspace icon

### LINKED ISSUE #9007: Cannot import from Notion without workspace icon

### Current Behavior

If your workspace in Notion does not have an uploaded icon, the Notion integration will fail during authentication. From the error logs below, I see this:

```
 ZodError: [
   {
     "validation": "url",
     "code": "invalid_string",
     "message": "Invalid url",
     "path": ["workspace_icon"]
   }
 ]
```

The workaround was to upload any icon for the workspace, then authentication was successful.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- Outline:
- Browser:
```

### Anything else?

_No response_

hmacr:
Closes #9007 

Notion workspace icon can either be an emoji, their internal icon, or an uploaded image.
This PR fixes a bug where parsing failed in case of anything other than a url-like string.


shared/i18n/locales/en_US/translation.json
```diff
@@ -520,7 +520,6 @@
   "Groups": "Groups",
   "Shared Links": "Shared Links",
   "Import": "Import",
-  "Self Hosted": "Self Hosted",
```
hmacr, line 523: Unrelated translations missed in https://github.com/outline/outline/pull/9001

plugins/notion/server/notion.ts
```diff
@@ -37,7 +41,15 @@ const AccessTokenResponseSchema = z.object({
   bot_id: z.string(),
   workspace_id: z.string(),
   workspace_name: z.string().nullish(),
-  workspace_icon: z.string().url().nullish(),
+  workspace_icon: z
+    .string()
+    .nullish()
+    .transform((val) => {
+      if (val && (isUrl(val) || emojiRegexp.test(val))) {
```
tommoor, line 48: With this regex it's best to just call it inline rather than assign in module scope

Issue #6407, opened 2024-01-19
Linear integration


tommoor:
After #6406 the second authenticated editor integration will be Linear.

Pranav2612000:
@tommoor I'll like to work on this issue.

tommoor:
Hi, we'd like to establish patterns in https://github.com/outline/outline/issues/6406 first (It's currently being worked on)

Pranav2612000:
Understood! I'll keep on following the PR so that I can pick up linear quickly once we decide to implement it.
Or let me know if there's some other way I can help with #6406

almereyda:
When it's shown how to integrate with Linear, it will be nice to adapt the example for the Linear FLOSS clone [plane](https://github.com/makeplane/plane/).

hotdogtherapper:
Hi @tommoor I work with a small team doing React+Node work. There are some things I would like to automatically build/generate from a documentation standpoint. Outline could be a very good repository for that.

A nice integration with Linear would be especially useful for our use case (similar to the one that Notion has where Issues can be displayed and updated dynamically based on a pasted hyperlink).

If it makes sense to both sides, would it be possible for our team to help build this integration?

tommoor:
Hi @hotdogtherapper – as much as I want this integration we first need to finish off the foundations for having these types of interactive chips/mentions in the document, as we want the solution to be pluggable with other issue trackers as well.

almereyda:
Released with [v0.84.0](https://github.com/outline/outline/releases/tag/v0.84.0).

PR #9037, opened 2025-04-23
feat: Linear integration

hmacr:
Towards #6407 

This PR adds Linear integration with support for converting linear issues to mentions. Also, takes care of revoking access when the integration is deleted.

I haven't added the issue sources caching and issue creation here since the PR is already a bit huge - these can be follow-ons later when necessary.


app/utils/mention.ts
```diff
@@ -27,6 +27,16 @@ export const isURLMentionable = ({
       );
     }
 
+    case IntegrationService.Linear: {
+      const settings =
+        integration.settings as IntegrationSettings<IntegrationType.Embed>;
+
+      return (
+        hostname === "linear.app" &&
+        settings.linear?.workspace.name === pathParts[1] // ensure installed workspace name matches with the provided url.
```
tommoor, line 36: This doesn't work unfortunately, we'll need to store the workspace `key` and compare that against the first url part

plugins/linear/client/Icon.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from "react";
+
+type Props = {
+  /** The size of the icon, 24px is default to match standard icons */
+  size?: number;
+  /** The color of the icon, defaults to the current text color */
+  fill?: string;
+};
+
+export default function Icon({ size = 24, fill = "currentColor" }: Props) {
```
tommoor, line 10: The logo is a bit large in the space, I can fix as we get closer to merging though. Needs to be the same visual size as GitHub

plugins/linear/server/linear.ts
```diff
@@ -0,0 +1,160 @@
...
+   * @returns {object} Containing resource identifiers - `workspace`, `type`, `id` and `name`.
+   */
+  private static parseUrl(url: string) {
+    const { hostname, pathname } = new URL(url);
+    if (hostname !== "linear.app") {
+      return;
+    }
+
+    const parts = pathname.split("/");
+    const workspace = parts[1];
```
tommoor, line 149: Similarly, here this is called `workspaceKey` and should compare to `workspace.key`

shared/components/IssueStatusIcon/LinearIssueStatusIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+  Started = "started",
+  Completed = "completed",
+  Canceled = "canceled",
+}
+
+/** Renders an icon for a specific workflow state type, animating the change of state type and completion percentage. */
+export function LinearIssueStatusIcon(props: BaseIconProps) {
+  const theme = useTheme();
+  const { state } = props;
+  const percentage =
```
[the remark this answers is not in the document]
tommoor: `position` should represent the order, but perhaps only within a specific state type – could that be it?

plugins/linear/server/api/linear.ts
```diff
@@ -0,0 +1,109 @@
...
+      ctx.redirect(LinearUtils.errorUrl("invalid_state"));
+      return;
+    }
+
+    const { teamId } = parsedState;
+
+    // this code block accounts for the root domain being unable to
+    // access authentication for subdomains. We must forward to the appropriate
+    // subdomain to complete the oauth flow
+    if (!user) {
```
tommoor, line 41: We really need to figure out a nice refactor for these blocks that appear everywhere

shared/components/IssueStatusIcon/LinearIssueStatusIcon.tsx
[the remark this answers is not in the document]
hmacr: For pre-defined names like Todo, In Progress, the order feels natural.. but for custom names, it's not the case. Here's what I'm seeing.

| State name | State type | Position | |
| Backlog | backlog | 0 | |
| Todo | unstarted | 1 | |
| In Progress | started | 2 | |
| Dev Complete | started | 556.88 | Custom state |
| In Review | started | 1079.72 | Custom state |
| Test | unstarted | 2019.73 | Custom state |

Feels like it's some form of fractional index, but not scoped in the right order 🤔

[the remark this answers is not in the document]
hmacr: Oops, nvm. I got the hang of it. It's fractional indexing _scoped_ within a state type as you said.
Let me see if we can approximate the completion percentage.

plugins/linear/server/api/linear.ts
hmacr: yep, a middleware probably.

hmacr:
Pushed a couple of commits to fix workspace key, Linear icon, and completion percent.

tommoor (approved):

hmacr:
Yay, that was quick 🚀 😆

PR #9061, opened 2025-04-25
fix: Store Linear workspace logo in storage

### LINKED ISSUE #9054: Linear workspace logo needs to be moved to Outline storage
On connect we need to upload the image to our own storage as it's an authenticated presigned URL that does not last long.

Similar to when connecting a workspace, `FileStorage.storeFromUrl` can be used.

hmacr:
Closes #9054

### DIFF
--- a/plugins/linear/server/api/linear.ts
+++ b/plugins/linear/server/api/linear.ts
@@ -6,9 +6,9 @@ import auth from "@server/middlewares/authentication";
 import { transaction } from "@server/middlewares/transaction";
 import validate from "@server/middlewares/validate";
 import { IntegrationAuthentication, Integration, Team } from "@server/models";
-import UploadLinearWorkspaceLogoTask from "@server/queues/tasks/UploadLinearWorkspaceLogoTask";
 import { APIContext } from "@server/types";
 import { Linear } from "../linear";
+import UploadLinearWorkspaceLogoTask from "../tasks/UploadLinearWorkspaceLogoTask";
 import * as T from "./schema";
 import { LinearUtils } from "plugins/linear/shared/LinearUtils";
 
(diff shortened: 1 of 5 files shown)


plugins/linear/server/api/linear.ts
```diff
@@ -72,6 +75,18 @@ router.get(
     const oauth = await Linear.oauthAccess(code!);
     const workspace = await Linear.getInstalledWorkspace(oauth.access_token);
 
+    let logoUrl: string | undefined;
+
+    if (workspace.logoUrl) {
+      const res = await FileStorage.storeFromUrl(
```
hmacr, line 81: Contemplating doing this here vs in a background task.
This is simpler, however background task reduces the redirect latency a lot.

hmacr:
Has any customer started using Linear already? We might need patching the existing integrations in that case.

tommoor:
I'll email them and have them reconnect if any have been recreated, won't be many

plugins/linear/server/api/linear.ts
```diff
@@ -103,6 +106,13 @@ router.get(
       { transaction }
     );
 
+    transaction.afterCommit(async () => {
```
tommoor, line 109: We should probably generalize this soon, perhaps some `.scheduleWithCtx` or equivalent

plugins/linear/server/index.ts
```diff
@@ -15,6 +16,10 @@ if (enabled) {
       type: Hook.API,
       value: router,
     },
+    {
```
tommoor, line 19: I moved the task into the plugin, lmk if you saw something I didnt

hmacr: Nah, this is right - I totally missed it :)

Issue #7829, opened 2024-10-24
Code blocks should be capped at a height and have an option to expand


tommoor:
### Discussed in https://github.com/outline/outline/discussions/7828

<div type='discussions-op-text'>

<sup>Originally posted by **caffeinewriter** October 24, 2024</sup>
Having code embedded in documentation is extremely nice without having to rely on an external provider like Github Gists or Gitlab Snippets.

However, for larger code blocks, it would be nice to have the option to set a maximum number of lines shown at a time, with the rest either collapsed, or scrollable, or just in general, to be able to set a max height for the block. </div>

PR #9067, opened 2025-04-26
adding expandable code block

Rahma-sbei:
Heyy , this is to adress the enhancement in #7829 ,I though it was really great so I worked on it .
I kept is pretty simple, I gave code blocks a default max height of 20 lines and added a button to expand and collapse it , all using css, and toggle button that updates a new expanded attribute on the node. Like so :

https://github.com/user-attachments/assets/316838b7-e0e4-4033-bc3d-f5752da7d4ea

Hope this helps

### DIFF
--- a/app/editor/menus/code.tsx
+++ b/app/editor/menus/code.tsx
@@ -41,6 +41,15 @@ export default function codeMenuItems(
 
   return [
     {
+      name: "expandCodeBlock",
+      icon: <ExpandedIcon />,
+      tooltip: node.attrs.expanded ? "Collapse" : "Expand",
+    },
+    {
+      name: "separator",
+      visible: !readOnly,
+    },
+    {
       name: "copyToClipboard",
       icon: <CopyIcon />,
       label: readOnly ? dictionary.copy : undefined,
--- a/app/editor/menus/code.tsx
+++ b/app/editor/menus/code.tsx
@@ -43,7 +43,7 @@ export default function codeMenuItems(
     {
       name: "expandCodeBlock",
       icon: <ExpandedIcon />,
-      tooltip: node.attrs.expanded ? "Collapse" : "Expand",
+      tooltip: "Expand",
     },
     {
       name: "separator",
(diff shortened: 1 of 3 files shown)


tommoor:
Unfortunately this won't work well as anything in the `attrs` is synced between editors including other users. Whether the code block is collapsed or not should only affect the current browser window

Rahma-sbei:
Oh I understand, thank you for clearing that up! I'll see what I can do

Rahma-sbei:
Heyy, so I changed the logic a little bit here, I added a plugin to manage the decorations for expanding and collapsing , and made the button toggle a local flag that way it is sure to be user specific and won't sync across users. Hope this works.

Rahma-sbei:
The condition to change the code-block class to expanded includes the block being selected AND expanded being true , so even when multiple code blocks exist only the selected one would expand. Plus, the suggested solution of storing the flag in the node's attribute is exactly what we've een trying to avoid :) . Frankly I think this solution works well enough unless I am missing something else.

Rahma-sbei:
This is a new version, I tried to avoid all the previously discussed issues:
- It does not use the node's attrs do it does not sync across users.
- There is no one global flag for all nodes so there wouln't be any conflict.

Instead of the previous flag, I made a map that appends a seperate flag for every code-fence node, the flag still toggles with the expandCodeBlock command , and the plugin checks for all true expanded flags and assigns the expanded class to them.
I tried it and made sure it will be consistent for all cases, we can:
- have all blocks collapsed.
- have two or more expanded at the same time, and some others collapsed.
- have all blocks expanded.

I hope I am not missing anything else, please do let me know if so 😊.

Uriziel01:
@tommoor Do you think this is mergeable as is right now? I have some documents with multiple 100+ lines code blocks, scrolling through them is not really usable in my case.

I'm also thinking about different approach, could we just have a simple generic `/spoiler` block? Not dedicated to any particular content type like code block.  This way everything would be easily collapsible, long tables, quotations, extensive lists, large embeds. WDYT?

Uriziel01:
NVM apparently I'm blind, just found https://github.com/outline/outline/pull/8317 and this is EXACTLY what I need in my case.

matbrgz:
Any update on that?

Issue #8976, opened 2025-04-15
Misalignment between comment @mention and email @mention


includesec-erik:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

There is inconsistent behavior between how previously deleted users are presented visually. 

In the webUI resolved comments they show up as "@Unknown" but in the comment snippets sent via notification emails they are presented as "@First Last" which is the original name of the now deleted user. They should probably both be "@Unknown" to keep a consistent UX.

This is not a big deal AT ALL, it's just inconsistent behavior that I thought might want to be considered or addressed to keep the UX consistent.

### Expected Behavior

1) The email would show "@Unknown" in the same way that the page shows the now deleted user. 

### Steps To Reproduce

1) Create user ABC
2) User ABC comments on a page, user XYZ replies to that comment "@ABC if you have anything to add here!"
3) Outline admin deletes user ABC, as they are no longer part of the organization using Outline
4) The comment by @ABC is viewed by others the comment shows "@unknown" as the @mention to the now deleted user.
5) The comment is marked as resolved
6) Users subscribed to the page see "@ABC" in the email snippet, instead of "@Unknown" as is shown on the page

![Image](https://github.com/user-attachments/assets/e21b4db2-b62d-480f-9b9f-0cd563274558)

### Environment

```markdown
- Outline: getoutline.com
- Browser: Chrome
```

### Anything else?

This is a super low priority bug, it might make a great "First bug" for somebody looking to contribute. I'd expect the primary maintainers of Outline would categorize this bug as a low priority

tommoor:
When you mention a user the current name is embedded as a fallback into the document itself, email notifications only use this data vs the current data in the database for a user. Just an explainer as to why this happens

PR #9069, opened 2025-04-26
Fix misalignment between email and comment mentions

Tess99854:
This PR updates the `CommentMentionedEmail` template so that any user mentions referring to deleted users are displayed as “@unknown” in notification emails, in order to match the behavior of the webUI comments.

Related Issue: https://github.com/outline/outline/issues/8976

### DIFF
--- a/server/emails/templates/CommentMentionedEmail.tsx
+++ b/server/emails/templates/CommentMentionedEmail.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
-import { NotificationEventType } from "@shared/types";
-import { Collection, Comment, Document } from "@server/models";
+import { MentionType, NotificationEventType } from "@shared/types";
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks, I think this is roughly along the right lines but we'd want to refactor this logic out of the email template and into `ProsemirrorHelper` and then make sure it's used in all places that prosemirror data is sent via email – not just in this single location :)

Tess99854:
@tommoor thank you for the reply, from what I understood I can move this logic to a helper function inside of `ProsemirrorHelper`. Would you prefer we invoke it in `htmlForData` so it applies everywhere, or call it directly in `CommentMentionedEmail`? What do you think

tommoor:
`htmlForData` would be a great spot yea

tommoor:
closed in https://github.com/outline/outline/pull/9234

PR #9070, opened 2025-04-27
Extract subdomain auth redirect

### LINKED ISSUE #9055: Refactor subdomain auth redirects
As mentioned here:
https://github.com/outline/outline/pull/9037/files#r2057438224

We have a repeated pattern of forwarding from app domain to authed subdomain that is needed by all OAuth flows and has resulted in a lot of copy/paste code. 

Time to DRY it up

tommoor:
closes #9055
towards #8884

### DIFF
--- a/plugins/github/server/api/github.ts
+++ b/plugins/github/server/api/github.ts
@@ -1,13 +1,12 @@
 import Router from "koa-router";
 import find from "lodash/find";
 import { IntegrationService, IntegrationType } from "@shared/types";
-import { parseDomain } from "@shared/utils/domains";
 import { createContext } from "@server/context";
-import Logger from "@server/logging/Logger";
+import apexAuthRedirect from "@server/middlewares/apexAuthRedirect";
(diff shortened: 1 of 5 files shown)


tommoor:
lmk if you can see a way to further simplify, but I think this is about it

hmacr (approved):
I'd have done the same

plugins/github/server/api/github.ts
```diff
@@ -43,33 +49,6 @@ router.get(
       return;
     }
 
-    // this code block accounts for the root domain being unable to
```
hmacr, line 46: I haven't tested it, but it feels like this change would have fixed a _probable_ redirection issue for `install_request` flow now that the apex redirection is done before any other sub-domain specific redirects.

tommoor: Pretty sure there's no issue there as the installRequest method doesn't require any specific auth it can happen from any subdomain

PR #9079, opened 2025-04-28
Implement type-safe task scheduler

### LINKED ISSUE #9073: Ensure `schedule` method is type-safe
This is honestly the fault of `schedule` not being properly type-safe

_Originally posted by @tommoor in https://github.com/outline/outline/issues/9072#issuecomment-2833457916_

hmacr:
Closes #9073 

Explored a concept for adding type-safety to scheduling tasks.

### DIFF
--- a/plugins/linear/server/api/linear.ts
+++ b/plugins/linear/server/api/linear.ts
@@ -5,6 +5,7 @@ import auth from "@server/middlewares/authentication";
 import { transaction } from "@server/middlewares/transaction";
 import validate from "@server/middlewares/validate";
 import { IntegrationAuthentication, Integration } from "@server/models";
+import { TaskScheduler } from "@server/queues/TaskScheduler";
 import { APIContext } from "@server/types";
 import { Linear } from "../linear";
 import UploadLinearWorkspaceLogoTask from "../tasks/UploadLinearWorkspaceLogoTask";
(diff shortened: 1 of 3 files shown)


plugins/linear/server/api/linear.ts
```diff
@@ -77,14 +78,14 @@ router.get(
       { transaction }
     );
 
-    if (workspace.logoUrl) {
-      transaction.afterCommit(async () => {
-        await UploadLinearWorkspaceLogoTask.schedule({
+    transaction.afterCommit(async () => {
+      if (workspace.logoUrl) {
+        await TaskScheduler.schedule(UploadLinearWorkspaceLogoTask, {
```
tommoor, line 83: It's quite decent, I didn't think of introducing an additional helper class to act as a middleman. I think the alternative of introducing `schedule` as an instance method would also be something we can gradually opt-in to and be slightly more terse:

```typescript
await new UploadLinearWorkspaceLogoTask().schedule({
```

hmacr: I wanted to try and avoid creating an instance, but it seems it's unavoidable since we need the job options any way :) I'll make the change as you suggested.

tommoor: > I wanted to try and avoid creating an instance

That's why I wrote it this way initially thinking "I'll just make sure to pass the type at each call point", had too much faith.

hmacr: We've all been there 😄

tommoor (approved):

PR #9096, opened 2025-04-29
chore: Add unique index to notification user<->comment

### LINKED ISSUE #9075: Add unique index to notification-user-comment
Also need to include a query in the migration to remove any existing duplicates.

tommoor:
closes #9075

### DIFF
--- a/server/migrations/20250429001902-add-notifications-unique-index.js
+++ b/server/migrations/20250429001902-add-notifications-unique-index.js
@@ -3,6 +3,26 @@
 /** @type {import('sequelize-cli').Migration} */
 module.exports = {
   async up (queryInterface, Sequelize) {
+    if (
+      process.env.NODE_ENV !== "test" && process.env.DEPLOYMENT !== "hosted"
+    ) {
+      // Delete duplicate notifications, keeping only the most recent one for each user-comment pair
+      await queryInterface.sequelize.query(`
+        WITH duplicates AS (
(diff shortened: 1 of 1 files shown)


server/migrations/20250429001902-add-notifications-unique-index.js
```diff
@@ -0,0 +1,34 @@
+"use strict";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    if (
+      process.env.NODE_ENV !== "test" && process.env.DEPLOYMENT !== "hosted"
+    ) {
+      // Delete duplicate notifications, keeping only the most recent one for each user-comment pair
+      await queryInterface.sequelize.query(`
```
tommoor, line 10: I tested this query against production and it runs in 100ms, not an issue.

hmacr:
Introducing the unique index causes `SequelizeUniqueConstraintError` when user mentions are involved in a comment. The flows I found issues are,
1. Mention a user with access to the doc in a comment -> resolve comment -> unresolve comment -> resolve again.
2. Mention a user with access to the doc in a comment -> edit comment removing the mention -> edit again to include the mention. (This flow feels rare, just noting that it is an issue).

I think the fix would be to catch and suppress `SequelizeUniqueConstraintError` in `Notification.create()` calls.

PR #9099, opened 2025-04-29
fix: Notifications received for draft with access but no subscription

### LINKED ISSUE #8807: Comment notifications are sent to all users with read access

tommoor:
Finally solves this issue, it required the following to be true:

- There must be a draft without a collection
- Someone should comment on the draft
- Other users must have access to the draft through a user or group membership
- Other users must have document subscriptions to at least one other document
- If all the above is true they will be notified, when they should not have been

Prior to the [fix here](https://github.com/outline/outline/pull/8997/files#diff-0a9c587545d94867afe9683cdeb90d01295f6dd2f07b7a0b5f4470c7a93c7cb3R209-R212) they would be sent a notification for _every other document subscription_ in the db. In the worst case scenario with a large Outline instance and a large group given access to the draft document this could result in thousands of spurious notifications.

closes #8807

### DIFF
--- a/server/models/helpers/NotificationHelper.test.ts
+++ b/server/models/helpers/NotificationHelper.test.ts
@@ -1,7 +1,9 @@
-import { NotificationEventType } from "@shared/types";
+import { DocumentPermission, NotificationEventType } from "@shared/types";
+import { UserMembership } from "@server/models";
 import {
   buildComment,
   buildDocument,
+  buildDraftDocument,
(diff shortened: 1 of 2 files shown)


hmacr (approved):
Thanks 🙏 

I had a hunch workspace drafts could be the cause, but couldn't verify it.

PR #9104, opened 2025-04-29
Configure commenting feature per collection

MahmoudMuhammedAli:
## Description
This PR adds per-collection commenting configuration, so users can enable or disable comments on individual collections while still defaulting to the global team setting when no override is specified. Collections now store a nullable `commenting` flag (true = enabled, false = disabled, null = use workspace default)

## Changes: 
- Database: Migration add `commenting` field (nullable boolean) column to `collections` table and create its corresponding migration file.

- Server model: `Collection` updated with `commenting` field.

- API:

    - `collections.create` and collections.update now accept and persist commenting.

    - Zod schemas extended to allow commenting on create/update.

    - `presentCollection` includes commenting in JSON responses.

- Frontend form: CollectionForm adds a tri-state Commenting switch and seeds defaultValues.commenting.

- Store helper: Added `CollectionsStore.canComment(id)` to centralize override vs. default logic.

- UI & actions:

   - `openDocumentComments` and `DocumentMeta` use `canComment(...)` instead of inline checks.

- Types: Updated client types to include `commenting` on `Collection`.

---


app/stores/CollectionsStore.ts
```diff
@@ -21,6 +22,26 @@ export default class CollectionsStore extends Store<Collection> {
     super(rootStore, Collection);
   }
 
+  /**
+   * Return whether comments should be enabled, using the
+   * per-collection override if present, otherwise falling back
+   * to the team's workspace default.
+   */
+  canComment(collectionId?: string): boolean {
```
tommoor, line 30: Mostly the PR is straightforward, and I'm happy to include this functionality.

This method doesn't seem in quite the right place though– I think it would be better on the collection model `collection.canCreateComment` would work. Where is no `collectionId` available we should use the team preference as today.

tommoor (changes requested):
As mentioned this is pretty straightforward and a nice option to have, thanks.

We also need to make sure that if the commenting is disabled for a collection then the shortcut and toolbar icon no longer appear to create a comment. This can be achieved by making sure `onCreateCommentMark` is not passed inside `Document/components/Editor`

tommoor:
lmk if you're planning on addressing the feedback here or if someone else should do it, thanks

PR #9121, opened 2025-05-02
Add windows support to Outline

loyaltypollution:
fix: Swap `rm -rf` / `cp -R` for `fs‑extra` helpers
fix: Introduce `copy-i18n.js` for locale copying
fix: Use `cross-env` in Makefile to set NODE_ENV


server/scripts/copy-i18n.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-var-requires, no-console */
+const path = require("path");
+const fs = require("fs-extra");
+
+async function copyI18n() {
+  try {
+    const buildDir = path.join(__dirname, "../../build/shared/i18n");
```
tommoor, line 7: Split these into individual arguments?

server/utils/PluginManager.ts
```diff
@@ -118,11 +118,22 @@ export class PluginManager {
...
+        rootDir.replace(/\\/g, "/"),
+        "plugins/*/server/!(*.test|schema).[jt]s"
+      )
+      .replace(/\\/g, "/");
+
+    const files = glob.sync(globPattern, { windowsPathsNoEscape: true });
+
+    files.forEach((filePath: string) => {
+      // Use path.join for the actual require to ensure proper OS-specific path
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
```
tommoor, line 133: Why is this needed now? 🤔

tommoor (changes requested):
Not against this thanks 👍, but:

- You need to sign the SLA
- Remove the unrelated changes in translation.json
- A little tidying

PR #9141, opened 2025-05-06
perf: Remove `documentStructure` from default query select

tommoor:
While `documentStructure` was removed from collection responses, it was never removed from the SELECT statements for collection data. This PR updates that so it is only queried where necessary to return or update the structure

### DIFF
--- a/package.json
+++ b/package.json
@@ -228,6 +228,7 @@
     "sequelize": "^6.37.3",
     "sequelize-cli": "^6.6.2",
     "sequelize-encrypted": "^1.0.0",
+    "sequelize-strict-attributes": "^1.0.2",
     "sequelize-typescript": "^2.1.6",
     "slug": "^5.3.0",
     "slugify": "^1.6.6",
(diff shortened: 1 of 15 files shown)


tommoor:
This is worth it from a performance point of view but I'm still trying to find a way to guard against accidentally loading the collection without the `documentStructure` column and then saving it with a null value.

hmacr:
I think `BeforeSave`  hook could be useful here - check `model.changed("documentStructure")` is true and the column value is set to `null`. In that case, we reset the changed property for `documentStructure` column (we don't have any use-case where it could be `null` other than non-loaded). I think this ensures sequelize will skip the column in the update query.

tommoor:
Maybe, though this

https://github.com/outline/outline/blob/a5f9e61a15ff6287e22700622a3fc7b28d7e9f6a/server/models/Collection.ts#L784-L786

is the case that concerns me the most

hmacr:
Fair, I can think of a _hacky_ way that sets a local `isDocStructureLoaded` field in BeforeFind hook based on query attributes, but even that will need careful handling of special cases I think.

PR #9153, opened 2025-05-08
fix: CMD+F with in-app find interface open should open native find

tommoor:
CMD+F twice should open native browser find

Regressed in https://github.com/outline/outline/pull/8560


hmacr:
Ok 👍 Thought the in-app find/replace dialog should be enough, hence trapped the key press in #8560

PR #9169, opened 2025-05-11
Add ADFS Authentication

xadammr:
This pull request adds support for ADFS' slightly weird implementation of OIDC. ADFS' userinfo endpoint only provides the `sub` claim, which causes the generic OIDC authentication plugin to fail which expects more than just that. 

This pull request creates a new ADFS authentication plugin which pulls claim info (email, preferred_username, sub) directly from the decoded id_token. 

Fixes #5756


tommoor (changes requested):
Thanks, we will not accept additional authentication plugins for oddities of OIDC implementations though – it gets out of hand fast.

In this case I think we can alter the existing OIDC implementation to parse the token and consider the email available there?

xadammr:
Thanks @tommoor, that makes sense -- I'll close this one off and have a crack at enhancing the OIDC provider to look at the ID token if the profile doesn't have the claims required.

PR #9172, opened 2025-05-11
fix: fall back to id_token when OIDC userinfo endpoint is sparse

xadammr:
Supersedes #9169.

ADFS' userinfo endpoint only provides the sub claim, which causes the generic OIDC authentication plugin to fail, as it expects email, preferred_username etc.

This pull request enhances the OIDC authentication plugin by falling back to the OIDC id_token which could contain these claims. 

Regression tested against ADFS and Keycloak -- authentication works with both. 

(Typescript is not my bread and butter -- let me know if I need to rethink any of this!)

### DIFF
--- a/plugins/oidc/server/auth/oidc.ts
+++ b/plugins/oidc/server/auth/oidc.ts
@@ -81,12 +81,27 @@ if (
 
           // Some providers, namely ADFS, don't provide anything more than the `sub` claim in the userinfo endpoint
           // So, we'll decode the params.id_token and see if that contains what we need.
-          const id_token = JWT.decode(params.id_token);
-          const token = id_token as {
-            email?: string;
-            preferred_username?: string;
-            sub?: string;
-          };
+          const token = (() => {
+            try {
+              const decoded = JWT.decode(params.id_token);
+
+              if (!decoded || typeof decoded !== "object") {
+                throw AuthenticationError(
+                  `Decoded token is not a valid object.`
+                );
+              }
+
+              return decoded as {
+                email?: string;
+                preferred_username?: string;
+                sub?: string;
+              };
+            } catch (err) {
+              throw AuthenticationError(
+                `Unable to decode id_token from identity provider.`
+              );
+            }
+          })();
 
           const email = profile.email ?? token.email ?? null;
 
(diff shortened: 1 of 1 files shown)


plugins/oidc/server/auth/oidc.ts
```diff
@@ -78,14 +79,26 @@ if (
             accessToken
           );
 
-          if (!profile.email) {
+          // Some providers, namely ADFS, don't provide anything more than the `sub` claim in the userinfo endpoint
+          // So, we'll decode the params.id_token and see if that contains what we need.
+          const id_token = JWT.decode(params.id_token);
```
tommoor, line 84: I think there's even a world where we try the id_token first and if it has enough information then there is no need to hit up the userinfo endpoint at all.

One change I would make here is to wrap `JWT.decode` in error handling

xadammr: I've added some error handling to the JWT.decode in commit 454f000.

plugins/oidc/server/auth/oidc.ts
```diff
@@ -118,13 +146,16 @@ if (
 
           // Claim name can be overriden using an env variable.
           // Default is 'preferred_username' as per OIDC spec.
-          const username = get(profile, env.OIDC_USERNAME_CLAIM);
+          // This will default to the profile.preferred_username, but will fall back to preferred_username from the id_token
+          const username = get(profile, env.OIDC_USERNAME_CLAIM)
+            ? get(profile, env.OIDC_USERNAME_CLAIM)
+            : get(token, env.OIDC_USERNAME_CLAIM);
```
tommoor, line 152:
```suggestion
          const username = get(profile, env.OIDC_USERNAME_CLAIM) ?? get(token, env.OIDC_USERNAME_CLAIM);
```

plugins/oidc/server/auth/oidc.ts
```diff
@@ -78,14 +79,41 @@ if (
...
+                );
+              }
+
+              return decoded as {
+                email?: string;
+                preferred_username?: string;
+                sub?: string;
+              };
+            } catch (err) {
+              throw AuthenticationError(
```
tommoor, line 100: I think we should log instead of throw, I'm concerned that some providers may provide tokens that are not parsable and this would break login for existing deployments

xadammr: In commit d43de00 now logging rather than throwing

tommoor (approved):

PR #9173, opened 2025-05-11
Add custom script injection support

mkay1375:
I've implemented custom JS script injection in Outline.
By adding this feature we were able to add [Chainlit's copilot](https://docs.chainlit.io/deploy/copilot) to Outline:

image_2025-05-10_130052165(1)

Given [this sample project](https://github.com/mkay1375/chainlit-and-outline) is running, by adding these envs the copilot will show up:
```bash
CUSTOM_SCRIPT_BASE_URL=http://localhost:8000
CUSTOM_SCRIPT_PATHS=/copilot/index.js,/public/copilot/outline.js
```

### DIFF
--- a/server/env.ts
+++ b/server/env.ts
@@ -649,6 +649,47 @@ export class Environment {
   @Public
   public APP_NAME = "Outline";
 
+  @IsOptional()
+  @IsUrl({ require_tld: false })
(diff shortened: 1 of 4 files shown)


tommoor:
Hi, thanks for the PR however we do not want to support arbitrary script injection here –

"Ask in doc" is coming soon, and we already have the ability to search the knowledgebase with AI and RAG

PR #9191, opened 2025-05-12
Fix: Remove invalid prop from Popover component

nadia-ben-njima:
This PR addresses an issue where an invalid or unnecessary prop was passed to the Popover component, causing warnings and unexpected behavior. The fix ensures proper tooltip positioning and eliminates related React warnings.


tommoor (changes requested):
Please reduce this to a minimum set of changes required to fix the issue. In this case, including `place` before `...rest` is probably all that's required?

nadia-ben-njima:
Thanks for the feedback! I’ve simplified the changes to focus only on what’s needed:

+place is now destructured early (as you suggested) to keep it out of the DOM.
+Added type definitions for Reakit’s internal props—this fixes the TypeScript error while keeping all the existing functionality exactly the same.
No behavior changes, just cleaner type safety and no more warnings

Issue #9195, opened 2025-05-13
Flash of old value when editing document title in sidebar


tommoor:
Only happens for documents, not collections.

1. Edit title in sidebar
2. Hit enter
3. Notice flash of previous value

MehdiBouzouaya:
I adressed a PR #9197 to try to fix this issue

PR #9197, opened 2025-05-13
Fix flash of previous value after editing title in sidebar for documents

MehdiBouzouaya:
In this PR, I addressed an issue where editing the title in the sidebar and hitting "Enter" would cause a flash of the previous value, specifically for documents. issue #9195

Moved setIsEditing(false) after successful onSubmit to prevent premature UI updates.

Now properly exits edit mode on both success and error cases.

Eliminates the flash of old values when submitting with Enter.

https://github.com/user-attachments/assets/28058917-1210-46b7-9fc0-33d92a64bcbb

### DIFF
--- a/app/components/EditableTitle.tsx
+++ b/app/components/EditableTitle.tsx
@@ -69,7 +69,6 @@ function EditableTitle(
 
       if (trimmedValue === originalValue || trimmedValue.length === 0) {
         setValue(originalValue);
-        setIsEditing(false);
         onCancel?.();
         return;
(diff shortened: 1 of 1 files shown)


tommoor:
Ah, makes sense – lets just move it into a `finally` block?

MehdiBouzouaya:
> Ah, makes sense – lets just move it into a `finally` block?

Yes , good point.

tommoor (approved):

PR #9224, opened 2025-05-15
Refactor `Placeholder` plugin to make it configurable

### LINKED ISSUE #9223: Refactor editor's `Placeholder` plugin to make it configurable
Paragraph placeholders are scattered in different places, like in `app/editor/extensions/BlockMenu.tsx` and `shared/editor/extensions/Placeholder.ts`. It'd be nice to move(and keep) them all in `Placeholder` plugin in a configurable manner.

apoorv-mishra:
Closes #9223 

[Ref](https://github.com/outline/outline/pull/8317#pullrequestreview-2826775926)

> An empty content area should always show at least one paragraph with a placeholder so that you can see where you should be typing. We might be able to tweak/reuse the logic on the existing placeholder?


shared/editor/plugins/PlaceholderPlugin.ts
```diff
@@ -0,0 +1,77 @@
+import filter from "lodash/filter";
+import find from "lodash/find";
+import map from "lodash/map";
+import { Node } from "prosemirror-model";
+import { EditorState, Plugin } from "prosemirror-state";
+import { Decoration, DecorationSet } from "prosemirror-view";
+
+type Config = Array<{
+  /** Condition to meet for the placeholder to be applied to a node */
+  cond: (
```
tommoor, line 10:
```suggestion
  condition: (
```

Lets change this to the full word, we're not short on characters

app/editor/extensions/BlockMenu.tsx
```diff
@@ -69,33 +70,47 @@ export default class BlockMenuExtension extends Suggestion {
...
       }),
+      new PlaceholderPlugin([
+        {
+          cond: (
+            node: Node,
+            pos: number,
+            _parent: Node | null,
+            state: EditorState
+          ) => {
+            const $start = state.doc.resolve(pos + 1);
```
tommoor, line 87: Every single plugin condition is doing this calculation – can we move it outside and pass as an argument for perf?

tommoor (commented):
Small nits, overall I definitely prefer over the previous approach

tommoor (approved):

Issue #9078, opened 2025-04-28
Drag & Drop of collection only work upwards


Suplanus:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Drag indicator shows off but on drop the collection wont move if I drag from top to down.

### Expected Behavior

Collection should be moved to the indicated position.

### Steps To Reproduce

Drag first collection to bottom. Video:

https://github.com/user-attachments/assets/b84d4adb-f21a-43a6-addd-bc050c8ece56

### Environment

```markdown
- Outline: v0.83.0 self hosted
- Browser: Brave 1.76.74 Chromium: 134.0.6998.89 (arm64)
```

skander2904:
It seems that the behavior you're encountering is working as expected on my end. When dragging the collection down, it's important to ensure that a visible indicator (like a line) appears to show where the collection will be placed. This helps in positioning the collection correctly. If the line isn't appearing or the collection doesn't move, it might be related to how the position is managed during the drag-and-drop action.

To resolve the issue, I suggest ensuring that the indicator line is visible when dragging and that the collection position is adjusted based on this indicator. Without the indicator, the position change won't be recognized.

https://github.com/user-attachments/assets/b661b9bd-374e-4499-8925-9501b3e05c95

Suplanus:
I can reproduce it to 100% when adding a new collection and drag it to the end. Indicator (line) is visible at drop:

https://github.com/user-attachments/assets/aee85c3e-2e39-4b77-bffd-bbded538663f

tommoor:
Please check the browser console for errors and network errors related to this

Suplanus:
@tommoor: Oh, yes there it is:
```
authenticated.7dHpKjKc.js:2 Uncaught Error: , >= +z
    at fractionalIndex (authenticated.7dHpKjKc.js:2:68088)
    at Object.drop (authenticated.7dHpKjKc.js:2:111671)
    at DropTargetImpl.drop (authenticated.7dHpKjKc.js:2:39031)
    at determineDropResult (authenticated.7dHpKjKc.js:2:8959)
    at authenticated.7dHpKjKc.js:2:9157
    at Array.forEach (<anonymous>)
    at DragDropManagerImpl.drop (authenticated.7dHpKjKc.js:2:8863)
    at Object.drop (authenticated.7dHpKjKc.js:2:11178)
[...]

PR #9225, opened 2025-05-16
fix: Consider active collections only for index computation

hmacr:
Related #9078

### DIFF
--- a/app/components/Sidebar/components/Collections.tsx
+++ b/app/components/Sidebar/components/Collections.tsx
@@ -22,7 +22,7 @@ import SidebarContext from "./SidebarContext";
 function Collections() {
   const { documents, collections } = useStores();
   const { t } = useTranslation();
-  const orderedCollections = collections.orderedData;
+  const orderedCollections = collections.allActive;
 
   const params = React.useMemo(
     () => ({


hmacr:
We actually don't need to.

https://github.com/outline/outline/blob/main/server/models/Collection.ts#L400 - This ensures index collisions are resolved even though FE sends a used index of an archived collection.

Main problem this PR resolves is when a collection is moved to the bottom - sometimes `belowCollection` field was incorrectly determined in this case.

hmacr:
Repro steps when `belowCollection` field becomes an issue.
1. For a new user, create 3 collections.
2. Edit their indexes to be `+z`, `,` & `P`.
3. Archive the collection with index `+z`.
4. Move fails when moving the top most collection to the bottom.

tommoor:
> This ensures index collisions are resolved when dnd-ing a collection even though FE sends a used index of an archived collection.

But that only hits when the `index` changes, the case I'm wondering about is if you have a collection with index `P` that's active, and one that's archived and then you restore the archived collection – now you have two active collections with index `P`, TLDR we should run that logic when `archivedAt` is cleared too? 🤔

hmacr:
> TLDR we should run that logic when archivedAt is cleared too?

We can, yea - for resiliency, will create a PR.
But I'm not completely confident that 2 collections (active-active, active-archive, archive-archive) will have the same index after #8810.

PR #9229, opened 2025-05-16
Resolve index collision when restoring collection

hmacr:
Related #9078
Related #9225


server/routes/api/collections/collections.ts
```diff
@@ -920,6 +921,13 @@ router.post(
       }
     );
 
+    collection.index = await removeIndexCollision(
```
tommoor, line 924: Any particular reason I might have missed we can't just tweak `static async checkIndex` to account for `archivedAt` changing?

hmacr: nope, you're right 🙂

tommoor (approved):

PR #9231, opened 2025-05-17
fix: Remove # from stored filesystem keys

tommoor:
I can't reproduce the original issue, but it's straight forward and reasonable to sanitize this character


Its4Nik:
> I can't reproduce the original issue, but it's straight forward and reasonable to sanitize this character

thanks a lot :D

PR #9283, opened 2025-05-23
perf: Utilize plugin state in placeholder plugin

tommoor:
Adds some perf improvements to placeholders plugin - because it must be recalculated on selection change this only really helps with remote changes

### DIFF
--- a/shared/editor/plugins/PlaceholderPlugin.ts
+++ b/shared/editor/plugins/PlaceholderPlugin.ts
@@ -25,24 +25,29 @@ export class PlaceholderPlugin extends Plugin {
 
   constructor(config: Config) {
     super({
+      state: {
+        init: (_, state: EditorState) => ({
+          decorations: this.createDecorations(state, config),
+        }),
+        apply: (tr, pluginState, oldState, newState) => {
+          // Only recompute if doc or selection changed
+          if (tr.docChanged || tr.selectionSet) {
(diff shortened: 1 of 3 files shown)


shared/editor/plugins/PlaceholderPlugin.ts
```diff
@@ -25,24 +28,29 @@ export class PlaceholderPlugin extends Plugin {
...
-                "data-empty-text": placeholder.text,
-              })
-          );
-          return DecorationSet.create(state.doc, decorations);
+          const pluginState = this.getState(state);
+          return pluginState ? pluginState.decorations : null;
         },
       },
     });
     this.config = config;
```
apoorv-mishra, line 50: Can get rid of this I suppose?

apoorv-mishra, line 50:
```suggestion
```

apoorv-mishra: nit: `private config: Config;` above the constructor can be removed. GitHub wouldn't let me suggest it the way I suggested previous change 😬

apoorv-mishra (approved):

apoorv-mishra:
Screenshot 2025-05-23 at 7 37 38 PM

Oops! Looks like this isn't the placeholder we expect when the doc is empty?

PR #9397, opened 2025-06-05
fix: invalidate stale policy only

### LINKED ISSUE #9390: Draft automatically deleted

### Current Behavior

When attempting to create a new draft, it automatically gets deleted and displays this: 

Image

I had an Untitled blank documented that I deleted last week that was in the Trash and now all new drafts automatically go into the Trash and I cannot edit them. I tried Permanently deleting everything from the Trash but the issue still persists. I tried both Chrome and Safari on a Mac.

### Expected Behavior

New draft document that I could edit.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Outline:
- Browser: Chrome and Safari on Mac
```

### Anything else?

_No response_

hmacr:
Closes #9390

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -217,6 +217,12 @@ class WebsocketProvider extends Component<Props> {
               continue;
             }
 
+            if (event.invalidatedPolicies) {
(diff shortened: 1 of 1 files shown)


tommoor:
Makes sense, needs the same block under collectionIds too

PR #9414, opened 2025-06-09
Listen to GitHub webhooks to update `issueSources` cache

hmacr:
This PR adds the ability to update `issueSources` cache on receiving webhooks from GitHub.

We only need to process a tiny subset of events & action-types as listed below.

| Event name | Action type | When is it received? | How is it processed?
| installation | new_permissions_accepted | New permission is accepted for an existing installation. | Accessible repos (`issueSources`) and scopes are persisted in DB. |
| installation_repositories | added | Repo is created / given access to an installation. | Repos in the event are added to `issueSources`. |
| installation_repositories | removed | Repo is deleted / revoked access from an installation. | Repos in the event are removed from `issueSources`. |
| repository | renamed | Repo is renamed. | Corresponding repo is renamed in `issueSources`. |

### Deployment actions
- Generate a secret key for signing webhooks.
- Configure the webhook url in GitHub app as `<host>/api/github.webhooks` (use the previously generated secret key in the config).
- Add a new env variable in the server - `GITHUB_WEBHOOK_SECRET`
- Post deployment, GitHub app's permission needs to be changed.
  - `write` permission for issues (needed for creating issue from Outline).
  - subscribe to `repository` event (needed for receiving repo renamed event)

---

Towards #8506 
Related #8951


server/middlewares/validateWebhook.ts
```diff
@@ -0,0 +1,40 @@
+import crypto from "crypto";
+import { Next } from "koa";
+import { APIContext } from "@server/types";
+
+export default function validateWebhook({
```
tommoor, line 6: Use the official one?

https://github.com/octokit/webhooks.js/?tab=readme-ov-file#webhooksverify

server/queues/tasks/CacheIssueSourcesTask.ts
```diff
@@ -16,6 +14,7 @@ export default class CacheIssueSourcesTask extends BaseTask<Props> {
       return;
     }
 
+    const plugins = PluginManager.getHooks(Hook.IssueProvider);
```
tommoor, line 17: Moved on purpose?

plugins/github/server/api/github.ts
```diff
@@ -93,4 +101,30 @@ router.get(
...
+      const signature = Array.isArray(signatureHeader)
+        ? signatureHeader[0]
+        : signatureHeader;
+      return signature?.split("=")[1];
+    },
+  }),
+  async (ctx: APIContext) => {
+    const { headers, body } = ctx.request;
+
+    await new IssueProviderWebhookTask().schedule({
```
tommoor: I'm not 100% sold that there is much value in an abstraction here, as the internals of every providers hooks would be completely different anyway. Might as well have `GitHubWebhookTask` inside of plugins/github 🤔

server/middlewares/validateWebhook.ts
hmacr: That was the other option, but I wrote the middleware so it can be used for Linear later 😉 Thoughts?

server/queues/tasks/CacheIssueSourcesTask.ts
hmacr: yea - In rare instances, I noticed this file is loaded before hooks are registered.

plugins/github/server/api/github.ts
hmacr: fair yea, idk what I was thinking here - abstraction seems unnecessary since the API will be specific to each service.

hmacr: I'd still like to have [handleWebhook](https://github.com/outline/outline/pull/9414/files#diff-5eac8bcf8a92ba4c6cd90cae8dd8f10e3d951cec651d69579b3e637668f90961R16) as part of `BaseIssueProvider` - this establishes hard guarantees for an ideal _issue provider_ (since we want to update the local cache). Thoughts on this?

tommoor: Sure thing

server/middlewares/validateWebhook.ts
tommoor: Definitely making some assumptions about these signatures that might not hold true, but lets see!

hmacr: ~Hmm, it follows the same guidelines that are in [Linear docs](https://linear.app/developers/webhooks#securing-webhooks)~ 🤔 

nvm, json parsing might be the difference.. will use the official one here.

hmacr: Retaining this code for now since `octokit-webhooks` is a [pure ESM module](https://github.com/octokit/webhooks.js?tab=readme-ov-file#typescript) which needs a decent bit of change in tsconfig to import in the server. I tried a couple of workarounds that didn't work (or) were convoluted.

tommoor: Ok

server/middlewares/validateWebhook.ts
```diff
@@ -0,0 +1,40 @@
...
+      return;
+    }
+
+    const computedSignature = crypto
+      .createHmac("sha256", secretKey)
+      .update(JSON.stringify(body))
+      .digest("hex");
+
+    if (
+      !crypto.timingSafeEqual(
```
tommoor: We do have the `safeEqual` wrapper method for this too

tommoor (approved):

plugins/github/server/GitHubIssueProvider.ts
```diff
@@ -37,4 +50,215 @@ export class GitHubIssueProvider extends BaseIssueProvider {
...
+    const action = payload.action as string;
+
+    if (!eventName || !action) {
+      Logger.warn(
+        `Received GitHub webhook without event name or action; hookId: ${hookId}, eventName: ${eventName}, action: ${action}`
+      );
+      return;
+    }
+
+    const actions = this.interestedActionsForEvent[eventName];
```
tommoor: Could we just switch on the various action below instead? This feels like misdirection. e.g.

```
switch (action) {
  case "new_permissions_accepted":
    ...
}
```

hmacr: Using action name directly will be problematic since multiple events will have same action name.. for e.g. `renamed` action for `repository` & `member` event. That said, I've replaced `interestedActionsForEvent` mapper with simpler conditionals.

tommoor (approved):

PR #9421, opened 2025-06-10
fix: Persist document icon & color in import flow

### LINKED ISSUE #9420: Page icons and emojis don't get restored

### Current Behavior

When I export a collection in which some of the items have either an icon or an emoij set, I can see that data in the `icon` fields in the JSON file inside the ZIP (with a short text description for “homemade” icons, resp. the literal emoji as the value).

However, when I delete and re-import this collection from the exported ZIP file, both “homemade” icons and emoji disappear.

### Expected Behavior

Both `"icon": "buildingBlocks"` and `"icon":"🐿️"` data in the exported JSON should be honored at restore time.

### Steps To Reproduce

1. Install Outline version 0.84.0 somewhere
2. Create a a collection
3. Create a few pages with icons
4. Export in JSON format
5. Delete collection
6. Restore from ZIP produced at step 4

### Environment

```markdown
- Outline: 0.84.0
- Browser: (Doesn't matter, but) Firefox 139.0.1, Mac OS X Sequoia 15.5 (24F74)
```

### Anything else?

_No response_

hmacr:
Closes #9420

### DIFF
--- a/server/queues/tasks/ImportTask.ts
+++ b/server/queues/tasks/ImportTask.ts
@@ -462,6 +462,8 @@ export default abstract class ImportTask extends BaseTask<Props> {
               urlId: item.urlId,
               text,
               content: item.data,
+              icon: item.icon,
+              color: item.color,
               collectionId: item.collectionId,
               createdAt: item.createdAt,
               updatedAt: item.updatedAt ?? item.createdAt,


domq:
Thanks for the quick fix, this is awesome!

image

PR #9436, opened 2025-06-12
fix: Sporadic rate limiting errors from Notion

tommoor:
I can't find a logical error that would result in going past the 3req/min that's defined in the limiter. For reliability's sake I'd like to include this retry mechanism as well.


hmacr (approved):

hmacr:
> Perhaps if you triggered two imports simultaneously?

I'm not sure, they say rate-limit is scoped per access_token.. Interesting case 🤔 Time to add rate-limit retry based on response header I guess.

PR #9440, opened 2025-06-13
fix: Login via email does not properly redirect to desktop app

tommoor:


plugins/email/server/auth/schema.ts
```diff
@@ -14,12 +14,12 @@ export type EmailReq = z.infer<typeof EmailSchema>;
 export const EmailCallbackSchema = BaseSchema.extend({
   query: z.object({
     token: z.string().optional(),
-    client: z.nativeEnum(Client).default(Client.Web),
```
tommoor, line 17: This default meant that `Client.Web` always overrode `Client.Desktop`

PR #9470, opened 2025-06-19
fixed comments

TimeToCodeSomething:
issue #9300 

# Hide comment marks when commenting is disabled

## Summary
This change makes the comment plugin in the editor initialize conditionally based on the collection’s **Commenting** setting. When commenting is disabled (`TeamPreference.Commenting = false`), the `withComments` extension is not loaded, so no inline comment markers are ever rendered in the document.

## Changes
- Removed the global `extensions` declaration.
- In `DocumentEditor` (`app/scenes/Document/components/DocumentEditor.tsx`), after reading the `commentingEnabled` flag, compute:
  ```ts
  const extensions = withUIExtensions(
    commentingEnabled
      ? withComments(richExtensions)
      : richExtensions
  );
Only pass onCreateCommentMark and onDeleteCommentMark to <EditorComponent> when commentingEnabled && can.comment.

All other editor extensions and UI behavior remain unchanged.

How to Test
Commenting enabled (Team Preference → Commenting = true):

Inline markers should render as before.

Sidebar comments panel opens and functions normally.

Commenting disabled (Team Preference → Commenting = false):

No comment markers appear in the document (DOM contains no marker decorations).

Sidebar comments panel does not open.

Existing comment data in the database remains intact.

### DIFF
--- a/app/scenes/Document/components/Editor.tsx
+++ b/app/scenes/Document/components/Editor.tsx
@@ -34,8 +34,6 @@ import MultiplayerEditor from "./AsyncMultiplayerEditor";
 import DocumentMeta from "./DocumentMeta";
 import DocumentTitle from "./DocumentTitle";
 
-const extensions = withUIExtensions(withComments(richExtensions));
-
 type Props = Omit<EditorProps, "editorStyle"> & {
   onChangeTitle: (title: string) => void;
   onChangeIcon: (icon: string | null, color: string | null) => void;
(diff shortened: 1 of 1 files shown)


tommoor:
This is not the correct solution, it removes the ability for the editor to recognize comment nodes at all – which will cause them to get removed from the document entirely. Using `withComments` within render like this will also result in a fresh array each time and degradated performance

Issue #7878, opened 2024-11-02
Add notification for reactions


tommoor:
Related to #7137 

One detail here is that I think we should continue to update an existing notification if you get multiple reactions and the notification hasn't been read, for example it might start "Tom reacted 👍 to your comment", but could be updated to read "Tom and 5 others reacted to your comment" over time.

PR #9474, opened 2025-06-20
multiple reactions comments

TimeToCodeSomething:
issue #7878
### Summary

This PR adds support for updating an existing unread notification when multiple users react with the same emoji on a comment.  
Instead of creating a new notification each time, we now display messages like:

> `Tom and 5 others reacted 👍 to your comment`

### Changes

- **Notification model**
  - Added new columns:
    - `emoji: STRING`
    - `actorIds: UUID[]`
    - `message: TEXT`
  - Updated `server/models/Notification.ts` accordingly.

- **NotificationEventType**
  - Added new event type:
    ```ts
    ReactionToComment = "comments.reaction"
    ```
  - Registered it in `NotificationEventDefaults`.

- **comments.add_reaction handler**
  - After `Reaction.findOrCreate(...)`:
    - If an unread notification with same `commentId + emoji` exists:
      - Append actor to `actorIds`
      - Update `message`
    - Otherwise, create a new notification
  - Added necessary imports (`Notification`, `NotificationEventType`, `Op`)

---

### Problem

The project does **not compile** due to a type error:

TS1360: Type 'NotificationEventType.ReactionToComment' does not satisfy the expected type 'never'.

This likely means:
- the `NotificationEventType.ReactionToComment` enum is not recognized correctly in the `Notification` model
- or the shared enum types were not correctly linked at build time

### DIFF
--- a/server/routes/api/comments/comments.ts
+++ b/server/routes/api/comments/comments.ts
@@ -2,6 +2,7 @@ import Router from "koa-router";
 import difference from "lodash/difference";
 import { FindOptions, Op, WhereOptions } from "sequelize";
 import {
+  NotificationEventType,
   CommentStatusFilter,
   TeamPreference,
   MentionType,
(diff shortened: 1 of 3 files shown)


server/routes/api/comments/comments.ts
```diff
@@ -395,17 +399,54 @@ router.post(
...
-        commentId: id,
-      },
+      where: { emoji, userId: user.id, commentId: id },
       ...ctx.context,
     });
 
-    ctx.body = {
-      success: true,
-    };
+    if (comment.createdById !== user.id) {
```
tommoor, line 406: We don't create notifications inline in routes like this, please see the other notification types for examples of how to set this up on the job queue.

PR #9479, opened 2025-06-23
Allow replacing embed links

### LINKED ISSUE #8084: Allow replacing embed link without having to delete and reenter
A good idea, came in through support.

hmacr:
Closes #8084 

~Toolbar is now rendered for all embed types, which will have three actions - edit, copy, and open link.~
Uses the same approach as floating link toolbar.

### Screen capture

https://github.com/user-attachments/assets/f1810578-3037-4606-9b8d-c6d0b4f94e46


tommoor:
Unfortunately a _little_ bit of a wrong path here, this bottom bar is just kind of ugly so I don't want it as a chin for every single embed.

Ideally we can use the same approach as we have for issues/links where you can highlight the embed which will show the floating toolbar and have options the input to edit the url, remove, open link.

hmacr:
> you can highlight the embed which will show the floating toolbar

yep, that'll work.

I was thinking along the lines of how it works for images (mouse click shows a floating toolbar) which wouldn't work well for embeds that are interactive (YouTube, draw.io, etc), so I went with a default bar (doesn't look great as you said).

tommoor:
Agreed it's not as easy to get the highlight state (basically have to use keyboard) so I can see why you went this way, but we can tackle that separately.

Perhaps we can add a hover trigger for it later

tommoor (approved):

PR #9529, opened 2025-07-02
feat: Public sharing of collections

### LINKED ISSUE #1160: Public Share Link of Collections
Share one item isnt enough. I'm considering to implement the shareable collection link. How to start this implementation? Could some one help? Is possible to implement this on fast ring?

hmacr:
Closes #1160

General idea is to use `shares.info` endpoint to get the share data + tree + source model (could be either collection or document), which means the dependency on `documents.info` endpoint is removed w.r.t share model + tree. However it will continue to be used to get document data within share's context.
Moving forward, `shares.sitemap` will be used in place of `documents.sitemap`.

Keeping the existing code as-is for backward compatibility - they can be cleaned up post deployment.

### DIFF
--- a/server/commands/shareLoader.ts
+++ b/server/commands/shareLoader.ts
@@ -100,10 +100,14 @@ export async function loadPublicShare({
       rejectOnEmpty: true,
     });
 
-    const allIdsInSharedTree = share.includeChildDocuments
-      ? getAllIdsInSharedTree(sharedTree)
-      : [];
-    if (!allIdsInSharedTree.includes(document.id)) {
+    let isDocumentAccessible = share.documentId === document.id;
+
+    if (share.includeChildDocuments) {
+      const allIdsInSharedTree = getAllIdsInSharedTree(sharedTree);
+      isDocumentAccessible = allIdsInSharedTree.includes(document.id);
+    }
+
+    if (!isDocumentAccessible) {
       throw AuthorizationError();
     }
   } else {
--- a/server/commands/shareLoader.test.ts
+++ b/server/commands/shareLoader.test.ts
@@ -201,7 +201,7 @@ describe("shareLoader", () => {
       expect(result.parentShare?.id).toEqual(parentShare.id);
     });
 
-    it("should throw error when the requested document is not part of the share", async () => {
+    it("should throw error when the requested document is not part of the share (includeChildDocuments = true)", async () => {
       const user = await buildUser();
       const collection = await buildCollection({
         userId: user.id,
@@ -225,6 +225,61 @@ describe("shareLoader", () => {
 
       await expect(
         loadPublicShare({ id: share.id, documentId: anotherDocument.id })
+      ).rejects.toThrow();
+    });
+
+    it("should throw error when the requested document is not part of the share (includeChildDocuments = false)", async () => {
+      const user = await buildUser();
+      const collection = await buildCollection({
+        userId: user.id,
+        teamId: user.teamId,
+      });
+      const document = await buildDocument({
+        collectionId: collection.id,
+        userId: user.id,
+        teamId: user.teamId,
+      });
+      const anotherDocument = await buildDocument({
+        userId: user.id,
+        teamId: user.teamId,
+      });
+      const share = await buildShare({
+        includeChildDocuments: false,
+        userId: user.id,
+        teamId: user.teamId,
+        documentId: document.id,
+      });
+
+      await expect(
+        loadPublicShare({ id: share.id, documentId: anotherDocument.id })
+      ).rejects.toThrow();
+    });
+
+    it("should throw error when the child document is requested for a share with includeChildDocuments = false", async () => {
+      const user = await buildUser();
+      const collection = await buildCollection({
+        userId: user.id,
+        teamId: user.teamId,
+      });
+      const document = await buildDocument({
+        collectionId: collection.id,
+        userId: user.id,
+        teamId: user.teamId,
+      });
+      const childDocument = await buildDocument({
+        parentDocumentId: document.id,
+        userId: user.id,
+        teamId: user.teamId,
+      });
+      const share = await buildShare({
+        includeChildDocuments: false,
+        userId: user.id,
+        teamId: user.teamId,
+        documentId: document.id,
+      });
+
+      await expect(
+        loadPublicShare({ id: share.id, documentId: childDocument.id })
       ).rejects.toThrow();
     });
   });


app/models/Document.ts
```diff
@@ -453,6 +453,7 @@ export default class Document extends ArchivableModel implements Searchable {
   @action
   share = async () =>
     this.store.rootStore.shares.create({
+      type: "document",
```
tommoor, line 464:
```suggestion
```

Not needed?

app/models/Collection.ts
```diff
@@ -320,6 +320,13 @@ export default class Collection extends ParanoidModel {
     this.index = index;
   }
 
+  @action
+  share = async () =>
+    this.store.rootStore.shares.create({
+      type: "collection",
```
tommoor, line 326:
```suggestion
```

Not needed?

app/components/Sidebar/Shared.tsx
```diff
@@ -121,9 +131,7 @@ const ToggleWrapper = styled.div`
   right: 0;
   opacity: 0;
   transform: translateX(10px);
-  transition:
-    opacity 100ms ease-out,
-    transform 100ms ease-out;
+  transition: opacity 100ms ease-out, transform 100ms ease-out;
```
tommoor: This should change back when you pull `main` as it was due to prettier upgrade

server/routes/api/shares/shares.ts
```diff
@@ -1,96 +1,86 @@
...
+import { navigationNodeToSitemap } from "@server/utils/sitemap";
 import pagination from "../middlewares/pagination";
 import * as T from "./schema";
 
 const router = new Router();
 
 router.post(
   "shares.info",
-  auth(),
+  auth({ optional: true }),
```
tommoor, line 34: Is this needed? We're not accepting `shareId` here

server/presenters/share.ts
```diff
@@ -4,6 +4,9 @@ import { presentUser } from ".";
 export default function presentShare(share: Share, isAdmin = false) {
   const data = {
     id: share.id,
+    sourceTitle: share.collection?.name ?? share.document?.title,
+    sourceUrl: share.collection?.path ?? share.document?.path,
```
tommoor:
```suggestion
    sourcePath: share.collection?.path ?? share.document?.path,
```

Lets use "path" for new stuff, really trying to get away from the old usages of "url" where it's only a path

server/routes/api/shares/shares.ts
```diff
@@ -1,96 +1,86 @@
...
+          ? await presentCollection(ctx, collection, {
+              isPublic: isPublicCollection,
+              shareId: share.id,
+            })
+          : null,
+        document
+          ? await presentDocument(ctx, document, {
+              isPublic: isPublicDocument,
+              shareId: share.id,
+              includeUpdatedAt: share.showLastUpdated,
```
tommoor: Could we include for collection?

app/models/Document.ts
hmacr: Not used in the API, but this adds stricter type-checking in the store [here](https://github.com/outline/outline/pull/9529/files#diff-eb844bcc8069a598dc7096c047817f04a01295de2e1815626cc165b06e232819R53-R55)

server/routes/api/shares/shares.ts
hmacr: Yep, I had the server support initially, but removed with an incorrect assuming the UI meta component would need a separate effort.
Will add it back with UI comp.

server/routes/api/shares/shares.ts
hmacr: We need this since public shares will use this endpoint with `id` (equivalent to `shareId`) param, while the share popovers will pass `user` & `collectionId`/`documentId` params.

hmacr:
I noticed a couple of bugs here - ~will be fixing them~ fixed.
- [x] Images aren't loaded in collections overview.
- [x] Search does not work.

@tommoor - this is done. Search is scoped to documents for now, could be extended when we add collections search in general.

app/utils/ApiClient.ts
```diff
@@ -153,10 +153,12 @@ class ApiClient {
 
     // Handle 401, log out user
     if (response.status === 401) {
-      await stores.auth.logout({
-        savePath: true,
-        revokeToken: false,
-      });
+      if (!this.shareId) {
```
tommoor, line 156: Haven't seen any issues around this before, why did it change?

app/scenes/Shared/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+      </Helmet>
+      <TeamContext.Provider value={team}>
+        <ThemeProvider theme={theme}>
+          <DocumentContextProvider>
+            <Layout
+              title={pageTitle}
+              sidebar={
+                share.tree?.children.length ? (
+                  <Sidebar
+                    rootNode={share.tree}
```
tommoor: Probably just change this to pass `share` in and avoid the `isCollectionShare` weirdness

app/scenes/Shared/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+    <>
+      <Helmet>
+        <link
+          rel="canonical"
+          href={canonicalOrigin + location.pathname.replace(/\/$/, "")}
+        />
+        <link
+          rel="sitemap"
+          type="application/xml"
+          href={`${env.URL}/api/shares.sitemap?shareId=${shareId}`}
```
tommoor: Might be nice to pull this into `share.sitemapUrl`  getter

tommoor (commented):
Just realized I had these comments stuck in pending 🤦

app/utils/ApiClient.ts
hmacr: Shared scene now uses data from stores (previously it had its own internal state management). Without this change, stores get cleared as part of logout - this could happen for any `401` error (mentions in a public doc for e.g.)

app/scenes/Shared/index.tsx
hmacr: I'm a _little_ apprehensive about this.. this will mean `share` needs to be loaded before crawlers can act on it.. perhaps a static `Share.sitemapUrl(shareId)` will clean this up .

hmacr:
Just updated the PR with the recent feedback and resolved merge conflicts.

server/routes/app.ts
```diff
@@ -143,34 +143,38 @@ export const renderApp = async (
...
   const rootShareId = ctx.state?.rootShare?.id;
   const shareId = rootShareId ?? ctx.params.shareId;
+  const collectionSlug = ctx.params.collectionSlug;
   const documentSlug = ctx.params.documentSlug;
 
   // Find the share record if publicly published so that the document title
   // can be returned in the server-rendered HTML. This allows it to appear in
   // unfurls with more reliability
-  let share, document, team;
+  let share, collection, document, team;
```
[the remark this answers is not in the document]
tommoor: I'll modify the rules to avoid this one

hmacr:
@tommoor Finally got around to refactoring `shareLoader`. Code path has been split into two.
1. `loadPublicShare` - used when loading public links. It doesn't load memberships by default, so when an authenticated user is present, it'd mean an additional query to load memberships in the router layer (trade-off worth it for maintaining clean code).
2. `loadShareWithParent` - used in share popovers.

server/commands/shareLoader.ts
```diff
@@ -0,0 +1,231 @@
...
+    where.teamId = teamId;
+  } else {
+    where.id = id;
+  }
+
+  const share = await Share.findOne({
+    where,
+    include: [
+      {
+        model: Document.scope("withDrafts"),
```
tommoor, line 47: I don't think drafts should be included – what is shared should represent the tree in Outline?

server/commands/shareLoader.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  if (documentId && documentId !== share.documentId) {
+    document = await Document.findByPk(documentId, {
+      paranoid: false,
+      rejectOnEmpty: true,
+    });
+
+    const allIdsInSharedTree = share.includeChildDocuments
+      ? getAllIdsInSharedTree(sharedTree)
+      : [document.id];
```
tommoor: I believe this should be an empty array – if the share does not include child documents and `documentId !== shareId` then there would never be any access.

Tests are successful either way

app/scenes/Collection/components/Overview.tsx
```diff
@@ -94,7 +94,7 @@ function Overview({ collection, shareId }: Props) {
             maxLength={CollectionValidation.maxDescriptionLength}
             onCreateLink={onCreateLink}
             canUpdate={can.update}
-            readOnly={!can.update}
+            readOnly={!can.update || !!shareId}
```
hmacr, line 92: Thank you for finding this 🙏

server/commands/shareLoader.ts
hmacr: for logical correctness, yea - will push a commit.

server/routes/api/shares/shares.ts
```diff
@@ -311,12 +311,12 @@ router.post(
 
     if (published !== undefined) {
       share.published = published;
-      if (published && !!share.documentId) {
+      if (published) {
```
hmacr: Added this earlier, but for correctness, `includeChildDocuments` should be true for collection share as well.

On a related note, this fixes a bug on initial SSR load where title was not returned for a child doc link, for e.g. `s/{collection_share_id}/doc/{child_doc_id}`

server/commands/shareLoader.ts
hmacr: I'm afraid we need `[document.id]` to handle cases when the request `documentId` is a slug type. We can add a special conditional for this, but the current ones feel cleaner. wdyt?

tommoor:
Noted in DM, as a follow up we will:

- Use `loadPublicShare` instead of `documentLoader` when loading a `shareId` from documents endpoints. This was left out of this PR to reduce scope and problems.
- Refactor `loadShareWithParent` to just be a static model method, e.g. `Share.findWithParent`

tommoor (approved):

PR #9530, opened 2025-07-02
feat: add optional public URL for serving uploaded files from S3

EarMaster:
Our S3 endpoint is not publicly accessible. This is not a problem for attachments as they are requested by the VM outline is running on. But the avatars require the /public folder of the S3 endpoint to be accessible as the avatars are not handled the same way as attachments. This introduces a new env variable named `AWS_S3_PUBLIC_URL` to manually configure these URLs. If unset nothing changes for existing users. Also nothing changes for attachments as the new variable is only used for `^/public/` paths.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -96,6 +96,10 @@ AWS_REGION=xx-xxxx-x
 AWS_S3_ACCELERATE_URL=
 AWS_S3_UPLOAD_BUCKET_URL=http://s3:4569
 AWS_S3_UPLOAD_BUCKET_NAME=bucket_name_here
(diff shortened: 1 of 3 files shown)


tommoor:
> This is not a problem for attachments as they are requested by the VM outline is running on

I'm not sure I follow this part, they are still requested from the client – just with a signature added

EarMaster:
I noticed that while testing. I will continue to work on this.

EarMaster:
I am closing this. I don't have the time to work on it. Avatars aren't working for me, but that's fine.

PR #9547, opened 2025-07-04
Migrate share popovers to Radix

hmacr:
- Added a `Popover` primitive that has base styles with entry/exit animations.
- `Collaborators`, `IconPicker`, and `Notifications` were already ported to Radix. They now use the new primitive.
- Ported document and collection share popovers in this PR.

Towards #6456

### DIFF
--- a/app/scenes/Document/components/ShareButton.tsx
+++ b/app/scenes/Document/components/ShareButton.tsx
@@ -1,11 +1,15 @@
 import { observer } from "mobx-react";
 import { GlobeIcon } from "outline-icons";
+import { useCallback, useState } from "react";
 import { useTranslation } from "react-i18next";
-import { usePopoverState, PopoverDisclosure } from "reakit/Popover";
 import Document from "~/models/Document";
 import Button from "~/components/Button";
-import Popover from "~/components/Popover";
 import SharePopover from "~/components/Sharing/Document";
+import {
+  Popover,
+  PopoverTrigger,
+  PopoverContent,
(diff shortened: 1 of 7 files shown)


app/scenes/Document/components/ShareButton.tsx
```diff
@@ -16,47 +20,43 @@ type Props = {
...
-        scrollable={false}
+        side="bottom"
+        align="end"
       >
         <SharePopover
           document={document}
-          onRequestClose={popover.hide}
-          visible={popover.visible}
+          onRequestClose={closePopover}
+          visible={open}
```
tommoor, line 56: This doesn't seem to work quite the same as before, because `popover.visible` accounts for the animate out time.

tommoor (commented):
looks good apart from that one small thing on animate out

app/scenes/Document/components/ShareButton.tsx
hmacr: I'm not sure I get this. Previously, there wasn't any exit animation which could be why the timings are in sync 🤔 
Should we maintain the same in this case?

tommoor: We could remove the out anim for now, I don't mind if closing is extra snappy

hmacr: Got it, I removed the exit anim and used a faster curve for entry anim (feels extra snappy). Lmk if that needs to be reverted.

tommoor (approved):

PR #9556, opened 2025-07-07
feat: Add OTP sign-in for PWA

tommoor:
It is not currently possible to sign-in to a PWA with email as the callback redirect occurs in the default browser, this PR introduces an alternative flow that uses an OTP to allow the auth to continue in the original tab.


hmacr:
Works great!

PR #9578, opened 2025-07-08
feat: Radix dropdown menu

hmacr:
This PR adds Radix-backed dropdown menu. As part of this setup, new action types are introduced which will used for context menu and kbar in future. Sub-menu support is not added here.. it'll be the next item to tackle.

### Contents
- Added Radix dropdown primitive.
- Added `DropdownMenu` component that renders items in a dropdown on desktop, and drawer on mobile.
- Migrated three components - NewTemplatesMenu, ApiKeyMenu, FileOperationMenu.

Towards #6456


app/menus/ApiKeyMenu.tsx
```diff
@@ -1,44 +1,25 @@
...
-  const handleRevoke = useCallback(() => {
-    dialogs.openModal({
-      title: t("Revoke token"),
-      content: (
-        <ApiKeyRevokeDialog onSubmit={dialogs.closeAllModals} apiKey={apiKey} />
-      ),
-    });
-  }, [t, dialogs, apiKey]);
+  const actions = [revokeApiKeyFactory({ apiKey })];
+  const rootAction = createRootMenuAction(actions);
```
tommoor: Should we have a `use` hook equivalent that memoizes the output – we need to consider how to make it so that rootAction doesn't invalidate every render

app/actions/index.ts
```diff
@@ -163,3 +171,193 @@ export async function performAction(action: Action, context: ActionContext) {
...
+export function createActionV2Group(
+  definition: Omit<ActionV2Group, "type">
+): ActionV2Group {
+  return {
+    ...definition,
+    type: "action_group",
+  };
+}
+
+export function createActionV2Separator(): ActionV2Separator {
```
tommoor: Not really an action, perhaps this could just be a static reusable object?

app/components/primitives/DropdownMenu.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+  background: ${s("menuBackground")};
+  box-shadow: ${s("menuShadow")};
+  border-radius: 6px;
+  padding: 6px;
+  outline: none;
+
+  transform-origin: var(--radix-dropdown-menu-content-transform-origin);
+
+  &[data-state="open"] {
```
tommoor, line 188: Haven't looked into it, but I'm not seeing any animation on the new menu?

hmacr: Timing function is super fast that it _feels_ instant for shorter menus (templates menu is relatively tall which shows the animation) - I'll adjust it.

Other option is to use different sets of timing function depending on menu height, but I think that's an unwanted complexity at this time.

app/menus/ApiKeyMenu.tsx
hmacr: `useMemo` should work here no? or I'm not understanding what you mean here? 🤔 
One case we need to be careful is when policies are involved in building the actions (should maintain reactivity there).

tommoor: It would. If we find ourselves using useMemo a lot it would be preferable to have a `useMenuAction` type thing that makes the memoized behavior default

app/hooks/useMenuAction.ts
```diff
@@ -0,0 +1,26 @@
...
+  ActionV2Group,
+  ActionV2Separator,
+  ActionV2Variant,
+  ActionV2WithChildren,
+} from "~/types";
+import usePrevious from "./usePrevious";
+
+type Actions = (ActionV2Variant | ActionV2Group | ActionV2Separator)[];
+
+export function useMenuAction(actions: Actions) {
```
hmacr, line 14: memo-ish hook that uses value identity instead of `useMemo`'s referential identity.. this ensures rootAction remains stable even when `actions` aren't memo-ed.

hmacr:
> I think we could test a bit better with a menu that has more items in it

I migrated `TemplatesMenu` for a much taller menu which helped uncover a scrolling bug in dropdown menu and has been fixed.

https://github.com/user-attachments/assets/e4e29617-95df-4e97-bd4b-c9f04ffa6658

app/menus/FileOperationMenu.tsx
```diff
@@ -17,41 +21,45 @@ type Props = {
...
+        url: fileOperation.downloadUrl,
+      }),
+      ActionV2Separator,
+      createActionV2({
+        name: t("Delete"),
+        icon: <TrashIcon />,
+        section,
+        visible: can.delete,
+        dangerous: true,
+        perform: () => onDelete,
```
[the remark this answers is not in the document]
tommoor: Good guy graphite

tommoor:
> I migrated TemplatesMenu for a much taller menu which helped uncover a scrolling bug in dropdown menu and has been fixed.

Nice, I think we should actually cap the menu at 75vh (this should be the case on the old one) – when it takes up the entire screen like that it looks a bit silly and requires a lot of mouse movement

hmacr:
> I think we should actually cap the menu at 75vh (this should be the case on the old one)

Agreed, just pushed a commit for this 🙂

PR #9629, opened 2025-07-14
build: try native plugin

TheAlexLichter:


tommoor:
Nice, ran locally too and everything seems to work fine

tommoor (approved):

PR #9650, opened 2025-07-16
chore: Move `Revision` event writing to model layer

### LINKED ISSUE #9609: Revision

hmacr:
Closes #9609 
Towards #7920

### DIFF
--- a/server/types.ts
+++ b/server/types.ts
@@ -258,7 +258,6 @@ export type EmptyTrashEvent = {
 export type RevisionEvent = BaseEvent<Revision> & {
   name: "revisions.create";
   documentId: string;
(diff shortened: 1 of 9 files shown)


server/types.ts
```diff
@@ -258,7 +258,6 @@ export type EmptyTrashEvent = {
 export type RevisionEvent = BaseEvent<Revision> & {
   name: "revisions.create";
   documentId: string;
-  collectionId: string;
```
hmacr, line 261: Removed it since `Revision` model does not have `collectionId` column.. this field isn't used in the event, so we're safe.

tommoor (approved):

PR #9662, opened 2025-07-17
fix: Skip overwriting non-transaction fields in `firstCollectionForTeam` query

hmacr:
In the file imports flow, we were incorrectly querying for team's first collection to resolve index collision (basically overwriting the `where` clause).


hmacr:
One mitigating factor is that it was isolated to file-based imports, so it's not as bad as it could have been..

PR #9663, opened 2025-07-17
chore: Move `Collection` event writing to model layer

### LINKED ISSUE #9611: Collection

hmacr:
Closes #9611 
Towards #7920


server/routes/api/collections/collections.ts
```diff
@@ -965,26 +942,19 @@ router.post(
...
-        transaction,
-      }
-    );
-    await Event.createFromContext(ctx, {
-      name: "collections.move",
-      collectionId: collection.id,
-      data: {
-        index: collection.index,
-      },
+    collection.index = await removeIndexCollision(collection.teamId, index, {
```
hmacr: This is unfortunate but we need the resolved index for the event.

server/routes/api/collections/collections.ts
```diff
@@ -653,14 +647,7 @@ router.post(
...
-      name: "collections.update",
-      collectionId: collection.id,
-      data: {
-        name,
-      },
-    });
+    await collection.saveWithCtx(ctx, undefined, { data: { name } });
 
     if (privacyChanged || sharingChanged) {
       await Event.createFromContext(ctx, {
```
hmacr: This feels like a special case that can be left as-is, rather than overload (or) introduce new `withCtx` methods. Thoughts?

server/commands/collectionDestroyer.ts
```diff
@@ -1,32 +1,23 @@
...
 
-export default async function collectionDestroyer({
-  collection,
-  transaction,
-  user,
-  ip,
-}: Props) {
-  await collection.destroy({ transaction });
-
+export default async function collectionDestroyer(
```
tommoor: We could also move `Document.update` to a beforeDestroy hook and remove `collectionDestroyer` command altogether

server/routes/api/collections/collections.ts
```diff
@@ -862,9 +844,16 @@ router.post(
 
     collection.archivedAt = new Date();
     collection.archivedById = user.id;
-    await collection.save({ transaction });
     collection.archivedBy = user;
 
+    await collection.saveWithCtx(ctx, undefined, {
+      name: "archive",
+      data: {
```
tommoor: Are there processors relying on this? Could just change them to use `changes` – we don't need consistency of `data` with past events.

Comment applies to a couple of other spots too

server/commands/collectionDestroyer.ts
hmacr: silly oversight, yep.

server/routes/api/collections/collections.ts
hmacr: This is moved to a `beforeUpdate` hook.

server/routes/api/collections/collections.ts
hmacr: Turns out this wasn't necessary at all, `changes` can be used to get the latest index in the processor.

server/queues/processors/NotificationsProcessor.ts
```diff
@@ -86,9 +86,8 @@ export default class NotificationsProcessor extends BaseProcessor {
   async collectionCreated(event: CollectionEvent) {
     // never send notifications when batch importing
     if (
-      "data" in event &&
-      "source" in event.data &&
-      event.data.source === "import"
+      !!event.changes?.attributes.apiImportId ||
+      !!event.changes?.attributes.importId
```
hmacr, line 90: This fixes a bug where collection creation notifications were sent in file import flow.

server/types.ts
```diff
@@ -307,44 +307,17 @@ export type DocumentGroupEvent = BaseEvent<GroupMembership> & {
...
-    | {
-        name: "collections.permission_changed";
-        collectionId: string;
-        data: {
-          privacyChanged: boolean;
-          sharingChanged: boolean;
-        };
-      }
-  );
+export type CollectionEvent = BaseEvent<Collection> & {
```
tommoor, line 310: Nice this is gradually getting cleaned up 👌

tommoor (approved):

PR #9666, opened 2025-07-17
fix: Frontend errors not properly filtered when source is minified

tommoor:
This regressed as part of #9523 – however, relying on this string comparison was bad practice to begin with, a further enhancement here might be to inherit all of these errors from a single ignored type?


tommoor:
image

😅 it worked

PR #9672, opened 2025-07-18
fix: Invariant violation with `addToArchive` called with non-model

### LINKED ISSUE #9667: Invariant Violation: No store found for model name "undefined"
Sentry Issue: [OUTLINE-CLOUD-986](https://outline-jl.sentry.io/issues/6752145054/?referrer=github_integration)

```
Invariant Violation: No store found for model name "undefined"
  at _Be.getStoreForModelName (../../../app/stores/RootStore.ts:125:5)
  at <anonymous> (../../../app/stores/base/Store.ts:224:36)
  at Map.forEach (<anonymous>)
  at Dy.addToArchive (../../../app/stores/base/Store.ts:223:22)
  at vl.<anonymous> (../../../app/components/WebsocketProvider.tsx:290:19)
...
(11 additional frame(s) were not displayed)
```

tommoor:
This is one of those "How did this ever work?" bugs, as far as I can tell a POJO has always been passed to this method that expects a model instead with `modelName` attribute.

closes #9667

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -285,7 +285,8 @@ class WebsocketProvider extends Component<Props> {
     this.socket.on(
       "documents.archive",
       action((event: PartialExcept<Document, "id">) => {
-        documents.addToArchive(event as Document);
+        const model = documents.add(event);
+        documents.addToArchive(model);
 
         if (event.collectionId) {
           const collection = collections.get(event.collectionId);


hmacr:
While I agree this change makes the code syntactically correct, I think we need to find why `modelName` is undefined for a configured relation [here](https://github.com/outline/outline/blob/ccac9379cae225801b32017db77c7749cca1dc78/app/models/decorators/Relation.ts#L46-L57) (input `item` isn't used in this flow except for updating the map).

tommoor:
It uses modelName on the incoming model for lookup, because it wasn't a model it's undefined.

PR #9679, opened 2025-07-19
feat: Option to export nested documents

### LINKED ISSUE #6884: Export as Markdown & Export as HTML does not export nested documents but only the root document
When you click on Download=>Markdown or Download=>HTML it will only export the top / root page but not the whole nested document structure into a zip file. Instead it only generates one html / md file. This feature was working flawlessly on older versions.

**To Reproduce**
1. Go to any document
2. Click on 'Download => HTML' or 'Download => Markdown'
3. Get only one file downloaded (the root file you selected)

**Expected behavior**
You should get a zip file containing all the files from that document in a nested manner.

**Screenshots**
Not applicable as it does not help to describe the issue.

- Install: self hosted
- Version: 4cb0f8bb94275f7ebf7f9056d6ca2a485dcc961f

 - OS: Arch Linux
 - Browser Thorium (Chromium) 
 - Version 123.0.6312.133

hmacr:
Closes #6884 

Document download menu is replaced with a dialog that has an option to include child documents.


app/actions/definitions/documents.tsx
```diff
@@ -415,84 +411,35 @@ export const shareDocument = createAction({
...
-  visible: ({ activeDocumentId, stores }) =>
-    !!activeDocumentId &&
-    stores.policies.abilities(activeDocumentId).download &&
-    env.PDF_EXPORT_ENABLED,
   perform: ({ activeDocumentId, t, stores }) => {
     if (!activeDocumentId) {
       return;
     }
 
-    const id = toast.loading(`${t("Exporting")}…`);
```
hmacr, line 558: It seems a toast is being rendered for PDF export.
This is removed in this PR - should this be maintained?

server/routes/api/documents/documents.ts
```diff
@@ -733,20 +742,67 @@ router.post(
       includeState: !accept?.includes("text/markdown"),
     });
 
+    authorize(user, "download", document);
```
hmacr, line 766: This policy check was missing in the API earlier.
Right now guests can download a document.. I'm thinking this policy should be changed to restrict them..?

hmacr:
@tommoor - resolved the merge conflicts here too ✌️

PR #9693, opened 2025-07-21
chore: Improve group membership dialogs

tommoor:
Remove fullscreen dialog styling


hmacr:
Much needed, something I've wanted to do as well.

PR #9715, opened 2025-07-23
chore: Update `rolldown-vite`

tommoor:
Allows removal of temporary resolutions rule


tommoor:
Urgh, unfortunately this breaks new and different things 😢

PR #9717, opened 2025-07-23
Allow user to add plausible domain and source for analyitics

motangpuar:
Add Plausible domain and source as environment variable.


tommoor:
Hi – thanks for the contribution. However new analytics providers should be added through the plugin system, `plugins/matomo` is a good example!

tommoor:
I'm closing this for now, please reopen if you take the other approach!

PR #9741, opened 2025-07-26
Conversion of `User` to event system

### LINKED ISSUE #9613: User

tommoor:
closes #9613
towards #7920


server/commands/teamProvisioner.ts
```diff
@@ -109,8 +104,7 @@ async function teamProvisioner({
 
   // We cannot find an existing team, so we create a new one
   const team = await sequelize.transaction((transaction) =>
-    teamCreator({
-      ctx: createContext({ ip, transaction }),
+    teamCreator(createContext({ transaction }), {
```
hmacr, line 107:
```suggestion
    teamCreator(ctx, {
```

server/commands/userDestroyer.ts
```diff
@@ -46,15 +46,5 @@ export default async function userDestroyer(
...
-    userId: user.id,
-    data: {
-      name: user.name,
-    },
-  });
-
-  return user.destroy({
-    transaction,
-  });
+  return user.destroyWithCtx(ctx);
```
hmacr: I think the `last_user` and `last_admin` checks can be moved to `BeforeDestroy` hook that will allow us to remove this file.

server/commands/userProvisioner.ts
```diff
@@ -163,7 +138,7 @@ export default async function userProvisioner({
           name,
           avatarUrl,
           lastActiveAt: new Date(),
-          lastActiveIp: ip,
+          lastActiveIp: ctx.ip,
         },
         {
           transaction,
```
hmacr, line 144: Any reason to not re-use transaction from ctx?

server/commands/userProvisioner.ts
```diff
@@ -239,25 +215,12 @@ export default async function userProvisioner({
...
-        data: {
-          name: user.name,
-        },
-        ip,
-      },
-      {
-        transaction,
       }
     );
     await transaction.commit();
```
hmacr, line 226: `const transaction = await User.sequelize!.transaction();` on L189  and its uses can be removed since we now use `withCtx` method.

server/commands/userSuspender.ts
```diff
@@ -1,60 +1,42 @@
...
-      suspendedById: actorId,
+      suspendedById,
       suspendedAt: new Date(),
     },
     {
-      transaction,
+      name: "suspend",
     }
   );
   await GroupUser.destroy({
```
hmacr: Opportunity to move this to a `BeforeUpdate` hook and remove this file entirely?

server/commands/userUnsuspender.ts
```diff
@@ -1,48 +1,31 @@
...
-  actorId,
-  transaction,
-  ip,
-}: Props): Promise<void> {
-  if (user.id === actorId) {
+export default async function userUnsuspender(
+  ctx: APIContext,
+  { user }: Props
+): Promise<void> {
+  if (user.id === ctx.state.auth.user.id) {
```
hmacr: wdyt of moving to router layer and removing this file?

server/commands/teamProvisioner.ts
tommoor: This is because the surrounding context doesn't have a transaction. Ideally would remove this here, but I was trying to reduce churn in the PR especially around transactions as it's easy to create deadlock

hmacr: Ah, my bad.
There are a couple other comments related to transactions, those become inapplicable too!

server/commands/userSuspender.ts
tommoor: I decided to remove this logic, it feels safer to retain group memberships on suspension

server/commands/userUnsuspender.ts
tommoor: Even better, moved it into the policy so it's covered by `authorize()` call

server/routes/api/users/users.ts
```diff
@@ -485,12 +463,16 @@ router.post(
...
+    await user.updateWithCtx(
+      ctx,
+      {
+        suspendedById: actor.id,
+        suspendedAt: new Date(),
+      },
+      {
+        name: "suspend",
+      }
+    );
```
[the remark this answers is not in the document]
tommoor: Incorrect - it's called from `UserDemotedProcessor` now

server/models/decorators/Fix.ts
```diff
@@ -30,10 +30,23 @@ export default function Fix(target: any): void {
 
         Object.defineProperty(this, propertyKey, {
           get() {
-            return this.getDataValue(propertyKey);
+            // Safety check for Jest serialization - getDataValue may not be available
```
tommoor, line 33: Unrelated, but this has been annoying me running local tests for a while

server/routes/api/users/users.ts
```diff
@@ -485,12 +461,16 @@ router.post(
...
```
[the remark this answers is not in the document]
tommoor: As already noted, this was purposefully removed

hmacr (approved):
Works great!

Issue #9641, opened 2025-07-15
Slack Auth doesn't work with OIDC


tommoor:
Hi everyone!

I'm trying to host Outline with OIDC and Slack authentication because some of my team members are not in Slack. I have followed all the steps to add Slack authentication and can confirm that it works perfectly if I disable OIDC authentication. However, the issue arises when I try using them simultaneously. This is the issue I got when login using Slack

{"error":"Invalid authentication","level":"error","message":"Error during authentication","stack":"UnauthorizedError: Invalid authentication\n    at InvalidAuthenticationError (/opt/outline/build/server/errors.js:52:34)\n    at teamProvisioner (/opt/outline/build/server/commands/teamProvisioner.js:74:50)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"}

Please take note that when I first created outline, I used my company domain for the super user of OIDC. However, when I implemented Slack auth, all of my team members are "@gmail.com". I thought this is the root of the issue so I have added "gmail.com" to the allowed domain in the app, but it doesn't seem to work.
![image](https://github.com/user-attachments/assets/f4345859-4d63-4169-928f-88f9e153f454)

Thank you very much for your attention. Every help is appreciated.

_Originally posted by @vo-d in https://github.com/outline/outline/discussions/9016_

PR #9742, opened 2025-07-27
fix: extract domain from user email in Slack authentication

naaa760:
### I made a PR!

#9641

### Bug Description

- Issue: Slack authentication fails with "Invalid authentication" error when both OIDC and Slack are enabled simultaneously.

- Root Cause: Slack authentication was not extracting the domain from the user's email address, unlike other auth providers (OIDC, Google, Azure, Discord).

- Fix: Added domain extraction from user's email in Slack auth to match the behavior of other authentication providers.

- Impact: Users can now authenticate with Slack even when OIDC is enabled, allowing organizations to use both authentication methods together.

### DIFF
--- a/plugins/slack/server/auth/slack.ts
+++ b/plugins/slack/server/auth/slack.ts
@@ -25,7 +25,6 @@ import {
   StateStore,
 } from "@server/utils/passport";
 import { parseEmail } from "@shared/utils/email";
-import { slugifyDomain } from "@shared/utils/domains";
 import env from "../env";
 import * as Slack from "../slack";
 import * as T from "./schema";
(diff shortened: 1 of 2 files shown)


plugins/slack/server/auth/slack.ts
```diff
@@ -82,12 +84,17 @@ if (env.SLACK_CLIENT_ID && env.SLACK_CLIENT_SECRET) {
...
+        const subdomain = slugifyDomain(domain);
+
         const result = await accountProvisioner({
           ip: ctx.ip,
           team: {
             teamId: team?.id,
             name: profile.team.name,
-            subdomain: profile.team.domain,
+            domain,
+            subdomain,
```
tommoor: It would be better to retain the previous logic for `subdomain` so that we match the Slack subdomain where possible

tommoor (changes requested):

plugins/slack/server/auth/slack.ts
nehaaprasad: I fixed Slack OIDC authentication to retain original subdomain logic (profile.team.domain) while adding domain extraction from user email.

tommoor: No you didnt?

nehaaprasad: ohh let me recheck again.

nehaaprasad: @tommoor 
please check?

Issue #9753, opened 2025-07-28
Inconsistent UI when editing document in a collection vs when editing root document of the collection


Wint3rmute:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

When opening the root document of a collection (via clicking on the collection name on the document tree), the "Edit" button is not visible. The root document's UI flow regarding switching in and out of edit mode is different than every other document.

## Root document

<img width="1404" height="372" alt="Image" src="https://github.com/user-attachments/assets/5d0d10d0-f1d8-4422-ba38-109fdcb40521" />

## "Normal" document

<img width="1101" height="238" alt="Image" src="https://github.com/user-attachments/assets/7e770325-339b-48d3-953d-163bb8a1279a" />

### Expected Behavior

The UI flow of switching in and out of edit mode should be the same across all documents.

### Steps To Reproduce

**This is written assuming full privileges on the Outline instance.**

1. Create a new collection
2. Create a new page within the collection
3. Click on the new page within the collection
4. Notice the "Edit" button at the top-right corner
5. Click on the root page of the collection
6. Notice that there's no "Edit" button

### Environment

```markdown
- Outline: 0.83.0
- Browser: Chromium
```

patelajay745:
i would love solve this issue.

nehaaprasad:
Hello! @Wint3rmute 
added Edit button for collection root documents to match document editing flow resolves UI inconsistency between collection and document editing modes

PR #9764, opened 2025-07-29
feat: Add Edit button for collection root documents to match document editing flow

naaa760:
#9753 

### Problem
When viewing a collection (root document), there was no "Edit" button visible in the header, unlike regular documents which have a clear "Edit" button. This created an inconsistent UI flow between editing collection root documents and regular documents.

### Solution
- Added `collectionEditPath()` helper function to create edit paths for collections
- Added `/collection/:id/edit` route to handle collection editing
- Modified Collection scene to detect edit mode and show appropriate UI:
  - Shows "Done editing" button when in edit mode
  - Hides tabs when editing (focuses on content)
  - Passes `isEditing` prop to Overview component
- Updated Overview component to be read-only when not in edit mode
- Added "Edit" button to Collection Actions component with proper permissions

### Changes
- `app/utils/routeHelpers.ts`: Added `collectionEditPath()` function
- `app/routes/authenticated.tsx`: Added collection edit route
- `app/scenes/Collection/index.tsx`: Added edit mode detection and conditional UI
- `app/scenes/Collection/components/Overview.tsx`: Added `isEditing` prop support
- `app/scenes/Collection/components/Actions.tsx`: Added Edit button

- Fixes the enhancement request for consistent UI flow across all documents.

### DIFF
--- a/app/scenes/Collection/index.tsx
+++ b/app/scenes/Collection/index.tsx
@@ -47,6 +47,9 @@ import MembershipPreview from "./components/MembershipPreview";
 import Notices from "./components/Notices";
 import Overview from "./components/Overview";
 import ShareButton from "./components/ShareButton";
+import Tooltip from "~/components/Tooltip";
+import Button from "~/components/Button";
+import { Link } from "react-router-dom";
 
 const IconPicker = lazy(() => import("~/components/IconPicker"));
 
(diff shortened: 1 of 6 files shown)


tommoor:
There is a user preference `user.separateEditMode` that chooses whether an edit button appears on documents, or whether it is seamless always-editing. We should respect that here too

nehaaprasad:
@tommoor 
Fixed! Now respects user.separateEditMode preference - Edit button only shows when separate edit mode is enabled, and collection editing behavior matches documents exactly.

app/scenes/Collection/index.tsx
```diff
@@ -219,31 +249,32 @@ const CollectionScene = observer(function _CollectionScene() {
...
-                <Tab {...tabProps(CollectionPath.Overview)}>
-                  {t("Overview")}
-                </Tab>
-              )}
-              <Tab {...tabProps(CollectionPath.Recent)}>{t("Documents")}</Tab>
-              {!collection.isArchived && (
-                <>
-                  <Tab {...tabProps(CollectionPath.Updated)}>
-                    {t("Recently updated")}
+          {!isEditing && (
```
tommoor, line 252: This hides all of the UI when not editing?

PR #9772, opened 2025-07-29
fix(i18n): replace "Sammlung" with "Kollektion" for collections in German translation

quickstar:
`Sammlung` is a literal translation of `collection`, but in German it is primarily used for physical or curated collections (like art) and sounds strange and formal in digital contexts. For software and wikis, terms like `Kollektion` are more modern and intuitive for users, improving clarity and usability.


tommoor:
Translations cannot be provided through code directly as they will be overridden by the translation provider.

Please see the docs here for details on how to contribute:

https://github.com/outline/outline/blob/main/docs/TRANSLATION.md

Issue #9776, opened 2025-07-30
Full width tables don't automatically go "full-width"


Its4Nik:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Hey there, so right now if you edit a Table and remove for example one column and set it to full width, the table just moves over to the left and does not span across the whole available editor width.

<img width="1814" height="622" alt="Image" src="https://github.com/user-attachments/assets/c86f887d-24d5-4420-95f4-a414c35e3575" />

(just ignore the text it does not have to do with this issue and translating won't help you :D )

### Expected Behavior

I expected the Table to just span like Images when set to full width.

### Steps To Reproduce

1. Create Table
  a. Delete a Column (didn't have time to test)
2. Set to full width

### Environment

```markdown
- Outline: v0.85.1
- Browser: Latest Firefox and friends of mine had the same issue with chrome / opera / ...
```

### Anything else?

Would be cool if there was an additional table setting like "Dynamic Table" which always goes as big as the editor width (not full width) because right now you always have to move the column borders to make it fit without an annoying scrollbar.

nehaaprasad:
Hi @tommoor  and @Its4Nik 
I fixed table full-width CSS transform to match image implementation removed extra padding that was causing tables to not span entire editor width

PR #9777, opened 2025-07-30
fix: Full-width tables now properly span entire editor width

naaa760:
### I made a PR!

FIXES: #9776

### Problem
When users set a table to full-width mode, the table would only move to the left instead of expanding to fill the entire editor width. This was inconsistent with how images behave in full-width mode.

### Root Cause
The CSS transform calculation for table full-width included an extra padding value (`${EditorStyleHelper.padding}px`) that caused incorrect positioning, unlike the working image full-width implementation.

### Solution
- Removed the extra padding from the table full-width transform calculation
- Updated the scrollable area styles to ensure proper full-width behavior
- Tables now use the same positioning logic as images for consistency

### DIFF
--- a/shared/editor/components/Styles.ts
+++ b/shared/editor/components/Styles.ts
@@ -605,11 +605,15 @@ iframe.embed {
 }
 
 .${EditorStyleHelper.tableFullWidth} {
-  transform: translateX(calc(50% + ${
-    EditorStyleHelper.padding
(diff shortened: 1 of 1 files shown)


tommoor:
Hi, I'm afraid this doesn't fix the issue – it removes the padding that is neccessary to use the table handles. The issue is not solveable with css. Thanks for the attempt

PR #9787, opened 2025-08-01
Add sub-menu support for dropdown

hmacr:
- Added sub-menu support for dropdown
- Migrated `UserMenu` and `AccountMenu` that uses sub-menus
- Migrated `TeamMenu` for fixing type errors (it uses one of the V2 actions)

Towards #6456 
Follow-on from #9578


app/components/primitives/Drawer.tsx
```diff
@@ -22,15 +24,25 @@ const DrawerContent = React.forwardRef<
   React.ComponentPropsWithoutRef<typeof DrawerPrimitive.Content>
 >((props, ref) => {
   const { children, ...rest } = props;
+  const [measureRef, bounds] = useMeasure();
```
hmacr, line 27: We do have `useComponentSize` but for reasons unknown, it doesn't correctly measure the div size (took a decent time to debug vaul/framer).  So, ended up adding battle-tested `react-use-measure` (tiny library).

tommoor: Maybe we can just remove `useComponentSize` after then

PR #9806, opened 2025-08-03
fix: duplicate highlight shortcut key label

k2on:
Fixes this bug!

image


tommoor:
Interesting, the keyboard shortcut on mac is actually `cmd` + `ctrl` + `h` – but clearly this isn't portable to Windows. The actual shortcut needs changing, rather than just this UI

PR #9822, opened 2025-08-04
Use `shareLoader` for loading public docs in `documents.info` endpoint

### LINKED ISSUE #9801: Use `loadPublicShare` instead of `documentLoader` when loading a `shareId` from documents endpoints
_Originally posted by @tommoor in https://github.com/outline/outline/issues/9529#issuecomment-3148542909_

hmacr:
Closes #9801


server/commands/documentLoader.ts
```diff
@@ -1,228 +1,50 @@
...
+}: Props): Promise<Document> {
+  if (!user) {
+    throw AuthenticationError(`Authentication required`);
   }
 
-  const shareUrlId =
-    shareId && !isUUID(shareId) && UrlHelper.SHARE_URL_SLUG_REGEX.test(shareId)
-      ? shareId
-      : undefined;
+  const document = await Document.findByPk(id, {
```
hmacr, line 16: Diff is messy here, but the only change is removing the "if" block for share specific code.

app/stores/DocumentsStore.ts
```diff
@@ -515,74 +515,6 @@ export default class DocumentsStore extends Store<Document> {
...
-        return {
-          document: doc,
-        };
-      }
-
-      if (
-        doc &&
-        options.shareId &&
-        !options.force &&
-        this.sharedCache.has(options.shareId)
```
tommoor, line 546: Can `sharedCache` in documents store go entirely?

hmacr: yep :)

hmacr: Pushing `shareId` cleanup in authenticated path params..

server/commands/documentLoader.ts
```diff
@@ -1,228 +1,50 @@
...
-import { authorize, can } from "@server/policies";
+import { Document, User } from "@server/models";
+import { authorize } from "@server/policies";
 
 type Props = {
-  id?: string;
-  shareId?: string;
-  teamId?: string;
+  id: string;
   user?: User;
```
tommoor: Can documentLoader require `user` now

hmacr: I left it as is since `documentLoader` is used in export API which has an optional auth, but it feels like it shouldn't have been the case to start with since `download` policy restricts it to auth-ed user anyway (and loader would have always thrown error too).
Safe to make auth mandatory in `documents.export`?

server/routes/api/documents/documents.ts
```diff
@@ -699,34 +711,6 @@ router.post(
   }
 );
```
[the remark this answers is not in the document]
hmacr: We have `shares.sitemap` as a replacement now.

server/commands/documentLoader.ts
tommoor: Traced it back to the very first version of that endpoint, 5 years ago has always supported `shareId`:

https://github.com/outline/outline/commit/b51d818db3740068bf289fded02f80d350f6c8ad#diff-ce4bb28c7d77e61b5a4348fcda078c27194ddb25b68df1d0b2e34fdc3ef26ad9R420

So, I guess it should be kept.

hmacr: At some point, it stopped supporting `shareId` but auth wasn't updated. I'm more in favor of fixing this now and require `user` in documentLoader as you suggested. wdyt?

tommoor: Ok, finally tracked down the [commit](https://github.com/outline/outline/commit/1c9eee2134e8bacd96d5b6200d07a53c11b5b1c7) from a while ago where this was changed – agreed, back to my original suggestion then!

tommoor (approved):

Issue #9836, opened 2025-08-05
Self hosting - Editing collection name - Focus not set in search field for Icons & Emojis


Amtech281:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

1: I edit a collection
2: I want to change an icon and search in the search line (see screenshot)
3: The focus is not in the text-field and I cannot enter text to search an icon

When I switch to emojis and back and enter, then I see that in the back it flickers and if by coincidende the focus is then set in the search field.

Version v0.85.1 (however it was in the versions before too)

### Expected Behavior

I can search for icons

### Steps To Reproduce

see above

### Environment

```markdown
- Outline:0.85.1
- Browser: Firefox, Chrome, Opera
```

### Anything else?

![Image](https://github.com/user-attachments/assets/65c1d108-19f1-4036-ae4b-b001b295bced)

Amtech281:
thanks, so I should create an account to read the answer?

one addition: Same problem with setting the color code

hmacr:
This has been fixed in #9576 and will work as expected in the next release.

Amtech281:
thats great!

PR #9839, opened 2025-08-05
Autofocus search input when opening icon picker

hmacr:
Related #9836

### DIFF
--- a/app/components/IconPicker/components/EmojiPanel.tsx
+++ b/app/components/IconPicker/components/EmojiPanel.tsx
@@ -136,11 +136,12 @@ const EmojiPanel = ({
         freqEmojis,
       });
 
-  React.useEffect(() => {
-    if (scrollableRef.current) {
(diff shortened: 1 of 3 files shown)


app/components/IconPicker/components/EmojiPanel.tsx
```diff
@@ -136,11 +136,12 @@ const EmojiPanel = ({
...
-  React.useEffect(() => {
-    if (scrollableRef.current) {
-      scrollableRef.current.scrollTop = 0;
+  React.useLayoutEffect(() => {
+    if (!panelActive) {
+      return;
     }
-    searchRef.current?.focus();
+    scrollableRef.current?.scroll({ top: 0 });
+    requestAnimationFrame(() => searchRef.current?.focus());
```
hmacr, line 144: Delaying by a frame since the default focus timing conflicts with Radix Tab's internal focus management

PR #9844, opened 2025-08-05
Migrate `SearchPopover` to Radix

hmacr:
Turns out `PopoverAnchor` can be used in place of a trigger, rather than build a combobox.

Supersedes #9751


tommoor (approved):
Tested ✅

PR #9850, opened 2025-08-06
fix: Various issues with JSON import

tommoor:


app/components/List/Item.tsx
```diff
@@ -231,6 +231,7 @@ const Subtitle = styled.p<{ $small?: boolean; $selected?: boolean }>`
   font-size: ${(props) => (props.$small ? 13 : 14)}px;
   color: ${s("textTertiary")};
   margin-top: -2px;
+  overflow-wrap: break-word;
```
tommoor, line 234: Long errors extended outside the list

server/utils/fs.ts
```diff
@@ -31,18 +31,57 @@ export function stringByteLength(str: string): number {
...
- * @param length The maximum length of the file name.
+ * @param length The maximum length of the file name in bytes.
  * @returns The trimmed file name.
  */
 export function trimFileAndExt(text: string, length: number): string {
-  if (stringByteLength(text) > length) {
+  if (Buffer.byteLength(text, "utf8") > length) {
     const ext = path.extname(text);
     const name = path.basename(text, ext);
-    return name.slice(0, length - stringByteLength(ext)) + ext;
```
tommoor, line 45: Mismatch here between measurement of bytelength but slicing on character length

server/utils/ZipHelper.ts
```diff
@@ -163,19 +163,22 @@ export default class ZipHelper {
...
 
+                      const fileName = trimFileAndExt(
+                        path.basename(filePath),
+                        MAX_FILE_NAME_LENGTH
+                      );
+
                       const location = trimFileAndExt(
-                        path.join(
-                          outputDir,
-                          trimFileAndExt(fileName, MAX_FILE_NAME_LENGTH)
```
tommoor, line 175: `fileName` was badly named here, it was actually the entire file path – so we were trimming it to a max of 255 characters causing issues.

package.json
```diff
@@ -269,7 +269,7 @@
     "y-indexeddb": "^9.0.11",
     "y-prosemirror": "^1.3.7",
     "y-protocols": "^1.0.6",
-    "yauzl": "^2.10.0",
```
tommoor, line 272: Update zip reading lib while we're here, drops support for Node 12

Issue #7504, opened 2024-09-01
`documents.move` websocket events are not send down enough channels


tommoor:
`documents.move` event should be received by every client with access to the document in question, currently these events are only sent down collection channels.

hmacr:
Maybe solved by #7745..

tommoor:
Noted, will review this week 😉

PR #9854, opened 2025-08-06
Show correct child structure for shared documents

### LINKED ISSUE #7502: Sub-documents within shared parent are not ordered correctly

hmacr:
Ensures collection sort order is used in the shared section.
Realtime move updates are not handled here.. at least better than what we have currently.

Closes #7502

### DIFF
--- a/app/models/base/NavigableModel.ts
+++ b/app/models/base/NavigableModel.ts
@@ -1,0 +1,83 @@
+import invariant from "invariant";
+import { observable, runInAction } from "mobx";
+import { JSONObject, NavigationNode } from "@shared/types";
+import { client } from "~/utils/ApiClient";
+import ParanoidModel from "./ParanoidModel";
+
+export default abstract class NavigableModel extends ParanoidModel {
+  private isFetching = false;
+
+  /** The child documents structure of the model. */
+  @observable
+  documents?: NavigationNode[];
+
+  /**
+   * Fetches the child documents structure from the server.
+   */
+  async fetchDocuments(options: {
+    path: string;
+    params: JSONObject;
+    force?: boolean;
+  }) {
+    if (this.isFetching) {
+      return;
+    }
+
+    if (this.documents && options.force !== true) {
+      return;
+    }
+
+    try {
+      this.isFetching = true;
+      const res = await client.post(options.path, options.params);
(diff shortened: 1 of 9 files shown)


app/models/base/NavigableModel.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /** The child documents structure of the model. */
+  @observable
+  documents?: NavigationNode[];
+
+  /**
+   * Fetches the child documents structure from the server.
+   */
+  async fetchDocuments(options: {
+    path: string;
```
hmacr, line 16: Opportunity to make `Collection` a `NavigableModel` as well. Customising path and param is needed in that case.

app/models/base/NavigableModel.ts
```diff
@@ -0,0 +1,83 @@
...
+    } finally {
+      this.isFetching = false;
+    }
+  }
+
+  /**
+   * Returns the document path from the original document shared with this membership.
+   */
+  pathToDocument(documentId: string) {
+    let path: NavigationNode[] | undefined = [];
```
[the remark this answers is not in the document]
hmacr: There's a semicolon 🤷‍♂️

hmacr:
@tommoor This is ready for review.

Remaining part is to figure out sending move events for these shared docs (#7504).

app/scenes/Document/components/References.tsx
```diff
@@ -29,9 +29,13 @@ function References({ document }: Props) {
...
-  const children = collection
-    ? collection.getChildrenForDocument(document.id)
-    : [];
+  const membership =
+    userMemberships.getByDocumentId(document.id) ??
+    groupMemberships.getByDocumentId(document.id);
+  const children =
+    collection?.getChildrenForDocument(document.id) ??
+    membership?.getChildrenForDocument(document.id) ??
+    [];
```
tommoor: What do you think about collapsing this inside a `document.getChildren` method?

hmacr: neat, yep.

PR #9862, opened 2025-08-06
chore: Setup missing oxlint configs

hmacr:
Noticed that oxlint was missing some configs, so did the following.
- Added oxlint config for app, server and shared packages.
- Removed vestigial eslintrc files.
- Renamed comment directives, this doesn't functionally change anything except for semantic correctness.


tommoor (approved):
Ok, the hundreds of warnings about `any` are back – this makes sense

PR #9863, opened 2025-08-06
Move `ColorPicker` to Radix popover

hmacr:
This should have been in a popover to start with, but alas, it was using the context menu primitive.


tommoor:
Nice, now the open direction adjusts automatically for the window bounds

tommoor (approved):

PR #9865, opened 2025-08-07
fix: Admins unable to revoke workspace API keys created by other users

### LINKED ISSUE #9864: Unable to revoke API keys of other users

### Current Behavior

I am unable to revoke and delete some API keys, receiving an "authorization error". Some of these keys were created over 2 years ago and are still active, others were more recently created but are expired. I am an admin, and our wiki is hosted by outline.

apiKeys.delete 
```
{
    "ok": false,
    "error": "authorization_error",
    "status": 403,
    "message": "Authorization error"
}
```

### Expected Behavior

As an admin, I want to be able to revoke and delete any API keys. 

### Steps To Reproduce

1. Go to API Keys list
2. Revoke API key

### Environment

```markdown
- Outline: 0.86.0
- Browser: Edge on macos
```

### Anything else?

_No response_

tommoor:
closes #9864

### DIFF
--- a/server/policies/apiKey.ts
+++ b/server/policies/apiKey.ts
@@ -33,8 +33,13 @@ allow(User, "listApiKeys", Team, (actor, team) =>
 
 allow(User, ["read", "update", "delete"], ApiKey, (actor, apiKey) =>
   and(
-    or(isOwner(actor, apiKey), actor.teamId === apiKey?.user.teamId),
-    actor.isAdmin ||
-      !!actor.team?.getPreference(TeamPreference.MembersCanCreateApiKey)
+    isTeamModel(actor, apiKey?.user),
+    or(
+      actor.isAdmin,
+      and(
+        isOwner(actor, apiKey),
+        !!actor.team?.getPreference(TeamPreference.MembersCanCreateApiKey)
+      )
+    )
   )
 );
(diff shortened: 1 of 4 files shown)


server/policies/apiKey.ts
```diff
@@ -32,7 +33,7 @@ allow(User, "listApiKeys", Team, (actor, team) =>
 
 allow(User, ["read", "update", "delete"], ApiKey, (actor, apiKey) =>
   and(
-    isOwner(actor, apiKey),
+    or(isOwner(actor, apiKey), actor.teamId === apiKey?.user.teamId),
```
hmacr: I understand this policy works now since list endpoint ensures non-admins can't see others' api keys, but I'm wary of the relaxed policy here which allows non-admins to take action on anyone's key as long as they are allowed to create a key.

How about a stricter policy like below?
```ts
and(
  isTeamModel(actor, apiKey?.user),
  or(
    actor.isAdmin,
    or(
      isOwner(actor, apiKey),
      !!actor.team?.getPreference(TeamPreference.MembersCanCreateApiKey)
    )
  )
)
```

tommoor: Good call, I'll add some tests around this

tommoor: Yours wasn't quite right – that last condition needs to be an `and` but tests caught it :)

hmacr (approved):

server/policies/apiKey.ts
hmacr: Damn yea, thankfully tests worked :)

PR #9866, opened 2025-08-07
chore: Move `SkinTonePicker` to Radix popover

tommoor:
Another one that was technically a menu, we gain a bit of nice animation here too with the switch

towards #6456

### DIFF
--- a/app/components/IconPicker/components/SkinTonePicker.tsx
+++ b/app/components/IconPicker/components/SkinTonePicker.tsx
@@ -36,16 +36,21 @@ const SkinTonePicker = ({
 
   const menuItems = useMemo(
     () =>
-      Object.entries(handEmojiVariants).map(([eskin, emoji]) => (
-        <IconButton
-          key={emoji.value}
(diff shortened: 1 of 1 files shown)


app/components/IconPicker/components/SkinTonePicker.tsx
```diff
@@ -19,62 +22,61 @@ const SkinTonePicker = ({
...
-          >
-            {handEmojiVariants[skinTone]!.value}
-          </StyledMenuButton>
-        )}
-      </MenuButton>
-      <Menu {...menu} aria-label={t("Choose default skin tone")}>
-        {(props) => <MenuContainer {...props}>{menuItems}</MenuContainer>}
-      </Menu>
-    </>
+    <Popover open={open} onOpenChange={setOpen}>
```
hmacr, line 58: Another one down!

Only issue I see here is nested popovers are messing up the focus management in comment reaction picker - notice that once `SkinTonePicker` is opened and closed, search box becomes unfocusable on mouse click (keyboard nav works though).
Quite a few issues are open for this bug on Radix, I have't found a solution yet 😞

Issue #9810, opened 2025-08-04
Width and height of images `undefined` when pasting HTML


tommoor:
Width and height is also `undefined`, so something certainly needs fixing here

_Originally posted by @tommoor in https://github.com/outline/outline/discussions/9809#discussioncomment-13981086_

nehaaprasad:
Hi @tommoor 
fix undefined image dimensions by parsing CSS styles when HTML attributes are missing during paste operations

PR #9868, opened 2025-08-07
fix: undefined width/height when pasting HTML images

nehaaprasad:
- fixes issue where image dimensions were undefined when pasting HTML content that contained images with CSS style dimensions instead of HTML attributes.

fix: #9810 

**Problem:**
- When pasting HTML content, images often have dimensions set via CSS styles (`style="width: 300px; height: 200px"`) rather than HTML attributes (`width="300" height="200"`)
- The image parser only checked HTML attributes, leaving width/height undefined for CSS-styled images

**Solution:**
- Enhanced the `parseDOM` logic in `Image.tsx` to fallback to CSS style dimensions when HTML attributes are not available
- Added support for parsing `img.style.width` and `img.style.height` (px values)
- Maintains backward compatibility with existing HTML attribute parsing

**Testing:**
- Images pasted from various sources (web pages, documents) now retain their dimensions
- Both HTML attributes and CSS styles are properly parsed

### DIFF
--- a/shared/editor/components/Styles.ts
+++ b/shared/editor/components/Styles.ts
@@ -609,6 +609,7 @@ iframe.embed {
     EditorStyleHelper.padding
   }px + var(--container-width) * -0.5 + var(--full-width-transform-offset)));
 
+  .${EditorStyleHelper.tableScrollable},
   table {
     width: calc(var(--container-width) - ${EditorStyleHelper.padding * 2}px);
   }


shared/editor/nodes/Image.tsx
```diff
@@ -122,8 +122,24 @@ export default class Image extends SimpleImage {
...
+              if (styleWidth.endsWith('px')) {
+                width = styleWidth.slice(0, -2);
+              }
+            }
+            if (!height && img?.style.height) {
+              const styleHeight = img.style.height;
+              if (styleHeight.endsWith('px')) {
+                height = styleHeight.slice(0, -2);
+              }
+            }
```
tommoor: We can lose this, as the parsing here is for our internal representation `div.image` and will never have style tags.

shared/editor/components/Styles.ts
```diff
@@ -605,11 +605,15 @@ iframe.embed {
 }
 
 .${EditorStyleHelper.tableFullWidth} {
-  transform: translateX(calc(50% + ${
-    EditorStyleHelper.padding
-  }px + var(--container-width) * -0.5 + var(--full-width-transform-offset)));
+  transform: translateX(calc(50% + var(--container-width) * -0.5 + var(--full-width-transform-offset)));
```
tommoor: Please revert these changes

tommoor (commented):
Looks reasonable!

nehaaprasad:
@tommoor 
fix undefined image dimensions by parsing CSS styles in img tag parser for external HTML paste operations

tommoor (approved):

PR #9871, opened 2025-08-07
chore: Move more dropdowns to Radix

hmacr:
#### Menus migrated
- Collection group member
- Comment
- Group member
- Group menu
- Insights
- Member
- Notifications
- Share


app/menus/CollectionGroupMemberMenu.tsx
```diff
@@ -12,34 +14,31 @@ type Props = {
 
 function CollectionGroupMemberMenu({ onMembers, onRemove }: Props) {
```
hmacr: This component and `MemberMenu`  aren't used in the app, is it safe to delete them provided they aren't used in Enterprise edition?

tommoor: Yep, unused in enterprise too!

hmacr: Pushed a commit that removes them.

tommoor (approved):

PR #9874, opened 2025-08-08
chore: Move `CollectionMenu` to Radix

hmacr:
Towards #6456


app/components/ActionButton.tsx
```diff
@@ -1,17 +1,22 @@
...
+} from "~/types";
 
 export type Props = React.HTMLAttributes<HTMLButtonElement> & {
   /** Show the button in a disabled state */
   disabled?: boolean;
   /** Hide the button entirely if action is not applicable */
   hideOnActionDisabled?: boolean;
   /** Action to use on button */
-  action?: Action;
+  action?: Action | Exclude<ActionV2Variant, ActionV2WithChildren>;
```
hmacr, line 19: Action button now supports v2 :)

As a follow-on to Radix migration, we can port all action button touchpoints to v2 as well.

tommoor: I guess we have to to get rid of the old `Action` type, can't have it hanging around!

tommoor:
image

It's allowing me to have multiple menus open at once. Other than that, looks good!

hmacr:
> It's allowing me to have multiple menus open at once. Other than that, looks good!

This has been fixed - sidebar links weren't respecting disabled pointer events which Radix sets on opening a menu (notice that pointer events don't occur in other places).

Also, separators now span the whole menu - being on dark mode, I didn't notice it was affected by the padding.

tommoor (approved):

PR #9879, opened 2025-08-09
chore: Move `DocumentMenu` to Radix

hmacr:
Towards #6456


app/actions/definitions/collections.tsx
```diff
@@ -235,6 +235,7 @@ export const subscribeCollection = createActionV2({
     const collection = stores.collections.get(activeCollectionId);
 
     return (
+      !!collection?.isActive &&
```
hmacr, line 238: Unrelated - hides subscribe/unsubscribe menu items for archived collections & documents.

Feels like we should probably do the same for star/unstar too, but there's an argument to be made that they're just archived for future reference, so starring is fine 🤔

app/components/Menu/DropdownMenu.tsx
```diff
@@ -43,19 +45,28 @@ export const DropdownMenu = observer(function DropdownMenu({
...
     context ??
     useActionContext({
       isContextMenu: true,
     });
-  const menuItems = (action.children as ActionV2Variant[]).map((childAction) =>
-    actionV2ToMenuItem(childAction, actionContext)
-  );
+
+  const menuItems = useComputed(() => {
+    if (!open) {
```
hmacr, line 70: perf: actions don't re-run unnecessarily 📈

tommoor:
image

"Unsubscribe" is stuck disabled here

tommoor (approved):

PR #9880, opened 2025-08-09
chore: Move more dropdowns to Radix (part 2)

hmacr:
_Marking as draft as there could merge conflicts with #9879._

Only a couple more to go after this PR.

#### Menus migrated
- Revision
- New child document
- Import
- OAuth authentication
- OAuth client

Towards #6456
Follow-on #9871


app/menus/NewChildDocumentMenu.tsx
```diff
@@ -44,37 +40,51 @@ function NewChildDocumentMenu({ document, label }: Props) {
...
+  const rootAction = useMenuAction(actions);
 
   return (
-    <>
-      <MenuButton {...menu}>{label}</MenuButton>
-      <ContextMenu {...menu} aria-label={t("New child document")}>
-        <Template {...menu} items={items} />
-      </ContextMenu>
-    </>
+    <Tooltip content={t("New document")} shortcut="n" placement="bottom">
```
tommoor, line 77: Don't you think its better/more flexible to keep the menu definition separate to the trigger in cases like this?

hmacr: I was a bit apprehensive about touching the tooltip component since it's used across the app (needs to forward dropdown trigger ref and props), but I agree that's the ideal way to go here - I'll modify the impl.

tommoor: We can also leave a `// TODO` in that case, as it's not used in more than one place yet

hmacr: That works too - handle it in a separate effort rather than increase scope here.

PR #9881, opened 2025-08-09
fix: Use shared link for collection in public breadcrumb

hmacr:
Couple of tiny bug fixes related to public breadcrumbs
- Clicking on collection was redirecting to auth page
- Breadcrumb wasn't shown for the first level document


tommoor:
I actually removed the top level document on purpose as there are plenty of ways to get back to the "top"

hmacr:
> I actually removed the top level document on purpose

In that case, I suppose this PR ends up being unnecessary - we won't ever show the collection in breadcrumb.

PR #9882, opened 2025-08-09
chore: Move `TableOfContentsMenu` & `BreadcrumbMenu` to Radix

hmacr:
And with that, all dropdowns have been migrated to Radix.

There are two components that use reakit context menu but don't naturally translate to dropdown in Radix world.
- ToolbarMenu - Need to ensure focus doesn't shift out of the editor which Radix dropdown doesn't support, so need to creative here with something like popover + arrow key navigation.
- FilterOptions - This most likely needs a combobox component to support search + options list.

Towards #6456


app/components/CollectionBreadcrumb.tsx
```diff
@@ -14,32 +15,24 @@ type Props = {
...
-  }, [collection, t]);
-
-  return <Breadcrumb items={items} highlightFirstItem />;
+  const actions = React.useMemo(
+    () => [
+      createInternalLinkActionV2({
+        name: t("Archive"),
+        section: ActiveCollectionSection,
+        icon: <ArchiveIcon />,
+        visible: collection.isArchived,
```
tommoor, line 24: So much nicer like this

tommoor (approved):

app/components/CollectionBreadcrumb.tsx
hmacr: Helped clean up quite a few `if` statements in document breadcrumb :)

PR #9914, opened 2025-08-13
fix: Standardize request filtering between cloud / self-hosted

tommoor:
Currently `request-filtering-agent` is only used in cloud hosted deployments, this PR makes it available to self-hosted deployments by default.

Note: It is not compatible with https tunneling if an additional https proxy is used.


server/utils/fetch.ts
```diff
@@ -60,3 +73,114 @@ export default async function fetch(
...
+
+    if (parsedProxyURL.tunnelMethod?.startsWith("httpOver")) {
+      const proxyURL = new URL(parsedProxyURL.href);
+      proxyURL.pathname = (parsedURL.protocol ?? "")
+        .concat("//")
+        .concat(parsedURL.host ?? "")
+        .concat(parsedURL.pathname + parsedURL.search);
+      if (parsedProxyURL.username || parsedProxyURL.password) {
+        proxyURL.username = parsedProxyURL.username;
+        proxyURL.password = parsedProxyURL.password;
```
ikkisoft, line 164: I assume you're not **** this field since it's http? I would still do that tbh.

server/utils/fetch.ts
```diff
@@ -1,9 +1,26 @@
...
+  tunnelMethod?: string;
+}
+
+const DefaultOptions = {
+  keepAlive: true,
+  timeout: 1000,
+  keepAliveMsecs: 500,
+  maxSockets: 200,
+  maxFreeSockets: 5,
+  maxCachedSessions: 500,
```
ikkisoft, line 22: Some people might need to fetch stuff from servers with self-signed certs so you might need to add `rejectUnauthorized: false`

For security, it would be better without but I can already see people complaining.

ikkisoft (commented):
There are two more places that use the standard `node-fetch` directly instead of the secure wrapper:

* shared/editor/plugins/UploadPlugin.ts # 75
* plugins/webhooks/server/tasks/DeliverWebhookTask.ts # 718

tommoor:
> There are two more places that use the standard node-fetch directly instead of the secure wrapper:

> shared/editor/plugins/UploadPlugin.ts # 75
> plugins/webhooks/server/tasks/DeliverWebhookTask.ts # 718

`UploadPlugin` is only used in the browser, the spot in `DeliverWebhookTask` is using the correct wrapped fetch – there is a lint rule to make sure unwrapped is never used on server.

server/utils/fetch.ts
tommoor: Can't quite parse what you mean here :)

ikkisoft: Never mind, I thought it was part of the `Logger.silly`

ikkisoft:
Ok, I was quickly grepping through the entire codebase and I've misread things. Yes, seems ok now - besides the corner case mention in Slack. We will need to understand what's going on with that.

ikkisoft:
Mistery solved: we should upgrade [request-filtering-agent](https://github.com/azu/request-filtering-agent) to `^2.0.1`

tommoor:
Upgraded separately in https://github.com/outline/outline/pull/9937

PR #9926, opened 2025-08-14
fix: Remove `mime-types` usage from the browser

tommoor:
Fixes warnings in dev/vite about extname usage.
Upgrade mime-types while we're here


shared/editor/plugins/UploadPlugin.ts
```diff
@@ -76,17 +75,13 @@ export class UploadPlugin extends Plugin {
                 .then((response) => response.blob())
                 .then((blob) => {
                   const fileName = fileNameFromUrl(imageSrc) ?? "pasted-image";
-                  const ext = extension(blob.type) ?? "png";
```
tommoor, line 79: Honestly, this wasn't particularly useful logic anyway – and only hits in rare cases that the uploaded file has no extension

PR #9962, opened 2025-08-18
chore: Remove focused comment state from router

### LINKED ISSUE #9960: Decouple comments from router
Currently the active comment is stored in the router state. While this was nice for accessibility, it leads to some odd and unperformant behavior overall.

Lets move this to react state - document context hook perhaps?, we should still allow the initial state to be loaded from query string for notifications.

Also unlocks #9728

tommoor:
closes #9960

### DIFF
--- a/app/scenes/Document/components/CommentSortMenu.tsx
+++ b/app/scenes/Document/components/CommentSortMenu.tsx
@@ -1,12 +1,10 @@
 import React from "react";
 import { useTranslation } from "react-i18next";
-import { useHistory, useLocation } from "react-router-dom";
 import styled from "styled-components";
 import { s } from "@shared/styles";
 import { UserPreference } from "@shared/types";
 import { InputSelect, Option } from "~/components/InputSelect";
 import useCurrentUser from "~/hooks/useCurrentUser";
-import { useLocationSidebarContext } from "~/hooks/useLocationSidebarContext";
 import { CommentSortType } from "~/types";
 
 type Props = {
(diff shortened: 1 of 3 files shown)


app/scenes/Document/components/CommentSortMenu.tsx
```diff
@@ -8,59 +7,45 @@ import { UserPreference } from "@shared/types";
...
-          ...queryString.parse(location.search),
-          resolved: undefined,
-        }),
-        pathname: location.pathname,
-        state: { sidebarContext },
-      });
+      onChange?.(val);
     },
-    [history, location, sidebarContext, user, preferredSortType]
+    [history, location, sidebarContext, user, onChange, preferredSortType]
```
hmacr: Can remove unused deps? (likewise in `Editor.tsx`)

app/scenes/Document/components/Editor.tsx
```diff
@@ -95,14 +95,7 @@ function DocumentEditor(props: Props, ref: React.RefObject<any>) {
...
       ) {
-        history.replace({
-          search: focusedComment.isResolved ? "resolved=" : "",
-          pathname: location.pathname,
-          state: {
-            commentId: focusedComment.id,
-            sidebarContext,
-          },
-        });
+        setFocusedCommentId(focusedComment.id);
```
hmacr, line 98: I'm afraid we're losing the search param update here - it ensures the correct view is opened when comment states changes between link generation vs link use (bug fixed here - https://github.com/outline/outline/pull/7959)

PR #10011, opened 2025-08-26
Use `leafText` when converting mention nodes to its text content

### LINKED ISSUE #10000: Table sorting does not work correctly on column with backlinks
Reported by customer – need to look at toPlainText behavior here.

hmacr:
Closes #10000

Keeping it scoped to the bug here - As a follow-on, we should remove the custom `toPlainText` method in favor of prosemirror's `leafText`.

### DIFF
--- a/shared/editor/nodes/Mention.tsx
+++ b/shared/editor/nodes/Mention.tsx
@@ -120,6 +120,7 @@ export default class Mention extends Node {
         toPlainText(node),
       ],
       toPlainText,
+      leafText: toPlainText,
     };
   }
 


tommoor:
Wasn't aware of leafText, must be newish?

hmacr:
> Wasn't aware of leafText, must be newish?

Stumbled on it when looking at prosemirror docs - looks like it's been available for quite sometime now ([v1.18.0](https://github.com/ProseMirror/prosemirror-model/releases/tag/1.18.0))

PR #10047, opened 2025-08-30
chore: Store refresh tokens for Linear integration

### LINKED ISSUE #10032: Update Linear implementation to handle refresh tokens
Linear OAuth is updating to issue short lived tokens with refresh tokens. We need to update the implementation to handle - should be quick.

tommoor:
closes #10032
ref https://linear.app/developers/oauth-2-0-authentication#refresh-an-access-token

### DIFF
--- a/plugins/linear/server/api/linear.ts
+++ b/plugins/linear/server/api/linear.ts
@@ -11,7 +11,7 @@ import { Linear } from "../linear";
 import UploadLinearWorkspaceLogoTask from "../tasks/UploadLinearWorkspaceLogoTask";
 import * as T from "./schema";
 import { LinearUtils } from "plugins/linear/shared/LinearUtils";
-import { toDate } from "@shared/utils/time";
+import { addSeconds } from "date-fns";
 
 const router = new Router();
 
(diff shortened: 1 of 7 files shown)


plugins/linear/server/linear.ts
```diff
@@ -95,6 +105,8 @@ export class Linear {
     try {
       const client = new LinearClient({
         accessToken: integration.authentication.token,
+        // TODO, waiting on SDK support
+        // refreshToken: integration.authentication.refreshToken,
```
hmacr: Do we need mutex to guard race conditions here? Unlike user auth, multiple mentions in a document would fire separate requests at the same time.

tommoor: Yea that sounds wise a- I'm going to let this sit until Linear SDK updates next week

plugins/linear/server/api/linear.ts
```diff
@@ -44,14 +44,21 @@ router.get(
...
 
       const authentication = await IntegrationAuthentication.create(
         {
           service: IntegrationService.Linear,
           userId: user.id,
           teamId: user.teamId,
           token: oauth.access_token,
+          refreshToken: oauth.refresh_token,
+          expiresAt: oauth.expires_in
+            ? new Date(Date.now() + oauth.expires_in * 1000)
```
apoorv-mishra:
```suggestion
            ? new Date(Date.now() + oauth.expires_in * Second.ms)
```

Issue #10040, opened 2025-08-29
Headings on pages imported from confluence are raised by one level


perlfisch:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Any H2 headings on confluence pages imported to Outline are turned into H1. H3 is turned into H2 etc.
H1 stays as H1.

Thus, a lot of the document structure is lost.

### Expected Behavior

All headings stay at their previous level

### Steps To Reproduce

Have H2 and H3 in a confluence Page
Export as HTML
Import into Outline

### Environment

```markdown
- Outline: v0.86.1 Enterprise
- Browser: Firefox ESR
```

### Anything else?

The HTML after the export is:

```
<h1 id="ExportTest-ThisisH1">This is H1</h1><h2 id="ExportTest-ThisisH2">This is H2</h2><h3 id="ExportTest-ThisisH3">This is H3</h3><h4 id="ExportTest-ThisisH4">This is H4</h4>
```

In Confluence:

<img width="308" height="209" alt="Image" src="https://github.com/user-attachments/assets/00347ff6-998e-45cb-93fe-cb37fe0d8a25" />

In Outline:

<img width="400" height="288" alt="Image" src="https://github.com/user-attachments/assets/109ad5ce-3b6e-4ff5-97e7-3772edb3c740" />

nehaaprasad:
@tommoor 
I've fixed it by removing the offset calculation that was incorrectly shifting heading levels up by 1, so now H2 stays H2, H3 stays H3, etc. 
Could you test importing a confluence page again to confirm the fix works?

PR #10048, opened 2025-08-30
fix: Preserve heading levels during HTML import from confluence

naaa760:
### **Fix: Preserve heading levels during HTML import from Confluence**

fixes: #10040

### **Problem**
Headings imported from Confluence HTML were being raised by one level:
- H2 → H1 (incorrect)
- H3 → H2 (incorrect) 
- H4 → H3 (incorrect)

This caused loss of document structure and hierarchy.

### **Solution**
Removed the offset calculation in the Heading node's `toDOM` method that was incorrectly shifting heading levels during HTML parsing.

**Before:** `h${node.attrs.level + (this.options.offset || 0)}`
**After:** `h${node.attrs.level}`

### **Result**
- H1 stays H1 ✅
- H2 stays H2 ✅  
- H3 stays H3 ✅
- H4 stays H4 ✅

### **Files Changed**
- `shared/editor/nodes/Heading.ts`


tommoor:
This doesn't fix anything it just removes the option and makes a bunch of other random changes

PR #10126, opened 2025-09-08
Update .nvmrc

exploreui:


exploreui:
Changed Node version from 22 to 20 in .nvmrc to fix deployment errors on Render. Compatible with Outline 0.82.0.

tommoor:
I don't know what issues you're seeing on render, but downgrading Node for everyone that uses Outline isn't the solution. 22 is LTS which is why it's the chosen version.

PR #10131, opened 2025-09-08
Update .env.sample with required variables section

ferhanmm:
A previous commit removed the "REQUIRED" section, adding it back.


tommoor:
Most of the variables in this section are not required, so the heading no longer makes sense

ferhanmm:
Understood. The documentation still references the Required section so I found it a bit confusing. 
https://docs.getoutline.com/s/hosting

tommoor:
Thanks – removed that note in the docs

Issue #10045, opened 2025-08-30
Add sidebar context menus


tommoor:
Now that we have context menus as a primitive (https://github.com/outline/outline/issues/9883) it's time to add them in places that are missing. An obvious starter is the documents and collections in sidebar.

nehaaprasad:
@tommoor 
fixes sidebar context menus for documents and collections by adding right-click support using the existing ContextMenu primitive

tommoor:
I took a pass at this and it was reasonably straightforward but I think it will be much better if we refactor `useActionContext` to be an actual context first, I don't like how it's passed as props everywhere when you should be able to augment the context at any point in the tree.

PR #10140, opened 2025-09-09
chore: Refactor `useActionContext` to use React context

tommoor:
- Reduces prop piping, no need to pass `context` directly to menus and buttons
- `isContextMenu` -> `isMenu` now that we differentiate internally between dropdown and context menus

Last remaining obvious refactor is to abstract away `activeDocumentId` 

towards #10045


app/components/ActionButton.tsx
```diff
@@ -28,22 +22,23 @@ export type Props = React.HTMLAttributes<HTMLButtonElement> & {
...
   ) {
+    const actionContext = useActionContext({
+      isButton: true,
+    });
     const isMounted = useIsMounted();
     const [executing, setExecuting] = React.useState(false);
     const disabled = rest.disabled;
 
-    if (action && !context) {
+    if (action && !actionContext) {
```
hmacr: Can remove this `if`? (Likewise remove `!actionContext` in the next one)
`actionContext` cannot be falsy here.

hmacr (commented):
Actions framework is becoming much cleaner 🔥

PR #10148, opened 2025-09-10
Regression in Lightbox swipe gestures

apoorv-mishra:
Regressed [here](https://github.com/outline/outline/pull/9704/commits/ad88bcc8ddd3c8504aeaa1d8f37c0a1e5a9c0271).


tommoor:
I think Graphite was correct here though, we shouldn't have floating `let` inside component as they get cleared on every render potentially interrupting swipe detection

PR #10153, opened 2025-09-11
Enable swipe actions on "Image failed to load" component within Lightbox

apoorv-mishra:
Enable swiping on Lightbox `Error` component.


apoorv-mishra:
Also, keeping copy and download action buttons disabled until the image is loaded.

tommoor:
Why? They're not affected by the image being loaded at all

apoorv-mishra:
Yes, you are right. Reverted!

PR #10179, opened 2025-09-14
fix: code cannot contain LaTeX syntax

### LINKED ISSUE #10172: Inline code has a bug

### Current Behavior

Can't type the code of latex inside inline code. 

https://github.com/user-attachments/assets/794dd1f2-b820-4431-9768-823c2fa5f6ab

### Expected Behavior

I think it should act like Vscode. It keeps the original code of latex.

Image

### Steps To Reproduce

Type `$$\sqrt{a}$$`

### Environment

```markdown
- Outline:Outline 0.87.3
- Browser:chrome
```

### Anything else?

_No response_

tommoor:
closes #10172

### DIFF
--- a/shared/editor/nodes/Math.ts
+++ b/shared/editor/nodes/Math.ts
@@ -1,7 +1,6 @@
 import {
   mathBackspaceCmd,
   insertMathCmd,
-  makeInlineMathInputRule,
   mathSchemaSpec,
 } from "@benrbray/prosemirror-math";
 import { PluginSimple } from "markdown-it";
@@ -16,6 +15,8 @@ import MathPlugin from "../extensions/Math";
 import { MarkdownSerializerState } from "../lib/markdown/serializer";
 import mathRule, { REGEX_INLINE_MATH_DOLLARS } from "../rules/math";
 import Node from "./Node";
+import { InputRule } from "prosemirror-inputrules";
+import { isInCode } from "../queries/isInCode";
 
 export default class Math extends Node {
   get name() {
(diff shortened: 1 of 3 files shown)


shared/editor/queries/isInCode.ts
```diff
@@ -20,31 +20,29 @@ type Options = {
  */
 export function isInCode(state: EditorState, options?: Options): boolean {
   const { nodes, marks } = state.schema;
+  const opts =
```
tommoor, line 23: Subtle perfmance and behavior difference avoids passing `{ inclusive: undefined }` here

shared/editor/queries/getMarksBetween.ts
```diff
@@ -17,14 +17,19 @@ export function getMarksBetween(
...
   state.doc.nodesBetween(start, end, (node, pos) => {
-    marks = [
-      ...marks,
-      ...node.marks.map((mark) => ({
-        start: pos,
-        end: pos + node.nodeSize,
-        mark,
-      })),
-    ];
+    if (node.isText) {
```
tommoor, line 20: Perf: Avoid mutations for nodes that will never have marks

shared/editor/nodes/Math.ts
```diff
@@ -35,10 +36,34 @@ export default class Math extends Node {
 
   inputRules({ schema }: { schema: Schema }) {
     return [
-      makeInlineMathInputRule(
-        REGEX_INLINE_MATH_DOLLARS,
-        schema.nodes.math_inline
-      ),
+      new InputRule(REGEX_INLINE_MATH_DOLLARS, (state, match, start, end) => {
+        if (isInCode(state)) {
```
tommoor, line 40: 🍖 added check

Issue #10171, opened 2025-09-14
Printing with tables broken


ddauer:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Open a page with some tables and some text before and after the tables. Hit CMD+P. 

Print cuts off tables horizontally and vertically. Also only one page is printed, even when table is longer and some text follows after table.

### Expected Behavior

- Tables are properly scaled to page width
- Compelte tables are printed
- Compelte page is printed

### Steps To Reproduce

1. Create table, text, long table whatever
2. Hit print
3. Look at print preview

### Environment

```markdown
- Outline: 0.87.3
- Browser: Firefox
```

nehaaprasad:
@ddauer and @tommoor 
I just ppushed a fix for the table printing issues tables should now scale properly and print across multiple pages without clipping.

ddauer:
Also emails that include table (previews) have similar table formatting issues.

nehaaprasad:
@ddauer 
issue was in documentImporter.ts where line 66 was removing the first H1 heading during import and I removed that line so headings now preserve their correct levels (H2 stays H2, H3 stays H3, etc.).

salihudickson:
There's a few things going on here:
- The behavior where only one page is printed and the table is cut off vertically seems to be a firefox specific issue. And it happens with all content not just tables.
- The table being cut off horizontally doesn't seem to be a formatting issue. It seems to be an unavoidable side effect of the browser trying to fit content that is wider than the page into a static document. Perhaps we can explore shrinking the table a bit when it makes sense, but at some level I think we just have to let the browser cut out some part of it.

@tommoor, what do you think?

tommoor:
That sounds right to me, I haven't checked out the PR yet – there are a bunch of unrelated changes on there so it's not mergeable.

PR #10184, opened 2025-09-15
fix: table printing issues with clipping and pagination

naaa760:
fixes: #10171

**Issues Fixed:**
- Tables now properly scale to page width instead of being clipped
- Removed horizontal scrolling behavior in print media
- Hidden interactive table elements (grips, shadows, controls) from print view
- Fixed page break handling to allow multi-page printing of long tables and content
- Improved text wrapping in table cells for better print layout

**Changes:**
- Added comprehensive print media CSS rules for tables in `Styles.ts`
- Ensured tables use full page width with `table-layout: auto`
- Proper page break management for table rows
- Enhanced cell content overflow handling

### DIFF
--- a/shared/editor/components/Styles.ts
+++ b/shared/editor/components/Styles.ts
@@ -1958,6 +1958,61 @@ del[data-operation-index] {
   blockquote {
     font-family: "SF Pro Text", ${props.theme.fontFamily};
   }
+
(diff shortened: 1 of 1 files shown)


salihudickson:
hey @naaa760, could you add a before and after screen recording of the fix that this PR implements.

Issue #10086, opened 2025-09-03
Allow embedding PDF


tommoor:
It should be a regular attachment node with an additional property .e.g `preview: true`

- It should be resizable in the same way as videos and images
- Restrict option to only pdf for now
- We need a new icon to represent this toggle.

salihudickson:
hey @tommoor, mind if i work on this?

tommoor:
Sure thing

seamon67:
This is hype!

ImaCrea:
G R E A T (!) This will be perfect to embed presentations for example.

hdoo42:
I already make kind of this feature for private use, but code is kind of dirty so that I didn't make a pull request. if you want to try, you can see my code here. https://github.com/42Gyeongsan/outline

<img width="1022" height="1262" alt="Image" src="https://github.com/user-attachments/assets/18f65e96-b398-4056-96cb-655c409ee9cc" />

<img width="954" height="200" alt="Image" src="https://github.com/user-attachments/assets/28f415c3-1539-4ee4-87bc-67d1a277ef6c" />

PR #10198, opened 2025-09-17
PDF embed

salihudickson:
This PR aims to embed an interactive PDF viewer in the Outline editor.

Fixes (#10086)

Demo:

https://github.com/user-attachments/assets/c5399105-3e0f-44c8-8036-d5ba20eb9993

### DIFF
--- a/shared/editor/nodes/Attachment.tsx
+++ b/shared/editor/nodes/Attachment.tsx
@@ -112,6 +112,7 @@ export default class Attachment extends Node {
     };
 
   component = (props: ComponentProps) => {
+    const { embedsDisabled } = this.editor.props;
     const { isSelected, isEditable, node } = props;
     const context = node.attrs.href ? (
       bytesToHumanReadable(node.attrs.size || "0")
@@ -122,6 +123,7 @@ export default class Attachment extends Node {
     );
 
     return node.attrs.preview &&
+      !embedsDisabled &&
       node.attrs.contentType === "application/pdf" ? (
       <PdfViewer
         icon={<FileExtension title={node.attrs.title} />}
--- a/shared/editor/components/DisabledEmbed.tsx
+++ b/shared/editor/components/DisabledEmbed.tsx
@@ -1,14 +1,11 @@
 import { OpenIcon } from "outline-icons";
-import * as React from "react";
-import { DefaultTheme, ThemeProps } from "styled-components";
 import { EmbedProps as Props } from "../embeds";
 import Widget from "./Widget";
 
 export default function DisabledEmbed(
-  props: Omit<Props, "matches" | "attrs"> &
-    ThemeProps<DefaultTheme> & {
-      href: string;
-    }
+  props: Omit<Props, "matches" | "attrs"> & {
+    href: string;
+  }
 ) {
   return (
     <Widget
@@ -17,7 +14,6 @@ export default function DisabledEmbed(
       icon={props.embed.icon}
       context={props.href.replace(/^https?:\/\//, "")}
       isSelected={props.isSelected}
-      theme={props.theme}
     >
       <OpenIcon size={20} />
     </Widget>
--- a/shared/editor/components/Embed.tsx
+++ b/shared/editor/components/Embed.tsx
@@ -67,7 +67,7 @@ const Embed = (props: Props) => {
 
 const InnerEmbed = React.forwardRef<HTMLIFrameElement, Props>(
   function InnerEmbed_(
-    { isEditable, isSelected, theme, node, embeds, embedsDisabled, style },
+    { isEditable, isSelected, node, embeds, embedsDisabled, style },
     ref
   ) {
     const cache = React.useMemo(
@@ -88,7 +88,6 @@ const InnerEmbed = React.forwardRef<HTMLIFrameElement, Props>(
           embed={embed}
           isEditable={isEditable}
           isSelected={isSelected}
-          theme={theme}
         />
       );
     }
@@ -120,7 +119,6 @@ const InnerEmbed = React.forwardRef<HTMLIFrameElement, Props>(
           isEditable={isEditable}
           isSelected={isSelected}
           embed={embed}
-          theme={theme}
         />
       );
     }


salihudickson:
hey @tommoor, I still have a bit of work to do on this. But I'd appreciate any feedback you could give me on my current implementation.

server/routes/api/attachments/attachments.ts
```diff
@@ -291,15 +291,29 @@ const handleAttachmentsRedirect = async (
...
+
+  if (
+    attachment.contentType === "application/pdf" &&
+    !!ctx.input.query.preview
+  ) {
+    const buffer = await FileStorage.getFileBuffer(attachment.key);
+    const base64String = buffer.toString("base64");
+
+    ctx.set("Content-Type", "application/json");
+    ctx.body = { data: base64String };
```
tommoor: I would prefer to avoid these changes on the API – serving potentially hundreds of mb's of data as base64 encoded is a problem. Why can we not make the signed url just get passed directly into `<object>`? I assume you were dealing with a problem/error

salihudickson: 😅 Honestly, I totally missed that the URL was right there, and that I could’ve just been sending that instead.

The performance issue was top of mind, but this was a quick rework of my original approach. At first, I was planning to use a custom PDF viewer to keep things consistent across browsers and devices. I scrapped that though, since it started to feel too heavy and messy for an initial version.

Previously, I was streaming the PDF directly from storage without loading it into memory. But when I reworked it, fetching the stream kept causing page redirects. As a temporary workaround, I decided to convert it to base64 on the backend.

Really appreciate you checking this out so quickly, you definitely saved me some time there🙏🏾

salihudickson:
hey @tommoor, I think this is just about ready for the first round of reviews.

tommoor:
Thanks! One aspect of the PR was actually closer before, this should _not_ be modeled as a new node type for a couple of reasons:

- New nodes are breaking changes we try to make as rarely as possible
- I want to enable easy conversion between attachments that are shown with a preview or not (PDF is just the first, potentially)

So it's better to stick with the existing node and just an attribute for whether preview is enabled or not

salihudickson:
That makes sense. I'll modify the PR to reflect this.

salihudickson:
hey @tommoor, is there any more work that needs to be done on this?

salihudickson (commented):
hey @tommoor, can i get another review on this when you have the time?

server/routes/api/attachments/attachments.ts
```diff
@@ -291,15 +291,26 @@ const handleAttachmentsRedirect = async (
     }
   );
 
-  if (attachment.isPrivate) {
-    ctx.set(
-      "Cache-Control",
-      `max-age=${BaseStorage.defaultSignedUrlExpires}, immutable`
-    );
-    ctx.redirect(await attachment.signedUrl);
+  const { url, maxAge } = attachment.isPrivate
```
tommoor: I don't understand why these changes are needed at all still, I updated the component to just pass the attachment URL into the iframe directly and it worked fine… there should be no backend changes needed.

app/hooks/useDictionary.ts
```diff
@@ -49,6 +49,7 @@ export default function useDictionary() {
       em: t("Italic"),
       embedInvalidLink: t("Sorry, that link won’t work for this embed type"),
       file: t("File attachment"),
+      pdf: t("Upload PDF"),
```
tommoor:
```suggestion
      pdf: t("Embed PDF"),
```

Perhaps more descriptive

server/routes/api/attachments/schema.ts
```diff
@@ -73,6 +73,7 @@ export const AttachmentsRedirectSchema = BaseSchema.extend({
   query: z.object({
     /** Id of the attachment to be deleted */
     id: z.string().uuid().optional(),
+    preview: z.string().optional(),
```
tommoor: @copilot 
```suggestion
```

shared/editor/nodes/Attachment.tsx
```diff
@@ -38,6 +39,23 @@ export default class Attachment extends Node {
...
+        width: {
+          default: 300,
+        },
+        height: {
+          default: 424,
+        },
+        type: {
+          default: null,
+          validate: "string|null",
+        },
```
tommoor:
```suggestion
        contentType: {
          default: null,
          validate: "string|null",
        },
```

Lets just store the contentType for all future attachments, feels easier

shared/editor/components/PDF.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [node.attrs.width]);
+
+  return (
+    <div
+      className={layoutClass ? `pdf pdf-${layoutClass}` : "pdf"}
+      contentEditable={false}
+    >
+      <iframe
```
tommoor: Using `<embed>` would be better and avoid a border.

shared/editor/components/PDF.tsx
```diff
@@ -0,0 +1,76 @@
...
+  );
+
+  useEffect(() => {
+    const url = href + (href.includes("?") ? "&" : "?") + "preview=true";
+    fetch(url)
+      .then((res) => res.json())
+      .then((res) => {
+        setData(res.url);
+      });
+  }, [href]);
```
tommoor: Not needed

shared/editor/components/PDF.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export default function PdfViewer(props: Props) {
+  const { node, isEditable, onChangeSize, isSelected } = props;
+  const { href, name, layoutClass } = node.attrs;
+  const [data, setData] = useState<string>();
+  const iframeRef = useRef<HTMLIFrameElement>(null);
+  const isFullWidth = layoutClass === "full-width";
+
+  const { width, height, setSize, handlePointerDown, dragging } = useDragResize(
```
tommoor, line 30: Something is going on with the resize logic, it's very hard to use – and keeps flickering about in ways that don't happen on images and videos

app/editor/menus/attachment.tsx
```diff
@@ -1,13 +1,80 @@
...
+  });
+  const isFullWidthAligned = isNodeActive(schema.nodes.attachment, {
+    layoutClass: "full-width",
+  });
+  const isPreview = isNodeActive(schema.nodes.attachment, {
+    preview: true,
+  });
+
   return [
+    {
```
tommoor, line 22: I would remove all these alignment options in a first version, resizing is enough – also means we don't need to bring across the legacy `layoutClass` stuff from images.

shared/editor/nodes/Attachment.tsx
```diff
@@ -170,13 +220,59 @@ export default class Attachment extends Node {
...
+          dispatch?.(transaction.setSelection(new NodeSelection($pos)));
+          return true;
+        },
     };
   }
 
   toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
     state.ensureNewLine();
+
+    const attrs = [
```
tommoor: No changes needed in this method – revert

shared/editor/nodes/Attachment.tsx
```diff
@@ -188,6 +284,10 @@ export default class Attachment extends Node {
         href: tok.attrGet("href"),
         title: tok.attrGet("title"),
         size: tok.attrGet("size"),
+        width: tok.attrGet("width"),
+        height: tok.attrGet("height"),
+        preview: tok.attrGet("preview"),
+        layoutClass: tok.attrGet("layoutClass"),
```
tommoor: Revert!

shared/editor/nodes/Image.tsx
```diff
@@ -433,11 +433,14 @@ export default class Image extends SimpleImage {
         if (!(state.selection instanceof NodeSelection)) {
           return false;
         }
+
+        const { node } = state.selection;
         const attrs = {
-          ...state.selection.node.attrs,
-          title: null,
+          ...node.attrs,
+          title: node.type.name === "attachment" ? node.attrs.title : null,
```
tommoor: We shouldn't have logic related to the `attachment` node inside the image node... 

As previous comment lets lose these alignment options for now anyway.

tommoor (changes requested):
It's still pretty hard to use to be honest, especially resizing. Are you seeing the issue where resizing causes it to jump around? The iframe has a giant ugly border on it, I would use the embed element instead to avoid this.

server/routes/api/attachments/attachments.ts
salihudickson: If we pass the URL directly into the iframe, we run into CORS, and I’d need to modify the CSP to allow embeds from anywhere, like we're doing for images. That could introduce some extra security concerns, which I was trying to avoid.

But, I’ll go ahead and pass the URL directly.

tommoor: > modify the CSP to allow embeds from anywhere

These files are hosted on our own storage though, so it's not _anywhere_

salihudickson: Good point. Should probably just allow URLs from whatever URL the storage is.

shared/editor/components/PDF.tsx
salihudickson: Just noticed that, on chrome. I initially only tested on firefox.

salihudickson: Hey @tommor, I’ve made the resize UX smoother in general. Firefox still behaves a bit better overall since it handles embeds differently. 

I've debounced the re-renders that caused most of the flickering on chrome, but it still flickers a bit because the embed re-centers on resize. I’m continuing to look into that, but give it a try and let me know if it’s good enough for now.

Still looking into further tweaks.

Issue #10145, opened 2025-09-10
Add ability to update URL of embedded images


baditaflorin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

<img width="1116" height="426" alt="Image" src="https://github.com/user-attachments/assets/7c3dc23c-0287-4e75-a44d-df1a8de5646c" />

# Bug Report: Unable to Update Embedded Images via gui or Code in Outline
Please add support for editing embedded image URLs directly. This will align Outline with common documentation practices and make it more compatible with technical workflows.

## Summary

When embedding dynamic images (such as Shields.io badges) in Outline documents, users cannot update or modify the image source directly via code. Instead, the editor only provides a **Replace image** option, which triggers a file upload dialog. This behavior breaks common use cases where the image URL should remain editable and point to a dynamic resource (e.g., badges, charts, or version indicators).

## Expected Behavior

* Users should be able to edit the image source directly from the document code (Markdown mode or link editor).
* The editor should recognize dynamic image URLs and allow updates without requiring file uploads.

## Actual Behavior

* The inserted image is treated as a static file.
* The only available option is **Replace image**, which opens a file selector for uploading a new file.
* There is no way to modify the URL of an already embedded image.

## Why This Matters

* Dynamic images (e.g., Shields.io, build status badges, GitHub stats) are extremely common in forums, READMEs, and documentation.
* Preventing users from editing the URL directly makes Outline unsuitable for technical documentation workflows.
* This forces users into manual workarounds or repeated deletions/re-uploads.

## Proposed Fix

* Allow editing of embedded image URLs in Markdown mode or through a link editor.
* Preserve the existing **Replace image** functionality for file uploads, but provide an **Edit URL** option for web-linked images.

## Environment

* Outline version: latest (as of bug filing)
* Browser: Chrome/Firefox/Safari (all affected)
* Platform: macOS/Linux/Windows

### Steps To Reproduce

1. Create a new document in Outline.
2. Insert an image using Markdown syntax:

   ```markdown
   ![Outline](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/USERNAME/shields-version-badges/main/outline-status.json)
   ```
3. Save the document.
4. Attempt to update the image URL (e.g., change `USERNAME` or point to another badge).

### Environment

```markdown
- Outline:
- Browser:
```

salihudickson:
[...]

PR #10258, opened 2025-09-26
ehancement: allow users edit image links

salihudickson:
## Description
This PR allows users to dynamically edit the src of embedded images. 

fixes: #10145 

## Video Demo:

https://github.com/user-attachments/assets/99858c55-d24c-42cf-b15e-3a0cef6c7019


tommoor:
Looks pretty reasonable, I think from a UI perspective we should not introduce another toolbar button though – instead we can combine the "Replace" option and have a menu, so that you can choose to replace by editing URL or uploading a new image?

All of the structure for this exists, we have menus from the toolbar for e.g. code selection, and highlight selection.

shared/editor/nodes/Image.tsx
```diff
@@ -475,6 +479,30 @@ export default class Image extends SimpleImage {
           dispatch?.(tr.setSelection(new NodeSelection($pos)));
           return true;
         },
+      editLink: (): Command => (state, dispatch) => {
```
tommoor:
```suggestion
      editImageUrl: (): Command => (state, dispatch) => {
```

These commands are "global" to the editor, so should be more specific

salihudickson: I was sort of thinking we can just use this one command if this feature needs to be applied to other types of embedded media in the future. But I guess we can just abstract this into a stand alone command later, if we need to.

tommoor: I would keep it specific –  it's often a different attribute anyway, e.g. `url`, `src`, `href`

tommoor (changes requested):
Lets just combine edit and replace under a single icon with two menu options and then this is good!

salihudickson:
hey @tommoor, I've combined the edit and replace, like you asked. The video demo in the description has been updated to reflect this.

tommoor:
I think you need to push up the changes here

salihudickson:
😅 sorry, didn't realize that I hadn't pushed the changes.

app/editor/components/EmbededMediaLinkEditor.tsx
```diff
@@ -20,54 +16,43 @@ type Props = {
   dictionary: Dictionary;
 };
 
-export function EmbedLinkEditor({ node, view, dictionary }: Props) {
-  const { t } = useTranslation();
-  const embeds = useEmbeds();
-
-  const url = node.attrs.href as string;
+export function EmbededMediaLinkEditor({ node, view, dictionary }: Props) {
```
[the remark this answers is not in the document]
salihudickson: this is not consequential

app/editor/components/EmbeddedMediaLinkEditor.tsx
```diff
@@ -89,26 +74,56 @@ export function EmbedLinkEditor({ node, view, dictionary }: Props) {
...
           <OpenIcon />
         </ToolbarButton>
       </Tooltip>
       {view.editable && (
-        <Tooltip content={dictionary.deleteEmbed}>
-          <ToolbarButton onClick={removeEmbed}>
+        <Tooltip
+          content={
+            node.type.name === "image"
+              ? dictionary.deleteImage
```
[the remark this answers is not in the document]
salihudickson: I have confirmed that the deleteImage key is defined in the `useDictionary.ts` file.

shared/editor/nodes/Image.tsx
```diff
@@ -114,6 +114,10 @@ export default class Image extends SimpleImage {
           default: null,
           validate: "string|null",
         },
+        editing: {
```
tommoor: I just noticed this - Unfortunately we should not store this state in the node attributes – it will become part of the document, potentially getting stuck in an editing state and will also sync to everyone else viewing the doc.

Generally I would try and keep this type of state inside the toolbar itself

salihudickson: done 👍🏾

app/editor/components/MediaLinkEditor.tsx
```diff
@@ -20,54 +16,43 @@ type Props = {
...
     );
-    const selection = nextSelection ?? TextSelection.create(state.tr.doc, 0);
 
+    const selection = nextSelection ?? TextSelection.create(state.tr.doc, 0);
     dispatch(state.tr.setSelection(selection));
     view.focus();
   }, [view]);
 
-  const openEmbed = useCallback(() => {
+  const openLink = useCallback(() => {
```
tommoor, line 36: I'm not seeing open link or remove work correctly on the link editor, are you? 

Honestly they aren't really needed in that context though.

app/editor/components/SelectionToolbar.tsx
```diff
@@ -270,15 +276,21 @@ export default function SelectionToolbar(props: Props) {
           onClickLink={props.onClickLink}
           onSelectLink={handleOnSelectLink}
         />
-      ) : isEmbedSelection ? (
-        <EmbedLinkEditor
+      ) : isEditingMedia ? (
```
tommoor, line 235: I'd like to figure out a way that it can inherit the existing width so that the toolbar doesn't jump position when switching modes, if you have ideas

salihudickson: currently the width is set by the content inside it. Perhaps it might be simpler to just pick a static width and use that?

app/editor/components/MediaLinkEditor.tsx
salihudickson: yeah, didn't really think it through before, but can just have the input box when for the image.

app/editor/components/SelectionToolbar.tsx
tommoor: Probably the pragmatic way forward is to set it to the same width

salihudickson: check out my current implementation.

tommoor (approved):

Issue #10261, opened 2025-09-28
Downloading images with CORS error


Ali-ovo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

When I download the image in the link I shared, I report an error CORS

### Expected Behavior

I hope I can download the pictures properly

### Steps To Reproduce

https://test1102.getoutline.com/s/6099d629-2a71-41c0-bc3d-0351199296a0

 https://github.com/user-attachments/assets/6cc6bb6c-c2e2-4cbb-8664-a6e4b8881850

### Environment

```markdown
- Outline: 0.87.4
- Browser: Chrome
```

Ali-ovo:
Interesting find, when I open the network's `Disable cache`, the image can be downloaded normally

https://github.com/user-attachments/assets/49bcfa13-f403-4812-93fa-418e95e18fb4

nehaaprasad:
Hi! @Ali-ovo 
I fixed it by adding the necessary CORS headers to the attachment redirect responses, so images should now download properly without any browser blocks

let me know

salihudickson:
Hey @Ali-ovo,

This issue isn’t actually related to the Outline backend, since this image isn’t being served from there. It’s more likely caused by how Google Chrome handles cached responses.

In short, when Chrome first requests the image from the S3 bucket to display it, it makes a non-CORS request. Because of that, the browser doesn’t include any Access-Control headers, and S3 returns a response without them. Chrome then caches that version and reuses it for future requests.

When you later try to download the image, Chrome reuses the cached response, but this time a CORS header is required. Since it’s missing, the browser throws a CORS error.

Disabling the cache forces Chrome to fetch a fresh copy (with the right headers), which is why it works then. Firefox doesn’t have this problem, so you won't have this same issue there.

The proper fix is to ensure the image is always served with the appropriate Access-Control headers needed for downloads. A good long-term solution would be to serve images through Amazon CloudFront, which gives you more control over caching and headers.

@tommoor, you might want to weigh in, incase I've missed something.

tommoor:
Ah, if it's served by the service worker cache then it could be just that the headers need to be mirrored correctly 🤔

Ali-ovo:
> Hey [@Ali-ovo](https://github.com/Ali-ovo),
> 
[...]

PR #10269, opened 2025-09-29
fix: CORS headers for image downloads

nehaaprasad:
### PR Description:

**Problem:** 
- Users were experiencing CORS errors when downloading images from Outline. The error occurred because redirect responses for attachments were missing proper CORS headers, causing browsers to block cross-origin requests.

fixes: #10261

**Solution:**
- Added CORS headers to attachment redirect responses in two key endpoints:
- `server/routes/api/attachments/attachments.ts` - Fixed `handleAttachmentsRedirect` function
- `server/routes/api/fileOperations/fileOperations.ts` - Fixed `handleFileOperationsRedirect` function

**Changes:**
- Added `Access-Control-Allow-Origin: *` to allow cross-origin requests
- Added `Access-Control-Allow-Methods: GET, OPTIONS` for proper HTTP method handling  
- Added `Access-Control-Allow-Headers: Content-Type, Authorization` for required headers

**Result:**
Images now download properly without CORS errors. This resolves the issue where disabling browser cache temporarily fixed the problem, confirming that the issue was specifically with CORS headers on redirect responses.

**Files Modified:**
- `server/routes/api/attachments/attachments.ts`
- `server/routes/api/fileOperations/fileOperations.ts`

### DIFF
--- a/server/routes/api/fileOperations/fileOperations.ts
+++ b/server/routes/api/fileOperations/fileOperations.ts
@@ -87,6 +87,9 @@ const handleFileOperationsRedirect = async (
   }
 
   const accessUrl = await FileStorage.getSignedUrl(fileOperation.key);
+  ctx.set("Access-Control-Allow-Origin", "*");
+  ctx.set("Access-Control-Allow-Methods", "GET, OPTIONS");
+  ctx.set("Access-Control-Allow-Headers", "Content-Type, Authorization");
   ctx.redirect(accessUrl);
 };
 


salihudickson (commented):
hey @naaa760, i think there are a few things wrong here;

- The download request is not sent to the outline backend, it is sent to the storage.
- The request that is sent here to generate the image blob has the same origin so the Acess control header won't change anything.
- Setting the access control header to allow any origin might be a security risk that isn't needed and doesn't actually solve the problem.

tommoor:
In the issue the request isn't even to Outline storage – it's a third party image url, I'm not sure we can trigger a download for that at all, might be best to just remove the button in that case.

PR #10279, opened 2025-09-30
fix CORS err on img download

salihudickson:
## Description
This PR address the chrome specific issue where downloading an image can lead to a CORS error if the image isn't properly cached.

It does this by catching the error and rerunning the download action with an added instruction to force an update on the image cache.

It also adds a fallback to open the image on a separate tab if the image cannot be downloaded for some reason or another.

fixes: #10261
 
## Video Demo:
Before: 

https://github.com/user-attachments/assets/59f0e545-bf00-48e6-ad5c-d5ee1e40d7e8

After: 

https://github.com/user-attachments/assets/97720566-cd41-4048-bd4d-89633f58f88e

### DIFF
--- a/shared/editor/components/Image.tsx
+++ b/shared/editor/components/Image.tsx
@@ -21,10 +21,12 @@ type Props = ComponentProps & {
   /** The editor view */
   view: EditorView;
   children?: React.ReactElement;
+  isDownloading?: boolean;
 };
 
 const Image = (props: Props) => {
-  const { isSelected, node, isEditable, onChangeSize, onClick } = props;
+  const { isSelected, node, isEditable, onChangeSize, onClick, isDownloading } =
+    props;
   const { src, layoutClass } = node.attrs;
   const { t } = useTranslation();
   const className = layoutClass ? `image image-${layoutClass}` : "image";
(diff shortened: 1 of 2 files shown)


tommoor:
I'm wondering if we shouldn't just simplify to set `cache: reload` always. In your screencast it appears that multiple download animations happen, do you know what that is?

salihudickson:
@tommoor, 

> In your screencast it appears that multiple download animations happen, do you know what that is?

In the screencast I clicked the download button a second time, immediately the first download was completed. I wanted to show that the second download would be quicker since the image would now be cached correctly.

> I'm wondering if we shouldn't just simplify to set cache: reload always.

I chose this approach since the always reload option would hurt performance, especially on slower connections. The issue only occurs when the cached image is missing access control headers, usually right after upload or when the cache is cleared and the first request comes from an  tag, and only on chrome.

Since that’s probably not the majority scenario and forcing reload would always skip valid cache entries, letting the browser handle caching by default seems more efficient overall,. The tradeoff is an occasional slower download on Chrome (and maybe other Chromium browsers), but that seems minor compared to the performance hit of disabling caching entirely.

salihudickson:
But taking another look at this, I see how the fact that the first download might be a bit slower if it fails and has to retry with `cache: reload` and this can cause users to click multiple times and get multiple downloads. So I'll add an `isDownloading` flag to check for that.

tommoor:
Check is good, lets just disable the button too now we have the state to pass through?

salihudickson:
sure!

tommoor (approved):

PR #10283, opened 2025-10-01
Add option to choose default TOC visibility on public shares

tommoor:
Artisinally hand-coded on an international airliner


app/scenes/Document/components/Contents.tsx
```diff
@@ -56,7 +56,7 @@ function Contents() {
 
   return (
     <StickyWrapper>
-      <Heading>{t("Contents")}</Heading>
+      <Heading>{t("On this page")}</Heading>
```
Frazl: Should Contents be removed from `app\menus\TableOfContentsMenu.tsx` too? And if so, then there will be no occurence of "Contents" in the i8n file so it could possibly be removed then too.

Hope you had a nice vacation! 🏖️

tommoor: I'll revert this as it's unrelated to the changes

apoorv-mishra:
> Artisinally hand-coded on an international airliner

Wing, rudder, fuselage, engine,...??? And they didn't fine you?

PR #10315, opened 2025-10-06
fix: change list children to match list parent when list style changes

### LINKED ISSUE #9382: Changing list type only works at top level

### Current Behavior

When converting a numbered list that contains nested items into a bulleted list, only the top-level items change. The nested items remain numbered and must be converted individually.

### Expected Behavior

Selecting the entire list, including nested items, and switching to bullets should convert every item to a bullet "style" in one step.

### Steps To Reproduce

Image

### Environment

```markdown
- Outline: Latest hosted cloud version as of 6/2/25
- Browser: Safari 18.5
```

### Anything else?

_No response_

salihudickson:
## Description
This PR aims to fix the issue where changing a list type only applies to the list items at the top level and not the nested lists.

fixes #9382 

## Video Demo
https://github.com/user-attachments/assets/1e2a714d-102a-4559-96ad-2f438b304001

### DIFF
--- a/shared/editor/commands/toggleList.ts
+++ b/shared/editor/commands/toggleList.ts
@@ -39,7 +39,7 @@ export default function toggleList(
       const currentItemType = parentList.node.content.firstChild?.type;
       const differentType = currentItemType && currentItemType !== itemType;
 
-      if (differentType || differentListStyle) {
+      if (differentType) {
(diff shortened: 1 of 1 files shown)


tommoor (approved):
Nice, definitely better behavior than today

PR #10318, opened 2025-10-07
fix: replace uuid package with standard module

AlexHladin:
This PR replaces the uuid package with standard crypto module.
* [Web documentation](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)
* [Node.js documentation](https://nodejs.org/api/crypto.html#cryptorandomuuidoptions)


tommoor:
Hi @AlexHladin – thanks, you'll need to sign the CLA here, anything specifically that triggered this?

AlexHladin:
Hi @tommoor,
The CLA is signed.

PR #10327, opened 2025-10-09
Mention chip for regular URLs

apoorv-mishra:
> Add the "Mention" chip display for regular unfurls – we have it for GitHub and Linear links, but we can make it an option for any unfurl.

### DIFF
--- a/server/presenters/unfurl.ts
+++ b/server/presenters/unfurl.ts
@@ -30,8 +30,8 @@ const presentURL = (
   url: data.url,
   title: data.meta.title,
   description: data.meta.description,
-  thumbnailUrl: data.links.thumbnail[0].href,
-  faviconUrl: data.links.icon[0].href,
+  thumbnailUrl: (data.links.thumbnail ?? [])[0]?.href ?? "",
+  faviconUrl: (data.links.icon ?? [])[0]?.href ?? "",
 });
 
 const presentMention = async (
--- a/shared/editor/components/Mentions.tsx
+++ b/shared/editor/components/Mentions.tsx
@@ -28,6 +28,7 @@ import {
 } from "../../types";
 import { cn } from "../styles/utils";
 import { ComponentProps } from "../types";
+import { sanitizeUrl } from "@shared/utils/urls";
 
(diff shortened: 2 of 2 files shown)


apoorv-mishra:
- [x] Favicon for URL provider

shared/editor/nodes/Mention.tsx
```diff
@@ -47,7 +48,7 @@ export default class Mention extends Node {
     return {
       attrs: {
         type: {
-          default: MentionType.User,
+          default: undefined,
```
tommoor: We need to keep this for backwards compat if possible. Just add a new enum value for URL

apoorv-mishra: Sure, thanks for the spot!

server/presenters/unfurl.ts
```diff
@@ -19,18 +19,19 @@ async function presentUnfurl(
     case UnfurlResourceType.Issue:
       return presentIssue(data);
     default:
-      return presentOEmbed(data);
+      return presentURL(data);
   }
 }
 
-const presentOEmbed = (
+const presentURL = (
```
tommoor, line 26: Mostly works well, but these links are not always present…

image

shared/editor/components/Mentions.tsx
```diff
@@ -190,10 +191,9 @@ export const MentionURL = (props: ComponentProps) => {
       rel="noopener noreferrer nofollow"
     >
       <Flex align="center" gap={6}>
-        <img
-          src={unfurl.faviconUrl}
-          style={{ width: "16px", height: "16px" }}
-        />
+        {unfurl.faviconUrl ? (
+          <Logo src={sanitizeUrl(unfurl.faviconUrl)} alt="" />
```
tommoor, line 195: I guess it doesn't hurt, but sanitize not necessary when rendering through React, not sure if there was anything in particular that made you think to add it?

apoorv-mishra: Just out of general concern, no particular reason as such.

Issue #10307, opened 2025-10-06
Allow @group mentions


tommoor:
### Discussed in https://github.com/outline/outline/discussions/8527

<div type='discussions-op-text'>

<sup>Originally posted by **Jeeltilva** February 21, 2025</sup>
Hi All, Can I request a feature where you can do `@` and mention a group similar to how you mention a member? It should send notification to members part of that group. </div>

yashgoyal0110:
Hey @tommoor 
Assign please!

PR #10331, opened 2025-10-10
feat: add group mentions

salihudickson:
## Description
This PR adds the ability for users to mention entire groups in a document. This will also notify users in the group that they have been mentioned in a document.

fixes: #10307 

### Mentioning:

https://github.com/user-attachments/assets/7bf10631-7131-4c1a-bc8b-6690d6f6662c

### Notification:

https://github.com/user-attachments/assets/b4960e8d-3643-47da-88fe-8291622b9b3d

### DIFF
--- a/shared/components/Squircle.tsx
+++ b/shared/components/Squircle.tsx
@@ -31,7 +31,13 @@ const Squircle: React.FC<Props> = ({
     className={className}
     style={style}
   >
-    <svg width={size} height={size} fill={color} viewBox="0 0 28 28">
+    <svg
+      width={size}
+      height={size}
+      fill={color}
+      viewBox="0 0 28 28"
+      data-fixed-color
+    >
       <path d="M0 11.1776C0 1.97285 1.97285 0 11.1776 0H16.8224C26.0272 0 28 1.97285 28 11.1776V16.8224C28 26.0272 26.0272 28 16.8224 28H11.1776C1.97285 28 0 26.0272 0 16.8224V11.1776Z" />
     </svg>
     <Content>{children}</Content>


salihudickson:
hey @tommoor, I still have a bit of work to do on this, but could you take a look and let me know what you think of my approach.

tommoor:
Approach looks solid. It's probably worth a new notification type to distinguish between a personal and a group mention

salihudickson:
Thanks.

I also wanted to ask, is there a particular reason that there isn't consistent component or e2e testing across the client-side. Or is it just something you haven't gotten to yet?

tommoor:
Mostly I don't care for it, more hassle than it's worth. Open to suggestions though

tommoor:
Note https://github.com/outline/outline/pull/10318 just merged, so will affect use of `v4()` in this PR

salihudickson:
> Mostly I don't care for it, more hassle than it's worth. Open to suggestions though.

🤔 often i find that the biggest benefit of these kinds of tests is really just making sure we don't accidentally break something important. Since you've gotten this far without them, if you don't feel they're necessary, I'm inclined to agree.

That said, would you be up for opening a discussion about adding some high-level E2E tests for only top-level core functionality? This would let us zero in on areas where the benefit clearly outweighs the maintenance cost. We wouldn't be trying to hit full test coverage, just making sure we don't have to manually check that core features are still working after multi-module changes or refactoring.

salihudickson:
hey @tommoor, I think this is ready for another round of reviews.

server/queues/tasks/CommentCreatedNotificationsTask.ts
```diff
@@ -77,6 +83,58 @@ export default class CommentCreatedNotificationsTask extends BaseTask<CommentEve
       }
     }
 
+    // send notifications to users in mentioned groups
```
tommoor, line 86: Approving for v1 but I feel like we should split this into it's own task for resiliency, especially as groups can have many members.

tommoor (approved):

Issue #10330, opened 2025-10-10
The left sidebar does not update after a new document is created


Ali-ovo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

When I create a new document under a collection that someone has `shared with me`, the document on the left side doesn't update and has to wait for me to refresh the browser page before it will update.

### Expected Behavior

Expected to update the left sidebar properly

### Steps To Reproduce

https://github.com/user-attachments/assets/fff07f27-e758-476e-896c-d1304b614c6b

### Environment

```markdown
- Outline: 0.87.4
- Browser: Chrome
```

### Anything else?

form here https://github.com/outline/outline/issues/10328#issuecomment-3388084543

Ali-ovo:
Also, the left sidebar doesn't refresh when I delete a document

pushtisonawala:
Hi! 👋
I'd like to take this issue as part of Hacktoberfest.
It looks really interesting and I’d love to contribute to it.
Could you please assign it to me?

Thanks in advance! 🙌

Ali-ovo:
> Hi! 👋 I'd like to take this issue as part of Hacktoberfest. It looks really interesting and I’d love to contribute to it. Could you please assign it to me?
> 
> Thanks in advance! 🙌

You should be looking for this guy. @tommoor

Ali-ovo:
@tommoo

PR #10356, opened 2025-10-13
fix: Sidebar doesn't refresh when creating or deleting shared documents

Ali-ovo:
fixed: https://github.com/outline/outline/issues/10330


Ali-ovo:
I think this may not be the best way to fix it, but for now he works fine

Let me know if there's a better way to do this. @tommoor

PR #10370, opened 2025-10-13
fix: Display fallback instead of error if cannot unfurl URL

tommoor:


apoorv-mishra:
Screenshot 2025-10-14 at 9 16 28 PM

Link icon appearing feeble, becomes almost invisible when hovered.

tommoor:
Yea, not great in dark mode – maybe I'll just give it white background here as it's not possible to be dynamic while in an `img` tag

PR #10379, opened 2025-10-14
Added plantuml embedding

patroldo:
Added plantuml embedding

Link to test
https://editor.plantuml.com/uml/SoWkIImgAStDuNBCoKnELT2rKt3AJx9IS2mjoKZDAybCJYp9pCzJ24ejB4qjBk42oYde0jM05MDHLLoGdrUSoeLkM5u-K5sHGY9sGo6ARNHr2QY66kwGcfS2SZ00


shared/editor/components/Frame.tsx
```diff
@@ -101,6 +102,11 @@ const Frame = ({
               <OpenIcon size={18} /> Open
             </Open>
           )}
+          {editUrl && (
```
tommoor: I wouldn't introduce an additional prop here, `Open` can easily go to the edit url?

shared/editor/embeds/PlantUml.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const title = props.attrs.href.split("/uml/")[1];
+  var finalUrl = "https://img.plantuml.biz/plantuml/svg/" + title;
+  return (
+    <Frame
+      {...props}
+      src={finalUrl}
+      icon={
+        <Image
+          src="/images/plantuml.png"
+          alt="Diagrams.net"
```
tommoor: Incorrect alt

shared/editor/embeds/PlantUml.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+import Frame from "../components/Frame";
+import Image from "../components/Img";
+import { EmbedProps as Props } from ".";
+
+function PlantUmlDiagrams({ matches, ...props }: Props) {
+  const title = props.attrs.href.split("/uml/")[1];
+  var finalUrl = "https://img.plantuml.biz/plantuml/svg/" + title;
```
tommoor: What is this other domain, how does it relate to `plantuml.com`?

shared/editor/components/Frame.tsx
patroldo: the idea here is that it shall be possible quite easy to open svg image, like make it bigger. Maybe it's possible to implement by clicking on the iframe itself, i will check

shared/editor/embeds/PlantUml.tsx
patroldo: ye, you're right. Plantuml is exposing link to render svg. I used wrong one from their GUI

patroldo:
Hey, @tommoor 

There is another approach, we can use library which interpreter text of and drawing diagram from that instead of relying onto plantuml service. Which approach you would like to choose?

The idea is following:
When you inserting plantUml in markdown, there is a button, which changes text to drawing, or drawing to text. As guideline we can use following
https://plantuml.com/code-javascript-asynchronous

What you think about it?

shared/editor/embeds/PlantUml.tsx
tommoor: So there is an alternative using .com? I'm not clear

tommoor:
> When you inserting plantUml in markdown, there is a button, which changes text to drawing, or drawing to text. As guideline we can use following

I don't want to bring in more custom JS for this, we already have multiple diagraming methods. This is how OSS projects become a mess. I can live with an embed though

shared/editor/components/Frame.tsx
patroldo: ok. I removed editurl. By clicking "open" it is opening image. And link which is inserted into embedding will lead to edit mode of plantuml

shared/editor/embeds/PlantUml.tsx
patroldo: Fixed

shared/editor/embeds/PlantUml.tsx
patroldo: I've corrected a link in new commit

tommoor (approved):

patroldo:
Hi, @tommoor 

Thanks for merging. I checked contribution docs, but didn't find(or maybe missed it) - i think it would make sense for people to give a clue how to use plantuml embedding. Could you please point out what is the best place to describe it in documentation? Or it's quite self-explanatory and thus is not require any explanation?

BR
Yurii

whuwangyong:
> Hi, @tommoor
> 
> Thanks for merging. I checked contribution docs, but didn't find(or maybe missed it) - i think it would make sense for people to give a clue how to use plantuml embedding. Could you please point out what is the best place to describe it in documentation? Or it's quite self-explanatory and thus is not require any explanation?
> 
> BR Yurii

You're right. I just don't know how to use it. Can I embed a link from self-hosted plantuml server?

strangerinusall:
@tommoor, does this support local PlantUML server as well, or only links from https://editor.plantuml.com supported? Would like to know this before I deploy my own server. Thanks!

tommoor:
Looks like it works only with the cloud version right now

patroldo:
Hi. I will investigate and try options with self hosted plantuml

@tommoor - are you ok if for this embedding i will add 1(or in worst case 2) env variables in order to provide self-hosting options?

tommoor:
We don't do this through env anymore – it should be an integration of type `IntegrationType.Embed`, so you'd get a page in Settings to input the url.

patroldo:
> We don't do this through env anymore – it should be an integration of type `IntegrationType.Embed`, so you'd get a page in Settings to input the url.

Got it, thanks. This or next week i will prepare PR and see where it will go

arnaudroques:
Hi @tommoor, @patroldo, I am the PlantUML maintainer, thanks for shipping this and sorry for the drive-by.

I read through the thread, including the "no more custom JS" call and the direction toward `IntegrationType.Embed` with a configurable URL for self-hosting. Not here to relitigate that, just wanted to flag something that might make the self-hosted side of that easier without adding anything to Outline's own JS footprint.

PlantUML now publishes a pure-JS build of its rendering engine (no Java, no server process) as a small static page that renders a diagram straight from its URL, same shape as `img.plantuml.biz`, just self-hostable as a plain static file (or usable as-is, hosted by us, for free): https://plantuml.github.io/plantuml/js-plantuml/index.html

So whenever that Settings URL field lands, "self-hosted PlantUML" could mean "point at this" rather than "stand up a Java service", still just a URL Outline embeds, nothing custom to maintain on your side. Happy to adjust the URL format on our end if that'd make it a cleaner fit for whatever you land on.

PR #10398, opened 2025-10-16
refactor: remove chalk module

AnastasiyaHladina:
This PR replaces the chalk module with [util.styleText](https://nodejs.org/docs/latest-v22.x/api/util.html#utilstyletextformat-text-options).

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,6 @@
     "babel-plugin-transform-class-properties": "^6.24.1",
     "body-scroll-lock": "^4.0.0-beta.0",
     "bull": "^4.16.5",
-    "chalk": "^4.1.0",
     "class-validator": "^0.14.2",
     "command-score": "^0.1.2",
     "compressorjs": "^1.2.1",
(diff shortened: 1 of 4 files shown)


tommoor:
Looks like `styleText` was only added in Node `v20.12.0`, so we'd need to make that the minimum version

AnastasiyaHladina:
> Looks like `styleText` was only added in Node `v20.12.0`, so we'd need to make that the minimum version

Yes, the feature was introduced in Node.js v20.12.0 and stabilized in Node.js v22.17.0. As far as I see, the server application already uses [version 22](https://github.com/outline/outline/blob/main/Dockerfile#L9)

tommoor:
> As far as I see, the server application already uses [version 22](https://github.com/outline/outline/blob/main/Dockerfile#L9)

Right, but we need to make sure that the `engines` field never resolves to less than that patch version

tommoor:
I'd probably try rebase against `main` and see if the tests pass

tommoor:
Let me know if you plan on taking this across the finish line

PR #10400, opened 2025-10-17
fix: Clicking around image should close lightbox

tommoor:
Makes it so that the area around the image shows a "zoom-out" icon and clicking closes the lightbox, I've found myself missing this interaction since moving to the new behavior.

### DIFF
--- a/app/components/Lightbox.tsx
+++ b/app/components/Lightbox.tsx
@@ -49,6 +49,7 @@ import {
   ReactZoomPanPinchRef,
 } from "react-zoom-pan-pinch";
 import { transparentize } from "polished";
+import { mergeRefs } from "react-merge-refs";
 
 export enum LightboxStatus {
   READY_TO_OPEN,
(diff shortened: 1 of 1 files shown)


apoorv-mishra:
IMO, we should restrict this behavior to the zoomed out state of image. I just observed Lightbox closing because I accidentally clicked _outside_ the zoomed in image while panning.

tommoor:
Good call, updated – I also made it so that the draggable cursor will show in this area to indicate that you can pan from there, and it will change to the "grabbing" cursor correctly now.

app/components/Lightbox.tsx
```diff
@@ -138,10 +174,7 @@ function usePanning() {
 
   const onPanningStart: ComponentProps<
     typeof TransformWrapper
-  >["onPanningStart"] = (ref, event) => {
-    if (!(event.target instanceof HTMLImageElement)) {
```
tommoor, line 142: This check was preventing `isPanning` from getting set correctly when you start the pan operation from outside the image bounds. The library already only triggers this callback when it's actually the case, I can't find anything else this guard was protecting 🤔

apoorv-mishra: IIRC, I thought of not showing the "grab" cursor outside of image. But I think that wasn't well justified since you could still pan by click-and-drag outside. I feel we can safely lose this.

app/components/Lightbox.tsx
```diff
@@ -998,7 +1029,7 @@ const StyledImg = styled.img<{
   object-fit: contain;
   cursor: ${(props) =>
     props.$panning
-      ? "grab"
+      ? "grabbing"
```
apoorv-mishra, line 1032: ~Oh, this resembles "closed fist". I feel `grab` was better(image wiping your monitor screen). Not too picky with this though, just a personal choice.~ Actually feels nicer after I tried it, so keep it 😄

PR #10403, opened 2025-10-17
chore: update minimall Node.js version

AnastasiyaHladina:
This PR updates the minimal `Node.js` engine in package.json to `22`. This PR is part of #10398.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "url": "https://github.com/sponsors/outline"
   },
   "engines": {
-    "node": ">=20.12 || 22"
+    "node": ">=20.12 <21 || 22"
   },
   "repository": {
     "type": "git",


AnastasiyaHladina:
> There is no need to remove 20 entirely

Ok, reverted.

tommoor:
The feature was introduced in `v20.12.0` so that should be the minimum version, I'd like to keep support for the v20 line as it is maintenance LTS still

AnastasiyaHladina:
> The feature was introduced in `v20.12.0` so that should be the minimum version, I'd like to keep support for the v20 line as it is maintenance LTS still

Thanks for noticing it. I've updated the PR.

AnastasiyaHladina:
> Nope, this still allows Node 21 which is unfortunately broken with the Outline codebase.

Thanks for catching! I've updated the PR, now it forbids Node.js 21.

PR #10462, opened 2025-10-23
fix: prevent list conversion inside heading nodes

huiseo:
## Description
Fixes a bug where typing list syntax (e.g., `1. `, `* `, `[ ]`) inside heading nodes would incorrectly trigger list conversion.

## Problem
When a user selected H1 from the "/" menu and typed "1. " followed by a space, the OrderedList inputRule would attempt to convert the heading into an ordered list, causing a conflict since headings can only contain inline content.

## Solution
- Add `isInHeading` utility to detect if selection is inside a heading
- Create `safeWrappingInputRule` wrapper that prevents list conversion when inside heading nodes
- Apply the fix to OrderedList, BulletList, and CheckboxList nodes

## Testing
1. Open editor
2. Type "/" and select "H1 Heading"
3. Type "1. " and press space
4. Expected: Text "1. " is preserved
5. Before fix: Attempted to convert to ordered list

This ensures that list markdown syntax is preserved as plain text when typed within headings, matching expected editor behavior.

## Files Changed
- `shared/editor/queries/isInHeading.ts` (new)
- `shared/editor/nodes/OrderedList.ts`
- `shared/editor/nodes/BulletList.ts`
- `shared/editor/nodes/CheckboxList.ts`

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,0 +1,13 @@
+URL=https://local.outline.dev:3000
+
+DATABASE_URL=postgres://user:pass@127.0.0.1:5432/outline
+REDIS_URL=redis://127.0.0.1:6379
+
+SMTP_FROM_EMAIL=hello@example.com
+
+# Enable unsafe-inline in script-src CSP directive
+# Setting it to true allows React dev tools add-on in Firefox to successfully detect the project
+DEVELOPMENT_UNSAFE_INLINE_CSP=true
+
+# Increase the log level to debug for development
+LOG_LEVEL=debug
--- a/shared/editor/nodes/BulletList.ts
+++ b/shared/editor/nodes/BulletList.ts
@@ -1,40 +1,13 @@
-import { wrappingInputRule, InputRule } from "prosemirror-inputrules";
 import {
   Schema,
   NodeType,
   NodeSpec,
   Node as ProsemirrorModel,
-  Attrs,
 } from "prosemirror-model";
 import toggleList from "../commands/toggleList";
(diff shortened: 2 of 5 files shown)


.env.development
```diff
@@ -1,13 +0,0 @@
-URL=https://local.outline.dev:3000
```
tommoor, line 1: Restore this file

shared/editor/nodes/BulletList.ts
```diff
@@ -1,14 +1,41 @@
...
 import toggleList from "../commands/toggleList";
 import { MarkdownSerializerState } from "../lib/markdown/serializer";
+import { isInHeading } from "../queries/isInHeading";
 import Node from "./Node";
 
+/**
+ * A wrapper for wrappingInputRule that prevents execution inside heading nodes.
+ * This fixes the bug where typing "* " or "- " inside a heading would trigger bullet list conversion.
+ */
+function safeWrappingInputRule(
```
tommoor: I would refactor this method to `listWrappingInputRule` and move it so there is only a single implementation – too much repetition

.env.development
huiseo: Restored .env.development as requested. No content changes.

shared/editor/nodes/BulletList.ts
huiseo: Refactored to a single listWrappingInputRule helper and updated Ordered/Bullet/Checkbox list nodes. Please take another look.

huiseo:
Thanks for merging!

PR #10468, opened 2025-10-24
Fix/persian translation

professor-1101:
Enhanced Farsi Translation + i18n consistency test case script


tommoor:
Please see the [docs for submitting translations](https://github.com/outline/outline/blob/main/docs/TRANSLATION.md) – the files cannot be edited directly in the codebase like this as they will be overwritten.

PR #10469, opened 2025-10-24
custom emoji

salihudickson:
## Description
This PR allows users to create and use custom emojis.

## Video Demo
https://github.com/user-attachments/assets/20550341-c6ad-4053-a68e-4d0cbf184d60


salihudickson:
hey @tommoor, I still have a bit of work to do on this. But could you check this out so far and let me know if there are any modifications I need to make.

I'm also thinking of splitting the PR into three smaller ones to make it easier to review and to keep each focused on a specific part. The first would handle emoji creation, the second would cover usage in collection and document titles, and the third would implement it in the editor. What do you think?

tommoor:
> But could you check this out so far and let me know if there are any modifications I need to make.

Will do in the next 24h

> I'm also thinking of splitting the PR into three smaller ones to make it easier to review and to keep each focused on a specific part. The first would handle emoji creation, the second would cover usage in collection and document titles, and the third would implement it in the editor.

Sounds good, maybe make this one the base branch add then do additional PR's into this one?

server/routes/api/emojis/emojis.ts
```diff
@@ -0,0 +1,156 @@
...
+    const emoji = await Emoji.createWithCtx(ctx, {
+      name,
+      url,
+      teamId: user.teamId,
+      createdById: user.id,
+      createdBy: user,
+    });
+
+    ctx.body = {
+      data: presentEmoji(emoji!),
```
tommoor, line 127: Surprised bang is needed here? I think that's just bad code coming across from codegen

server/routes/api/emojis/emojis.ts
```diff
@@ -0,0 +1,156 @@
...
+      policies: presentPolicies(user, [emoji!]),
+    };
+  }
+);
+
+router.post(
+  "emojis.delete",
```
tommoor, line 134: There's definitely a question as to what to do when an emoji record is deleted as it'll be referenced in many spots.

I'd probably make sure that references continue to load but serve up a `?` icon or something…

server/models/Emoji.ts
```diff
@@ -0,0 +1,75 @@
...
+class Emoji extends IdModel<
+  InferAttributes<Emoji>,
+  Partial<InferCreationAttributes<Emoji>>
+> {
+  @Length({
+    max: EmojiValidation.maxNameLength,
+    msg: `emoji name must be  less than ${EmojiValidation.maxNameLength} characters`,
+  })
+  @Column(DataType.STRING)
+  name: string;
```
tommoor, line 34: I think we'll need some additional validations to ensure that there are no special characters, something like `[a-z0-9-_]`

app/components/EmojiCreateDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+      Logger.error("Failed to create emoji", error);
+    } finally {
+      setIsUploading(false);
+    }
+  };
+
+  const handleNameChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value
+      .toLowerCase()
+      .replace(/[^a-z0-9_]/g, "")
```
tommoor, line 83: Lets move this to `shared/validations` and share it with the backend model field validation

salihudickson:
> Sounds good, maybe make this one the base branch add then do additional PR's into this one?

I'd need permissions to create a branch on the Outline repo.

server/routes/api/emojis/emojis.ts
salihudickson: yh. I still need to properly vet the code I copied from codegen.

server/routes/api/emojis/emojis.ts
salihudickson: I thought the default browser placeholder for the img tag would be fine.

But If you think it would be better to have a custom placeholder icon I could work on that.

salihudickson:
I'm reopening this here: https://github.com/outline/outline/pull/10513, with a branch on this repo.

PR #10474, opened 2025-10-26
Enable commenting on images

### LINKED ISSUE #7385: No inline comment on images
It is not possible to create an inline comment on an image description

**To Reproduce**
1. Select a part of the text of an image description
2. See that no formatting toolbar appears

**Expected behavior**

When selecting the text, the editing toolbar with reduced features appears that shows only the comment button and which allows to comment on the selection.

**Screenshots**

Outline still looks so good. Here we are discussing font choices for a website.

**Outline:**
- Install: self hosted
- Version: v0.78.0

apoorv-mishra:
Closes #7385

### DIFF
--- a/app/editor/menus/image.tsx
+++ b/app/editor/menus/image.tsx
@@ -75,15 +75,6 @@ export default function imageMenuItems(
       name: "separator",
     },
     {
-      name: "commentOnImage",
-      tooltip: dictionary.comment,
-      shortcut: `${metaDisplay}+⌥+M`,
-      icon: <CommentIcon />,
-    },
-    {
-      name: "separator",
-    },
-    {
       name: "downloadImage",
       tooltip: dictionary.downloadImage,
       icon: <DownloadIcon />,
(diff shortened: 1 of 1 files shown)


app/editor/menus/image.tsx
```diff
@@ -72,6 +74,15 @@ export default function imageMenuItems(
     {
       name: "separator",
     },
+    {
+      name: "commentOnImage",
```
apoorv-mishra: This reflects `commentOn${nodeType}` format. Can follow the same pattern for enabling comments on other node types.

app/editor/index.tsx
```diff
@@ -683,13 +685,33 @@ export class Editor extends React.PureComponent<
       }
     });
 
+    state.doc.descendants((node, pos) => {
```
tommoor: Lets combine this loop with the one above so we're not looping over descendants multiple times

app/editor/index.tsx
```diff
@@ -715,11 +737,26 @@ export class Editor extends React.PureComponent<
           ...mark.attrs,
           ...attrs,
         });
-
         tr.removeMark(from, to, mark).addMark(from, to, newMark);
       }
     });
 
+    state.doc.descendants((node, pos) => {
```
tommoor: Same re combining loop

shared/editor/components/Styles.ts
```diff
@@ -733,7 +733,7 @@ img.ProseMirror-separator {
   }
 }
 
-.heading-name {
+.scroll-to-anchor-heading, .scroll-to-anchor-image {
```
tommoor: Not really sure why the rename to include `scroll-to-`. Probably best to just move the class name into `EditorStyleHelper` as a constant as it's used across files 👍

Issue #9278, opened 2025-05-22
Allow adding custom emojis


tommoor:

salihudickson:
Hey @tommoor, just wondering if this feature is still in demand. I’ve done something similar in a past project and I'd like to work on it here.

tommoor:
Definitely still in demand, but there's a lot of details to get right – I had AI get started on it here previously and abandoned:
https://github.com/outline/outline/pull/9299, but the models/presenters/policies etc are pretty much correct and could be repurposed.

A couple of thoughts:

- One thing I want to make sure is that custom emojis can also be used as icons and small inline images in the editor – so they should be available when triggering the emoji picker in the editor with `:` as well as the emoji picker
- Probably just add a third tab in the main picker for custom
- The settings can use the same table pattern, with a search filter as groups and users
- Probably need to preload custom emoji when the app loads, upto a point – though I'm sure some orgs would create thousands.

salihudickson:
Sure, sounds workable.

tommoor:
Reopening until support in documents is added

PR #10514, opened 2025-10-29
Feat: Custom Emoji In editor

salihudickson:
## Description
This PR allows users to use custom emojis in the editor.

fixes: #9278

## Video Demo
https://github.com/user-attachments/assets/28a44d4d-7110-491d-9b09-7bd953340194

### DIFF
--- a/app/editor/components/EmojiMenu.tsx
+++ b/app/editor/components/EmojiMenu.tsx
@@ -8,6 +8,7 @@ import SuggestionsMenu, {
   Props as SuggestionsMenuProps,
 } from "./SuggestionsMenu";
 import { isInternalUrl } from "@shared/utils/urls";
+import useStores from "~/hooks/useStores";
 
 type Emoji = {
   name: string;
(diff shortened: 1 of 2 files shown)


app/editor/components/EmojiMenu.tsx
```diff
@@ -38,12 +41,21 @@ const EmojiMenu = (props: Props) => {
             title: emoji,
             description: capitalize(item.name.toLowerCase()),
             emoji,
-            attrs: { markup: shortcode, "data-name": shortcode },
+            attrs: {
+              markup: shortcode,
+              "data-name": isInternalUrl(emoji) ? emoji : shortcode,
```
tommoor: Weirds me out for `data-name` to be a url, feels too hacky. A new attribute?

app/editor/components/EmojiMenuItem.tsx
```diff
@@ -19,7 +21,13 @@ export default function EmojiMenuItem({ emoji, ...rest }: EmojiMenuItemProps) {
   return (
     <SuggestionsMenuItem
       {...rest}
-      icon={<Emoji className="emoji">{emoji}</Emoji>}
+      icon={
```
tommoor, line 24: I think we should update the `<Emoji>` icon to encompass this logic

salihudickson: Sure!

salihudickson:
hey @tommoor, could you take another look at this?

shared/editor/nodes/Emoji.tsx
```diff
@@ -87,15 +109,34 @@ export default class Emoji extends Extension {
 
   toMarkdown(state: MarkdownSerializerState, node: ProsemirrorNode) {
     const name = node.attrs["data-name"];
-    if (name) {
+    const url = node.attrs["data-url"];
+
+    if (url) {
```
salihudickson, line 110: wasn't sure which way to go with the markdown format for custom emojis. Do you a preference for what you want this to look like?

The one issue i have with this approach is that there's a chance it could overlap with regular images, and we probably want to limit any chance of that happening.

shared/editor/nodes/Emoji.tsx
```diff
@@ -48,20 +54,36 @@ export default class Emoji extends Extension {
...
-          getEmojiFromName(name),
-        ];
+        if (type === "custom" && !!url) {
+          return [
+            "img",
+            {
+              class: `emoji custom-emoji ${name}`,
+              "data-name": name,
+              src: url,
+              style:
```
tommoor: Don't inline styles like this, they should be in `editor/Styles.tsx`

shared/editor/nodes/Emoji.tsx
```diff
@@ -28,6 +29,11 @@ export default class Emoji extends Extension {
           default: "grey_question",
           validate: "string",
         },
+        "data-url": {
+          default: null,
+          validate: (url: string) => !url || isInternalUrl(url),
+        },
+        type: { default: "emoji", validate: "string" },
```
tommoor: This doesn't really seem to be used for anything?

shared/editor/nodes/Emoji.tsx
tommoor: Markdown export is considered lossy for us at this point, so I think they could look like regular inline images.

shared/editor/rules/emoji.ts
```diff
@@ -1,12 +1,46 @@
-import MarkdownIt from "markdown-it";
+import MarkdownIt, { StateCore } from "markdown-it";
 import { full as emojiPlugin } from "markdown-it-emoji";
 import { nameToEmoji } from "../lib/emoji";
 
 type Options = MarkdownIt.Options & {
   emoji: boolean;
 };
 
+function customEmojiRule(state: StateCore) {
```
tommoor: RE: Lossy I don't think we need to have a way to import these from markdown at all, would remove this complexity

shared/editor/nodes/Emoji.tsx
salihudickson: The toDOM method is using it to check if the emoji is a custom emoji. But now that there is a dedicated `data-url` attr for the url, It is a bit redundant so I'll take it out.

PR #10522, opened 2025-10-30
Link bar cleanup

apoorv-mishra:
> What do you think is natural? I take it partly back, to me it should be:
Enter saves
Click outside saves
Esc discards changes
Empty input removes comment mark

### DIFF
--- a/app/editor/components/LinkEditor.tsx
+++ b/app/editor/components/LinkEditor.tsx
@@ -146,6 +146,11 @@ const LinkEditor: React.FC<Props> = ({
         if (!initialValue) {
           handleRemoveLink();
         }
+
+        // Moving selection to end causes editor state to change,
+        // forcing a re-render of the top-level editor component. As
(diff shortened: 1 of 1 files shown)


tommoor (changes requested):
Good overall, but there is one case that isn't working – when you select one of the suggested docs from the dropdown the link that's inserted is incorrect.

app/editor/components/LinkEditor.tsx
```diff
@@ -183,8 +183,8 @@ const LinkEditor: React.FC<Props> = ({
   const hasResults = !!results.length;
 
   return (
-    <>
-      <Wrapper ref={wrapperRef}>
+    <div ref={wrapperRef}>
```
apoorv-mishra, line 192: `stopPropagation` inside `SuggestionsMenuItem`'s `onClick` wasn't helpful because `useOnClickOutside` uses `onpointerdown`.

Issue #10499, opened 2025-10-28
Invisible documents under "shared with me" when sharing more than 9 docs


dadaxr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

I have a user group "employees" with 1 user inside, and a user group "administrators"

As an admin I created a collection "Customers"

Inside the same collection, I want to have some docs to be viewable & editable by the "employees", and some unreadable (ie invisible). 

As the permissions set on the collection level cannot be overrode to be less permissive, I enable the "edit" permission for the "employees" on the collection level, and then remove doc by doc this permission... 

So one way to achieve that is by setting the following permissions : 
- "All members" (default existing group) to "no access",
- "Administrators" to "can manage"

and then, doc by doc, set "employees" permission to "edit". 

Even if not ideal, it "works" (I can even override the "edit" permission to "no access" inside a sub doc by removing the permission inheritance) .

Then, when logged in as an employee, I can see the doc bellow "shared with me".

But there is a limit ... Only the 9 shared doc are listed here, If more doc are shared, the first ones start  to disappear... 

More annoying, If I start to reorganize afterward my structure to respect this limit and stay bellow the 9 doc shared... any doc which has been "hidden", will stay hidden, even when reseting the permission... (it appears briefly until the next browser reload...)

### Expected Behavior

Ideally it would be awesome to manage this at the collection level instead of "hacking" permission on the "doc" level.

But at least, I guess, all doc "shared with me" should be listed (ie : removing the hard limit of 9), of course, for performance or ux reasons, it could be ok to have a "show more" (or "show all") button.

And also fixing the "invisible" (but readable) doc problem... 

### Steps To Reproduce

create 1 employees group, 
create a collection with "no acces" to everybody
create 12 docs (or more) at the root of the collection, and share them all to the "employees" group
connect as an employee, and you can only see 9 docs

as an admin, remove some of the visible doc... and previously hidden docs wont appear... 

### Environment

```markdown
- outline: v0.87.4 (self hosted)
- Browser:brave (chromium)
```

tommoor:
> But there is a limit ... Only the 9 shared doc are listed here, If more doc are shared, the first ones start to disappear...

This is of course a bug, will take a look

salihudickson:
> Ideally it would be awesome to manage this at the collection level instead of "hacking" permission on the "doc" level.

@tommoor
[...]

PR #10539, opened 2025-10-31
Fix: show more button for "Shared with me" documents

salihudickson:
## Description
Documents under the "shared with me" heading on the sidebar will now have a "show more" button so that users can see all the documents that have been shared with them. This fixes the issue where some of the documents are not displayed.

Fixes: #10499 

### Before
Screenshot 2025-10-31 at 21 16 05

### After
https://github.com/user-attachments/assets/d1fae737-bbdd-436c-8d31-46f08c86d159


app/components/Sidebar/components/GroupLink.tsx
```diff
@@ -50,6 +63,14 @@ const GroupLink: React.FC<Props> = ({ group }) => {
               depth={1}
             />
           ))}
+          {!end && (
```
tommoor, line 66: There is already a "Show more…" button on L127 of `SharedWithMe` component – the logic should be combined to reuse this otherwise we'll potentially end up with both of these showing. 

Nothing should change inside of GroupLink

tommoor (changes requested):

app/components/Sidebar/components/GroupLink.tsx
tommoor: Actually, after testing more I take this back – I see why it makes sense with the nesting.

app/components/Sidebar/components/SharedWithMe.tsx
```diff
@@ -30,7 +30,9 @@ function SharedWithMe() {
   const history = useHistory();
   const locationSidebarContext = useLocationSidebarContext();
 
-  usePaginatedRequest<GroupMembership>(groupMemberships.fetchAll);
+  const gmResponse = usePaginatedRequest<GroupMembership>(
+    groupMemberships.fetchAll
```
tommoor, line 34: We're already calling `fetchAll` here, so all memberships should be available, I think the problem is that there is no `total` returned in the pagination response so it doesn't know how many pages to load.

Looking into that…

tommoor: closed in #10589

salihudickson:
hadn't realized that. Thanks for the review 🙏🏾

Issue #10520, opened 2025-10-30
Report that admin can edit docs with view only permission


tommoor:
> I am an admin in our workspace. If I have access to a collection through All Members -> View Only -> I can edit any document in this collection. If I get explicit View Only permission on my user, and All Members has no access, I get only View Only access. Is this intentional?

Needs investigation

AshtakaOOf:
Shouldn't admins have permissions to modify any documents? 
I'm not sure why it wouldn't be allowed as it can be useful in many cases, for example when a user accidentally left sensitive information in their document but aren't available to fix it.

tommoor:
Admins can change permissions on collections and documents, so ultimately they are able to fix things in an emergency but I don't think that means they should have write access to everything by default. This is the middleground we've gone for.

PR #10542, opened 2025-11-01
Fix: admin permissions bypass

salihudickson:
## Description
This PR fixes the issue that allows admins to edit documents on collections that they should only have "view only" access to.

fixes: #10520

### DIFF
--- a/server/policies/document.test.ts
+++ b/server/policies/document.test.ts
@@ -31,16 +31,14 @@ describe("read_write collection", () => {
     const abilities = serialize(user, document);
     expect(abilities.read).toBeTruthy();
     expect(abilities.download).toBeTruthy();
+    expect(abilities.update).toBeTruthy();
+    expect(abilities.createChildDocument).toBeTruthy();
+    expect(abilities.manageUsers).toBeTruthy();
+    expect(abilities.archive).toBeTruthy();
+    expect(abilities.delete).toBeTruthy();
     expect(abilities.share).toBeTruthy();
+    expect(abilities.move).toBeTruthy();
     expect(abilities.comment).toBeTruthy();
-
-    // members without edit permissions cannot
-    expect(abilities.archive).toBeFalsy();
-    expect(abilities.manageUsers).toBeFalsy();
-    expect(abilities.createChildDocument).toBeFalsy();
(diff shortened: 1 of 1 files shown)


server/policies/collection.test.ts
```diff
@@ -9,23 +9,6 @@ import {
 import { serialize } from "./index";
 
 describe("admin", () => {
-  it("should allow team admin to archive collection", async () => {
```
tommoor, line 12: Awkward I know, but we want to keep this ability – admins should be able to manage collections from the outside. Just not write docs within

salihudickson: I didn't actually remove the ability for admins to manage permissions. I just removed that test block because it seems to be a duplicate of the one that now starts on line 35. 

I can restore it if you'd like.

server/policies/collection.test.ts
```diff
@@ -49,7 +32,7 @@ describe("admin", () => {
     expect(abilities.archive).toBeTruthy();
   });
 
-  it("should allow updating documents in view only collection", async () => {
+  it("should have correct permissions in view only collection", async () => {
```
salihudickson, line 35: I removed the duplicate test block. The only distinction was that this one explicitly set the collection’s permissions to `read`, while the default behavior of createCollection is `read_write`, which implicitly includes `read`.

server/policies/collection.ts
```diff
@@ -101,11 +101,8 @@ allow(User, "updateDocument", Collection, (user, collection) => {
     return false;
   }
 
-  if (!collection.isPrivate && user.isAdmin) {
-    return true;
-  }
-
   if (
+    !collection.isPrivate ||
```
tommoor: Something about this still doesn't feel quite right, this additional OR case is saying that the membership should be considered if the collection is public, irregardless of the `collection.permission`. 

It _technically_ works right now because the only other permission is ReadWrite but I don't think it's neccessary. The only _needed_ change is the removal of the admin exception above

salihudickson: Yh, you're absolutely right. I hadn't considered that the `includesMembership` check already covers a private collection.

tommoor (approved):

Issue #10538, opened 2025-10-31
Report of nested documents losing ordering during duplication


tommoor:
Customer reported, haven't triaged yet but it sounds distinctly possible

salihudickson:
I’ve confirmed this behavior. My initial observation is that it’s likely caused by the document duplicator copying child documents in reverse order based on the createdAt column. I'll have to take a closer look.

tommoor:
The canonical order is defined by the position in the tree stored on `collection` btw

salihudickson:
🤔 I'm thinking we simply need to have the duplicator take the order into account when duplicating the `childDocuments`.

PR #10543, opened 2025-11-01
Fix: nested document order when duplicating

salihudickson:
## Description
This PR fixes the bug that breaks the document nested document order when duplicating a document

fixes: #10538

## Video Demo
https://github.com/user-attachments/assets/c9d92dc6-a26b-43a5-9165-880937cd2233

### DIFF
--- a/server/commands/documentDuplicator.test.ts
+++ b/server/commands/documentDuplicator.test.ts
@@ -1,8 +1,11 @@
 import { createContext } from "@server/context";
 import { sequelize } from "@server/storage/database";
-import { buildDocument, buildUser } from "@server/test/factories";
+import {
+  buildCollection,
+  buildDocument,
+  buildUser,
+} from "@server/test/factories";
 import documentDuplicator from "./documentDuplicator";
-import { Collection } from "@server/models";
 
 describe("documentDuplicator", () => {
   it("should duplicate existing document", async () => {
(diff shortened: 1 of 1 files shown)


server/commands/documentDuplicator.ts
```diff
@@ -76,10 +76,26 @@ export default async function documentDuplicator({
               [Op.eq]: null,
             },
       },
-      ctx
+      {
+        ...ctx,
+        include: {
```
tommoor: I think we should just fetch the associated collection separately rather than adding join complexity. By definition all the documents will be in the same collection

salihudickson: yh. I'll fix that.

salihudickson:
hey @tommoor, I think this is just about ready for another review.

tommoor:
Seems like your tests are failing…

salihudickson:
I took out the bit that was failing. Should be working fine now.

tommoor:
> I took out the bit that was failing

Isn't that the test that was actually checking the order?

salihudickson:
my bad, I actually replaced it with a 2 different tests where i checked the length of the docTitles array and then verified each item in the array. I must've accidentally taken out the second part of that while removing the failing test. I'll push a fix now.

For some reason the strict equal was acting weird on the CI test. Still not sure why, but since the actual functionality seems to be fine even when testing in a few different scenarios, i decided not to spend too much time on trying to figure it out.

tommoor (approved):

PR #10547, opened 2025-11-02
fix: 'shared with me' optimistic updates

salihudickson:
## Description
This PR adds optimistic state updates to documents under the 'shared with me' section, in the sidebar.

fixes: #10504

### DIFF
--- a/app/models/base/NavigableModel.ts
+++ b/app/models/base/NavigableModel.ts
@@ -2,7 +2,7 @@ import { action, computed, observable, runInAction } from "mobx";
 import { JSONObject, type NavigationNode } from "@shared/types";
 import { client } from "~/utils/ApiClient";
 import ParanoidModel from "./ParanoidModel";
-import Document from "../Document";
+import type Document from "../Document";
 
 export default abstract class NavigableModel extends ParanoidModel {
   private isFetching = false;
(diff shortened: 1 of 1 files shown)


app/models/GroupMembership.ts
```diff
@@ -47,6 +51,31 @@ class GroupMembership extends NavigableModel {
...
 
+  /**
+   * Adds the document identified by the given id to the group membership in
+   * memory. Does not add the document to the database or store.
+   *
+   * @param document The document to add.
+   * @param parentDocumentId The id of the document to add the new document to.
+   */
+  @action
+  addDocument(document: Document, parentDocumentId?: string) {
```
tommoor: Seems good but both of these inherit from `NavigableModel` so we could probably just put it there once

salihudickson: 👍🏾 sure!

salihudickson: hey @tommoor, I think this is just about ready for another review.

app/models/base/NavigableModel.ts
```diff
@@ -45,6 +46,14 @@ export default abstract class NavigableModel extends ParanoidModel {
     return this.node?.children;
   }
 
+  set documents(value: NavigationNode[] | undefined) {
```
tommoor: Any reason we can't use `@action` decorator here? Not sure if you tried

app/models/base/NavigableModel.ts
```diff
@@ -1,7 +1,8 @@
-import { computed, observable, runInAction } from "mobx";
+import { action, computed, observable, runInAction } from "mobx";
 import { JSONObject, type NavigationNode } from "@shared/types";
 import { client } from "~/utils/ApiClient";
 import ParanoidModel from "./ParanoidModel";
+import Document from "../Document";
```
tommoor:
```suggestion
import type Document from "../Document";
```

app/models/base/NavigableModel.ts
salihudickson: That was actually an oversight on my part. Fixed it 🙏🏾

Issue #7420, opened 2024-08-20
Editor appears to allow linked images with CMD+K but link is not persisted


microbearlogist:
Hello,

When an http link is attributed to a image (using ctrl +k shortcut for exemple), a left clic on the picture from the "private" wiki page redirect to the appropriate address.

However, if the page is accessed via a public link (generated using "share / web" menu), the http link is no longer present, and only the image zoom option appears.

**Outline (please complete the following information):**
- Install: self hosted
- Version: 0.78.0

**Desktop (please complete the following information):**
 - OS: Win 11
 - Browser : Firefox 129.0.1

tommoor:
The actual issue here seems to be that links are not supported for images right now, though the editor gives the impression they are.

microbearlogist:
You're right, my bad.

When a link is “forcibly” added to an image via keyboard shortcuts, it works even when you exit edition mode, but only for the current session.

The link does not work in edit mode, but only once this mode is exited and until the page is reloaded.

tommoor:
This is due to an upstream bug, https://github.com/yjs/y-prosemirror/issues/47 – I do think we should add support though, so will probably mean adding `href` as an optional attribute to images.

PR #10556, opened 2025-11-03
Link bar refactor to use `commands`

apoorv-mishra:
- [x] Refactor to use commands
- [x] Pressing "Enter" on empty link bar should remove the link mark

Towards #7420

### DIFF
--- a/app/editor/components/LinkEditor.tsx
+++ b/app/editor/components/LinkEditor.tsx
@@ -30,12 +30,6 @@ type Props = {
   view: EditorView;
 };
 
-enum Action {
-  OPEN_LINK = "openLink",
-  UPDATE_LINK = "updateLink",
-  REMOVE_LINK = "removeLink",
-}
-
 const LinkEditor: React.FC<Props> = ({ mark, dictionary, view }) => {
   const getHref = () => sanitizeUrl(mark?.attrs.href) ?? "";
   const initialValue = getHref();
@@ -79,18 +73,18 @@ const LinkEditor: React.FC<Props> = ({ mark, dictionary, view }) => {
   });
 
   const openLink = React.useCallback(() => {
-    commands[Action.OPEN_LINK]();
+    commands["openLink"]();
   }, []);
 
   const removeLink = React.useCallback(() => {
-    commands[Action.REMOVE_LINK]();
+    commands["removeLink"]();
   }, []);
 
   const updateLink = (link: string) => {
     if (!link) {
       return;
     }
-    commands[Action.UPDATE_LINK]({ href: sanitizeUrl(link) ?? "" });
+    commands["updateLink"]({ href: sanitizeUrl(link) ?? "" });
   };
 
   const moveSelectionToEnd = () => {


app/editor/components/LinkEditor.tsx
```diff
@@ -196,26 +197,27 @@ const LinkEditor: React.FC<Props> = ({
           value={query}
           placeholder={dictionary.searchOrPasteLink}
           onKeyDown={handleKeyDown}
-          onPaste={handlePaste}
```
apoorv-mishra, line 199: I believe it's safe to get rid of this? Personally, I don't miss it from a UX point of view, given that a click outside already saves input. It was anyways defunct.

tommoor: Yes, seems fine. Subtly different but not in a terrible way

shared/editor/marks/Link.tsx
```diff
@@ -138,6 +146,72 @@ export default class Link extends Mark {
     };
   }
 
+  commands({ type }: { type: MarkType }) {
+    return {
+      addLink: (attrs: Attrs): Command => addMark(type, attrs),
```
tommoor, line 151: Not sure we need this duplicative command when `link` already works in the same style as other marks?

app/editor/components/LinkEditor.tsx
```diff
@@ -23,36 +22,52 @@ import SuggestionsMenuItem from "./SuggestionsMenuItem";
...
+  OPEN_LINK = "openLink",
+  UPDATE_LINK = "updateLink",
+  REMOVE_LINK = "removeLink",
+}
+
+type ActionDetails = {
+  [key in Action]: MenuItem;
+};
+
+const getActions = (
```
tommoor: I don't really think this data structure is adding much, just a bunch of indirection – would probably just inline these things

shared/editor/marks/Link.tsx
apoorv-mishra: Yeah, it's forward looking. Further, for image selections it'll be replaced by 👇 
```js
export const addLink = (attrs: Attrs): Command =>
  chainCommands(addLinkTextSelection(attrs), addLinkNodeSelection(attrs));
```

to accommodate node selections, in general. The default `toggleMark` command in the super class `Mark` won't be very useful in that case making override necessary. This will be very similar to the existing `addComment` override for comment marks.

app/editor/components/LinkEditor.tsx
```diff
@@ -78,35 +65,41 @@ const LinkEditor: React.FC<Props> = ({
...
-    }
-
     // If the link is totally empty or only spaces then remove the mark
     if (!trimmedQuery) {
-      return handleRemoveLink();
+      return removeLink();
+    }
+
+    // If the link in input is non-empty and same as it was when the editor opened, nothing to do
+    if (trimmedQuery === initialValue) {
```
apoorv-mishra, line 68: Unrelated change–moving this particular if statement down helps get rid of the extra `trimmedQuery.length` check, so it's just a cleanup.

app/editor/components/LinkEditor.tsx
```diff
@@ -23,36 +22,24 @@ import SuggestionsMenuItem from "./SuggestionsMenuItem";
...
-const LinkEditor: React.FC<Props> = ({
-  mark,
-  from,
-  to,
-  dictionary,
-  onSelectLink,
-  onClickLink,
-  view,
-}) => {
+enum Action {
```
tommoor: Lets remove and inline this enum too – then approved 👍

PR #10562, opened 2025-11-04
Include cookies with S3 POST upload

### LINKED ISSUE #10561: File upload does not include cookies in requests

### Current Behavior

The `files.ts` constructs an XMLHttpRequest, but it [does not set withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) and thus won't send _any_ cookies with the CORS preflight. If you are using a proxy service (such as [CloudFlare's Web Application proxy](https://developers.cloudflare.com/cloudflare-one/access-controls/applications/http-apps/)) that requires a cookie to be present on all requests (including CORS) this makes file upload fail.

This problem only affects file upload as it's the [only place XMLHttpRequest is used](https://github.com/search?q=repo%3Aoutline%2Foutline+XMLHttpRequest&type=code) in the project.

### Expected Behavior

File uploads should include cookies with the CORS and POST request, so that intermediate proxies that require session cookies don't redirect to login pages and fail.

### Steps To Reproduce

1. Open the Chrome dev console
2. Upload a file
3. Observe that no cookies are sent with the POST request to the S3 upload URL.

### Environment

```markdown
- Outline: v1.0.1
- Browser: Chrome Version 141.0.7390.123
```

### Anything else?

Let's say you have `outline.example.com` and `s3.example.com` both configured as [related Cloudflare web applications](https://developers.cloudflare.com/cloudflare-one/access-controls/applications/http-apps/). The first time you visit `wiki.example.com` and authenticate with Cloudflare's proxy, you'll get a `CF_Authorization` cookie valid for _both_ `outline.example.com` and `s3.example.com`. You'll then be redirected to the normal login page for Outline.

Downloading S3 files is handled by modern fetch requests, so cookies are included. Here we can see the network chain hitting the `attachments.redirect` endpoint, which redirects to s3.example.com to download the file. It includes the CF_Authorization cookie and succeeds.

Image

[Uploads are handled by the older XmlHttpRequest](https://github.com/outline/outline/blob/e4268c9a1fc04bac03c7c410b5423bdb71bc2422/app/utils/files.ts#L77-L92) for progress bar support. Without `withCredentials` we can see that no request cookies are sent, so we get a response cookie from Cloudflare creating a new session.

Image

Along with a 302 redirect to the cloudflare login page, which results in a CORS error since it's a totally different site.

Image

Based on searching through previous discussions it's likely I'm not the first person to encounter this, however the previously discussed solutions around opening my S3 storage system to the public internet are not possible in my network configuraton.

Cellivar:
Fixes #10561

I'm currently running Outline using Ceph Object Gateway to provide the S3 storage, and running behind Cloudflare's application proxy system. Pretty much everything works except for file uploading.

Because the XMLHttpRequest doesn't include cookies by default the `CF_Authorization` cookie is not sent in the POST to the S3 endpoint. No `CF_Authorization` means a redirect to the Cloudflare login page, breaking uploading. Because this is the only place XMLHttpRequest is used this is the only broken feature.

This one-line change includes cookies in the request, resolving the issue.

### DIFF
--- a/app/utils/files.ts
+++ b/app/utils/files.ts
@@ -88,6 +88,7 @@ export const uploadFile = async (
     xhr.addEventListener("loadend", () => {
       resolve(xhr.readyState === 4 && xhr.status >= 200 && xhr.status < 400);
     });
+    xhr.withCredentials = true;
     xhr.open("POST", data.uploadUrl, true);
     xhr.send(formData);
   });


tommoor:
This makes sense, I need to double check what the implications are for uploads to AWS before merging

Cellivar:
> I need to double check what the implications are for uploads to AWS before merging

Yep thanks, I can't test that myself. I also had some trouble getting a local Docker build to thoroughly test using my own setup. If you can push a test Docker image up I can deploy it into my cluster and confirm.

tommoor:
Ok, it makes no difference on S3 as there are no credentials for the bucket domain.

tommoor (approved):

PR #10565, opened 2025-11-04
feat: Adjust line-height depending on script

tommoor:
I can't decide if this is a genius level of detail and care… or insane.

It will be nice to have an idea in the database of which languages are being used.

### DIFF
--- a/server/__mocks__/franc.js
+++ b/server/__mocks__/franc.js
@@ -1,0 +1,10 @@
+// Mock for franc language detection library
+const franc = jest.fn((text) => {
+  // Return 'eng' (English) by default, or 'und' (undetermined) for empty text
+  if (!text || text.trim().length === 0) {
+    return "und";
+  }
+  return "eng";
+});
+
+module.exports = { franc };
--- a/server/__mocks__/iso-639-3.js
+++ b/server/__mocks__/iso-639-3.js
@@ -1,0 +1,18 @@
+// Mock for iso-639-3 language code conversion library
+const iso6393To1 = {
+  eng: "en",
+  fra: "fr",
+  deu: "de",
+  spa: "es",
+  ita: "it",
+  por: "pt",
+  rus: "ru",
+  jpn: "ja",
+  zho: "zh",
+  ara: "ar",
+  hin: "hi",
+  ben: "bn",
+  und: undefined, // undetermined
+};
+
+module.exports = { iso6393To1 };


tommoor:
Oh, cool. An esm-only module what fun

coekfung:
It is weird that the franc library will use some non-macrolanguage code which cannot be converted to iso6391. For example, `cmn` (cannot be converted to iso6391) for chinese but not `zho` (can be converted to `zh`).
Maybe we should maintain an extra conversion table to handle this situation? Since the franc library has been out-of-maintaince for a long time.

tommoor:
It uses iso-639-3,just a different standard

coekfung:
> It uses iso-639-3,just a different standard

Thanks for your great work and fast reply!

You’re right that franc returns language codes in ISO 639-3, and our pipeline uses `iso6393To1` to convert them to ISO 639-1 before writing to the database. My understanding is that we do this to make the stored language value compatible with the CSS `:lang()` pseudo-class, which expects a BCP 47 language tag (where ISO 639-1 codes are commonly used as the primary language subtag).

However, this conversion step breaks for some very common cases. For example, franc returns `cmn` for Chinese. `cmn` (Mandarin Chinese) is a valid ISO 639-3 code, but it has no ISO 639-1 equivalent, so `iso6393To1('cmn')` yields `undefined`. In practice, we would want to store `zh`, because `cmn` is encompassed by the ISO 639-3 macrolanguage `zho`, whose ISO 639-1 counterpart is `zh`. Due to this gap in the mapping, we end up saving `undefined` instead of `zh`.

Because of this, I suggest we add a small override/conversion table (or fallback logic) to handle these special cases—for example mapping `cmn` -> `zho` -> `zh` (or directly `cmn` -> `zh`)—so that the stored value remains usable for `:lang()`.

See also: https://iso639-3.sil.org/code/cmn

tommoor:
I'm ok with this, although I'd rather this mapping existed in `iso6393To1` project 😬 

Looking at https://www.npmjs.com/package/iso-639-3 it's never going to happen there as it's strictly mapping the spec. You're right that an internal remapping would be fine.

coekfung:
I found a mapping table https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3-macrolanguages.tab (thanks grok). I hope this would be useful.

In my opinion, a potential way is,
1. lookup the `iso6393To1` table, if not undefined, return
2. lookup the individual language to macrolanguage table, if mapping exists, lookup the `iso6393To1` using macrolanguage again, if not undefined, return
3. else, return undefined

As the conversion table is relatively small, maybe we can write them directly in the code?
Hope for your advise, maybe I can draft a pr for this.

PR #10567, opened 2025-11-04
Feat/custom emoji as document and collection icon

salihudickson:
## Dcescription 
This PR adds custom emoji as a document and collection icon

## Video Demo
https://github.com/user-attachments/assets/dc4bcef9-2cb7-4616-b31f-5149dec0e16c

### DIFF
--- a/app/components/IconPicker/components/CustomEmojiPanel.tsx
+++ b/app/components/IconPicker/components/CustomEmojiPanel.tsx
@@ -5,7 +5,6 @@ import useStores from "~/hooks/useStores";
 import GridTemplate, { DataNode, EmojiNode } from "./GridTemplate";
 import { IconType } from "@shared/types";
 import { DisplayCategory } from "../utils";
-import { isInternalUrl } from "@shared/utils/urls";
 import { StyledInputSearch, UserInputContainer } from "./Components";
 import { useIconState } from "../useIconState";
 import Emoji from "~/models/Emoji";
@@ -33,6 +32,7 @@ const CustomEmojiPanel = ({
   const searchRef = React.useRef<HTMLInputElement | null>(null);
   const scrollableRef = React.useRef<HTMLDivElement | null>(null);
   const [searchData, setSearchData] = useState<DataNode[]>([]);
+  const [freqEmojis, setFreqEmojis] = useState<EmojiNode[]>([]);
   const { getFrequentIcons, incrementIconCount } = useIconState(
     IconType.Custom
   );
@@ -68,12 +68,6 @@ const CustomEmojiPanel = ({
         })
         .then((data) => {
           if (data.length) {
-            const toIcon = (emoji: Emoji): EmojiNode => ({
-              type: IconType.Custom,
-              id: emoji.name,
-              value: emoji.url,
-            });
-
             const iconMap = new Map([
               ...initialData.map((emoji): [string, EmojiNode] => [
                 emoji.name,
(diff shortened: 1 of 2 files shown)


tommoor:
Seems to be a repeat of #10514 again

salihudickson:
No, that one is for the use of custom emojis in the editor. This one is for the use of custom emojis as documents and collection icons. The way i named them is just a bit confusing, I'll fix that :)

salihudickson:
hey @tommoor, this is just about ready for another round of reviews.

app/components/IconPicker/components/CustomEmojiPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+        width={panelWidth}
+        height={height - 48}
+        data={searchData.length ? searchData : templateData}
+        onIconSelect={handleEmojiSelection}
+      />
+    </Flex>
+  );
+};
+
+const getFrequentEmojis = (
```
tommoor: We should refactor this so that the core of the functionality is shared between `CustomEmojiPanel` and `EmojiPanel`

app/components/IconPicker/components/CustomEmojiPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+    const value = lastEmoji.substring(index + 1);
+
+    emojis.pop();
+    emojis.push({
+      id,
+      value,
+    });
+  }
+
+  return emojis.filter((emoji) => isInternalUrl(emoji.value));
```
tommoor: This last filter is probably the only difference?

app/components/IconPicker/components/CustomEmojiPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+};
+
+const StyledInputSearch = styled(InputSearch)`
```
tommoor: Similar, can we share these instead of repeating? just create `/Components.tsx` file locally is fine

shared/components/Icon.tsx
```diff
@@ -60,6 +61,10 @@ const Icon = ({
       );
     }
 
+    if (iconType === IconType.Custom) {
+      return <EmojiImage src={icon} style={{ width: 24, height: 24 }} />;
```
tommoor: Visual sizing is wrong, should probably be 16px or 18px in a 24px container

image

shared/utils/emoji.ts
```diff
@@ -214,6 +220,23 @@ export const search = ({
   });
 };
 
+export const getCustomEmojis = async (
```
tommoor: This should be in the `EmojiStore` like other data fetching, and store the results in the store

tommoor (commented):
Mostly good 👍

shared/components/Icon.tsx
salihudickson: I'm not sure I follow. should I just set it to 18px, or should I wrap it in another container inside the icon component, which is already acting as a container?

tommoor: The image should be shown at 16x16 in a 24x24 bounding box whenever it's displayed as an icon

salihudickson: Okay, got it. 

Just to confirm, should this apply to the document and collection title as well? Or should the style be different there?

tommoor: It should scale appropriately, as an icon would – does that make sense?

salihudickson: oh yeah,  thanks for clarifying.

tommoor (approved):

Issue #10582, opened 2025-11-07
Cannot attach files from Android devices


rstepinski:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Whenever I try to attach an image, video or any other file from my Android device, after selecting the file in media picker, nothing happens. There's no errors in the console visible in devtools, there's nothing in container logs. The feature works as expected on desktop.

### Expected Behavior

Adding an image or any other file should work normally.

### Steps To Reproduce

1. Open a document in Outline on an Android device
2. Try to insert an image, video or file
3. Select a file from media picker and continue.
4. Observe that nothing is added.

### Environment

```markdown
- Outline: 1.0.1
- Browser: Chrome and Firefox
- OS: Android 15 and 13
```

### Anything else?

If there's already an image in the document, it is possible to use the "replace" feature and it does work correctly also on Android.

salihudickson:
seems to be an issue on mobile devices, related to how the editor handles the view focus on those devices when there is no content on that line. 

thanks for reporting, i'll take a closer look.

tommoor:
Possibly related #10605

salihudickson:
🤔 very likely.

beastarman74:
If you switch on desktop mode in the mobile browser I have found that image uploads work as expected. I don't know if this helps at all with diagnosis?

Issue #10605, opened 2025-11-09
@Mentions always appear at top of document when added via mobile


jondoran:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Using Firefox on an Android phone, when I create an @Mention (for example, a link to another document) in the body of the document using the popup context menu, the created link always appears at the top of the document.

### Expected Behavior

The styled @Mention link should appear inline where it was created, as it does in the desktop editor.

### Steps To Reproduce

1. Open a document using the Firefox app on an Android phone
2. Add some content
3. Insert an @Mention and use the context menu to select a document or user
4. The styled link to the document will appear at the top of the page

### Environment

```markdown
- Outline:1.01
- Browser:Firefox Android
```

PR #10608, opened 2025-11-09
fix: uploading attachments on mobile devices

salihudickson:
## Description
This PR fixes a bug that was preventing users from uploading attachments on mobile devices.

fixes: #10582  #10605

## Video Demo
https://github.com/user-attachments/assets/aad41291-b8cf-4ec3-aba5-196983d1f8ed

### DIFF
--- a/app/editor/components/PasteMenu.tsx
+++ b/app/editor/components/PasteMenu.tsx
@@ -15,7 +15,6 @@ import SuggestionsMenu, {
   Props as SuggestionsMenuProps,
 } from "./SuggestionsMenu";
 import SuggestionsMenuItem from "./SuggestionsMenuItem";
-import { useEditor } from "./EditorContext";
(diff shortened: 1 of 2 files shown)


salihudickson:
@tommoor 

The problem seems to be that we were counting on the editor to refocus to the right spot after clicking a suggestion in the menu. Sometimes it works fine, but other times it doesn’t.

My fix is to explicitly handle the refocus by passing down the initial selection and resetting it after a suggestion is clicked.

I didn’t add this fix to the paste menu, though, since the editor already seems to handle that part well on its own, and doing it manually there actually causes more problems than it solves.

tommoor:
Something must have changed, perhaps in one of the recent prosemirror upgrades… this was all working fine previously, hence that existing comment in the source about mobile focus

PR #10636, opened 2025-11-13
chore: Move user suspension check inside `getUserForJWT`

tommoor:


server/middlewares/authentication.ts
```diff
@@ -247,12 +247,9 @@ async function validateAuthentication(
   }
 
   if (user.isSuspended) {
-    const suspendingAdmin = await User.findOne({
-      where: {
-        id: user.suspendedById!,
-      },
-      paranoid: false,
-    });
+    const suspendingAdmin = user.suspendedById
```
tommoor, line 250: Still need this one as it also covers OAuth and API key auth

Issue #10545, opened 2025-11-02
Add "Distribute Column Width Evenly" for table columns


huiseo:
## Summary

Add a feature to evenly distribute column widths in tables, similar to spreadsheet applications like Excel, Google Sheets, and Notion.

## Problem / Use Case

When working with tables in Outline, users often need to create uniform column layouts for better readability and professional appearance. Currently, achieving equal column widths requires:

- Manually adjusting each column individually
- Eyeballing the measurements (difficult to be precise)
- Time-consuming, especially for tables with many columns

**Example scenario:**

Creating a comparison table with 5 product features. After adding content, columns have uneven widths (220px, 180px, 240px, 160px, 200px). To make it look professional, all columns should be equal width (200px each). Currently requires tedious manual adjustment of each column.

### UI/UX

- Add menu item "Distribute column width evenly" in the column header menu
- Only show when 2 or more columns are selected
- One-click to redistribute selected columns to equal widths
- Preserve total table width (no layout shifts)

### User Flow

1. Select multiple columns (Shift+click on headers)
2. Open column menu
3. Click "Distribute column width evenly"
4. Columns become equal width

### Example

**Before**: [200px, 150px, 250px] → Total: 600px
**After**: [200px, 200px, 200px] → Total: 600px (preserved)

### User Benefits

- **Faster workflow**: One click vs. multiple manual adjustments
- **Professional appearance**: Perfect column uniformity
- **Familiar**: Standard feature in Excel, Notion, Confluence, Coda
- **Better UX**: No more "eyeballing" widths

### For the Project

- **Small addition**: Focused feature, minimal code
- **No breaking changes**: Purely additive
- **Low maintenance**: Self-contained functionality
- **Industry standard**: Users expect this feature

## Industry Comparison

This feature is standard in modern productivity tools:

- **Microsoft Word/Excel**: "Distribute Columns Evenly"
- **Google Sheets**: Auto-resize columns
- **Notion**: Column width distribution
- **Confluence**: Table column formatting
- **Coda**: Distribute columns

Users coming from these tools expect this functionality.

## Current Workarounds

Users currently have to:

1. Manually drag each column border
2. Visually estimate equal widths
3. Accept uneven columns

All workarounds are time-consuming and imprecise.

## Implementation Notes

I have a working prototype that:

- Follows ProseMirror command patterns (similar to existing table commands)
- Integrates into existing menu system
- Includes i18n for all supported languages
- Handles edge cases properly

[...]

PR #10645, opened 2025-11-15
feat: Distribute table columns evenly

salihudickson:
## Desc
This PR allows users to automatically, and  evenly space the columns of the tables in their documents.

fixes: #10545


shared/editor/commands/table.ts
```diff
@@ -174,6 +176,144 @@ export function exportTable({
...
+      const { tr } = state;
+      const { map } = rect;
+
+      const selectedColumns = getAllSelectedColumns(state).sort(
+        (a, b) => a - b
+      );
+
+      const columnCount = map.width;
+      const isLastColSelected = selectedColumns.includes(columnCount - 1);
+      const totalSelectedWidth = Math.max(
```
salihudickson: there are still some issues with the way the table width is calculated. Most times the larger width is correct, but a some edge cases exist.

salihudickson:
Hey @tommoor,

I’m a bit unsure about the best UI direction for this, mind taking a look? And I’m still ironing out some bugs, but I’d appreciate your thoughts on the general approach as well. Thanks!

shared/editor/nodes/TableRow.ts
```diff
@@ -12,7 +12,13 @@ export default class TableRow extends Node {
       tableRole: "row",
       parseDOM: [{ tag: "tr" }],
       toDOM() {
-        return ["tr", 0];
+        return [
+          "tr",
+          {
+            style: "word-break: break-word;",
```
tommoor: No styles here, should be in `Styles.tsx`

app/editor/menus/table.tsx
```diff
@@ -45,5 +50,15 @@ export default function tableMenuItems(
...
       icon: <DownloadIcon />,
     },
+    {
+      name: "separator",
+    },
+    {
+      name: "spaceColumnsEvenly",
+      tooltip: dictionary.spaceColumnsEvenly,
+      label: "space evenly",
+      attrs: { isFullWidth, documentIsFullWidth: documents.active?.fullWidth },
```
tommoor: `attrs` should directly map to properties we're storing on the node sending `documentIsFullWidth` here

app/editor/menus/tableCol.tsx
```diff
@@ -156,6 +166,19 @@ export default function tableColMenuItems(
           label: dictionary.deleteColumn,
           icon: <TrashIcon />,
         },
+        {
+          name: "separator",
+        },
+        {
+          name: "spaceColumnsEvenly",
```
tommoor: UI wise this is fine for now, but as the comment on the issue i think we should (separately) change this entire toolbar to be a menu instead and then it will be much clearer.

salihudickson: Yh, that's probably a good idea.

app/editor/menus/table.tsx
```diff
@@ -45,5 +48,13 @@ export default function tableMenuItems(
       attrs: { format: "csv", fileName: `${window.document.title}.csv` },
       icon: <DownloadIcon />,
     },
+    {
+      name: "separator",
+    },
+    {
+      name: "distributeWidthEvenly",
+      tooltip: dictionary.distributeWidthEvenly,
+      icon: <FontAwesomeIcon icon={faArrowsLeftRight} />,
```
tommoor: I added `TableColumnsDistributeIcon` in `outline-icons@3.14.0` for use here

salihudickson: Thanks, I'll replace it.

salihudickson:
@tommoor, the build seems to be failing due to an outdated yarn.lock file, since I updated the icons package. Not sure if you'd want me to update the lock file here though, especially since that could lead to some dependency conflicts.

tommoor:
You should update the lock file. just run yarn install

package.json
```diff
@@ -382,7 +382,10 @@
     "node-fetch": "^2.7.0",
     "js-yaml": "^3.14.1",
     "qs": "6.9.7",
-    "prismjs": "1.30.0"
+    "prismjs": "1.30.0",
+    "strip-ansi": "6.0.1",
+    "string-width": "4.2.3",
+    "wrap-ansi": "7.0.0"
```
tommoor: Why, this is unrelated?

app/editor/menus/table.tsx
```diff
@@ -45,5 +51,13 @@ export default function tableMenuItems(
       attrs: { format: "csv", fileName: `${window.document.title}.csv` },
       icon: <DownloadIcon />,
     },
+    {
+      name: "separator",
+    },
+    {
```
tommoor: Lets move this to the second item after alignFullWidth, no need for a separator

package.json
salihudickson: When I run `yarn install`, these dev dependencies resolve to newer versions, and their new versions are ES6 which causes some of the dependencies to break.

tommoor: Need to remove these changes and make sure only thing that updates in lockfile is the `outline-icons` dep

salihudickson: done

tommoor (approved):

PR #10647, opened 2025-11-15
fix: STARTTLS is not used

b11p:
STARTTLS is different from TLS. According to the doc of Nodemailer (https://nodemailer.com/smtp), the `ignoreTLS` actually disables STARTTLS. Outline disables STARTTLS when `SMTP_SECURE` is set to `false`. However, my SMTP provider use STARTTLS, not TLS. We need a separate control of STARTTLS.

### DIFF
--- a/app.json
+++ b/app.json
@@ -194,6 +194,11 @@
       "description": "Use a secure SMTP connection (optional)",
       "required": false
     },
+    "SMTP_DISABLE_STARTTLS": {
+      "value": "false",
+      "description": "Disable STARTTLS even if the server supports it (optional)",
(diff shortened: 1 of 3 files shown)


app.json
```diff
@@ -194,6 +194,11 @@
       "description": "Use a secure SMTP connection (optional)",
       "required": false
     },
+    "SMTP_NO_STARTTLS": {
```
tommoor: Lets call it `SMTP_DISABLE_STARTTLS`

b11p: Sure! I've changed it.

PR #10649, opened 2025-11-16
fix: Pointer down handling on mobile devices

tommoor:
closes https://github.com/outline/outline/issues/10582
closes https://github.com/outline/outline/issues/10605

### DIFF
--- a/app/editor/components/SuggestionsMenuItem.tsx
+++ b/app/editor/components/SuggestionsMenuItem.tsx
@@ -15,7 +15,7 @@ export type Props = {
   /** Whether the item is disabled */
   disabled?: boolean;
   /** Callback when the item is clicked */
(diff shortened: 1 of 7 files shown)


tommoor:
@salihudickson alt to https://github.com/outline/outline/pull/10608 - more directly fixes how the problem was introduced with new radix menus

app/editor/components/SuggestionsMenuItem.tsx
```diff
@@ -60,7 +60,7 @@ function SuggestionsMenuItem({
     <MenuButton
       ref={ref}
       disabled={disabled}
-      onClick={onClick}
+      onPointerDown={onPointerDown}
```
salihudickson, line 63: 🤔 I think it's a bit interesting that the editor can still focus on the correct node, just not with the click event.

PR #10650, opened 2025-11-16
Add optional S3 PUT uploads + per-folder reports for fashion model

oleksandr-diudiun:
@tommoor about #10609 

1. add an opt-in S3_SUPPORT_PUT_UPLOAD flag that switches presigned uploads from POST to PUT for S3-compatible backends like Cloudflare R2

2. propagate the new upload metadata through FileStorage.getPresignedPost, the attachments API, and the client uploader so PUT flows work transparently

### DIFF
--- a/app/utils/files.ts
+++ b/app/utils/files.ts
@@ -57,19 +57,8 @@ export const uploadFile = async (
   invariant(response, "Response should be available");
   const data = response.data;
   const attachment = data.attachment;
-  const formData = new FormData();
-
-  for (const key in data.form) {
-    formData.append(key, data.form[key]);
-  }
-
-  // @ts-expect-error ts-migrate(2339) FIXME: Property 'blob' does not exist on type 'File | Blo... Remove this comment to see the full error message
-  if (file.blob) {
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'file' does not exist on type 'File | Blo... Remove this comment to see the full error message
-    formData.append("file", file.file);
-  } else {
-    formData.append("file", file);
-  }
+  const method = (data.method ?? "POST").toUpperCase();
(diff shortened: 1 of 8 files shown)


tommoor:
I don't want to maintain additional configuration for this. S3, Minio, etc also supports the PUT verb for uploads from JS so I think as long as the local file storage endpoint is updated to support both that we can change the default to always be PUT and it will be fully backwards compatible without additional config or code branches.

oleksandr-diudiun:
> I don't want to maintain additional configuration for this. S3, Minio, etc also supports the PUT verb for uploads from JS so I think as long as the local file storage endpoint is updated to support both that we can change the default to always be PUT and it will be fully backwards compatible without additional config or code branches.

Totally agree with you, in this pull request it is optional feature lest say "experimental" optional feature so how about to try it out with official build and I will migrate normally from AWS to R2 my outlines and then will invest little bit time on next week to test it with other S3 services to be sure that it can be really simplified there and then will send one more pull request .

tommoor:
Sounds good, I won't merge until we have that simplified solution though

SapphoSys:
@oleksandr-diudiun any update / luck with other S3 services? i'd be down to test this with Backblaze B2 if needed

trungdlp:
We ran into this exact issue trying to use Cloudflare R2 with Outline — R2 doesn't support `CreatePresignedPost` (returns `501 Not Implemented`), so uploads fail completely. Since R2 does support presigned PUT, switching the default upload mechanism to PUT would fix R2 compatibility out of the box.

@tommoor's suggestion of making PUT the universal default (rather than an opt-in flag) sounds like the right call — S3, MinIO, and most S3-compatible providers all support presigned PUT. Is there anything blocking a follow-up PR taking that approach? Happy to help test if someone picks this up.

Issue #10500, opened 2025-10-28
Backlink to toolbar menu from link toolbar


apoorv-mishra:
> You can't actually comment(or trigger any other toolbar function for that matter) on a text which has a link attached to it because the toolbar menu itself doesn't open upon clicking the text. Every time you click the text, it opens up the link editor component.

PR #10659, opened 2025-11-18
Enable attributing links to images

### LINKED ISSUE #7420: Editor appears to allow linked images with CMD+K but link is not persisted

apoorv-mishra:
Closes #7420

### DIFF
--- a/shared/editor/components/Image.tsx
+++ b/shared/editor/components/Image.tsx
@@ -11,6 +11,7 @@ import { ResizeLeft, ResizeRight } from "./ResizeHandle";
 import useDragResize from "./hooks/useDragResize";
 import { useTranslation } from "react-i18next";
 import find from "lodash/find";
+import { Separator } from "~/components/Actions";
 
 type Props = ComponentProps & {
   /** Callback triggered when the image is clicked */
@@ -116,19 +117,25 @@ const Image = (props: Props) => {
         {!dragging && width > 60 && isDownloadable && (
           <Actions>
             {isExternalUrl(src) && (
-              <Button onClick={handleOpen} aria-label={t("Open")}>
-                <GlobeIcon />
-              </Button>
+              <>
+                <Button onClick={handleOpen} aria-label={t("Open")}>
+                  <GlobeIcon />
+                </Button>
+                <Separator height={24} />
+              </>
             )}
             {imgLink && !isSelected && (
-              <Button
-                // `mousedown` on ancestor `div.ProseMirror` was preventing the `onClick` handler from firing
-                onMouseDown={(e) => e.stopPropagation()}
-                onClick={props.onZoomIn}
-                aria-label={t("Zoom In")}
-              >
-                <ZoomInIcon />
-              </Button>
+              <>
+                <Button
+                  // `mousedown` on ancestor `div.ProseMirror` was preventing the `onClick` handler from firing
+                  onMouseDown={(e) => e.stopPropagation()}
+                  onClick={props.onZoomIn}
+                  aria-label={t("Zoom In")}
+                >
+                  <ZoomInIcon />
+                </Button>
+                <Separator height={24} />
+              </>
             )}
             <Button
               // `mousedown` on ancestor `div.ProseMirror` was preventing the `onClick` handler from firing
@@ -241,7 +248,6 @@ const Actions = styled.div`
   display: flex;
   align-items: center;
   position: absolute;
-  gap: 1px;
   top: 8px;
   right: 8px;
   opacity: 0;
--- a/app/components/Actions.ts
+++ b/app/components/Actions.ts
@@ -15,10 +15,10 @@ export const Action = styled(Flex)`
   }
 `;
 
-export const Separator = styled.div`
+export const Separator = styled.div<{ height?: number }>`
   flex-shrink: 0;
   width: 1px;
-  height: 28px;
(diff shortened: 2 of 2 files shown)


shared/editor/components/Image.tsx
```diff
@@ -164,18 +181,56 @@ const Image = (props: Props) => {
...
+                  ? "none"
+                  : "all",
+            }}
+            src={sanitizedSrc}
+            alt={node.attrs.alt || ""}
+            onError={() => {
+              setError(true);
+              setLoaded(true);
+            }}
+            onLoad={(ev: React.SyntheticEvent<HTMLImageElement>) => {
```
tommoor: I don't like all this duplication – is the img tag different in any notable way?

tommoor (commented):
Works well, but now the image toolbar is totally hidden if you have a link on the image – I think we need to wrap the work from https://github.com/outline/outline/issues/10500 into here.

apoorv-mishra:
> I think we need to wrap the work from https://github.com/outline/outline/issues/10500 into here

You suggesting that #10500 should precede this? Makes sense actually because you can't take other actions on an image with a link.

tommoor:
Maybe build ontop of this branch with this branch as base, would be best.

shared/editor/components/Image.tsx
apoorv-mishra: Me neither. The conditional rendering is a hack to hopefully cover all pointer related cases–for example, if you're in write mode and editor isn't focused, the cursor should become `pointer` upon hover and click should navigate to link. Again, if the editor is focused, `pointer-events` on `img` should be `none` so that a click selects the image.

Following are the cases:
- Editor
	- Write mode
		- Focused
			- Img node selected
				- `pointer-events: all`
				- `cursor: zoom-in`
			- Img node not selected
				- `pointer-events: none`
		- Not focused
			- Img with link
				- `pointer-events: all`
				- `cursor: pointer`
			- Img without link
				- `pointer-events: none`
	- Read-only mode
		- Img with link
			- `pointer-events: all`
			- `cursor: pointer`
		- Img without link
			- `pointer-events: all`
			- `cursor: zoom-in`

apoorv-mishra: There's an alt way that I explored, which won't need the `a` tag wrapping. But I'm not sure if that's any better. It's expressed as CSS mainly, as follows:

```css
.ProseMirror[contenteditable="true"] {
  &.ProseMirror-focused {
    .image-wrapper {
      &.ProseMirror-selectednode > img {
        pointer-events: all;
        cursor: zoom-in;
        /*border: 3px solid red;*/
      }
      &:not(.ProseMirror-selectednode) > img {
        pointer-events: none;
        /*border: 3px solid aqua;*/
      }
    }
    /*border: 3px solid red;*/
  }
  &:not(.ProseMirror-focused) {
    .image-wrapper  {
       & > img[data-url] {
          pointer-events: all;
          cursor: pointer;
          /*border: 3px solid orange;*/
       }
       & > img:not([data-url]) {
          pointer-events: none;
          /*border: 3px solid pink;*/
       }
    }
    /*border: 3px solid green;*/
  }
}

.ProseMirror[contenteditable="false"] {
  .image-wrapper  {
    & > img[data-url] {
      pointer-events: all;
      cursor: pointer;
      /*border: 3px solid orange;*/
    }
    & > img:not([data-url]) {
      pointer-events: all;
      cursor: zoom-in;
      /*border: 3px solid pink;*/
    }
  }
  /*border: 3px solid blue;*/
}

```

tommoor: I would like to retain the a tag as it's more correct. But perhaps the pointer cases can be covered with less duplication of unrelated props

apoorv-mishra: Sure, will take a look 👀

apoorv-mishra:
Considering edge cases, a review would be helpful here before I use this as base for #10500 @tommoor

shared/editor/components/Styles.ts
```diff
@@ -895,6 +892,41 @@ h6:not(.placeholder)::before {
...
+.ProseMirror[contenteditable="false"] {
+  .image-wrapper  {
+    & > a[href] {
+      cursor: pointer;
+    }
+    & > a:not([href]) {
+      cursor: zoom-in;
+    }
+  }
+}
```
apoorv-mishra, line 928: Made the decision to keep all this in shared styles instead of the image component to easily toggle pointer styles based on editor focus. The rationale was that the image component's re-rendering doesn't need to trigger based on editor focus change.

tommoor:
Noting a small styling issue here:
image

In edit mode this zoom button doesn't appear making it impossible to zoom in, probably just have it in both cases?

apoorv-mishra:
You can zoom in when the image is selected in edit mode.

tommoor:
You can, but it's very difficult to select when there is a link

tommoor:
Superseded by https://github.com/outline/outline/pull/10762

PR #10688, opened 2025-11-23
fix: Forward compatibility for upcoming custom emoji logic

tommoor:
It is a good idea to get this out to clients ahead of custom emoji so they have a better fallback once that branch lands

### DIFF
--- a/shared/components/EmojiIcon.tsx
+++ b/shared/components/EmojiIcon.tsx
@@ -15,9 +15,11 @@ type Props = {
  * in a way that can be used wherever an Icon would be.
  */
 export default function EmojiIcon({ size = 24, emoji, ...rest }: Props) {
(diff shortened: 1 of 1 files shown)


shared/components/EmojiIcon.tsx
```diff
@@ -15,9 +15,11 @@ type Props = {
  * in a way that can be used wherever an Icon would be.
  */
 export default function EmojiIcon({ size = 24, emoji, ...rest }: Props) {
+  const isUrl = emoji.includes("/");
+
   return (
     <Span $size={size} {...rest}>
-      <SVG size={size} emoji={emoji} />
+      <SVG size={size} emoji={isUrl ? "�" : emoji} />
```
tommoor, line 22: See: https://en.wikipedia.org/wiki/Specials_(Unicode_block)

Issue #10712, opened 2025-11-24
When you tap the name of the collection it doesn't just collapse, it also closes the sidebar.


tommoor:
I found one minor issue on mobile specifically, which is that when you tap the name of the collection that you're already viewing (to collapse the tree), it doesn't just collapse, it also closes the sidebar. Could you make the sidebar remain open in this case?

_Originally posted by @RyanEwen in https://github.com/outline/outline/discussions/10482#discussioncomment-15065358_

salihudickson:
🤔 Seems more like a design enhancement than a bug, because nothing is actually behaving incorrectly, it just wasn’t part of the original design. 

I’ll put together a PR for this.

PR #10713, opened 2025-11-24
fix: click name to toggle expanded when on collection or document page

salihudickson:
closes: #10712

### DIFF
--- a/app/components/Sidebar/components/NavLink.tsx
+++ b/app/components/Sidebar/components/NavLink.tsx
@@ -162,13 +162,11 @@ const NavLink = ({
 
   const handleClick = React.useCallback(
     (event: React.MouseEvent<HTMLAnchorElement>) => {
-      const toPathname = typeof to === "string" ? to : to?.pathname;
-
-      if (toPathname && location.pathname?.startsWith(toPathname)) {
+      if (isActive) {
         event.preventDefault();
       }
     },
-    [to, location.pathname]
+    [isActive]
   );
 
   React.useEffect(() => {


tommoor:
The toggle expanded part already works correctly, that was all the work here:
https://github.com/outline/outline/pull/10684

The bug is related to the sidebar closing on mobile in this scenario. Hoping we can achieve this without an additional location check

tommoor:
I will take a look too as I was literally just working on this

salihudickson:
😓 my bad. Hadn't updated done a fetch so I didn't realize this was already done.

salihudickson:
> Hoping we can achieve this without an additional location check.

For documents, yeah. But for collections, can't really think of a way around it. Assuming you only want to keep it open when on the collection page, not just when the collection is active.

app/components/Sidebar/components/CollectionLink.tsx
```diff
@@ -121,17 +122,22 @@ const CollectionLink: React.FC<Props> = ({
...
     <ActionContextProvider value={{ activeCollectionId: collection.id }}>
       <Relative ref={mergeRefs([parentRef, dropRef])}>
         <DropToImport collectionId={collection.id}>
           <SidebarLink
             onClick={onClick}
-            to={{
-              pathname: collection.path,
-              state: { sidebarContext },
-            }}
+            to={
```
tommoor: What if we push this logic down into `NavLink`? It already has `location`

salihudickson: Definitely makes more sense. Consider it done 👍🏾

app/components/Sidebar/components/NavLink.tsx
```diff
@@ -160,6 +160,17 @@ const NavLink = ({
     [onClick, navigateTo, isActive, shouldFastClick]
   );
 
+  const handleClick = React.useCallback(
+    (event: React.MouseEvent<HTMLAnchorElement>) => {
+      const toPathname = typeof to === "string" ? to : to?.pathname;
+
+      if (toPathname && location.pathname?.startsWith(toPathname)) {
```
salihudickson: This is admittedly a bit broad, since it applies to all NavLinks. But outline doesn't use nested pathnames so I think it should be fine.

Any thoughts?

tommoor: `isActive` exists in scope too ;)

tommoor: Maybe it's as simple as `to={isActive ? undefined : to}`

salihudickson: Still not sure why, but when i first tested the `isActive` prop i got a few false negatives and just thought it was for something else. 

The `to` prop cannot be undefined. And conditional rendering seemed like more work (cause of styling). From what I can tell, the only  way to deactivate it would be `preventDefault` or to disable pointer events with css

PR #10721, opened 2025-11-25
feat: Add popularity scoring

### LINKED ISSUE #10720: Add cached `popularityScore` to documents
- Add a new `popularityScore` float field to `documents` model (SkipChangeset)
- Add a scheduled task that will run once an hour and recalculate scores
- The score should take into account revisions, comments, and views with a precipitous falloff based on time down to zero after a month.
- Scores only need to be calculated for documents that have received interaction in the last month

Once this is landed a follow up will be to create a "Popular" tab on home/collections and enable sorting by popularity in search

tommoor:
Just speed-ran this feature in 45mins using Opus 👌 – adds:

- `popularityScore` to documents, calculated hourly for docs with recent comments,views,revisions
- `popularityScore` as a sort open in API
- "Popular" tab to Home and Collection pages
- Boosts search result rankings based on score

closes #10720

### DIFF
--- a/server/queues/tasks/UpdateDocumentsPopularityScoreTask.ts
+++ b/server/queues/tasks/UpdateDocumentsPopularityScoreTask.ts
@@ -1,20 +1,26 @@
+import crypto from "crypto";
 import { subWeeks } from "date-fns";
-import { Op } from "sequelize";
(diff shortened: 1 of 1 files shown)


jaustin-ant:
Thank you so much Tom!

PR #10736, opened 2025-11-27
chore: Add cron task partitioning

### LINKED ISSUE #10733: Add method to partition scheduled tasks
We're starting to have more and more tasks that take a long time to run on the large data set on cloud. We need a way to keep them scheduled on a regular basis but partition over a longer period of time instead of running everything on the hour. 

This became particularly pressing with the task for updating popularity

tommoor:
The partitioning system enables horizontal scaling by distributing database queries across time and worker instances. It divides the UUID primary key space into equal ranges, and then spreading execution across a defined time range with each individual task run computing a portion of the work based on the partitioned ranged.

I ended up wrapping in some small refactors here that made sense once `CronTask` is broken out into it's own class.

closes #10733

### DIFF
--- a/server/queues/tasks/CleanupExpiredFileOperationsTask.ts
+++ b/server/queues/tasks/CleanupExpiredFileOperationsTask.ts
@@ -7,14 +7,8 @@ import { TaskPriority } from "./base/BaseTask";
 import { CronTask, Props, TaskInterval } from "./base/CronTask";
 
 export default class CleanupExpiredFileOperationsTask extends CronTask {
-  public async perform({ limit, partition }: Props) {
-    const partitionInfo = partition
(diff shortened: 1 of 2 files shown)


server/services/cron.ts
```diff
@@ -1,27 +1,36 @@
...
-  setInterval(() => void run(TaskSchedule.Minute), Minute.ms);
+  setInterval(() => void run(TaskInterval.Day), Day.ms);
+  setInterval(() => void run(TaskInterval.Hour), Hour.ms);
 
   // Just give everything time to startup before running the first time. Not
   // _technically_ required to function.
   setTimeout(() => {
-    void run(TaskSchedule.Day);
-    void run(TaskSchedule.Hour);
-    void run(TaskSchedule.Minute);
```
tommoor, line 25: Never used, removing to avoid confusion

Issue #10681, opened 2025-11-21
Unable to delete text in a table


Ali-ovo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

When I write the text and paste the image, the first word of the text cannot be deleted

### Expected Behavior

Can be deleted normall

### Steps To Reproduce

https://github.com/user-attachments/assets/93058106-ae11-4a25-b662-0536b7523270

### Environment

```markdown
- Outline: 1.1.0
- Browser: Chrome
```

tommoor:
Very strange, as a workaround you can highlight the letter and delete

villAsh:
Can I work on this one?

tommoor:
If you can see the issue go for it, first to a PR wins :)

PR #10513, opened 2025-10-29
feat: Custom emoji

salihudickson:
## Description
This PR allows users to create and use custom emojis.

fixes: #9278
replaces: #10469

## Video Demo
https://github.com/user-attachments/assets/38dac3d1-5635-4c3f-a95e-5fdfa93517cd

PR #10758, opened 2025-12-01
feat: Custom emojis in editor

### LINKED ISSUE #9278: Allow adding custom emojis

tommoor:
## Todo

- [x] Test all text sizes
- [x] Copy and paste behavior
- [x] Refactor to reduce `isUUID` calls
- [x] Public shares
- [x] Click on emoji atom
- [x] Editing issues with atom as first character in header

Follow on from #10513 – ~there is still some editing finickiness around custom emojis at the beginning of nodes. It is very similar to #10681 so probably makes sense to tackle separately.~

closes #9278

### DIFF
--- a/app/scenes/Settings/components/EmojisTable.tsx
+++ b/app/scenes/Settings/components/EmojisTable.tsx
@@ -80,7 +80,7 @@ const EmojisTable = observer(function EmojisTable({
           width: "50px",
         },
       ]),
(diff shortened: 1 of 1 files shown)


tommoor:
Damn, the editing issue I mentioned in the description is this one – https://github.com/ProseMirror/prosemirror/issues/1189

PR #10759, opened 2025-12-01
 fix: Unable to delete last character before inline image (#10681)

### LINKED ISSUE #10681: Unable to delete text in a table

nwleedev:
## Summary

Fixes #10681 

When text and an inline image exist in the same paragraph (inside a table cell or anywhere else), the last character before the image cannot be deleted using Delete or Backspace keys.

Added a new extension `DeleteNearAtom` that explicitly handles Delete/Backspace.
- Delete: cursor is at `|x[image]` position (1 character before atom)
- Backspace: cursor is at `x|[image]` position (1 character before atom)

## How to validate

- Create a table and insert an image using `/image` command
- Move cursor before the image and type some text (e.g., "Hello")
  1. Position cursor before the character `H` and press Delete repeatedly
  2. Verify all characters including the last one are deleted

- Move cursor before the image and type another text like "World"
  1. Position the cursor after the character `d` and press Backspace repeatedly.
  2. Test the same scenario outside of tables (regular paragraph)
- Verify normal text deletion still works when no atom node is adjacent

### DIFF
--- a/shared/editor/extensions/DeleteNearAtom.ts
+++ b/shared/editor/extensions/DeleteNearAtom.ts
@@ -1,0 +1,90 @@
+import { Command, TextSelection } from "prosemirror-state";
+import Extension from "../lib/Extension";
+
+/**
+ * GitHub Issue: https://github.com/outline/outline/issues/10681
+ */
+export default class DeleteNearAtom extends Extension {
+  get name() {
+    return "deleteNearAtom";
+  }
+
+  keys(): Record<string, Command> {
+    return {
+      Delete: deleteForwardNearAtom(),
+      Backspace: deleteBackwardNearAtom(),
+    };
+  }
+}
+
+function deleteForwardNearAtom(): Command {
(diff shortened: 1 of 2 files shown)


tommoor (commented):
This does seem to work well, though I'm not sure why it's required at all.

Have you dealt with this issue before? Or was there AI assisted?

nwleedev:
@tommoor 

I have tried to find alternatives to Notion for better user experiences and lower cost. 

After using it for hours, several issues were discovered. Some of them are a little frustrating such as table editing.
That's why I began contributing to this repository by working on smaller issues.

PR #10769, opened 2025-12-03
feat: Allow PKCE clients to refresh tokens

### LINKED ISSUE #10768: Allow PKCE clients to refresh tokens
Allow PKCE (public) clients to refresh access tokens without providing a client_secret.

### Why
PKCE clients are public clients (e.g., mobile apps, SPAs) that cannot safely store secrets. They authenticate via code_verifier during the initial token exchange, but the current OAuth2 server configuration requires client authentication for all grant types, including refresh_token.

This blocks PKCE clients from refreshing their tokens since they have no secret to provide. The fix configures oauth2-server to skip client authentication for the refresh_token grant while still requiring it for authorization_code.

tommoor:
Introduces the concept of a client type, which can be switched between confidential and public.

Public clients are not expected to include a client secret in authentication flows. All existing clients default to confidential for safety.

closes #10768

### DIFF
--- a/server/routes/oauth/index.ts
+++ b/server/routes/oauth/index.ts
@@ -93,6 +93,9 @@ router.post(
           "Missing refresh_token for refresh_token grant type"
         );
       }
+      if (!clientId) {
+        throw ValidationError("Missing client_id for refresh_token grant type");
+      }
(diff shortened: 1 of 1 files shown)


tommoor:
cc @olu-an

Issue #9855, opened 2025-08-06
Copy as Markdown: Image links should use absolute URLs instead of relative paths


tommoor:
### Discussed in https://github.com/outline/outline/discussions/9852

<div type='discussions-op-text'>

<sup>Originally posted by **fandych** August  6, 2025</sup>
Hi Outline team,

I'm using a self-hosted Outline instance. When I use the "Copy as Markdown" feature, I notice that image links in the generated Markdown are using relative paths (e.g., /uploads/xxxx.png). This causes issues when I paste the content into other systems (such as my own AI application), because the images cannot be displayed unless the base URL is the same as my Outline instance.

**Expected behavior**:
It would be very helpful if the "Copy as Markdown" feature could generate image links with absolute URLs (e.g., https://my-outline-instance.com/uploads/xxxx.png). This way, when I use the copied Markdown in other applications, the images will display correctly.

**Use case:**
I am integrating Outline content into my own AI system. When the AI returns Markdown content, I want to ensure that images are visible to users, regardless of where the Markdown is rendered.

**Steps to reproduce**:

- Open a document with images in your self-hosted Outline.
- Click "Copy as Markdown".
- Paste the Markdown into another application.
- Notice that image links are relative and do not display unless the base URL matches Outline.

**Suggested solution**:
Please consider updating the "Copy as Markdown" feature to use absolute URLs for images, based on the current Outline instance's base URL.

Thank you for your great work!

</div>

fandych:
Hi team,

Thank you for implementing the feature to generate absolute URLs for images in the "Copy as Markdown" function. This has improved the integration experience with other systems.

However, I have encountered a new issue:

Currently, the absolute image URLs generated by "Copy as Markdown" use the /api/attachments.redirect endpoint, for example:

```
![](http://my-outline-instance.com/api/attachments.redirect?id=ff1ec1c8-4467-4ebe-a943-0d3bc82b62b1)
```

When I use these links in external applications or share the Markdown content, the images cannot be displayed unless the user is authenticated with Outline. Accessing these URLs without being logged in returns a 401 Unauthorized error.

#### Expected behavior:
It would be very helpful if there was a way to allow public (or token-based) access to images referenced in Markdown exports, so that they can be displayed in external systems without requiring Outline authentication.

#### Use case:
I am integrating Outline content into my own AI system. When the AI returns Markdown content, I want to ensure that images are visible to users, regardless of whether they are logged into Outline.

#### Steps to reproduce:

[...]

PR #10821, opened 2025-12-07
feat: Use signed urls in "copy as markdown"

libnewton:
This uses signed URLs in images, attachments, etc. for the "Copy Markdown" feature.
As complained in https://github.com/outline/outline/issues/9855#issuecomment-3173245557, the `` URLs do not work in external markdown programs.

The "Copy as Markdown" feature now uses the `documents.export` route with a special option to enable signing the attachment URLs. The TTL is configurable with `FILE_STORAGE_SIGNED_URL_DURATION`, defaults to 1h.


server/env.ts
```diff
@@ -639,6 +639,14 @@ export class Environment {
...
     1000000;
 
+  /**
+   * The duration in seconds for which signed URLs are valid.
+   */
+  @IsOptional()
+  @IsNumber()
+  public FILE_STORAGE_SIGNED_URL_DURATION =
+    this.toOptionalNumber(environment.FILE_STORAGE_SIGNED_URL_DURATION) ?? 3600;
+
```
tommoor: This isn't deserving of an installation level configuration, we should just choose a reasonable value.

app/actions/definitions/documents.tsx
```diff
@@ -614,10 +615,11 @@ export const copyDocumentAsMarkdown = createAction({
       ? stores.documents.get(activeDocumentId)
       : undefined;
     if (document) {
-      const { ProsemirrorHelper } = await import(
-        "~/models/helpers/ProsemirrorHelper"
-      );
-      copy(ProsemirrorHelper.toMarkdown(document));
+      const res = await client.post("/documents.export", {
+        id: document.id,
+        signedUrls: true,
```
tommoor:
```suggestion
        signedUrls: 3600,
```

We could pass a number through here.

server/routes/api/documents/documents.ts
```diff
@@ -760,6 +762,25 @@ router.post(
     let contentType: string;
     let content: string;
 
+    const toMarkdown = async () => {
+      if (signedUrls) {
+        const data = await DocumentHelper.toJSON(document, {
```
tommoor: Push this down into the existing `DocumentHelper.toMarkdown` method so it against a `signedUrls` option

server/routes/api/documents/schema.ts
```diff
@@ -152,7 +152,9 @@ export const DocumentsInfoSchema = BaseSchema.extend({
 export type DocumentsInfoReq = z.infer<typeof DocumentsInfoSchema>;
 
 export const DocumentsExportSchema = BaseSchema.extend({
-  body: BaseIdSchema,
+  body: BaseIdSchema.extend({
+    signedUrls: z.boolean().optional(),
```
tommoor: Or number

tommoor (changes requested):
Thanks for the effort

server/routes/api/documents/documents.ts
tommoor: This was marked as resolved but is not yet resolved

tommoor: @libnewton did you want to address the last feedback here to make this mergable

libnewton: @tommoor sure, did that. This made `toMarkdown()` async, thus more file changes, but that should be fine.

libnewton: any updates on this @tommoor

tommoor: The tests are failing here

libnewton: yes because the interface changed. It expected promise, but we now have to await as signing urls nessesitates async. Thus it now resolves beforehand. Should be an one-word fix in the testcase.

libnewton: @tommoor Fixed all tests/issues. Also rebased with upstream to make it work with revision exporting too.

PR #10822, opened 2025-12-07
fix: overflow tables and math on (pdf) print

libnewton:
This fixes tables and math overflow. Only affects print mode (media query enabled).

**Math overflow**:
When creating large math equations, they overflow into a scrollable element. Letting it overflow a bit is fine, as one usually does not create longer math equations anyway. Right now, in printing, they are cut. This is bad.

**Tables**
A more involved approach is needed. Only in print mode, the table is scaled down to fit the width.
before:
image

after:
image


libnewton:
bump

PR #10827, opened 2025-12-08
Utilize GitHub integration to fetch information about public issues/PRs

### LINKED ISSUE #10518: Enable GitHub integration to fetch information from public repositories
### Discussed in https://github.com/outline/outline/discussions/10516

<sup>Originally posted by **stephanrenggli** October 30, 2025</sup>
Currently the GitHub integration seems to only fetch information from the user/organization that created the application for the integration.
So if other GithHub links are mentioned in a document, the hover-over does not show any information. The `url.unfurl` request for public repos also returns a 204, while the one for a repo in the connected organization returns a 200.

image

Im not sure this is the intended behavior, but the flavor text when configuring the GitHub integration suggests this is intended.

> Enable previews of GitHub issues and pull requests in documents by connecting a GitHub organization or specific repositories to Outline.

It would be nice, if information on public issues and PRs were also displayed in the hover-over.</div>

apoorv-mishra:
Closes #10518

### DIFF
--- a/plugins/github/server/github.ts
+++ b/plugins/github/server/github.ts
@@ -239,11 +239,14 @@ export class GitHub {
       order: [
         [
           Sequelize.literal(
-            `CASE WHEN "settings"->'github'->'installation'->'account'->>'name' = '${resource.owner.replace(/'/g, "''")}' THEN 0 ELSE 1 END`
+            `CASE WHEN "settings"->'github'->'installation'->'account'->>'name' = :owner THEN 0 ELSE 1 END`
           ),
           "ASC",
         ],
       ],
+      replacements: {
+        owner: resource.owner,
+      },
     })) as Integration<IntegrationType.Embed>;
 
     if (!integration) {


shared/editor/nodes/Mention.tsx
```diff
@@ -148,13 +149,38 @@ export default class Mention extends Node {
             onChangeUnfurl={this.handleChangeUnfurl(props)}
           />
         );
-      case MentionType.URL:
+      case MentionType.URL: {
+        const href = props.node.attrs.href;
+        // This is for backwards compatibility so that older public github issue/pr links
+        // can be accordingly dealt with, instead of being rendered as generic mention URLs
```
apoorv-mishra: Kinda weird, open for suggestions.

tommoor: It's probably fine to leave these unmodified if they're already in the document 🤷

apoorv-mishra: You can't display issue status icon and id then

Screenshot 2025-12-08 at 8 46 47 PM

tommoor: I understand, but there won't be very many existing mentions like this and this code has to be maintained forever – lets just respect the existing data in the attributes

plugins/github/server/github.ts
```diff
@@ -233,7 +233,6 @@ export class GitHub {
       where: {
         service: IntegrationService.GitHub,
         teamId: actor.teamId,
-        "settings.github.installation.account.name": resource.owner,
```
tommoor, line 236: This works if you have one installation, but lets say you have two installs it will always return the first.

So we want to _prefer_ the one here the `resource.owner` matches and fallback to any

apoorv-mishra: Good catch! Thanks 🙏

plugins/github/server/github.ts
```diff
@@ -229,11 +230,20 @@ export class GitHub {
...
 
+    // Find integration, prioritizing one where the installation account matches the resource owner
     const integration = (await Integration.findOne({
       where: {
         service: IntegrationService.GitHub,
         teamId: actor.teamId,
       },
+      order: [
+        [
+          Sequelize.literal(
```
tommoor, line 241: Too dangerous to inject the owner into a literal like this. If we can't use replacements then I'd rather query all the integrations and match in js

PR #10840, opened 2025-12-09
Apply `react/rules-of-hooks`

apoorv-mishra:
[Ref](https://github.com/outline/outline/pull/10836#issue-3707716199)

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -94,7 +94,8 @@
             "args": "after-used",
             "ignoreRestSiblings": true
           }
-        ]
+        ],
+        "react/rules-of-hooks": "error"
       },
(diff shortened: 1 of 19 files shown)


tommoor:
Did you try? It doesn't seem to make a difference – pretty sure this rule is already enabled as part of `react` plugin

apoorv-mishra:
https://github.com/user-attachments/assets/b6524787-6b34-4d44-a6eb-0fc047c73968

tommoor:
Weird, will try again

apoorv-mishra:
Also, I don't see the rule turned on by default in the list [here](https://oxc.rs/docs/guide/usage/linter/rules.html).

PR #10762, opened 2025-12-02
Backlink to toolbar menu

apoorv-mishra:
Closes #10500, #7420, #10659

PR #10844, opened 2025-12-10
fix: Code language picker missing

### LINKED ISSUE #10843: Language picker drop-down and copy button are not appearing for code blocks

### Current Behavior

In Edit Mode, when I click a code block, no coding-language drop-down menu appears.  When not in Edit Mode, when I click a code block, no "Copy" button appears.

### Expected Behavior

The language drop-down menu and "Copy" button should appear when clicking any code block.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Outline: the hosted version
- Browser: Chrome version 143.0.7499.41 (arm64) on macOS
```

### Anything else?

_No response_

tommoor:
closes #10843
regressed in https://github.com/outline/outline/pull/10762

### DIFF
--- a/app/editor/components/SelectionToolbar.tsx
+++ b/app/editor/components/SelectionToolbar.tsx
@@ -82,6 +82,7 @@ export function SelectionToolbar(props: Props) {
   );
 
   React.useEffect(() => {
+    const { selection } = state;
     const linkMark =
       selection instanceof NodeSelection
         ? getMarkRangeNodeSelection(selection, state.schema.marks.link)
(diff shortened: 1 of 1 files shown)


app/editor/components/SelectionToolbar.tsx
```diff
@@ -254,10 +259,6 @@ export function SelectionToolbar(props: Props) {
     setActiveToolbar(null);
   };
 
-  if (!activeToolbar || !items.length) {
```
tommoor, line 257: This also fixes a bug that I think was existing where the first click into a code block would not show the menu due to `menuRef` being undefined

app/editor/components/SelectionToolbar.tsx
apoorv-mishra, line 1: This seems to have broken the `Cmd-k` shortcut and requires two clicks to open link editor. Plus, if you close the empty link editor, you end up with a `FloatingToolbar` with no items, that looks like a downward arrow in UI.

tommoor: Yea I realized :( 

This stuff is such a mess

PR #10850, opened 2025-12-10
Add SOCKS_PROXY config option

TimRepke:
This PR adds an option to configure a proxy for the SMTP server for users who host outline on a server that blocks common SMTP ports and the option to set up an SSH tunnel to forward the traffic.

Would be great to also have this available upstream so I don't always have to merge updates and also to make this available to others.

### DIFF
--- a/app.json
+++ b/app.json
@@ -203,6 +203,10 @@
       "description": "Override SMTP cipher configuration (optional)",
       "required": false
     },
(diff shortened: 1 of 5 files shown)


TimRepke:
Yes, this runs on a server like this and happily sends emails through the proxy.

tommoor:
There's an outstanding issue with types

Issue #6795, opened 2024-04-12
GitLab integration


tommoor:
After https://github.com/outline/outline/issues/6406 similar to #4427

almereyda:
To note, there is a basic GitLab integration _snippets_ from the hosted instance at `gitlab.com`, which was contributed with #6217, #6290 and https://github.com/outline/outline/commit/8b6a760e02a5a30d4f08db9e34682fb4d6840f0a

The use case for self-hosted snippets is documented in:

- https://github.com/outline/outline/discussions/6427

with some recent activity in the linked

- #6317

that leads us to

- https://github.com/IamTaoChen/outline/tree/gitlab_new

thedarkside:
That is what we are waiting for! 

I would like to outline our requirements for a good integration here. 

Besides the ability to embed issues, epics and merge requests from a selfhosted gitlab instance it would be beneficial when the embeds **update automatically** showing the current status like if an issue is open or closed, when was the last change etc. 

Because our instance is private only, outline would need an api key to fetch the information.

Is this kind of integration planned?

almereyda:
#6406 has landed with #6414, which could provide as a nice blue print for the API surface of the integration to cover on Outline's side. It should be not too hard to find equivalents for the API surface used on GitHub's side in GitLab.

JoaquinLop:
It would be great to have this functionality.

rausub:
i concur it would be nice to have this

almereyda:
@tommoor re: https://github.com/outline/outline/issues/6407#issuecomment-2652602393

In case one wanted to continue here from the community-side of things, should we also wait for this "solution to be pluggable" or might GitLab rather just be the next work item bringing with it the decoupling mentioned in the linked comment, also paving the way for Linear and sorts?

tommoor:
> should we also wait for this "solution to be pluggable"

Yep, not sure on the priority of GitLab support itself for the maintainer team, it's not requested all that often

almereyda:
For us it is a priority and I would put in significant effort into porting any existing integration to GitLab, which is mostly copy and paste and reading the documentation about API endpoints.

Will return here when the "pluggable solution" exists and can be adapted from.

almereyda:
With #6406 and #6407 having landed in v0.84.0, the required prior art for an adaptation to GitLab _and potentially other systems_ #6507 has arrived.

alphanso:
[...]

PR #10861, opened 2025-12-11
feat: GitLab integration

salihudickson:
closes: #6795

### DIFF
--- a/plugins/gitlab/shared/GitLabUtils.ts
+++ b/plugins/gitlab/shared/GitLabUtils.ts
@@ -4,6 +4,8 @@ import { integrationSettingsPath } from "@shared/utils/routeHelpers";
 import { UnfurlResourceType } from "@shared/types";
 
 export class GitLabUtils {
+  public static defaultGitlabUrl = "https://gitlab.com";
+
   private static supportedResources = [
     UnfurlResourceType.Issue,
     UnfurlResourceType.PR,
@@ -16,7 +18,7 @@ export class GitLabUtils {
    * @returns The GitLab URL to use.
    */
   private static getGitlabUrl(customUrl?: string): string {
-    return customUrl || "https://gitlab.com";
+    return customUrl || this.defaultGitlabUrl;
   }
 
   /**


plugins/gitlab/client/index.tsx
tommoor, line 1: Capitalization of this file is incorrect - _must_ be lowercase.

plugins/gitlab/server/env.ts
```diff
@@ -0,0 +1,36 @@
...
+   * GitLab OAuth2 client credentials. To enable integration with GitLab.
+   */
+  @Public
+  @IsOptional()
+  public GITLAB_CLIENT_ID = this.toOptionalString(environment.GITLAB_CLIENT_ID);
+
+  @Public
+  @IsOptional()
+  @CannotUseWithout("GITLAB_CLIENT_ID")
+  public GITLAB_APP_NAME = this.toOptionalString(environment.GITLAB_APP_NAME);
```
tommoor: No need for this

plugins/gitlab/server/env.ts
```diff
@@ -0,0 +1,36 @@
+import { IsOptional } from "class-validator";
+import { Environment } from "@server/env";
+import { Public } from "@server/utils/decorators/Public";
+import environment from "@server/utils/environment";
+import { CannotUseWithout } from "@server/utils/validators";
+
+class GitLabPluginEnvironment extends Environment {
```
tommoor, line 7: We could add `GITLAB_URL` to support self-hosted instances, default to cloud url

plugins/gitlab/server/index.ts
```diff
@@ -0,0 +1,32 @@
...
+      value: new GitLabIssueProvider(),
+    },
+    {
+      type: Hook.UnfurlProvider,
+      value: { unfurl: GitLab.unfurl, cacheExpiry: Minute.seconds },
+    },
+    {
+      type: Hook.Uninstall,
+      value: uninstall,
+    },
```
tommoor: No need for the fake hook, just move the comment here that uninstall is not required.

plugins/gitlab/client/components/Icon.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+
+type Props = {
+  /** The size of the icon, 24px is default to match standard icons */
+  size?: number;
+  /** The color of the icon, defaults to the current text color */
+  fill?: string;
+};
+
+export default function Icon({ size = 24, fill = "currentColor" }: Props) {
```
tommoor, line 10: This SVG is pretty broken visually. I can sort it if you're not able

plugins/gitlab/shared/GitLabUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+   * @returns {string} - The full URL to redirect the user to GitLab's OAuth authorization page.
+   */
+  static authUrl(state: string): string {
+    const baseUrl = `https://gitlab.com/oauth/authorize`;
+    const params = {
+      client_id: this.clientId,
+      redirect_uri: this.callbackUrl(),
+      response_type: "code",
+      state,
+      scope: "api read_api read_user",
```
tommoor: Is this the minimum scopes? I think `api` means write access 🤔

plugins/gitlab/shared/GitLabUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Generates the authorization URL for GitLab OAuth.
+   *
+   * @param {string} state - A unique state string to prevent CSRF attacks and maintain state between the request and callback.
+   * @returns {string} - The full URL to redirect the user to GitLab's OAuth authorization page.
+   */
+  static authUrl(state: string): string {
+    const baseUrl = `https://gitlab.com/oauth/authorize`;
```
tommoor: We'd use `GITLAB_URL` env here, so needs `Public` decorator

tommoor (changes requested):
image

~Haven't gotten it to work yet, but might be something on my end~ Missing `api` scope

plugins/gitlab/shared/GitLabUtils.ts
tommoor: Ah this scope is why it was failing for me as I didn't whitelist in application registration

plugins/gitlab/server/env.ts
```diff
@@ -0,0 +1,36 @@
...
+  @CannotUseWithout("GITLAB_CLIENT_ID")
+  public GITLAB_CLIENT_SECRET = this.toOptionalString(
+    environment.GITLAB_CLIENT_SECRET
+  );
+
+  @IsOptional()
+  @CannotUseWithout("GITLAB_CLIENT_ID")
+  public GITLAB_WEBHOOK_SECRET = this.toOptionalString(
+    environment.GITLAB_WEBHOOK_SECRET
+  );
```
tommoor: Not needed

plugins/gitlab/server/gitlab.ts
```diff
@@ -0,0 +1,430 @@
...
+      return;
+    }
+
+    // GitLab URLs: /owner/repo/-/issues/123 or /owner/repo/-/merge_requests/123
+    const parts = pathname.split("/").filter(Boolean);
+    if (parts.length < 4) {
+      return;
+    }
+
+    const owner = parts[0];
```
tommoor: This is not reliably the owner, e.g. in my case it's a group:

image

plugins/gitlab/server/gitlab.ts
```diff
@@ -0,0 +1,430 @@
...
+    const resource = GitLab.parseUrl(url);
+    if (!resource) {
+      return;
+    }
+
+    const integration = (await Integration.findOne({
+      where: {
+        service: IntegrationService.GitLab,
+        teamId: actor.teamId,
+        "settings.gitlab.installation.account.name": resource.owner,
```
tommoor:
```suggestion
```

We can remove this, it's not possible to detect the owner from the url, just have it try and load the data.

plugins/gitlab/client/index.tsx
salihudickson: 😅 spent a few hours trying to figure this out earlier. It's fixed locally though, not sure why the recent commit didn't fix it here.

tommoor: Because git is case insensitive by default – easiest workaround is rename it to something different and back

plugins/gitlab/server/index.ts
salihudickson: Will do 👍🏾

plugins/gitlab/client/components/Icon.tsx
salihudickson: I just generated that with AI. I've already replaced it locally with their official logo.

plugins/gitlab/shared/GitLabUtils.ts
salihudickson: You're right. I kept hitting 401s so I added that, but I think it's cause I gave too many permissions to my application.

I'll take it out.

plugins/gitlab/server/gitlab.ts
salihudickson: The repo url still follows the format So the owner, in this case, would refer to the group.

plugins/gitlab/server/gitlab.ts
salihudickson: oh yeah. I thought you were referring to a url issue.

salihudickson:
Thanks for the early review!

Still have quite a bit of clean up to do here. And there are some instances where the functionality still breaks that I need to straighten out.

plugins/gitlab/server/gitlab.ts
salihudickson: I was just working on webhook event handling and I noticed that we don’t have a good way to update the correct integration. Since we’re not getting owner data, we can’t use the user ID to identify which integration to update.

Would it be better to store the owner on the installation as the org or group (namespace) the repo is under? That should make it easier to accurately identify the integration.

tommoor: Whatever is neccessary, it doesn't need to match GitHub 1-1, so if the namespace makes sense for GitLab that's totally fine

plugins/gitlab/server/GitLabIssueProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+    payload: Record<string, any>,
+    eventName: string
+  ) {
+    let integration: Integration<IntegrationType.Embed> | null = null;
+
+    for (const owner of payload.owners) {
+      integration = (await Integration.findOne({
+        where: {
+          service: IntegrationService.GitLab,
+          "settings.github.installation.account.name": owner.name,
```
salihudickson: @tommoor, this is an admittedly a flawed way to retrieve the integration, since this value is just a display name and users can (and do) change it, which would break the lookup.

An alternative would be to update all instances of the project across all GitLab integrations, though that approach has its own drawbacks. Gitlab doesn't really provide much identifying information for the project owners in these hooks.

thoughts?

tommoor:
Kind of waiting for all the tests etc to be passing here before reviewing

salihudickson:
done!

.env.development
```diff
@@ -1,4 +1,4 @@
-URL=https://local.outline.dev:3000
+URL=https://local.outline.dev:4500
```
tommoor: Please restore, use `.env.local` if you need to change things for your own environment

.env.test
```diff
@@ -17,6 +17,10 @@ GITHUB_CLIENT_ID=123;
 GITHUB_CLIENT_SECRET=123;
 GITHUB_APP_NAME=outline-test;
 
+GITLAB_URL=http://localhost
```
tommoor:
```suggestion
GITLAB_URL=
```

Better to leave empty in test I would think

plugins/gitlab/server/GitLabIssueProvider.ts
```diff
@@ -0,0 +1,293 @@
...
+      }
+    });
+  }
+
+  private async destroyProject(payload: Record<string, any>) {
+    await sequelize.transaction(async (transaction) => {
+      const integrations = await Integration.findAll({
+        where: {
+          service: IntegrationService.GitLab,
+          [Op.and]: Sequelize.literal(
```
tommoor: Dangerous SQL injection - never pass untrusted content into `Sequelize.literal`

plugins/gitlab/server/gitlab.ts
```diff
@@ -0,0 +1,219 @@
...
+    const resource = GitLabUtils.parseUrl(url);
+    if (!resource) {
+      return;
+    }
+
+    const integration = (await Integration.findOne({
+      where: {
+        service: IntegrationService.GitLab,
+        teamId: actor.teamId,
+        [Op.and]: Sequelize.literal(
```
tommoor: Dangerous SQL injection

plugins/gitlab/server/GitLabIssueProvider.ts
```diff
@@ -0,0 +1,293 @@
...
+      default:
+        break;
+    }
+  }
+
+  private async updateNamespace(payload: Record<string, any>) {
+    const name = payload.old_full_path ?? payload.old_username;
+    const where = {
+      service: IntegrationService.GitLab,
+      [Op.and]: Sequelize.literal(
```
tommoor: Dangerous SQL injection

plugins/gitlab/server/GitLabIssueProvider.ts
```diff
@@ -0,0 +1,293 @@
...
+      await integration.save({ transaction });
+    });
+  }
+
+  private async updateProject(payload: Record<string, any>) {
+    await sequelize.transaction(async (transaction) => {
+      const integrations = await Integration.findAll({
+        where: {
+          service: IntegrationService.GitLab,
+          [Op.and]: Sequelize.literal(
```
tommoor: Dangerous SQL injection

plugins/gitlab/shared/GitLabUtils.ts
```diff
@@ -0,0 +1,308 @@
...
+  /**
+   * Makes an authenticated API request to GitLab.
+   *
+   * @param accessToken - The access token for authentication.
+   * @param endpoint - The API endpoint path.
+   * @param params - Additional fetch options.
+   * @param query - Query parameters to include in the request.
+   * @returns The response data from the GitLab API.
+   */
+  public static async apiRequest({
```
tommoor: So many fetch calls, we should just use https://github.com/jdalrymple/gitbeaker perhaps

shared/components/IssueStatusIcon/index.tsx
```diff
@@ -33,6 +33,8 @@ function getIcon(props: Props) {
       return <GitHubIssueStatusIcon {...props} />;
     case IntegrationService.Linear:
       return <LinearIssueStatusIcon {...props} />;
+    case IntegrationService.GitLab:
+      return <GitHubIssueStatusIcon {...props} />;
```
tommoor: This is one case I'd rather you copy/paste the component  as `GitLabIssueStatusIcon` rather than reuse.

tommoor:
Seeing a big canceled icon in the hovercard, here - not sure where that's coming from?

image

plugins/gitlab/server/GitLabIssueProvider.ts
salihudickson: My bad. My sql is pretty rusty, completely forgot to clean this up.

plugins/gitlab/shared/GitLabUtils.ts
salihudickson: I considered that, but I thought it might be best to try to limit deps. But I'll go ahead and use it.

tommoor: It can be, for sure. I'm more open to it for backend stuff... and if there is more than one call to the service.

plugins/gitlab/client/Settings.tsx
```diff
@@ -0,0 +1,159 @@
...
+        </>
+      ) : (
+        <Notice>
+          <Trans>
+            The GitLab integration is currently disabled. Please set the
+            associated environment variables and restart the server to enable
+            the integration.{" "}
+            <a href="#" target="_blank" rel="noopener noreferrer">
+              Learn more
+            </a>
```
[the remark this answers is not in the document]
salihudickson: @tommoor, we're going to need to add some documentation for this feature so I can link it here.

salihudickson:
Hey @tommor, I've implemented the requested changes here, and I think this is about ready for another review.

N0YU0:
Hi, is there any update on this?

shared/editor/components/Mentions.tsx
```diff
@@ -311,12 +311,14 @@ export const MentionIssue = observer((props: IssuePrProps) => {
...
-
   const url = new URL(issue.url);
+
   const service =
     url.hostname === "github.com"
       ? IntegrationService.GitHub
-      : IntegrationService.Linear;
+      : url.hostname === "linear.app"
+        ? IntegrationService.Linear
+        : IntegrationService.GitLab;
```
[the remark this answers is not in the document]
tommoor: I know, it sucks, but works for today.

IamTaoChen:
I met a problem https://github.com/outline/outline/issues/11635#issuecomment-3997706115

tommoor:
Can you see what is the format of the Invalid URL?

IamTaoChen:
I tried to mention `https://gitlab.kyzdt.com/aipin/aipin/-/issues`

PR #10863, opened 2025-12-12
perf: Policy evaluation

tommoor:
Removes several O(n) calculations during policy evaluation


tommoor:
Well, I think it's fair to say that made quite a big difference

image

PR #10930, opened 2025-12-16
Choose table cell background

### LINKED ISSUE #9271: Allow choosing the background color of table cells
It should be stored in the cell attrs. ~Probably makes sense to tackle cell split/merge first as the same menu will be needed on cell selection~ Split/merge is done now

apoorv-mishra:
Closes #9271

### DIFF
--- a/app/editor/components/ToolbarMenu.tsx
+++ b/app/editor/components/ToolbarMenu.tsx
@@ -35,9 +35,9 @@ type ToolbarDropdownProps = {
 function ToolbarDropdown(props: ToolbarDropdownProps) {
   const { commands, view } = useEditor();
   const { t } = useTranslation();
-  const { item } = props;
+  const { item, shortcut, tooltip } = props;
   const { state } = view;
-  const [isOpen, setIsOpen] = useState<boolean | null>(null);
+  const [isOpen, setIsOpen] = useState<boolean>(false);
 
   const handleOpenChange = useCallback((open: boolean) => {
     setIsOpen(open);
@@ -105,23 +105,25 @@ function ToolbarDropdown(props: ToolbarDropdownProps) {
   }, []);
 
   return (
-    <MenuProvider variant="dropdown">
-      <Menu onOpenChange={handleOpenChange}>
-        <MenuTrigger>
-          <ToolbarButton aria-label={item.label ? undefined : item.tooltip}>
-            {item.label && <Label>{item.label}</Label>}
-            {item.icon}
-          </ToolbarButton>
-        </MenuTrigger>
-        <MenuContent
-          align="end"
-          aria-label={item.tooltip || t("More options")}
-          onCloseAutoFocus={handleCloseAutoFocus}
-        >
-          <EventBoundary>{toMenuItems(items)}</EventBoundary>
-        </MenuContent>
-      </Menu>
-    </MenuProvider>
+    <Tooltip shortcut={shortcut} content={tooltip} disabled={isOpen}>
+      <MenuProvider variant="dropdown">
+        <Menu open={isOpen} onOpenChange={handleOpenChange}>
+          <MenuTrigger>
+            <ToolbarButton aria-label={item.label ? undefined : item.tooltip}>
+              {item.label && <Label>{item.label}</Label>}
+              {item.icon}
+            </ToolbarButton>
+          </MenuTrigger>
+          <MenuContent
+            align="end"
+            aria-label={item.tooltip || t("More options")}
+            onCloseAutoFocus={handleCloseAutoFocus}
+          >
+            <EventBoundary>{toMenuItems(items)}</EventBoundary>
+          </MenuContent>
+        </Menu>
+      </MenuProvider>
+    </Tooltip>
   );
 }
 


tommoor:
Some things to work through here – I don't think that combining the menu item for text highlight and cell background color is the right thing to do, it just makes it hard to find and the default colors should be quite different.

- Add a separate menu item for "Background" color
- It should have much lighter defaults, you can lighten the existing colors by 50% as a good start
- "Background" item should be available in the row and column menus
- "rainbow" swatch only makes sense when the current selection encompasses multiple background colors, if there is a single cell selected it should show the correct color

Lets see how far this gets us!

app/editor/components/ToolbarMenu.tsx
```diff
@@ -48,32 +53,53 @@ function ToolbarDropdown(props: { active: boolean; item: MenuItem }) {
...
+          dangerous: child.dangerous,
+          visible: child.visible,
+          selected:
+            child.active !== undefined ? child.active(state) : undefined,
+          onClick: handleClick(child),
+        };
+      });
+
+    return item.children ? mapChildren(item.children) : [];
+  }, [isOpen, commands]);
```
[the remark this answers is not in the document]
apoorv-mishra: We don't want to update `items` when the list is open. This prevents the list from re-rendering and hence prevents the color picker from closing.

shared/editor/presetColors.ts
```diff
@@ -0,0 +1,17 @@
+export const presetColors = [
```
tommoor, line 1: Probably just move this to `shared/utils/color` –  I know this came from before but would be worth combining this into a single object so color and name are colocated

app/editor/menus/tableCol.tsx
```diff
@@ -169,3 +264,10 @@ export default function tableColMenuItems(
     },
   ];
 }
+
+const DottedCircleIcon = styled(CircleIcon)`
```
tommoor: Used twice now, lets centralize in `components/Icons`

shared/editor/components/Styles.ts
```diff
@@ -1891,14 +1891,26 @@ table {
...
-  .selectedCell {
-    background: ${
-      props.readOnly ? "inherit" : props.theme.tableSelectedBackground
-    };
+  td[data-bgcolor] {
+    color: var(--cell-text-color);
 
-    /* fixes Firefox background color painting over border:
-     * https://bugzilla.mozilla.org/show_bug.cgi?id=688556 */
-    background-clip: padding-box;
```
tommoor, line 1901: We lost this fix?

shared/editor/lib/table.ts
```diff
@@ -70,6 +82,17 @@ export function setCellAttrs(node: Node): Attrs {
...
   }
+  if (node.attrs.marks) {
+    const backgroundMark = node.attrs.marks.find(
+      (mark: NodeAttrMark) => mark.type === "background"
+    );
+    if (backgroundMark) {
+      attrs["data-bgcolor"] = backgroundMark.attrs.color;
+      attrs.style =
+        (attrs.style ?? "") +
+        `background-color: ${backgroundMark.attrs.color}; --cell-text-color: ${readableColor(backgroundMark.attrs.color)};`;
```
tommoor, line 93: TIL `readableColor` is in polished 👌

shared/editor/commands/highlight.ts
```diff
@@ -0,0 +1,84 @@
+import type { Attrs } from "prosemirror-model";
+import {
+  TextSelection,
+  type Command,
+  type Transaction,
+} from "prosemirror-state";
+import { getMarkRange } from "../queries/getMarkRange";
+
+const createTextHighlight = (attrs: Attrs, tr: Transaction): Transaction => {
```
tommoor: Did all of these commands need to be added? Toggle text highlight worked fine before, wondering if this part can be reverted

shared/editor/commands/table.ts
```diff
@@ -822,3 +826,210 @@ function addRowWithAlignment(
...
+  const existingMarks = cell.attrs.marks ?? [];
+  const newMark = {
+    type: "background",
+    attrs,
+  };
+  const updatedMarks = [...existingMarks, newMark];
+  return tr.setNodeAttribute(pos, "marks", updatedMarks);
+};
+
+const updateCellBackground = (
```
tommoor, line 881: Having update and create methods separate isn't ideal, it can result in duplicate marks if create is called incorrectly –  something that handles both would be better, e.g

```
const updateCellBackground = (
  cell: Node,
  pos: number,
  attrs: Attrs,
  tr: Transaction
): Transaction => {
  const existingMarks: NodeAttrMark[] = cell.attrs.marks ?? [];

  const idx = existingMarks.findIndex((m) => m.type === "background");

  const updatedMarks =
    idx === -1
      ? [...existingMarks, { type: "background", attrs }]
      : existingMarks.map((m, i) =>
          i === idx ? { ...m, attrs: { ...m.attrs, ...attrs } } : m
        );

  return tr.setNodeAttribute(pos, "marks", updatedMarks);
};```

File: app/editor/components/CellBackgroundColorPicker.tsx [hunk 8]
```diff
@@ -0,0 +1,26 @@
...
+  command: string;
+};
+
+function CellBackgroundColorPicker({ activeColor, command }: Props) {
+  const { commands } = useEditor();
+
+  const handleSelect = useCallback(
+    (color: string) => {
+      if (commands[command]) {
+        commands[command]({ color });
```
tommoor, line 17: My main UX question is if it's actually desirable to have these solid colors, it would be better to have transparent colors that work against both light and dark background... maybe we introduce an alpha control in the picker and default it to .e.g `.7`

thoughts?

shared/editor/presetColors.ts
apoorv-mishra: I attempted colocation, but it actually touches places which escape `tsc`, for examples, `attrs: { color }` would need to change to `attrs: { color: color.hex }`. So, I refrained for now. Maybe a separate PR/commit _after_ this stands approved?

app/editor/components/CellBackgroundColorPicker.tsx
apoorv-mishra: Makes sense. Pushed with default alpha setting, including the slider.

PR #10948, opened 2025-12-18
fix: prevent extra newlines in pasted code blocks

hdoo42:
## Description

This PR fixes an issue where pasting code blocks (especially those with newlines immediately following the opening fence) resulted in extra empty lines being inserted at the beginning of the block.

The issue was caused by `normalizePastedMarkdown` aggressively normalizing newlines to ensure paragraphs are respected. This change updates `normalizePastedMarkdown` to:
1. Identify code blocks using a robust regex.
2. Temporarily replace these blocks with placeholders.
3. Perform the newline normalization on the rest of the text.
4. Restore the original code blocks.

### Reproduction Steps
When copying and pasting a code block like:
````md
```bash
echo hello
```
````
**Before this change**, the editor would incorrectly add a newline:
image
**After this change**, the content is pasted correctly without the leading newline.
image

### DIFF
--- a/shared/editor/lib/markdown/normalize.ts
+++ b/shared/editor/lib/markdown/normalize.ts
@@ -6,6 +6,17 @@
  */
 export default function normalizePastedMarkdown(text: string): string {
   const CHECKBOX_REGEX = /^\s?(\[(X|\s|_|-)\]\s(.*)?)/gim;
+  const CODE_BLOCK_REGEX = /^ {0,3}(`{3,}|~{3,})[\s\S]*?^ {0,3}\1/gm;
+
+  const placeholders: string[] = [];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: shared/editor/lib/markdown/normalize.ts [hunk 1]
```diff
@@ -18,5 +29,11 @@ export default function normalizePastedMarkdown(text: string): string {
   // find single newlines and insert an extra to ensure they are treated as paragraphs
   text = text.replace(/\b\n\b/g, "\n\n");
```
tommoor, line 30: After various tests I'm happy to just remove this line instead.

In the future it's best if you file issues before PRs

hdoo42: Thank you for fast reviewing! I'll issue first next time!

Issue #10742, opened 2025-11-28
Relative links inside duplicated document tree should remap to duplicated documents


dmitribodiu-orange:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### This is not related to configuring Outline

- [X] The issue is not related to self-hosting config

### Current Behavior

When duplicating a group (tree) of documents, internal relative links inside that tree still point to the original pages, not to the copies that were created.

### Expected Behavior

When duplicating a document tree:

All relative internal links between pages inside that tree<br>should automatically be remapped to point to the copies created inside the duplicated tree.

So Child 1 page link should be changed from:

/doc/child-2-id   →   /doc/child-2-copy-id

Duplicating a doc tree should produce an isolated, self-contained copy where all internal navigation (back and forth) stays inside the copy. Links could be between siblings, from parent to child or vice versa. Note: it should only apply to domain-relative links. (URL without [https://domain](<https://domain>))

### Steps To Reproduce

1. Original structure:<br>Parent<br>├── Child 1<br>      contains a link →  /doc/child-2-id   (i.e., link to “Child 2”)<br>└── Child 2
2. Duplicate the whole Parent folder → Parent (copy).

Result:

Parent<br>├── Child 1<br>└── Child 2

Parent (copy)<br>├── Child 1 (copy)<br>└── Child 2 (copy)

Actual behavior:<br>Inside Child 1 (copy), the relative link still points to:<br>/doc/child-2-id<br>Which goes to the original “Child 2”, not “Child 2 (copy)”.

### Environment

```markdown
- Outline: 1.1.0
- Browser:Microsoft Edge
```

### Anything else?

*No response*

tommoor:
Makes sense to me 👍

PR #10950, opened 2025-12-18
chore: Relative links inside duplicated document tree should remap to…

dmitribodiu-orange:
### Summary

This PR fixes document tree duplication so the duplicated subtree becomes self-contained: any domain-relative internal links between documents within the duplicated tree are automatically rewritten to point to the newly created copies (including links with #hash and ?query#hash tails).

### Problem

When duplicating a document folder/tree, pages inside the copied tree could still contain relative links like:

/doc/child-2-<originalUrlId>

/doc/child-2-<originalUrlId>#h-section

/doc/child-2-<originalUrlId>?x=1#h-section

These links continued to navigate to the original documents, not to the copies, breaking “isolated copy” behavior.

### What changed

Prebuild mapping for the entire duplicated subtree

idMap: original document ID → new document ID

urlIdMap: original urlId → new urlId

Rewrite internal links during duplication

Rewrites ProseMirror link marks (marks[].attrs.href) for domain-relative /doc/<slug>-<urlId> URLs, preserving any tail (?query, #hash, or ?query#hash).

Absolute links (https://…) are not rewritten.

Keep deterministic child order (same as previous behavior)

Child documents are retrieved and ordered according to the collection documentStructure using DocumentHelper.sortDocumentsByStructure(...).reverse().

### Implementation notes

Duplication now uses a queue traversal and explicitly assigns new id and urlId for each created document.

sourceMetadata.originalDocumentId is preserved for root and all duplicated children.

Hash and query components are preserved in rewritten links.

### Tests

Added integration tests covering link remapping for:

parent → child

child → parent

sibling → sibling

absolute links remaining unchanged

### Related

Fixes the issue where duplicated trees keep navigating back to the original documents via internal relative links (Outline issue #10742).


dmitribodiu-orange:
@tommoor could you please have a look? If anything is needed from my side let me know please

PR #10999, opened 2025-12-24
fix: Image warp exiting lightbox

### LINKED ISSUE #10731: Full width image warp when leaving lightbox

### Current Behavior

When closing the lightbox of a full-width image the shrinking animation warps the image as it resizes to fit the full width view. 

https://github.com/user-attachments/assets/ef94219c-43fb-4d9d-a389-1d73f738713a

### Expected Behavior

_No response_

### Steps To Reproduce

1. set an image to full width
2. click image to open lightbox
3. press escape or close the image

### Environment

```markdown
- Outline:1.1.0
- Browser: Safari 26.1 / Chrome 142.0.7444.176
```

### Anything else?

_No response_

tommoor:
closes #10731

### DIFF
--- a/app/components/Lightbox.tsx
+++ b/app/components/Lightbox.tsx
@@ -503,18 +503,16 @@ function Lightbox({ images, activeImage, onUpdate, onClose }: Props) {
         const toTx = to.center.x - final.center.x;
         const toTy = to.center.y - final.center.y;
 
-        const fromSx = from.width / final.width;
-        const fromSy = from.height / final.height;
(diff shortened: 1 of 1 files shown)


tommoor:
@apoorv-mishra seems this simple 🤷

apoorv-mishra:
Sweet 👌 

Makes sense that scaling should respect aspect ratio by default. But I do recall this not working and me fighting it while I built it. Perhaps, it was something else, in a half broken state.

Merry Christmas 🎉 🎄 🎅

PR #11121, opened 2026-01-09
track access requests

salihudickson:
tracking access requests

### DIFF
--- a/app/components/Notifications/NotificationListItem.tsx
+++ b/app/components/Notifications/NotificationListItem.tsx
@@ -36,7 +36,7 @@ function NotificationListItem({ notification, onNavigate }: Props) {
   const collection = collectionId ? collections.get(collectionId) : undefined;
   const [processing, setProcessing] = React.useState(false);
   const [selectedPermission, setSelectedPermission] =
-    React.useState<DocumentPermission>(DocumentPermission.ReadWrite);
+    React.useState<DocumentPermission>(DocumentPermission.Read);
 
   const isAccessRequest =
     notification.event === NotificationEventType.RequestDocumentAccess &&
--- a/server/routes/api/accessRequests/accessRequests.test.ts
+++ b/server/routes/api/accessRequests/accessRequests.test.ts
@@ -304,7 +304,6 @@ describe("#accessRequests.approve", () => {
       teamId: team.id,
     });
 
-    // add non-manager to the document with editor access
     await UserMembership.create({
       userId: admin.id,
       documentId: document.id,
@@ -461,7 +460,6 @@ describe("#accessRequests.dismiss", () => {
       body: {
         token: admin.getJwtToken(),
         id: accessRequest.id,
-        permission: DocumentPermission.ReadWrite,
       },
     });
 
--- a/server/routes/api/accessRequests/schema.ts
+++ b/server/routes/api/accessRequests/schema.ts
@@ -35,7 +35,7 @@ export const AccessRequestsApproveSchema = BaseSchema.extend({
   body: BaseIdSchema.extend({
     permission: z
       .nativeEnum(DocumentPermission)
-      .default(DocumentPermission.ReadWrite),
+      .default(DocumentPermission.Read),
   }),
 });
 


server/models/AccessRequest.ts
```diff
@@ -0,0 +1,136 @@
...
+      required: false,
+    },
+  ],
+}))
+@Table({
+  tableName: "access_requests",
+  modelName: "access_request",
+})
+@Fix
+class AccessRequest extends Model<
```
tommoor:
```suggestion
class AccessRequest extends IdModel<
```

Extend IdModel and you can remove the definitions for timestamps and id.

server/models/AccessRequest.ts
```diff
@@ -0,0 +1,136 @@
...
+  @IsUUID(4)
+  @PrimaryKey
+  @Default(DataType.UUIDV4)
+  @Column(DataType.UUID)
+  id: string;
+
+  @Default(AccessRequestStatus.Pending)
+  @Column({
+    type: DataType.STRING,
+    validate: {
```
tommoor: Use `IsIn` validation, reference enum directly so this doesn't need updating if value changes, like:

https://github.com/outline/outline-enterprise/blob/b53d845128d4d1905c314ae710fcf454c1e9cebc/server/models/Collection.ts#L259

server/models/AccessRequest.ts
```diff
@@ -0,0 +1,136 @@
...
+  ): Promise<boolean> {
+    if (!documentId || !userId) {
+      return false;
+    }
+
+    const count = await this.count({
+      where: {
+        documentId,
+        userId,
+        status: AccessRequestStatus.Pending,
```
tommoor, line 128: Query on status here, but there is no index on status – need to add the index

server/policies/accessRequest.ts
```diff
@@ -0,0 +1,7 @@
+import { AccessRequest, User } from "@server/models";
+import { allow } from "./cancan";
+import { isTeamModel } from "./utils";
+
+allow(User, "read", AccessRequest, (actor, accessRequest) =>
+  isTeamModel(actor, accessRequest?.user)
```
tommoor: This allows anyone in the same team to view access requests, probably needs to be more like isTeamAdmin or is originator of request

server/queues/tasks/DocumentAccessRequestNotificationsTask.ts
```diff
@@ -40,12 +46,19 @@ export default class DocumentAccessRequestNotificationsTask extends BaseTask<Doc
         continue;
       }
 
+      const accessReq = await AccessRequest.create({
```
tommoor: This is backwards now, the access request itself is created inside the task for it's notification – need to switch this around so that the AccessRequest is created and that triggers the notification

server/routes/api/accessRequests/accessRequests.ts
```diff
@@ -0,0 +1,132 @@
...
+import { authorize } from "@server/policies";
+import { presentAccessRequest, presentPolicies } from "@server/presenters";
+import { APIContext } from "@server/types";
+import { RateLimiterStrategy } from "@server/utils/RateLimiter";
+import * as T from "./schema";
+
+const router = new Router();
+
+router.post(
+  "access_requests.info",
```
tommoor:
```suggestion
  "accessRequests.info",
```

Use camelCase for these

server/routes/api/accessRequests/accessRequests.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      accessReq = await AccessRequest.findOne({
+        where: {
+          documentId: document.id,
+          userId: user.id,
+        },
+      });
+    }
+
+    if (!accessReq) {
```
tommoor, line 73: No checking of authorization in this endpoint

server/routes/api/accessRequests/accessRequests.ts
```diff
@@ -0,0 +1,132 @@
...
+  auth(),
+  validate(T.AccessRequestsApproveSchema),
+  transaction(),
+  async (ctx: APIContext<T.AccessRequestsApproveReq>) => {
+    const { id, permission } = ctx.input.body;
+    const { user } = ctx.state.auth;
+    const { transaction } = ctx.state;
+
+    const accessRequest = await AccessRequest.findByPk(id, {
+      rejectOnEmpty: true,
```
tommoor, line 97: Should pass transaction and lock in query

server/routes/api/accessRequests/accessRequests.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const document = await Document.findByPk(accessRequest.documentId, {
+      userId: user.id,
+    });
+    authorize(user, "share", document);
+
+    accessRequest.status = AccessRequestStatus.Approved;
+    accessRequest.responderId = user.id;
+    accessRequest.respondedAt = new Date();
+    await accessRequest.save({ transaction });
```
tommoor:
```suggestion
    await accessRequest.saveWithCtx(ctx, { transaction });
```

Use `withCtx` variants so that audit event is created

server/routes/api/accessRequests/accessRequests.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    await UserMembership.create(
+      {
+        userId: accessRequest.userId,
+        documentId: accessRequest.documentId,
+        permission: permission || DocumentPermission.ReadWrite,
```
tommoor: We need to check that `permission` is not greater than the users permission

server/routes/api/accessRequests/accessRequests.ts
```diff
@@ -0,0 +1,132 @@
...
+  auth(),
+  validate(T.AccessRequestsDismissSchema),
+  transaction(),
+  async (ctx: APIContext<T.AccessRequestsDismissReq>) => {
+    const { id } = ctx.input.body;
+    const { user } = ctx.state.auth;
+    const { transaction } = ctx.state;
+
+    const accessRequest = await AccessRequest.findByPk(id, {
+      rejectOnEmpty: true,
```
tommoor, line 157: transaction, lock

tommoor (changes requested):

server/policies/accessRequest.ts
salihudickson: oh yeah, didn't wanna restrict to just admins but didn't think of this at the moment. TY!

server/routes/api/accessRequests/accessRequests.ts
salihudickson: I guess I did it this way cause of the way I initially wrote the server policy but your suggestion there also covers this. Will fix.

server/models/AccessRequest.ts
salihudickson: there's an index for the status field in the create-access-request migration.

PR #11157, opened 2026-01-12
fix : Markdown task markers inside pipe table cells 

### LINKED ISSUE #11009: checkmarks (and styling) in table cells from imported markdown files are escaped, not displayed as checkmarks

### Current Behavior

When doing this:
1. Update a document via api
2. The updated text contains a table. A cell within that table contains a check mark a.k.a`[x] text`
3. The resulting cell in the markdown table in outline will be escaped

why is it problematic:
→ When exporting the document again, it is visible: the checkmark became `\[x\] text`
→ No way to import a table with checkmarks in cells and have them be displayed as checkmarks
 
same problem with displaying list items `- text`

### Expected Behavior

a table cell that contains `[x] test` when imported
should be displayed as a ticked checkmark with text 'test' next to it
and a table cell that contains `\[x\] test`

### Steps To Reproduce

1. Update a document via api → provide the new markdown content as specified with json
2. The updated text contains a table. A cell within that table contains a check mark a.k.a`[x] text`
3. The resulting cell in the markdown table in outline will be escaped

### Environment

```markdown
- Outline: v1.1.0
- Browser: Arc Version 1.126.1 (72660) ; Chromium Engine Version 143.0.7499.170
```

### Anything else?

> A related problem that is another big pain:
> new lines in table cells in general and the way it is exported.
> I know that it is not officially supported by the markdown standard so that's why it's problematic anyways but the current export and import behaviour is still annoying because exporting cells with newlines will just export them with actual new lines → thus generating invalid markdown  because the | symbols for the columns are now in the next line
> 
> but then while importing you can't even import that same file again, because it will not read those new lines as part of that cell.
> 
> The only working thing i found is to replace newline with <br> but then i ran into the checkmark issue.
> 
> Love the project!

RudraPrasad001:
### Fix: task markers in table cells render nicely (Fixes #11009)

This PR improves how Markdown task markers inside **pipe table cells** are displayed.

**Problem**  
When users paste or update Markdown like `| - [ ] task |`, the checkbox text can end up escaped or displayed incorrectly, and some editor paste flows can error because the Markdown parser doesn’t support custom checkbox token types. [6]

**Change**  
Inside table cells only, convert task markers to simple symbols:
- `- [ ] task` → `☐ task`
- `- [x] task` → `☑ task`

**Why this approach**  
- Works with existing Markdown + editor parsing without introducing new token types.[6]
- Keeps export/round-trip stable and readable.[6]
- Not interactive by design—interactive checkboxes in table cells aren’t reliably supported by Markdown parsing/round-tripping, so this PR focuses on correct display.[6]

**Test**  
Paste:
```md
| Item | Status |
|---|---|
| A | - [ ] todo |
| B | - [x] done |
```
Expected: cells show `☐ todo` and `☑ done`.

### DIFF
--- a/server/editor/index.ts
+++ b/server/editor/index.ts
@@ -5,6 +5,7 @@ import {
   richExtensions,
   withComments,
 } from "@shared/editor/nodes";
+import Mention from "@shared/editor/nodes/Mention";
 
 const extensions = withComments(richExtensions);
 export const extensionManager = new ExtensionManager(extensions);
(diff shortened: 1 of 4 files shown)


RudraPrasad001:
https://github.com/user-attachments/assets/53dfbf2d-8f26-48c5-935a-f347d737d4ea

The Video Demonstration

PR #11170, opened 2026-01-13
feat(redis-key-prefix): configurable key prefix for redis usages

Lowaiz:
New Redis Adapter options to set a key prefix based on an environment variable.
Split the bull's redis usage to not use the option as described [by the docs](https://github.com/OptimalBits/bull/blob/develop/PATTERNS.md#reusing-redis-connections), making the implementation compatible with Redis clusters:
> do not set a keyPrefix on the connection you create, use bull's built-in prefix feature if you need a key prefix

> In summary, to make bull compatible with Redis cluster, use a queue prefix inside brackets.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -61,6 +61,10 @@ DATABASE_CONNECTION_POOL_MAX=
 # DOCS: https://docs.getoutline.com/s/hosting/doc/redis-LGM4BFXYp4
 REDIS_URL=redis://redis:6379
 
+# Optional key prefix for all Redis keys. Useful for key namespacing when
+# sharing a Redis instance between multiple applications.
+# REDIS_KEY_PREFIX=
+
 
 # ––––––––––––––––––––––––––––––––––––––
 # –––––––––––  FILE STORAGE  –––––––––––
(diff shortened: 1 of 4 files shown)


tommoor:
A problem here is you just added two more redis connections per-process.

Lowaiz:
I think it cannot work without new connections, as Bull does not support io-redis key prefix (I tested giving the prefix with the `REDIS_URL="ioredis://<base64>"` as described by [the doc](https://docs.getoutline.com/s/hosting/doc/redis-LGM4BFXYp4) resulting in bull fails, because some key contained the prefix some didn't), bull needs its own connections while the other parts of the app leverage on the io-redis key prefix.

A heavy workaround would be to manually add the prefix each time the redis is used beside Bull, and I don't think it's a good long term solution

tommoor:
We may not be able to move forward here then I'm afraid, adding multiple new redis connections per-process for existing installations is not acceptable

Lowaiz:
From what I see, Redis default max connections is 10 000, so I don't see why 2 connections per-process would cause any problems.
Furthermore, without the possibility of a prefix, bull is quite incompatible with Redis in cluster mode from what I understand. 

Lastly key name-spacing seems a good/standard practice when using cluster (Redis Cluster does not support multiple databases like the standalone version of Redis. It only supports database 0)  or when we do not use dedicated database index.

Anyway, if the number of Redis connections is critical to you, I'll understand that you close the PR

Issue #11073, opened 2026-01-06
Enable comment shortcut to work without text selection to leave document-level comment


tommoor:
Currently it does nothing, or opens the move dialog.

RudraPrasad001:
Kindly check it out, The requested functionality has been done

PR #11173, opened 2026-01-13
fix : Enable comment shortcut to work without text selection (Issue #11073)

RudraPrasad001:
https://github.com/user-attachments/assets/b43d8bc4-c6aa-41ab-a315-124208daa29a 

Fixes ( #11073 )
**Before**
Previously, the Ctrl+Alt+M keyboard shortcut only worked when text was selected, creating a comment mark on the selection. When pressed without text selected, nothing happened. However, the comments menu button could open the sidebar without any selection.
​

**Solution**
Modified the shortcut behavior to open the comments sidebar when no text is selected, matching the menu button functionality, and auto-focus the comment input field for Doc level comment.

**Files Changed**
1. shared/editor/commands/comment.ts

- Added check for empty selection in addCommentTextSelection function

- When selection is empty, dispatch a custom event openCommentsSidebar with autoFocus: true flag

- Existing behavior (creating comment marks on selected text) remains unchanged

2. app/scenes/Document/index.tsx

- Added event listener for openCommentsSidebar custom event

- Calls ui.toggleComments() to open the sidebar

- Implements retry logic to find and auto-focus the comment input field (ProseMirror editor) after sidebar opens


shared/editor/commands/comment.ts
```diff
@@ -13,55 +13,71 @@ export const addComment = (attrs: Attrs): Command =>
...
-    ) {
-      return false;
-    }
+      // Handle empty selection - open the sidebar
+      if (state.selection.empty) {
+        // Access the UI store from the editor view's DOM
+        const editorElement = view?.dom;
+        if (editorElement) {
+          // Trigger a custom event that the Document component can listen to
+          const event = new CustomEvent('openCommentsSidebar', {
```
tommoor: While this is the correct place to hook into the keypress, using custom events like this isn't really a pattern for us. I would follow the same way as `onClickCommentMark`

app/scenes/Document/index.tsx
```diff
@@ -35,13 +35,53 @@ export default function DocumentScene(props: Props) {
...
+  //  Listen for opening comments sidebar from keyboard shortcut
+  useEffect(() => {
+    const handleOpenCommentsSidebar = (event: Event) => {
+      ui.toggleComments();
+
+      const customEvent = event as CustomEvent;
+      if (customEvent.detail?.autoFocus) {
+        const focusComment = (attempts = 0) => {
+          // Get all ProseMirror editors
+          const allEditors = document.querySelectorAll('.ProseMirror[role="textbox"]');
```
tommoor: This is messy, we're in react – we don't need to query the dom

app/scenes/Document/index.tsx
```diff
@@ -35,13 +35,53 @@ export default function DocumentScene(props: Props) {
...
+            if (placeholder) {
+              commentInput = editor as HTMLDivElement;
+            }
+          });
+
+          if (commentInput) {
+            (commentInput as HTMLDivElement).focus();
+            (commentInput as HTMLDivElement).click();
+          } else if (attempts < 20) {
+            setTimeout(() => focusComment(attempts + 1), 100);
```
tommoor: Messy, random setTimeouts are unmaintainable

shared/editor/commands/comment.ts
```diff
@@ -13,55 +13,71 @@ export const addComment = (attrs: Attrs): Command =>
...
-  };
 
 const addCommentTextSelection =
   (attrs: Attrs): Command =>
-  (state, dispatch) => {
-    if (!(state.selection instanceof TextSelection)) {
-      return false;
-    }
+    (state, dispatch, view) => {
+      console.log("addCommentTextSelection", state.selection)
```
tommoor: Random logging

tommoor (changes requested):

RudraPrasad001:
I've changed `onClickCommentMark` to follow the existing pattern.
```typescript
view?.props?.onOpenCommentsSidebar?.();
```

app/scenes/Document/index.tsx
tommoor: This was marked as resolved, but all the code and event handling remains

RudraPrasad001: kindly check the latest commit, the custom events are not used

tommoor: Sure, but the code is still there – so that's even worse

RudraPrasad001: alright, I'll make a new PR

tommoor: Just remove the unused code? I don't understand... I don't have time for this back and forth I'm sorry, this is basic stuff

RudraPrasad001: I am sorry but i genuinely do not understand what you mean. I have cleared the random timeouts and custom events,can you please clarify what is needed to remove in the latest commit

RudraPrasad001:
The Unused Code is Removed.

RudraPrasad001:
We store the callback in a plugin's state instead of modifying ProseMirror's types . 

     // Handle empty selection - open the sidebar
      if (state.selection.empty) {
      const pluginState = commentPluginKey.getState(state);
      const callback = pluginState?.onOpenCommentsSidebar;
      
      if (callback) {
        callback();
        return true;
      }
      return false

tommoor:
It's not neccessary, it is available through `this.editor.props.onOpenCommentsSidebar` – see the other hooks for comments

RudraPrasad001:
Added `onOpenCommentsSidebar` prop to the Editor that gets passed through to the Comment mark's options.

tommoor:
Fixed here https://github.com/outline/outline/pull/11261 – sorry, too many revisions in this PR and still wasnt there

PR #11174, opened 2026-01-13
perf: Vendorize `rfc6902`

tommoor:
The upstream `rfc6902` package has a performance issue with the `diffArrays` function. For large arrays, the recursive memoization approach uses excessive memory and CPU, making it impractical for real-world use cases.

This vendorized version includes a fix from [PR #88](https://github.com/chbrown/rfc6902/pull/88) which replaces the recursive approach with an iterative dynamic programming implementation, significantly improving performance for large array comparisons.

### DIFF
--- a/shared/utils/rfc6902/diff.ts
+++ b/shared/utils/rfc6902/diff.ts
@@ -187,10 +187,10 @@ function isArrayRemove(
 }
 
 interface DynamicAlternative {
-  /**
-   * track prev key position for less memory usage
-   */
-  prev: string | null;
+  /** Previous i coordinate (-1 means no previous) */
+  prevI: number;
+  /** Previous j coordinate (-1 means no previous) */
+  prevJ: number;
   operation: ArrayOperation | null;
   /**
    * cost indicates the total cost of getting to this position.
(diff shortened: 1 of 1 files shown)


shared/utils/rfc6902/diff.ts
```diff
@@ -250,72 +249,72 @@ export function diffArrays<T>(
...
 ): Operation[] {
-  // set up cost matrix (very simple initialization: just a map)
+  // handle weird objects masquerading as Arrays that don't have proper length
+  // properties by using 0 for everything but positive numbers
   const input_length =
     isNaN(input.length) || input.length <= 0 ? 0 : input.length;
   const output_length =
     isNaN(output.length) || output.length <= 0 ? 0 : output.length;
+
+  // Skip matching prefix
```
tommoor, line 259: New: Skip matching prefixes

shared/utils/rfc6902/diff.ts
```diff
@@ -250,72 +249,72 @@ export function diffArrays<T>(
...
+  for (let i = 0; i <= subInput; i++) {
+    for (let j = 0; j <= subOutput; j++) {
       let memoized = memo[i][j];
       if (memoized) {
         continue;
       }
-      const add_prev_key = `${i},${j - 1}`;
-      const remove_prev_key = `${i - 1},${j}`;
-      const replace_prev_key = `${i - 1},${j - 1}`;
-      const remove_operation: ArrayRemove = {
```
tommoor, line 295: Avoid creating potentially unused objects

shared/utils/rfc6902/diff.ts
```diff
@@ -208,10 +208,9 @@ function buildOperations(
     throw new Error("invalid memo");
   }
   const operations: ArrayOperation[] = [];
-  while (memoized && memoized.prev && memoized.operation) {
+  while (memoized && memoized.prevI >= 0 && memoized.operation) {
     operations.push(memoized.operation);
-    const index = memoized.prev.split(",");
```
tommoor, line 213: Avoid expensive split call in hot path

PR #11184, opened 2026-01-15
fix: alt+shift header selection in Safari

### LINKED ISSUE #10633: Safari: alt/option shift arrow keys word selection does not work in header

### Current Behavior

I cannot select whole words with option/alt + shift + arrow keys (left/right) in the editor.

### Expected Behavior

I _can_ select whole words with option/alt + shift + arrow keys (left/right) in the editor, just as I can do everywhere else.

### Steps To Reproduce

1. edit a page
2. try option/alt+shift+left/right
3. nothing happens

### Environment

```markdown
- Outline: Cloud
- Browser: Safari
- OS: macOS
```

### Anything else?

_No response_

tommoor:
There is no single solution that works for all browsers – Safari is the exception for now.

closes #10633

### DIFF
--- a/shared/editor/nodes/Heading.ts
+++ b/shared/editor/nodes/Heading.ts
@@ -224,7 +224,7 @@ export default class Heading extends Node {
             Decoration.widget(
               // Safari requires the widget to be placed at the end of the node rather than the beginning
               // or caret selection is not correct, browser quirk – see issue #1234
-              isSafari() ? pos + node.nodeSize - 1 : pos + 1,
+              isSafari ? pos + node.nodeSize - 1 : pos + 1,
               container,
               {
                 side: -1,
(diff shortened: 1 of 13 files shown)


shared/editor/nodes/Heading.ts
```diff
@@ -220,12 +221,18 @@ export default class Heading extends Node {
...
-            Decoration.widget(pos + 1, container, {
-              side: -1,
-              ignoreSelection: true,
-              relaxedSide: true,
-              key: pos.toString(),
-            })
+            Decoration.widget(
+              // Safari requires the widget to be placed at the end of the node rather than the beginning
+              // or caret selection is not correct, browser quirk – see issue #1234
+              isSafari ? pos + node.nodeSize - 1 : pos + 1,
```
tommoor, line 227: As this is getting called a lot it seemed like the right time to refactor this from a method to a const.

PR #11198, opened 2026-01-16
fix: add date sorting support to table columns

lucawimmer:
Table columns with dates were being sorted alphabetically instead of chronologically. For example, "01.01.2025" would sort before "08.04.2025" by chance, but not in the correct chronological order.

### Solution
- Added date parsing support using the existing `date-fns` for common formats:
  - ISO: `yyyy-MM-dd`
  - European: `dd.MM.yyyy`, `dd/MM/yyyy`
  - US: `MM/dd/yyyy`
  - Dates without year: `dd.MM.`, `dd/MM`, etc.
- Dates are sorted chronologically by timestamp
- Falls back to existing number/text sorting when dates aren't detected

### Behavior
- If all non-empty cells in a column are valid dates → sort chronologically
- If all non-empty cells are numbers → sort numerically (like before)
- Otherwise → sort alphabetically (like before)
- Empty cells always move to the end

### DIFF
--- a/shared/editor/commands/table.ts
+++ b/shared/editor/commands/table.ts
@@ -1,4 +1,3 @@
-import { parse, isValid } from "date-fns";
 import { GapCursor } from "prosemirror-gapcursor";
 import type { Node, NodeType } from "prosemirror-model";
 import { Slice } from "prosemirror-model";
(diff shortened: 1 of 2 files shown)


tommoor:
I am interested in merging this but `lint` is failing, you can run `yarn lint --fix`. I'm guessing this was largely AI generated?

lucawimmer:
> I am interested in merging this but `lint` is failing, you can run `yarn lint --fix`. I'm guessing this was largely AI generated?

Oh no - sorry for overlooking it. I checked linting on our internal repo but missed it on this fork. Should be fixed now.

Let me know if anything else needs changing. (and yes, sorry for the AI lazyness)

tommoor:
I refactored a little so that ambiguous slash formats are considered depending on the users current locale

PR #11222, opened 2026-01-19
fix: skip ACL parameter for GCS uniform bucket-level access

### LINKED ISSUE #11221: S3Storage fails with GCS uniform bucket-level access due to ACL parameter
### Description
When using Google Cloud Storage with **uniform bucket-level access** enabled, file uploads fail with `InvalidArgument` error because the S3Storage class always passes an ACL parameter to the upload request.

### Error Message
```
InvalidArgument: Invalid argument.
Code: 'InvalidArgument',
Details: 'Cannot insert legacy ACL for an object when uniform bucket-level access is enabled. Read more at https://cloud.google.com/storage/docs/uniform-bucket-level-access'
```

### Stack Trace
```
at S3Storage.store (/opt/outline/build/server/storage/files/S3Storage.js:49:7)
at async /opt/outline/build/server/routes/api/documents/documents.js:1357:3
```

### Root Cause
In `server/storage/files/S3Storage.ts`, the `store()` method always includes `ACL` in the upload params, even when it should be omitted:

```typescript
const upload = new Upload({
  client: this.client,
  params: {
    ACL: acl as ObjectCannedACL,  // Always set, even if undefined
    // ...
  },
});
```

Similarly, `getPresignedPost()` always includes `acl` in the Fields object.

### Proposed Fix
Only include ACL parameter when it has a value:

```typescript
// In store()
params: {
  ...(acl && { ACL: acl as ObjectCannedACL }),
  // ...
}

// In getPresignedPost()
Fields: {
  "Content-Disposition": this.getContentDisposition(contentType),
  key,
  ...(acl && { acl }),
},
```

This allows users to set `AWS_S3_ACL=` (empty) in their environment to disable object-level ACLs, which is required for GCS buckets with uniform bucket-level access.

### Environment
- Outline version: Latest
- Storage: Google Cloud Storage (S3-compatible API)
- GCS bucket configuration: Uniform bucket-level access enabled

### Affected Endpoints
- `POST /api/documents.import`
- Any endpoint that uploads files via `S3Storage.store()`
- Client-side uploads via presigned URLs

maycuatroi1:
## Summary
- Makes ACL parameter conditional in `S3Storage.store()` and `S3Storage.getPresignedPost()` methods
- Allows users to set `AWS_S3_ACL=` (empty) to disable object-level ACLs
- Fixes compatibility with Google Cloud Storage buckets that have uniform bucket-level access enabled

## Problem
When using GCS with **uniform bucket-level access** enabled, file uploads fail with:
```
InvalidArgument: Cannot insert legacy ACL for an object when uniform bucket-level access is enabled
```

This happens because `S3Storage` always passes an ACL parameter, even when it should be omitted.

## Solution
Only include ACL in the upload params when it has a truthy value:
```typescript
...(acl && { ACL: acl as ObjectCannedACL })
```

## Test plan
- [x] Set `AWS_S3_ACL=` (empty) in environment
- [x] Upload a file via document import
- [x] Verify upload succeeds without ACL-related errors
- [x] Verify existing behavior unchanged when `AWS_S3_ACL=private` is set

Fixes #11221

### DIFF
--- a/server/storage/files/S3Storage.ts
+++ b/server/storage/files/S3Storage.ts
@@ -52,7 +52,7 @@ export default class S3Storage extends BaseStorage {
       Fields: {
         "Content-Disposition": this.getContentDisposition(contentType),
         key,
-        acl,
+        ...(acl && { acl }),
       },
       Expires: 3600,
     };
@@ -114,7 +114,7 @@ export default class S3Storage extends BaseStorage {
     const upload = new Upload({
       client: this.client,
       params: {
-        ACL: acl as ObjectCannedACL,
+        ...(acl && { ACL: acl as ObjectCannedACL }),
         Bucket: this.getBucket(),
         Key: key,
         ContentType: contentType,


maycuatroi1:
image
Stuff worked

PR #11225, opened 2026-01-20
feat: store avatars and logos in dedicated prefix for granular S3 access control

alborworld:
## Summary

This enables S3 bucket policies to restrict public-read ACLs to only the `avatars/` prefix, providing defense-in-depth security:

- Anonymous reads can be denied for all prefixes except `avatars/`
- Even if a document accidentally gets public-read ACL, it remains inaccessible to anonymous users

## Backward Compatibility

The change is backward compatible — existing uploads without preset continue to use the `uploads/` prefix.

Related discussion: https://github.com/outline/outline/discussions/11220

### DIFF
--- a/server/commands/attachmentCreator.ts
+++ b/server/commands/attachmentCreator.ts
@@ -50,6 +50,7 @@ export default async function attachmentCreator({
     id: randomUUID(),
     name,
     userId: user.id,
+    preset,
   });
 
   let attachment;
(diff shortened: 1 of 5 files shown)


tommoor:
I don't think this is necessary anymore, with recent changes no attachments are stored with public ACL

albority:
Well, it is for avatars/logos, not attachments. 

Logo uploads fail if `allow_public_acls = false`.

The change you mentioned only fixed serving, not uploading.

alborworld:
I double checked the code. While logos / avatars are public, documents/attachments stay private. 

Closing this. Thanks anyways!

PR #11229, opened 2026-01-20
fix: Rebuild of sourced permissions on document move

### LINKED ISSUE #11096: Docs Permissions are reset when drag/droping nested doc

### Current Behavior

I have the following collection "tree": 
- Collection "clients"
  - root-doc 
    - nested-doc-1
    - nested-doc-2
    - sensible-doc

 the "clients" collection has the following permissions : 
Image

the "root-doc" has the following permissions : 
Image

nested-doc-1 and nested-doc-2 has : 

Image

but for security purpose the "employee - edit" (inherited) permission has been removed for the "sensible-doc"

Image

Here's the bug : 
If I move nested-doc-2 within nested-doc-1 ... the whole permissions set in "root-doc" are reapplied on each nested docs ... (even the sensible-doc !)

Image

nested-doc-1 and nested-doc-2 being unrelated with "sensible-doc" I dont exptect any permission change on this document !
However, I would understand, the nested-doc-2 permissions would be "reset" to reuse the one of it's direct parent (ie: nested-doc-1)

### Expected Behavior

To me, only moved doc permissions should be altered, not the adjacent documents... 

Also, if nested-doc-1 already have another nested doc (ie nested-doc-3), with custom permission, moving nested-doc-2 within nested-doc-1 should not reset the permission from nested-doc-3. Once again, only the moved document should be altered.

### Steps To Reproduce

Juste reproduce the explanations above. 

### Environment

```markdown
- Outline:v1.1.0
- Browser: latest brave (chromium based)
```

### Anything else?

I checked the latest release note from the v1.2.0  https://github.com/outline/outline/releases/tag/v1.2.0 but none of the shipped fixs seems to be related to that bug.

salihudickson:
closes #11096

### DIFF
--- a/server/queues/processors/DocumentMovedProcessor.test.ts
+++ b/server/queues/processors/DocumentMovedProcessor.test.ts
@@ -471,5 +471,172 @@ describe("DocumentMovedProcessor", () => {
     });
     expect(childUserMemberships.length).toBe(0);
     expect(childGroupMemberships.length).toBe(0);
+
+    // Verify inherited permissions from the new parent are applied
+    const newChildUserMemberships = await UserMembership.findAll({
+      where: { documentId: childDocument.id, userId: user3.id },
+    });
+    const newChildGroupMemberships = await GroupMembership.findAll({
+      where: { documentId: childDocument.id, groupId: group2.id },
+    });
+    expect(newChildUserMemberships.length).toBe(1);
+    expect(newChildGroupMemberships.length).toBe(1);
+    expect(newChildUserMemberships[0].sourceId).toBeTruthy();
+    expect(newChildGroupMemberships[0].sourceId).toBeTruthy();
+  });
+
+  it("should remove all sourced permissions when a document is moved to root level", async () => {
+    const team = await buildTeam();
+    const user = await buildAdmin({ teamId: team.id });
+    const user2 = await buildUser({ teamId: team.id });
+    const group = await buildGroup({ teamId: team.id });
+    const collection = await buildCollection({
+      userId: user.id,
+      teamId: team.id,
+    });
+
+    const parentDocument = await buildDocument({
+      collectionId: collection.id,
+      teamId: team.id,
+    });
+    const childDocument = await buildDocument({
+      teamId: team.id,
+      parentDocumentId: parentDocument.id,
+    });
+
+    await UserMembership.create({
+      userId: user2.id,
+      documentId: parentDocument.id,
+      createdById: user.id,
+    });
+    await GroupMembership.create({
+      groupId: group.id,
+      documentId: parentDocument.id,
+      createdById: user.id,
+    });
+
+    // Verify sourced permissions exist on child
+    let childUserMemberships = await UserMembership.findAll({
+      where: { documentId: childDocument.id, userId: user2.id },
+    });
+    let childGroupMemberships = await GroupMembership.findAll({
+      where: { documentId: childDocument.id, groupId: group.id },
+    });
+    expect(childUserMemberships.length).toBe(1);
+    expect(childGroupMemberships.length).toBe(1);
+
+    // Move child to root level (no parent)
+    childDocument.parentDocumentId = null;
+    await childDocument.save();
+
+    const processor = new DocumentMovedProcessor();
+    await processor.perform({
+      name: "documents.move",
+      documentId: childDocument.id,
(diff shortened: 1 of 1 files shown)


tommoor:
Appreciated, will need tests though

salihudickson:
done!

salihudickson:
With the current functionality, the sourced permissions are copied from the top document, and not from the direct parent. Looks like it was done on purpose and is not a bug, so I didn't change it. 

Just wanted to point it out, incase that's something you'd like to change.

app/components/Sharing/Document/DocumentMemberList.tsx
```diff
@@ -87,7 +87,11 @@ function DocumentMemberList({ document, invitedInSession }: Props) {
   const members = React.useMemo(
     () =>
       orderBy(
-        document.members,
+        Array.from(
```
tommoor, line 90: I'm not following why there are changes in this file

server/queues/processors/DocumentMovedProcessor.test.ts
```diff
@@ -0,0 +1,387 @@
...
+  buildTeam,
+  buildAdmin,
+  buildGroup,
+} from "@server/test/factories";
+import DocumentMovedProcessor from "./DocumentMovedProcessor";
+
+const ip = "127.0.0.1";
+
+describe("DocumentMovedProcessor", () => {
+  describe("Bug fix - https://github.com/outline/outline/issues/11096", () => {
```
tommoor: Remove this describe wrapper please, git history captures where the fix came from

server/models/GroupMembership.ts
```diff
@@ -326,46 +326,50 @@ class GroupMembership extends ParanoidModel<
...
     for (const childDocumentId of childDocumentIds) {
+      await this.destroy({
+        where: {
+          groupId: model.groupId,
+          sourceId: model.id,
+          documentId: childDocumentId,
+        },
+        transaction,
+      });
+
```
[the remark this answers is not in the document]
tommoor: Agreed on this, the loss of the single delete operation concerned me

app/components/Sharing/Document/DocumentMemberList.tsx
salihudickson: It’s meant to filter out duplicates. When a document is moved, recalculating sourced permissions can create entries that appear different but represent the same sourced permission, and the user would need to refresh the page to clear them. I added filtering to prevent that.

tommoor: This is an existing bug?

salihudickson: Yes, it is.

salihudickson:
Hey @tommoor, is there anything else we need to fix here to make it ready?

tommoor:
No, I just need to test it thoroughly as the consequences of it being wrong are pretty catastrophic

salihudickson:
I see, I'll take a look at this.

dadaxr:
> With the current functionality, the sourced permissions are copied from the top document, and not from the direct parent. Looks like it was done on purpose and is not a bug, so I didn't change it.
> 
> Just wanted to point it out, incase that's something you'd like to change.

given the linked issue (#11096), what would be the result if this PR was merged ? 

If I understand it well, if I move "nested-doc-2" within "nested-doc-1", the permissions of nested-doc-2 would be the ones of the "root-doc", and not the ones of the "nested-doc-1" (ie: it's new direct parent).

Would not it be more expectable the other way around ? (ie copying permissions of the direct parent, as it's how inheriting mechanism works in general) 

Thanks :)

tommoor:
I'm still seeing that last bug I described, made sure to check out the last commit

salihudickson:
It seems the issue was that the sourced permissions from the moved document were being re-added instead of being recalculated from the new parent. I removed that logic because I didn’t see a reason to re-add them, given that the purpose of recalculation is to derive permissions from the parent.

That said, I’m not sure if I’m missing something, and if  there a specific reason they were being re-added.

salihudickson:
@dadaxr yeah, documents currently get their sourced permissions from the “root-doc,” not their direct parent.

This scope of this PR is not to change how sourced permissions behave. Perhaps you can create an issue for that, but it would need to be assessed and handled separately, as an enhancement.

PR #11230, opened 2026-01-20
Remove unused showActions attr in SidebarLink.Action; fixes console error

infogulch:
The property is retrieved from the `props` function arg. This attribute is not used, not necessary, and shows a React error in the console.


app/components/Sidebar/components/SidebarLink.tsx
```diff
@@ -201,7 +201,7 @@ const Content = styled.span`
   width: 100%;
 `;
 
-const Actions = styled(EventBoundary)<{ showActions?: boolean }>`
+const Actions = styled(EventBoundary)`
   display: inline-flex;
   visibility: ${(props) => (props.showActions ? "visible" : "hidden")};
```
tommoor, line 206: It's used right here....

PR #11242, opened 2026-01-22
feat: Sort on search screen

### LINKED ISSUE #11234: Add sort order to search
### Discussed in https://github.com/outline/outline/discussions/11233

<sup>Originally posted by **filiabel** January 21, 2026</sup>
It would be very useful to be able to sort the search results by created date, updated date, title (maybe other props) ascending/descending ✨ </div>

salihudickson:
fix #11234


app/scenes/Search/components/SortFilter.tsx
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  return (
+    <FilterOptions
+      noSort
+      showFilter={false}
+      options={options}
+      selectedKeys={[selectedKey]}
+      onSelect={handleSelect}
+      defaultLabel={t("Recently updated")}
```
tommoor: The default sort is `Relevance`, it should be represented in the list as selected by default

app/scenes/Search/Search.tsx
```diff
@@ -267,6 +281,15 @@ function Search() {
                 onSelect={(dateFilter) => handleFilterChange({ dateFilter })}
               />
             )}
+            {filterVisibility.sort && (
```
tommoor, line 302: Not sure exactly how yet, but will need to differentiate sort from filters – I can take a look once other feedback is addressed 👍

server/models/helpers/SearchHelper.ts
```diff
@@ -492,7 +518,23 @@ export default class SearchHelper {
         "searchRanking",
       ]);
       replacements["query"] = this.webSearchQuery(query);
-      order.unshift(["searchRanking", "DESC"]);
+      // Only prioritize search ranking if no custom sort is specified
+      if (!sort) {
+        order.push(["searchRanking", "DESC"]);
```
tommoor: Maybe we should always push `searchRanking, DESC` as the second sort order

app/scenes/Search/Search.tsx
salihudickson: 🤔 perhaps a sort icon to the left or the right of the search bar?

tommoor (approved):

PR #11250, opened 2026-01-23
Word import fix #11219

### LINKED ISSUE #11219: Bugs - Word Import

### Current Behavior

Hi,

First Bug:

If you import a Word document with a table inside. An additional table header row is added to the table.

It doesn't matter whether you configure: a table header in word or not.

--> I would expect that the word config is respected by the import.

Image

Second Bug:

You cannot add text below the picture, if the picture is the last element of the Word file. The empty line after the pic is not respected by the import.

https://github.com/user-attachments/assets/0994707a-4c24-4a96-bba9-ae9a9dedeab7

Third Bug:

As you can see in the video - I cannot move the picture to another line. 

Here is the word example:
[outline Test.docx](https://github.com/user-attachments/files/24711256/outline.Test.docx)

### Steps To Reproduce

_No response_

### Environment

```markdown
- Outline: 1.2.0
- Browser: Chrome (version 142.0.7444.163)
```

### Anything else?

_No response_

RudraPrasad001:
Fix for Word Document Imports ( #11219 )

Bug 1 (Table Header): Word imports handles adding an extra table header row.

Bug 2 (Cursor Trap): Cannot type after images at document end because ProseMirror needs a trailing paragraph when the last paragraph contains only images (no text).

Bug 3 (Image Dragging): Images are not draggable to other lines because draggable was set to false in the image node schema.

https://github.com/user-attachments/assets/905e7223-a792-48a8-8830-e26ae345ff62

Tested with the given Document aswell

Cheers.

### DIFF
--- a/server/utils/DocumentConverter.test.ts
+++ b/server/utils/DocumentConverter.test.ts
@@ -29,3 +29,66 @@ John;25
     expect(await DocumentConverter.csvToMarkdown(csv)).toEqual(markdown);
   });
 });
+
+describe("docXToMarkdown", () => {
+  describe("tables with Word-style headers", () => {
+    it("should detect h1 tags in first row as header", async () => {
+      const docxBuffer = Buffer.from(""); 
+      
+      const html = `<table><tbody><tr><td><h1>Col1</h1></td><td><h1>Col2</h1></td><td><h1>Col3</h1></td></tr><tr><td><p>Data1</p></td><td><p>Data2</p></td><td><p>Data3</p></td></tr></tbody></table>`;
+      
+      const turndownService = require("./turndown").default;
+      const result = turndownService.turndown(html);
+      
+      const lines = result.trim().split('\n').filter((line: string) => line.trim());
+      
+      expect(lines[0]).toContain('Col1');
+      expect(lines[0]).toContain('Col2');
+      expect(lines[1]).toMatch(/\|\s*:?---/); 
+      expect(lines[2]).toContain('Data1');
+      expect(lines.length).toBe(3);
+    });
+
+    it("should detect h2-h6 tags in first row as header", async () => {
+      const turndownService = require("./turndown").default;
+      
+      const html = `<table><tbody><tr><td><h2>Header1</h2></td><td><h3>Header2</h3></td></tr><tr><td><p>Row1</p></td><td><p>Row2</p></td></tr></tbody></table>`;
+      const result = turndownService.turndown(html);
+      
+      const lines = result.trim().split('\n').filter((line: string) => line.trim());
+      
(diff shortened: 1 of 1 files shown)


shared/editor/extensions/TrailingNode.ts
```diff
@@ -41,6 +41,12 @@ export default class TrailingNode extends Extension {
         state: {
           init: (_, state) => {
             const lastNode = state.tr.doc.lastChild;
+
```
tommoor, line 44: This change is mergable separately 👍

shared/editor/nodes/Image.tsx
```diff
@@ -131,7 +131,7 @@ export default class Image extends SimpleImage {
       group: "inline",
       selectable: true,
       // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1289000
-      draggable: false,
+      draggable: true,
```
tommoor, line 134: See comment above, this line fixes another issue

server/utils/turndown/tables.ts
```diff
@@ -109,25 +109,50 @@ rules.table = {
     // Ensure there are no blank lines
     content = content.replace(/\n+/g, "\n");
 
-    // If table has no heading, add an empty one so as to get a valid Markdown table
-    const secondLineParts = content.trim().split("\n");
-    let secondLine = "";
-    if (secondLineParts.length >= 2) {
-      secondLine = secondLineParts[1];
+    // Check if table has a proper heading row
```
tommoor, line 112: Would need test coverage for this one

shared/editor/extensions/TrailingNode.ts
RudraPrasad001: Okay :+1:

server/utils/turndown/tables.ts
RudraPrasad001: Added test coverage for the table header detection fix in `server/utils/DocumentConverter.test.ts`:

**Tests added:**
-  Detects h1 tags in first row as header
-  Detects h2-h6 tags in first row as header  
-  Adds empty header when no heading tags present (backward compatibility)
-  Respects existing th tags as headers

All tests passing. The fix correctly handles Word documents where Mammoth converts table headers to `<td><h1>` instead of `<th>` tags.

Hope this is fine.

shared/editor/nodes/Image.tsx
RudraPrasad001: Okay, I will see to it. Thanks

tommoor:
Thanks, this was now fixed by completely avoiding markdown as part of the `.docx` conversion process here – https://github.com/outline/outline/pull/11331

PR #11253, opened 2026-01-23
feat(linear): add option to hide embed in paste menu

sgaluza:
## Summary

Linear sets `frame-ancestors 'self' https://cms.linear.app` in their CSP headers, which means embedding Linear links via iframe simply doesn't work — the browser blocks it silently. Despite this, Outline still shows an "Embed" option in the paste menu when you paste a Linear link, which leads to a blank/broken embed.

This PR adds a simple toggle in Linear integration settings: "Hide embed option". When enabled, the broken "Embed" choice is hidden from the paste menu for Linear links, so users only see the "Link" option that actually works.

Related discussion: https://github.com/outline/outline/discussions/4427

You can verify Linear's CSP yourself:
```
curl -sI https://linear.app | grep frame-ancestors
# frame-ancestors 'self' https://cms.linear.app
```

### DIFF
--- a/app/editor/components/PasteMenu.tsx
+++ b/app/editor/components/PasteMenu.tsx
@@ -50,6 +50,20 @@ export const PasteMenu = observer(({ pastedText, embeds, ...props }: Props) => {
   );
 });
 
+/** Find integration for a URL if it's mentionable */
+function findIntegrationForUrl(
+  text: string,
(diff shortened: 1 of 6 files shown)


tommoor:
I'm afraid it doesn't make sense to have this as a setting, it would be reasonable to maintain an internal block list of domains we know cannot be embedded however.

PR #11266, opened 2026-01-25
feat: add configurable embed domain blocklist

sgaluza:
## Summary

Supersedes #11253 with a more flexible, configurable approach instead of hardcoding Linear.

Adds ability to configure which domains should not show the embed option in the paste menu. This gives admins control over preventing embeds from specific services (e.g., internal tools like Linear, GitHub issues) while keeping the link and mention options available.

CleanShot 2026-01-25 at 10 06 17@2x

CleanShot 2026-01-25 at 10 05 09@2x

CleanShot 2026-01-25 at 10 04 30@2x

If Rich Embed is disabled the new block is hidden too and if not in effect:
CleanShot 2026-01-25 at 11 22 00@2x

## Changes

- Add `blockedEmbedDomains` field to Team model
- Add admin UI in Settings → Security for managing blocked domains  
- Respect existing `documentEmbeds` team setting - when disabled globally, embed option is hidden
- Show paste menu with Mention option even when embeds are disabled for a domain
- Default blocklist includes: linear.app, github.com, notion.so, slack.com, atlassian.net, asana.com

## Implementation

- `isDomainBlocked()` utility checks if URL hostname matches or is subdomain of blocked domain
- Reuses existing `DomainManagement` component (also used for allowed email domains)
- Server-side validation with Zod schema
- Full test coverage for API and blocklist logic

## Test plan

- [x] Paste URL from blocked domain → Embed hidden, Mention visible
- [x] Paste URL from allowed domain → Both Embed and Mention visible
- [x] Disable `documentEmbeds` globally → Embed hidden for all URLs
- [x] Add/remove domains in Settings → Changes apply immediately
- [x] Empty blocklist → Default blocklist applies

### DIFF
--- a/app/editor/components/PasteMenu.tsx
+++ b/app/editor/components/PasteMenu.tsx
@@ -62,8 +62,9 @@ function useItems({
   const { t } = useTranslation();
   const { integrations } = useStores();
   const user = useCurrentUser({ rejectOnEmpty: false });
-  const team = useCurrentTeam();
-  const blockedDomains = team.blockedEmbedDomains ?? DEFAULT_BLOCKED_EMBED_DOMAINS;
+  const team = useCurrentTeam({ rejectOnEmpty: false });
+  const blockedDomains = team?.blockedEmbedDomains ?? DEFAULT_BLOCKED_EMBED_DOMAINS;
+  const documentEmbeds = team?.documentEmbeds ?? true;
 
   // single item is pasted.
   if (typeof pastedText === "string") {
@@ -107,7 +108,7 @@ function useItems({
         name: "embed",
         title: t("Embed"),
         visible:
-          team.documentEmbeds &&
+          documentEmbeds &&
           !!embed &&
           !isDomainBlocked(pastedText, blockedDomains),
         icon: embed?.icon,
@@ -164,7 +165,7 @@ function useItems({
 
   // Check if embed option would actually be visible (respects documentEmbeds setting)
   const embedVisible =
-    team.documentEmbeds &&
+    documentEmbeds &&
     convertibleToEmbedList &&
     !pastedText.some((text) => isDomainBlocked(text, blockedDomains));
 


tommoor:
It feels like a couple of concepts are getting mixed up here, what is your thinking for the reason for an admin wanting to block a specific domain from being embedded?

It seems especially confusing to show popular services in the list by default

sgaluza:
The main problem is CSP `frame-ancestors` restrictions. Many popular services block iframe embedding via security headers, so clicking "Embed" in Outline results in a blank/broken iframe.

Related discussion: https://github.com/outline/outline/discussions/4427

All domains in the default list block iframe embedding:

| Domain | Restriction |
|--------|-------------|
| linear.app | `frame-ancestors 'self' https://cms.linear.app` |
| github.com | `frame-ancestors 'none'` + `X-Frame-Options: deny` |
| notion.so | `frame-ancestors 'self'` + `X-Frame-Options: SAMEORIGIN` |
| slack.com | `X-Frame-Options: SAMEORIGIN` |
| asana.com | `frame-ancestors 'self'` + `X-Frame-Options: DENY` |
| atlassian.net | `X-Frame-Options: DENY` |

You can verify with curl:
```bash
curl -sI https://linear.app | grep -i "frame-ancestors\|x-frame-options"
# frame-ancestors 'self' https://cms.linear.app

curl -sI https://github.com | grep -i "frame-ancestors\|x-frame-options"
# x-frame-options: deny
# frame-ancestors 'none'

tommoor:
I agree, so it doesn't make sense to have this as a configurable "security" list – lets just hardcode a list of domains that we know cannot be embedded – especially those that are mentionable. Much simpler.

sgaluza:
Actually, there's another consideration - CSP can be bypassed on the client side:

1. **Enterprise browser policies** - organizations can configure browsers to ignore CSP via group policies altogether
2. **Proxies** - it's possible to strip CSP headers through their proxy infrastructure  
3. **Browser extensions** - can modify response headers

In these cases, embeds actually work, and SOME admins would want to keep the embed option available.

This is why a configurable list makes sense - different organizations have different setups. Hardcoding would break the experience for companies that have legitimately worked around CSP restrictions.

The default blocklist covers common cases where embedding doesn't work, but admins can remove domains if their setup allows embedding.

tommoor:
Thanks.

Went with an alternative solution checking whether the resource is embeddable in realtime, the end result is the same - that links which cannot be embedded do not have the option but without any maintenance.

PR #11272, opened 2026-01-26
Words separated by hyphens to be treated as a single unit for word diffing

apoorv-mishra:
Screenshot 2026-01-26 at 9 32 40 PM
vs
Screenshot 2026-01-26 at 9 32 57 PM

### DIFF
--- a/app/scenes/Developer/components/ExampleData.ts
+++ b/app/scenes/Developer/components/ExampleData.ts
@@ -14,6 +14,38 @@ export interface Example {
  */
 export const examples: Example[] = [
   {
+    id: "word-replacement",
+    name: "Word replacement (interleaved)",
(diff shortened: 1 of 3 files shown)


tommoor:
Do you remember what the limitation was? It's a good improvement, wouldn't mind taking a pass at getting it prod ready

apoorv-mishra:
This was fully Claude generated and I'm not sure if it works out there in the wild. It _seems_ to work. But there was a much simpler way to configure word-level diffing directly through the diffing library. But, that didn't work because when the diff was passed to `prosemirror-changeset`, it's own diffing algo optimized(minimized) changes and that messed up the final diffing and it didn't end up as expected, i.e., not a proper word-level diffing.

PR #11323, opened 2026-01-31
fix: Do not show embed option for unembeddable links

tommoor:
We now collect headers from the link to determine if the resource is embeddable and appropriately disable the option if so. This removes the need for a hardcoded or manually curated list of urls.

The result is cached for one day and has a short timeout.

Supersedes #11266


app/editor/components/SuggestionsMenu.tsx
```diff
@@ -292,6 +292,10 @@ function SuggestionsMenu<T extends MenuItem>(props: Props<T>) {
 
   const handleClickItem = React.useCallback(
     (item) => {
+      if (item.disabled) {
```
tommoor, line 295: Fixes an issue where disabled items were considered for click and arrow navigation

PR #11403, opened 2026-02-10
filter group members

### LINKED ISSUE #11363: Add a filter to the "Group members" dialog
There may be hundreds of users in a group, it is useful to quickly filter and find if a specific user is a member

salihudickson:
fixes #11363

### DIFF
--- a/app/scenes/Settings/components/GroupDialogs.tsx
+++ b/app/scenes/Settings/components/GroupDialogs.tsx
@@ -363,25 +363,27 @@ export const ViewGroupMembersDialog = observer(function ({
           />
         </Text>
       )}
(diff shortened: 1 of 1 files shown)


app/scenes/Settings/components/GroupDialogs.tsx
```diff
@@ -304,13 +363,38 @@ export const ViewGroupMembersDialog = observer(function ({
           />
         </Text>
       )}
+      <Flex gap={8}>
```
tommoor: Lets hide this section if the group is empty, otherwise looks good.

salihudickson: done

PR #11469, opened 2026-02-16
Toggle toggle block's state upon clicking its head

apoorv-mishra:
Only for read-only mode. Edit mode takes cursor selection.

### DIFF
--- a/shared/editor/nodes/ToggleBlockView.ts
+++ b/shared/editor/nodes/ToggleBlockView.ts
@@ -7,7 +7,6 @@ import {
   toggleEventPluginKey,
   toggleFoldPluginKey,
 } from "./ToggleBlock";
-import { isSelectionInToggleBlockHead } from "../queries/toggleBlock";
 
 /**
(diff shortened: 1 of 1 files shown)


tommoor:
It would be nice if there was a clue, we should probably add the pointer cursor at least?

It's tempting to render an actual invisible button in read-only mode...

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -589,6 +546,52 @@ export default class ToggleBlock extends Node {
     };
   }
 
+  private createDecorations(doc: ProsemirrorNode, foldedIds: Set<string>) {
```
apoorv-mishra, line 549: Moved this within the class so that it had access to `this.editor.props.readOnly`

tommoor:
There is a minor issue that the cursor decoration does not get recreated when switching between edit/readonly, but it does fix itself once other transactions flow through the editor. I'm merging, but we might want to come back to that for perfection

PR #11478, opened 2026-02-17
feat: add anchorText parameter to comments.create API endpoint

bazmcdawson:
## Summary

Add an optional `anchorText` parameter to the `POST /api/comments.create` endpoint that automatically anchors a comment to specific text in the document by applying a ProseMirror `comment` mark.

## Motivation

Currently, anchoring comments to specific text ranges in a document can only be done through the frontend editor. This makes it impossible for API consumers (integrations, bots, automation tools) to create anchored comments. This PR enables that use case with a simple, optional parameter.

## Changes

- **`server/routes/api/comments/schema.ts`**: Added optional `anchorText` field to `CommentsCreateSchema`
- **`server/routes/api/comments/comments.ts`**: After creating the comment, if `anchorText` is provided, finds the first occurrence of that text in the document's ProseMirror content and applies a `comment` mark with `{id, userId, resolved: false, draft: false}`
- **`shared/utils/ProsemirrorHelper.ts`**: Added `addCommentMark()` static method that operates on ProseMirror JSON data, splitting text nodes as needed to apply the mark to the exact matching range
- **Tests**: Added unit tests for `addCommentMark()` and integration tests for the API endpoint

## Behavior

- If `anchorText` matches text in the document, a comment mark is applied to the **first occurrence**
- If `anchorText` is not found, the comment is still created successfully (just without anchoring)
- The parameter is entirely optional and backward-compatible
- Existing marks on the text are preserved

### DIFF
--- a/server/routes/api/comments/comments.ts
+++ b/server/routes/api/comments/comments.ts
@@ -18,6 +18,7 @@ import validate from "@server/middlewares/validate";
 import { Document, Comment, Collection, Reaction, Emoji } from "@server/models";
 import { ProsemirrorHelper } from "@server/models/helpers/ProsemirrorHelper";
 import { ProsemirrorHelper as SharedProsemirrorHelper } from "@shared/utils/ProsemirrorHelper";
+import { ProsemirrorHelper as ServerProsemirrorHelper } from "@server/models/helpers/ProsemirrorHelper";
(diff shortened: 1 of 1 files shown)


bazmcdawson:
recheck

PR #11493, opened 2026-02-19
Perf: apply initial decorations early on for toggle blocks

apoorv-mishra:
~Editor plugins are registered/initialized during editor state init, which happens _before_ editor view init. Therefore, all `init()` calls in editor plugins have `this.editor.view` as undefined. Here, we introduce a Plugin Manager that sets the expected initial state _before_ editor state init and holds it. Then, the `init()` within the corresponding plugin simply references the Plugin Manager to fetch initial state. This gets rid of extra init state setting plugins from within nodes, for example the `initPlugin` from `ToggleBlock`. It also improves perf since the extra init plugins don't need to separately dispatch an "init" transaction(to apply decorations for example).~

Observe the toggle button animating to folding state upon refresh, in read-only mode...

https://github.com/user-attachments/assets/e4ca4054-28e6-4421-a648-29a7b27eb05e

After this change, it no longer does that...

https://github.com/user-attachments/assets/9f2b4802-6cdf-4714-b21b-f6557e841bcd

### DIFF
--- a/shared/editor/nodes/ToggleBlock.ts
+++ b/shared/editor/nodes/ToggleBlock.ts
@@ -7,7 +7,7 @@ import type {
   Node as ProsemirrorNode,
   Schema,
 } from "prosemirror-model";
-import type { Command, Transaction } from "prosemirror-state";
+import type { Command, EditorState, Transaction } from "prosemirror-state";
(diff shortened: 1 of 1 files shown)


apoorv-mishra:
This simple 😄 @tommoor

PR #11499, opened 2026-02-20
fix: Read-only collection editor does not remount correctly on nav

tommoor:
Interesting case when linking between two shared collections and navigating back and forth the editor would not remount the new content.


tommoor:
closes #11476

PR #11539, opened 2026-02-24
Fix: Right sidebar header misaligned

apoorv-mishra:
Notice the button overflowing onto scroll track:
Screenshot 2026-02-24 at 10 16 43 AM

### DIFF
--- a/app/components/Sidebar/Right.tsx
+++ b/app/components/Sidebar/Right.tsx
@@ -70,7 +70,7 @@ function Right({ children, border, className }: Props) {
   const style = React.useMemo(
     () => ({
       width: windowScrollbarWidth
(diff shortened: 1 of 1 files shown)


tommoor:
It looks correct when no scroll, we might want to detect the scrollbar

app/components/Sidebar/Right.tsx
```diff
@@ -67,9 +69,11 @@ function Right({ children, border, className }: Props) {
 
   const style = React.useMemo(
     () => ({
-      width: `${ui.sidebarRightWidth}px`,
+      width: windowScrollbarWidth
+        ? `${ui.sidebarRightWidth - 6}px`
```
apoorv-mishra: Resizing just enough so as not to overflow onto scrollbar.

Issue #11573, opened 2026-02-25
Notion import fails for large workspaces due to search cursor expiration


james-albedo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

Notion imports fail for workspaces with ~10,000+ pages. The import fails during the initial page discovery phase with:

```
Failed: The start_cursor provided is invalid: 4d49987c-a64c-4ab8-916e-d9aeaffa265b
```

`fetchRootPages()` in `plugins/notion/server/notion.ts` paginates through all workspace pages via the Notion search API. The search cursor has an undocumented TTL that expires before pagination completes. With `pageSize: 25`, a 15k page workspace requires ~600 API calls at 3 req/sec, taking over 3 minutes. The cursor seems to expire before that.

`fetchWithRetry()` handles rate limits and timeouts but not cursor expiration, so the import fails without recovery. This is 100% reproducible on any large workspace.

### Expected Behavior

The Notion import should complete successfully regardless of workspace size. If a search cursor expires mid-pagination, the import should recover and continue rather than failing outright.

### Steps To Reproduce

1. Connect a Notion workspace containing 10,000+ pages via Settings > Integrations > Notion
2. Start an import from Settings > Import > Notion
3. Import enters "processing" state, then fails with "The start_cursor provided is invalid"
4. Every subsequent import attempt fails at the same point

### Environment

```markdown
- Outline: 1.5.0 (Enterprise, self-hosted on ECS Fargate)
- Browser: N/A (server-side failure)
- Notion workspace: ~15,000 pages
```

### Anything else?

**Root cause:** The Notion search API has no server-side filter for `parent.type === "workspace"`, so `fetchRootPages()` must paginate through every page in the workspace to find workspace-level items. This is a Notion API limitation.

**Proposed fix (two changes to `plugins/notion/server/notion.ts`):**

1. Increase `pageSize` from 25 to 100 (Notion API maximum). Reduces API calls from ~600 to ~150 for a 15k page workspace.

2. Add cursor recovery to `fetchRootPages()`. When the cursor expires, restart the search from the beginning, skip already-collected pages via a `Set<string>` of seen IDs, and allow up to 3 restarts before failing. Log recovery events for observability.

The change is isolated to `fetchRootPages()` only. No interface changes, no other files modified. The method returns the same `Page[]` as before.

[...]

PR #11576, opened 2026-02-25
fix: Notion import fails for large workspaces due to search cursor expiration

james-albedo:
## Summary

`fetchRootPages()` paginates through all workspace pages via the Notion search API to find workspace-level items. The search cursor has an undocumented positional limit (~11,200 pages) that causes it to expire before pagination completes on large workspaces, failing the import with `The start_cursor provided is invalid`.

This PR adds a `NOTION_ROOT_PAGE_IDS` environment variable and improves the existing search path:

- **When `NOTION_ROOT_PAGE_IDS` is set:** fetches root pages directly via `pages.retrieve`/`databases.retrieve`, bypassing the search API entirely. The existing tree traversal in `parseChildPages()` handles all child content discovery, so only top-level entry points are needed.
- **When not set:** existing search behavior is preserved with `pageSize` increased from 25 to 100 (Notion API max, 4x fewer API calls). On cursor expiry, the error message now directs the admin to set `NOTION_ROOT_PAGE_IDS` instead of showing the raw Notion API error.

Changes are in `plugins/notion/server/notion.ts`, `plugins/notion/server/env.ts`, and `.env.sample`. No interface changes to `fetchRootPages()` — it returns the same `Page[]` either way.

## Context

See #11573 for investigation details and test results.

## Test plan

- [ ] Import from a Notion workspace under 10k pages (search path, should work as before)
- [ ] Import from a large Notion workspace (10k+) with `NOTION_ROOT_PAGE_IDS` set to workspace-level page/database IDs
- [ ] Verify `NOTION_ROOT_PAGE_IDS` with a mix of page and database IDs resolves both types correctly
- [ ] Verify invalid/inaccessible IDs in `NOTION_ROOT_PAGE_IDS` are skipped with a warning
- [ ] Verify cursor expiry without env var produces actionable error message

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -246,6 +246,14 @@ SENTRY_TUNNEL=
 NOTION_CLIENT_ID=
 NOTION_CLIENT_SECRET=
 
+# Comma-separated list of Notion page/database IDs to use as import roots.
+# Required for large workspaces (10k+ pages) where the search API cursor
+# expires before pagination completes. When set, root pages are fetched
+# directly by ID instead of scanning the workspace via search.
+# To find a page ID: open it in Notion > Share > Copy link. The ID is the
+# 32-char hex string at the end of the URL (before any ?v= parameter).
+# NOTION_ROOT_PAGE_IDS=
+
 # The Iframely integration allows previews of third-party content within Outline.
 # For example, hovering over an external link will show a preview.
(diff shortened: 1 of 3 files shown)


james-albedo:
Closed per comments in https://github.com/outline/outline/issues/11573

PR #11584, opened 2026-02-26
Fix: Horizontal overflow in notification email

apoorv-mishra:
Prevents horizontal overflow in comment created notif email.


apoorv-mishra:
Don't know if this is the accurate fix. I couldn't reproduce it locally since ethereal renders differently than Gmail. But in my investigation Gmail inlining body's 100vw width seems to be the root cause.

cc @tommoor

PR #11595, opened 2026-02-27
feat: Add support for configurable proxy IP header in environment settings

### LINKED ISSUE #6083: Cannot log in after upgrade. Validation isIP on lastActiveIp failed (lastActiveIp)
We've upgraded the wiki from 0.61.1
 - Most users cannot log in due to error in title
 - The ones who are already logged in cannot perform any write operation

Log is full of these errors:
```
{
"error":"Validation error: Validation isIP on lastActiveIp failed",
"level":"error",
"message":"Failed to update user activeAt",
"stack":"SequelizeValidationError: Validation error: Validation isIP on lastActiveIp failed\n at InstanceValidator._validate (/opt/outline/node_modules/sequelize/lib/instance-validator.js:50:13)\n at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n at async InstanceValidator.validate (/opt/outline/node_modules/sequelize/lib/instance-validator.js:54:12)\n at async User.save (/opt/outline/node_modules/sequelize/lib/model.js:2426:7)"
}
```

We're using a Slack as login provider.

I have gone through all the users in the DB and experimentally tried to manually fill in missing lastActiveIps. No effect. (Don't worry I have the DB backed up). 
Also, the error says "Failed to update user activeAt" although there is no `activeAt` column in the users table. I'm quite confused here.

**To Reproduce**
1. Have a 0.61.1 version installed
2. Updated to v0.72.2 (tried also 0.69.2)

**Expected behavior**
Users can log in.

**Screenshots**
Screenshot 2023-10-27 at 15 41 06
Screenshot 2023-10-27 at 15 43 45

- Install: self hosted
- Version: outlinewiki/outline:0.72.2 docker image

 - OS: OSX 11.7.4
 - Browser chrome

ashleysommer:
Some hosts (looking mostly at Azure App Service, and other Azure Cloud services) do not provide a valid IP address in the `X-Forwarded-For` header. Instead, it provides an IP and a Port like `1.2.3.4:55201`. This breaks Outline's IP validation rules when writing to the database. See discussion in #6083

This has been an issue of Azure and some other hosts for over a decade and for compatibility reasons will never be fixed. As a workaround, some hosts include a different header like `X-Real-IP` or `X-Client-IP` that has _just_ the IP address, and it maintains the same IP across multiple proxy hops. Azure App Service uses `X-Client-IP`.

Most web app frameworks have ways of accommodating this.
Outline's 'Koa' app framework has support for changing the proxied IP header read, called `app.proxyIpHeader`. However when running Outline in a container, there is no way to set this config value.

See: 
https://github.com/koajs/koa/blob/d3ea8bf9649d164b40c448b5ce0b40306b4dca90/docs/api/request.md?plain=1#L195

This PR adds an Environmental configuration option to Outline, that passes the header name to Koa's `app.proxyIpHeader` property. 

Fixes #6083

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -119,9 +119,9 @@ SSL_CERT=
 # false if you can be sure that SSL is terminated at an external loadbalancer.
 FORCE_HTTPS=true
 
-# When behind a reverse proxy, the header to use for the client IP. Set to
-# X-Real-IP if your proxy sends that instead of X-Forwarded-For. Leave unset
-# to use Koa default (X-Forwarded-For).
+# When behind a reverse proxy, the header to use for the client IP.
+# The default value is "X-Forwarded-For", common values are "X-Real-IP"
+# and "X-Client-IP".
 # PROXY_IP_HEADER=
 
 


server/env.ts
```diff
@@ -341,6 +341,18 @@ export class Environment {
   @IsBoolean()
   public FORCE_HTTPS = this.toBoolean(environment.FORCE_HTTPS ?? "true");
 
+  /**
+   * When the app is behind a proxy (app.proxy is true), sets the HTTP header
+   * name Koa uses for the client IP. Common values are "X-Forwarded-For"
+   * (Koa default) or "X-Real-IP". Azure App Service uses "X-Client-IP".
+   * Only has effect in production when proxy
+   * headers are trusted.
```
tommoor:
```suggestion
   * When the app is behind a proxy, sets the HTTP header used for the client IP. 
   * The default value is "X-Forwarded-For", common values are "X-Real-IP"
   * and "X-Client-IP".
```

.env.sample
```diff
@@ -119,6 +119,11 @@ SSL_CERT=
 # false if you can be sure that SSL is terminated at an external loadbalancer.
 FORCE_HTTPS=true
 
+# When behind a reverse proxy, the header to use for the client IP. Set to
+# X-Real-IP if your proxy sends that instead of X-Forwarded-For. Leave unset
+# to use Koa default (X-Forwarded-For).
```
tommoor:
```suggestion
# When behind a reverse proxy, the header to use for the client IP.
# The default value is "X-Forwarded-For", common values are "X-Real-IP"
# and "X-Client-IP".
```

tommoor (changes requested):
I'm ok with merging this but please remove mentions of Koa from the docs

ashleysommer:
Thanks @tommoor 
Removed mention of Koa as per suggestions.

tommoor (approved):

Issue #7362, opened 2024-08-07
Allow viewers to add inline comments


tommoor:
Tracking issue for this request – to achieve we'll need to have the backend edit the document, for that reason the implementation here ends up very similar to what is required to fix https://github.com/outline/outline/issues/6459

keyhana:
Hey,
I see #8605 has been marked as a duplicate of this issue. Thanks for pointing that out.
I'm still interested in this capability as it would enable more control over the permission of our docs (We don't want to give edit access on some of them). 
Is there any timeline or plan for implementing this feature?
Thanks!

amireh:
Occasionally I find myself sharing content in my private wiki with friends/colleagues, and was asked a few times why they couldn't just leave comments there. It'd be nice to allow readers to leave comments - not necessarily inline, just a thread at the bottom - without resorting to using a dedicated blog engine.

I realize it's a niche case, and would understand if the ask is out of scope, but this would be very welcome to me.

Thank you for the great product!

EduNetEurope:
For our use case, it would be very helpful if view-only users could could comment on specific text.

marijnvandevoorde:
Want to add a +1 on this issue. Not a dealbreaker, but it'd be very nice for sure. We have documents that we consider meeting reports or final versions, so it'd be cool if we could block further editing but still allow comments.

marijnvandevoorde:
@tommoor I'm actually up for having a go at creating an MR for this. However, I was wondering what you had in mind. Would you have a separate commenter role, in between viewer and editor, or just allow viewers to comment. In a way, I see little harm in allowing viewers to comment, but that's my use case of course.

If the idea is that viewers could add comments, I'm pretty much done: 

<img width="158" height="100" alt="Image" src="https://github.com/user-attachments/assets/2c060c17-863c-447a-9ec1-1ab34cdb6438" />

adrfantini:
This is also a deal breaker for us.
In large-ish companies it's unsafe to give everyone editing rights, but at the same time it's important that everyone, especially for internal documents, is able to comment inline on the resource, which can then get improved over time by the editors.

gbarbon:
Hi, any updates on this issue? For us it is quite important to give readers (with only view permissions) the capability to write inline comments.

AlexWHughes:
For my organisation having this implemented even as a seperate user permission (View/Comment) would get us out of the hole we're in while waiting for #https://github.com/outline/outline/issues/1084 to be added.

[...]

PR #11664, opened 2026-03-05
feat: add anchorText parameter to comments.create API

NullSense:
## Problem

Currently, the `comments.create` API endpoint can only create **document-level (unanchored) comments** — they appear in the sidebar but don't link to any specific text in the document. Anchoring a comment to text requires browser-based editor interaction, which makes it impossible for API consumers (AI agents, MCP servers, CI/CD pipelines, code review bots) to create contextual inline comments.

The API already supports *reading* anchor text (`includeAnchorText` on `comments.list` and `comments.info`), but has no way to *write* it.

## Solution

Add an optional `anchorText` string parameter to the `comments.create` endpoint. When provided:

1. The server creates the comment as normal
2. Finds the first occurrence of `anchorText` in the document's ProseMirror content
3. Wraps the matching text range with a `comment` mark containing the new comment's ID
4. Saves the updated document content

If the text is not found, the comment is created as a document-level comment (graceful degradation). Reply comments (`parentCommentId` set) ignore `anchorText` since they inherit the parent's anchor.

## Changes

- **`schema.ts`**: Add `anchorText` (optional, max 1000 chars) to `CommentsCreateSchema`
- **`ProsemirrorHelper.ts`**: Add `addCommentMark()` static method that finds text in a ProseMirror JSON tree and injects a comment mark — handles text node splitting, preserves existing marks, and only marks the first occurrence
- **`comments.ts`**: Update `comments.create` handler to apply the mark when `anchorText` is provided
- **Tests**: Unit tests for `addCommentMark()` (5 cases) and integration tests for the API endpoint (3 cases)

## API Usage

```json
POST /api/comments.create
{
  "documentId": "<uuid>",
  "text": "This section needs legal review.",
  "anchorText": "Non-Compete"
}
```

The comment will appear anchored to the text "Non-Compete" in the document, highlighted in the editor with the comment linked in the sidebar — identical to how it looks when created through the UI.

## Use Cases

- **AI document review**: Agents can annotate specific sections of documents with feedback
- **MCP servers**: Model Context Protocol integrations for Outline (e.g., `dts-outline-mcp-server`)
- **Import/migration tools**: Preserve inline comments when migrating from other platforms
- **CI/CD**: Automated documentation review pipelines
- **Relates to #7362** (backend editing of documents for comment anchoring)

## Testing

- Unit tests for `ProsemirrorHelper.addCommentMark()`: exact match, not found, preserve existing marks, first-occurrence-only, nested nodes
- Integration tests: anchor created, text not found (graceful), reply comments ignored

### DIFF
--- a/server/routes/api/comments/comments.test.ts
+++ b/server/routes/api/comments/comments.test.ts
@@ -773,6 +773,103 @@ describe("#comments.create", () => {
   });
 });
 
+describe("#comments.create anchorText", () => {
+  it("should anchor comment to text in document", async () => {
+    const team = await buildTeam();
+    const user = await buildUser({ teamId: team.id });
+    const document = await buildDocument({
+      userId: user.id,
+      teamId: user.teamId,
+      text: "# Heading\n\nThis is some paragraph text with important content here.",
+    });
+
+    const comment = await buildComment({
+      userId: user.id,
+      documentId: document.id,
+    });
+
+    const res = await server.post("/api/comments.create", {
+      body: {
+        token: user.getJwtToken(),
+        documentId: document.id,
+        data: comment.data,
+        anchorText: "important content",
+      },
+    });
+
+    const body = await res.json();
+    expect(res.status).toEqual(200);
+
+    // Verify the document now has the comment mark
+    await document.reload();
+    const content = document.content as any;
+    const hasCommentMark = JSON.stringify(content).includes(body.data.id);
+    expect(hasCommentMark).toBeTruthy();
+  });
+
+  it("should create unanchored comment when text not found", async () => {
+    const team = await buildTeam();
+    const user = await buildUser({ teamId: team.id });
+    const document = await buildDocument({
+      userId: user.id,
+      teamId: user.teamId,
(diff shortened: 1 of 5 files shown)


tommoor:
I've been working on something similar [here](https://github.com/outline/outline/compare/main...copilot/add-inline-comments-support) that uses the encoded relative position for anchoring, that way a specific position in the document can be defined - even during an active multiplayer session.

How does this handle if the selected text appears in multiple locations?

NullSense:
> I've been working on something similar [here](https://github.com/outline/outline/compare/main...copilot/add-inline-comments-support) that uses the encoded relative position for anchoring, that way a specific position in the document can be defined - even during an active multiplayer session.
> 
> How does this handle if the selected text appears in multiple locations?

First match only — so duplicates would anchor to the earliest occurrence. Your RelativePosition approach is fundamentally more robust since it handles both duplicates and multiplayer cursor drift.

Happy to close this in favor of yours. If you'd like an extra set of eyes on your branch at any point, I'd be glad to review — just let me know.

Issue #11358, opened 2026-02-04
Automatically collapse tall code blocks


tommoor:
Tall code blocks should be truncated with a control to Expand/Collapse
Clicking into the code block should automatically expand

RudraPrasad001:
Hey Tom , I am willing to work on this , I have a few queries on the requirements,

How many lines should considered be considered as a tall code block? 

Typically 20+ lines is considered as a tall code block.

Also, is using expand and collapse the best approach , wouldn't it  be better if we use a *scrollable code container* 

Cheers!

PR #11699, opened 2026-03-09
collapsible code blocks

salihudickson:
#11358

### DIFF
--- a/app/editor/menus/code.tsx
+++ b/app/editor/menus/code.tsx
@@ -1,6 +1,4 @@
 import { CopyIcon, EditIcon, ExpandedIcon, TextWrapIcon } from "outline-icons";
-import { faChevronDown, faChevronUp } from "@fortawesome/free-solid-svg-icons";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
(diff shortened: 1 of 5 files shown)


app/editor/menus/code.tsx
```diff
@@ -63,6 +65,21 @@ export default function codeMenuItems(
     {
       name: "separator",
     },
+    {
```
tommoor: Lets add a new UI element at the bottom of the code block with the control to expand / collapse - it makes more sense at the bottom for this.

We can also make it so that when clicking into a code block that's collapsed it auto expands

tommoor: image. Kind of like this read more control on GitHub

salihudickson: i see. Thanks for the review.

salihudickson:
hey @tommor can you please take another look at this?

tommoor:
The implementation stores the collapsed state on a node attribute which will cause the state to sync between multiplayer clients, this is not desirable - collapse should be per user.

In this case it will be easier to do a clean room re-implementation.

PR #11705, opened 2026-03-10
 Enhancement: Fixed ALLOWED_DOMAINS env var for managing team domain restrictions

akshat-kumar-singhal:
## Summary
`ALLOWED_DOMAINS` environment variable was used only in the migration and not in the subsequent flows - without any warning to the admin hence creating a confusion.
 - When `ALLOWED_DOMAINS` is set, the settings UI shows the configured domains as read-only and blocks API updates to prevent conflicts
 - Falls back to the existing database-managed domains when the env var is not set

 ## Changes
 - **`server/env.ts`** — Added `getAllowedDomains()` helper to parse the comma-separated `ALLOWED_DOMAINS` env var
 - **`server/models/Team.ts`** — Added `getEffectiveAllowedDomains()` that checks env var first, then falls back to DB; refactored `isDomainAllowed` to use it
 - **`server/presenters/team.ts`** — Exposes `domainsManagedByEnv` in the team API response so the client knows when domains are env-managed
 - **`server/routes/api/teams/teams.ts`** — Rejects `allowedDomains` updates when the env var is set, returning a validation error
 - **`app/models/Team.ts`** — Added `domainsManagedByEnv` observable field
 - **`app/scenes/Settings/components/DomainManagement.tsx`** — Shows a read-only view of env-managed domains instead of the editable form
 - **`shared/i18n/locales/*/translation.json`** — Added translations for the env-managed domains message across all 27 locales

### DIFF
--- a/app/models/Team.ts
+++ b/app/models/Team.ts
@@ -79,7 +79,7 @@ class Team extends Model {
   allowedDomains: string[] | null | undefined;
 
   @observable
-  domainsManagedByEnv: boolean;
+  domainsManagedByEnv?: string[];
 
   @computed
   get signinMethods(): string {
--- a/app/scenes/Settings/components/DomainManagement.tsx
+++ b/app/scenes/Settings/components/DomainManagement.tsx
@@ -17,7 +17,27 @@ type Props = {
   onSuccess: () => void;
 };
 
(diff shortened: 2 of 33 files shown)


tommoor:
As discussed we'll remove the env instead.

PR #11775, opened 2026-03-16
fix: guard against concurrent restore in documentPermanentDeleter

dearlordylord:
`documentPermanentDeleter` receives pre-queried `Document` objects from cleanup tasks (`CleanupDeletedDocumentsTask`, `EmptyTrashTask`). A user restoring a document between the initial query and the permanent deletion causes silent data loss - the restored document is destroyed because `documentPermanentDeleter` trusts the stale in-memory `deletedAt` without re-checking the database.

- [**`server/commands/documentPermanentDeleter.ts`**](https://github.com/dearlordylord/outline/blob/fix/cleanup-recheck-deleted-before-permanent-delete/server/commands/documentPermanentDeleter.ts): Re-query each document's `deletedAt` via `Document.unscoped().count()` before processing; skip restored documents with a log message. Add `deletedAt: { [Op.ne]: null }` guard to the final `destroy` WHERE clause as a secondary safety net against the narrow window between re-check and destroy.

### DIFF
--- a/server/commands/documentPermanentDeleter.test.ts
+++ b/server/commands/documentPermanentDeleter.test.ts
@@ -111,6 +111,33 @@ describe("documentPermanentDeleter", () => {
     ).toEqual(0);
   });
 
+  it("should not destroy a document restored between query and destroy", async () => {
+    const document = await buildDocument({
+      publishedAt: subDays(new Date(), 90),
+      deletedAt: subDays(new Date(), 60),
+    });
+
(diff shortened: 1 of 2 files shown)


tommoor:
With the re-query there is still a race condition - to fix this properly we should just bake the deletion check into the where condition

dearlordylord:
oof, I misguided myself indeed. the update should tackle this better

PR #11814, opened 2026-03-19
feat(s3): add CloudFront via S3 presigned redirect

oliinykdm:
**feat: Add optional CloudFront support for S3 presigned download URLs**

### Overview
This PR adds support for serving attachment download / preview URLs through **AWS CloudFront** while keeping the existing short-lived S3 presigned URL security model intact.

Instead of generating presigned URLs directly against the S3 bucket endpoint, the code now (when configured) replaces the S3 domain with the CloudFront distribution domain.  
CloudFront then forwards the request (with the original signature) to S3.

**Benefits**
- All download traffic goes through CloudFront edge locations → better performance, lower latency, global caching
- S3 data transfer costs are significantly reduced (CloudFront has flat-rate tiers and even free one, so lowering the  DoW attack)
- Security model unchanged: URLs remain short-lived (default 1 hour), access still controlled via `/api/attachments.redirect`
- No new dependencies, no CloudFront key pairs / signed cookies required
- Fully backward compatible: if `AWS_CLOUDFRONT_URL` not set → behaves exactly as before

### Changes
- Added new optional env variable: `AWS_CLOUDFRONT_URL` (e.g. `https://d1234567890abcdef.cloudfront.net`)
- Modified `server/storage/files/S3Storage.ts`:
  - In `getSignedUrl()` — replace S3 bucket URL with CloudFront domain after generating the presigned URL
  - In `getUrlForKey()` — return CloudFront base URL when configured (for consistency in other places)
- Updated `server/env.ts` — added `AWS_CLOUDFRONT_URL` with `@Public` and `@IsOptional()`

No changes to uploads, database schema, API responses, or authentication flow.

### Testing steps
1. Set `AWS_CLOUDFRONT_URL=https://your-dist.cloudfront.net` in `.env`
2. Make sure S3 bucket is private + OAC policy allows your CloudFront distribution
3. Upload an image / file in Outline
4. Open the attachment / preview → check Network tab:
   - `/api/attachments.redirect` → 302 to `https://your-dist.cloudfront.net/...?X-Amz-Algorithm=...&X-Amz-Expires=3600...`
5. Without `AWS_CLOUDFRONT_URL` → should fallback to direct S3 presigned URL

**Update (2026-03-19):**  
After testing, I found that replacing the domain in presigned S3 URLs with CloudFront does not work reliably when Origin Access Control (OAC) is enabled.  
S3 rejects the client-generated signature when the request comes through CloudFront (OAC expects CloudFront's own signature).

To fix this:  
- Removed domain replacement in `getSignedUrl`  
- Now, when `AWS_CLOUDFRONT_URL` is set, a plain CloudFront URL is returned (`https://d...cloudfront.net/key`)  
- Access control remains intact via `/api/attachments.redirect` + private bucket with OAC  

This aligns with AWS recommendations for private S3 + CloudFront + OAC setups.  
No breaking changes - fallback to S3 presigned still works if CloudFront is not configured.

Testing confirmed: downloads go through CloudFront, direct S3 access returns 403, and permissions are enforced correctly.

Thanks for reviewing!

### DIFF
--- a/server/storage/files/S3Storage.ts
+++ b/server/storage/files/S3Storage.ts
@@ -167,7 +167,7 @@ export default class S3Storage extends BaseStorage {
       );
 
       const command = new GetObjectCommand(params);
-      let url = await getSignedUrl(this.client, command, {
+      const url = await getSignedUrl(this.client, command, {
         expiresIn: clampedExpiresIn,
       });
 


tommoor:
`AWS_S3_ACCELERATE_URL` already covers this usecase, perhaps it could be better named

oliinykdm:
> `AWS_S3_ACCELERATE_URL` already covers this usecase, perhaps it could be better named

Thanks for the quick reply!

I did try setting `AWS_S3_ACCELERATE_URL` to my CloudFront distribution domain, but unfortunately it breaks **uploads** completely.

From what I see in `S3Storage.getSignedUrl` (and the presigned POST logic for uploads), the accelerate URL replaces the base endpoint for **both** signed GET downloads **and** presigned POST uploads.  
When I set `AWS_S3_ACCELERATE_URL=https://d123...cloudfront.net`, the presigned POST policy gets generated against the CloudFront domain -> browser upload fails with signature mismatch / 403, because CF doesn't accept presigned POST the same way S3 accelerate does.

CloudFront works great for downloads (presigned S3 URLs are forwarded correctly), but uploads must stay direct to S3 (or accelerate endpoint) to avoid breaking the workflow.

That's why a separate variable like `AWS_CLOUDFRONT_URL` (or similar) would be ideal - it would allow replacing the domain **only** in `getSignedUrl` / `getUrlForKey` for downloads/previews, while keeping uploads on the original S3/accelerate URL.

oliinykdm:
**Update:** Removed presigned URL domain replacement for CloudFront.  
Presigned S3 URLs are incompatible with OAC - switched to plain CloudFront URLs.  
Access control via redirect + OAC, short-lived via redirect TTL.

PR #11818, opened 2026-03-19
fix: Parse bullet lists in table cells from markdown input

Raphmatt:
## Summary

The serializer outputs bullet lists in table cells as `* Item 1 <br> * Item 2`, but when this markdown is parsed back (e.g. via the API), the `*` markers become literal text instead of `bullet_list` nodes.

This adds bullet list detection to the table cell parsing rule, using the same approach already in place for checkbox lists.

Additionally, the existing checkbox parsing had a pre-existing issue where inline markdown (links, bold, etc.) in labels was stored as raw text instead of being parsed. For example, `[link](url)` would appear literally instead of rendering as a clickable link. This is fixed by extracting a shared `makeInlineToken` helper that uses `parseInline` to resolve inline formatting, benefiting both bullet items and checkbox labels.

## Changes

- Detect `* item <br> * item` patterns in table cells and emit `bullet_list` > `list_item` tokens
- Extract `makeInlineToken` helper that runs `parseInline` on content, so links and formatting are properly resolved
- Apply `makeInlineToken` to both bullet items and existing checkbox labels

### DIFF
--- a/shared/editor/rules/tables.test.ts
+++ b/shared/editor/rules/tables.test.ts
@@ -1,0 +1,127 @@
+import markdownit from "markdown-it";
+import tablesRule from "./tables";
+
+describe("Tables Rule", () => {
+  const md = markdownit().use(tablesRule);
+
+  describe("bullet lists in table cells", () => {
+    it("should parse bullet list items separated by br tags", () => {
+      const result = md.parse(
+        "| Header |\n|---|\n| * Item 1<br>* Item 2<br>* Item 3 |",
+        {}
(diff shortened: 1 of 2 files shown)


tommoor:
Hi @Raphmatt , thanks for the contribution - please include test coverage, you can see `alphaLists.test.ts` as an example

Raphmatt:
Hey @tommoor, I have added the test coverage in the latest commit, following the same pattern as alphaLists.test.ts.

PR #11831, opened 2026-03-20
feat: text color support

NoWauu:
I've added the support of colored text in addition to text highlight
Here what it looks like :
image

The color menu is the same as for highlighting and is represented by a caps A in the quick menu
image

When a color is set on text, the following text will inherit the same color
I've put a different preset color palette in order to contrast with the highlight colors


Kincott:
actually useful i wish this was added

UmbraCi:
There are too many file changes; there shouldn't be so many. Please only submit the necessary core file changes. I'm also very interested in this PR. @NoWauu

PR #11841, opened 2026-03-22
Add GITHUB_URL environment variable for GitHub Enterprise Server support

cameron-mckain:
The GitHub integration currently hardcodes github.com in three places: the Octokit client (defaults to api.github.com), URL parsing for link unfurling (hostname !== "github.com"), and the OAuth install URL. This means the integration doesn't work with self-hosted GitHub Enterprise Server (or GHE.com).

This adds an optional GITHUB_URL env var that, when set, points everything at the configured instance:
- Octokit gets baseUrl set to {GITHUB_URL}/api/v3
- parseUrl matches against the configured hostname instead of github.com
- authUrl constructs the install link using the configured URL

When GITHUB_URL is unset, behavior is identical to before — all three fall back to github.com.

### DIFF
--- a/plugins/github/shared/GitHubUtils.ts
+++ b/plugins/github/shared/GitHubUtils.ts
@@ -33,7 +33,8 @@ export class GitHubUtils {
 
   static authUrl(state: string): string {
     const ghUrl = (env.GITHUB_URL ?? "https://github.com").replace(/\/$/, "");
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks, but enterprise integrations are not included in the community edition so this won't be merged at this time.

This is another reason contributing guidelines ask for an issue before PR

PR #11860, opened 2026-03-23
feat: Strip comments from presentation mode

### LINKED ISSUE #11859: Crash within present mode
Some content causes a react error boundary, I think it's probably inline comments.

tommoor:
## Summary
- Strips comment marks from document content before rendering slides in presentation mode, so annotations don't appear during presentations.
- Moves `removeMarks` from the server-only `ProsemirrorHelper` to the shared version so it can be used on the client.
- Makes the server `ProsemirrorHelper` extend the shared class, eliminating duplicate code and removing `SharedProsemirrorHelper` imports across 5 server files.

## Test plan
- [x] Open a document with comments and enter presentation mode — comments should not be visible
- [x] Verify existing server-side mark removal (e.g. public share links) still works
- [ ] Run `yarn test server/models/helpers/ProseMirrorHelper.test.ts server/models/helpers/TextHelper.test.ts shared/utils/ProsemirrorHelper.test.ts`

closes #11859

### DIFF
--- a/shared/utils/ProsemirrorHelper.test.ts
+++ b/shared/utils/ProsemirrorHelper.test.ts
@@ -234,4 +234,148 @@ describe("ProsemirrorHelper", () => {
       ]);
     });
   });
+
+  describe("removeMarks", () => {
+    it("should remove specified mark types from text nodes", () => {
+      const doc: ProsemirrorData = {
+        type: "doc",
+        content: [
+          {
+            type: "paragraph",
+            content: [
+              {
+                type: "text",
+                text: "hello",
+                marks: [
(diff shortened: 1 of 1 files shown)


server/models/helpers/ProsemirrorHelper.tsx
```diff
@@ -62,7 +63,7 @@ export type MentionAttrs = {
 };
 
 @trace()
-export class ProsemirrorHelper {
+export class ProsemirrorHelper extends SharedProsemirrorHelper {
```
tommoor, line 66: Time for a nice simplification

PR #11865, opened 2026-03-24
Discord Role (Group) Sync Provider

sebasptsch:
Use the new Group Sync Plugin featureset added in https://github.com/outline/outline/pull/11684 to add Discord Role syncing, extending the existing Discord Integration further.

This also uses a new environment variable `DISCORD_BOT_TOKEN` to get role metadata from the configured server.

### DIFF
--- a/plugins/discord/server/auth/discord.ts
+++ b/plugins/discord/server/auth/discord.ts
@@ -21,10 +21,7 @@ import {
 } from "@server/utils/passport";
 import config from "../../plugin.json";
 import env from "../env";
-import {
-  DiscordGuildError,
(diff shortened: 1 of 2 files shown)


tommoor:
Hi, group sync is a licensed-only feature - we will not be including these capabilities in the community edition

sebasptsch:
Hi @tommoor 

Is there anywhere that I can see the capabilities available to plugins so that I can plan ahead in the future and avoid using features that aren't suitable for the community edition?

tommoor:
Hey

You should file an issue first rather than submitting code, this is covered in the contributing guidelines in the README. This would be the best way in the future

PR #11874, opened 2026-03-26
Mermaid improvements

### LINKED ISSUE #11782: Mermaid diagram renders too big or too small

### Current Behavior

As of v1.6.0, Mermaid diagrams render with a canvas (viewBox) that is either too large or too small in an RDP (Windows Remote Desktop) environment.

For example, given the following diagram:

```
flowchart
  hello
```

this is the result I get:

Image

This issue may be related to Chrome browser (the Blink rendering engine).

However, the same diagram rendered correctly in Outline v1.5.0, so it seems likely that a change introduced in v1.6.0 caused the problem.

### Expected Behavior

Mermaid diagram renders the same way in both RDP and local environment.

Image

### Steps To Reproduce

1. Connect to the machine using RDP (Windows Remote Desktop).
2. Launch Chrome or Edge. If the browser is already running, close it completely and relaunch it. This step is important.
3. Open (Go to) Outline v1.6.0.
4. In a new document, create a Mermaid diagram and check whether it renders at the correct size.
5. This is the result I get. The diagram is rendered inside an excessively large canvas (viewport).

Image

6. Then log in to the same machine locally, rather than through RDP.
7. The diagram still appears too small, and refreshing the page does not fix it.
8. Close the browser and relaunch it.
9. Open Outline again.
10. The diagram now renders correctly.

Image

### Environment

```markdown
- Outline: Outline 1.6.0
- Browser: Chrome Version 146.0.7680.80 (Official Build) (64-bit) or Microsoft Edge 146.0.3856.59
```

### Anything else?

Google AI's suggestion: https://share.google/aimode/UBBS9iGlSngAKV7vu

tommoor:
- Upgraded Mermaid lib for minor fixes
- Refactored pre-render technique to potentially fix #11782
- Improved diagram caching between page reloads

Related #11782

### DIFF
--- a/shared/editor/extensions/Mermaid.ts
+++ b/shared/editor/extensions/Mermaid.ts
@@ -15,6 +15,7 @@ import { findParentNode } from "../queries/findParentNode";
 import type { NodeWithPos } from "../types";
 import type { Editor } from "../../../app/editor";
 import { LightboxImageFactory } from "../lib/Lightbox";
+import { hashString } from "../../utils/string";
 import { sanitizeUrl } from "../../utils/urls";
 
 export const pluginKey = new PluginKey("mermaid");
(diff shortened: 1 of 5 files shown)


shared/editor/extensions/Mermaid.ts
```diff
@@ -104,16 +104,20 @@ class MermaidRenderer {
...
       "offscreen-mermaid-" + Math.random().toString(36).substr(2, 9);
     renderElement.id = tempId;
-    renderElement.style.position = "absolute";
-    renderElement.style.left = "-9999px";
-    renderElement.style.top = "-9999px";
+    renderElement.style.position = "fixed";
+    renderElement.style.visibility = "hidden";
+    renderElement.style.top = "0";
+    renderElement.style.left = "0";
+    renderElement.style.width = "100%";
```
justnx, line 119: 100% binds rendering to viewport. Shouldn’t this use the editor width instead? Thinks this could break scaling/wide screens/virtual desktops.

PR #11895, opened 2026-03-29
Feat: Add mentions syntax for create/update API

mcdamo:
This adds a support for including links to other documents in wiki when using API to create/update documents. Related #11894 
- reads mentions from markdown in format `<mention:DOCID>` that match DOCIDs in the system.


tommoor:
Document and collection markdown support @mentions using the syntax: `@[Display Name](mention://user/userId)`. For example: `@[John Doe](mention://user/c9a1b2e3-...)`

mcdamo:
Unless i'm mistaken this mention syntax is not actually supported when using document create/update via API?

The links are not converted to mentions and remain as they were in the Markdown, `@[John Doe](mention://user/c9a1b2e3-...)`

tommoor:
Added here https://github.com/outline/outline/pull/11851 last week, it will be in the next release.

mcdamo:
Fantastic, Thankyou for the info.

PR #11920, opened 2026-04-01
add self-hosted gitlab snippet

IamTaoChen:
add self-hosted gitlab snippet


tommoor:
We shouldn't introduce another config for this, we can use the url that is registered when connecting GitLab in Integrations

IamTaoChen:
> We shouldn't introduce another config for this, we can use the url that is registered when connecting GitLab in Integrations

Hi, Tommoor. Now, it gets gitlab host from `intgrations.gitlab.setting`.

PR #11928, opened 2026-04-02
Use CSS highlights instead of editor decorations when available

tommoor:
Much improved performance by using native CSS highlights for find and replace highlighting instead of DOM mutations when the browser supports it.

Retains the decoration fallback for now as this is a relatively new API.


app/editor/extensions/FindAndReplace.tsx
```diff
@@ -407,13 +404,83 @@ export default class FindAndReplaceExtension extends Extension {
...
+    });
+  }
+
+  /**
+   * Create a DecorationSet from the current search results.
+   */
+  private createDecorationSet(doc: Node) {
     this.search(doc);
-    return this.decorations
+    return this.decorations.length
```
tommoor, line 428: Fix 😬

PR #11941, opened 2026-04-03
Add support for client_secret_basic authentication method in OIDC

yushaye:
# Add support for `client_secret_basic` authentication method in OIDC

## Problem

Some OIDC provider may require client_secret_basic authentication method. Currently, Outline only supports client_secret_post, which sends client credentials in the request body, making it impossible to authenticate without modifying the code.

## Solution

This PR adds support for the client_secret_basic authentication method:

### 1. OIDC Discovery Integration
- Added `token_endpoint_auth_methods_supported` field to OIDC configuration
- Automatically detects supported authentication methods from the provider's discovery endpoint

### 2. Configuration Options
- New environment variable: `OIDC_TOKEN_ENDPOINT_AUTH_METHOD`
- Accepts values: `client_secret_basic` or `client_secret_post`(default)

### 3. HTTP Basic Auth Implementation
- Modified `OIDCStrategy` to support HTTP Basic Auth header for token requests by overriding `_oauth2.getOAuthAccessToken` method
- Updated `OAuthClient.rotateToken` to use HTTP Basic Auth when refreshing tokens

## Configuration

The cat authentication method can be configured in three ways:

### Option 1: Environment Variable
```bash
OIDC_TOKEN_ENDPOINT_AUTH_METHOD=client_secret_basic
```

### Option 2: Automatic Discovery
```bash
OIDC_ISSUER_URL=https://your-provider.com
# Will automatically detect supported methods from /.well-known/openid-configuration
```

### Option 3: Manual Configuration
```bash
OIDC_AUTH_URI=https://provider.com/oauth/authorize
OIDC_TOKEN_URI=https://provider.com/oauth/token
OIDC_USERINFO_URI=https://provider.com/oauth/userinfo
OIDC_TOKEN_ENDPOINT_AUTH_METHOD=client_secret_basic
```

## Testing

Tested with:
- yarn test

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -179,6 +179,11 @@ OIDC_AUTH_URI=
 OIDC_TOKEN_URI=
 OIDC_USERINFO_URI=
 OIDC_LOGOUT_URI=
+# The authentication method to use when making requests to the token endpoint.
+# Supported values: "client_secret_basic" or "client_secret_post".
+# If not set, will be automatically determined from OIDC discovery response,
+# or default to "client_secret_post".
+OIDC_TOKEN_ENDPOINT_AUTH_METHOD=
 
 # Specify which claims to derive user information from
 # Supports any valid JSON path with the JWT payload
--- a/.env.test
+++ b/.env.test
@@ -26,6 +26,7 @@ OIDC_CLIENT_SECRET=client-secret
 OIDC_AUTH_URI=http://localhost/authorize
 OIDC_TOKEN_URI=http://localhost/token
 OIDC_USERINFO_URI=http://localhost/userinfo
+OIDC_TOKEN_ENDPOINT_AUTH_METHOD=
 
 IFRAMELY_API_KEY=123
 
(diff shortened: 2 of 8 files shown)


tommoor:
> Some OIDC provider may require client_secret_basic authentication method

Which? In over 6 years this is the first time this has come up.

Phylaris:
It's a self-hosted OIDC service we use internally. I've also reached out to them about adding client_secret_post support, but figured I'd submit this PR in the meantime.
I'm not surprised this hasn't come up before - most major OIDC providers support both methods. But since client_secret_basic is part of the OAuth 2.0 spec (RFC 6749), I thought having support for it could help others who might run into similar situations with their own internal services.

yushaye:
Thanks for the feedback.

I understand this is an uncommon use case that should ideally be addressed by the OIDC provider. I'll close this PR for now.

If there are other users who need this feature in the future, we can revisit it. Thanks for maintaining Outline!

PR #12039, opened 2026-04-14
Translate 'Print template' to '打印模板'

vean05:


tommoor:
Translations are provided through https://crowdin.com/project/outline – these files are generated

PR #12072, opened 2026-04-16
fix: Strip port from X-Forwarded-For IP when behind reverse proxy

eisber:
Some reverse proxies (e.g. Azure Application Gateway) append the client port to the X-Forwarded-For header, producing values like '1.2.3.4:56789'.

Koa exposes this as ctx.ip, which then fails Sequelize's isIP validation when persisting lastActiveIp and lastSignedInIp on the User model.

This adds a lightweight middleware that strips the trailing ':port' from IPv4 addresses before downstream handlers see them. IPv6 addresses (containing '::') are left untouched.

Fixes login failures (500) when self-hosting behind Azure App Gateway or similar proxies that include the port in forwarded headers.

### DIFF
--- a/server/services/web.ts
+++ b/server/services/web.ts
@@ -52,6 +52,34 @@ export default function init(app: Koa = new Koa(), server?: Server) {
     }
   }
 
+  // Some reverse proxies (e.g. Azure App Gateway) append the client port to
+  // X-Forwarded-For producing values like "1.2.3.4:56789". Koa exposes this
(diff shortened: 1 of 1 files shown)


tommoor:
This change won't land as the issue sits within Azure, not Outline - and becomes a maintenance and performance burden on every request.

I recommend updating your proxy configuration to omit the port. (see [add_x_forwarded_for_proxy](https://learn.microsoft.com/en-us/azure/application-gateway/rewrite-http-headers-url))

PR #12073, opened 2026-04-16
fix: Use configured URL domain for cookies when behind reverse proxy

eisber:
When Outline runs behind a reverse proxy (e.g. Azure App Gateway, nginx) that forwards requests using a backend hostname different from the public URL, the request hostname seen by Koa differs from the domain the browser navigates to.

This caused cookies to be set on the backend hostname instead of the public domain, breaking authentication because the browser never sent the cookies back on subsequent requests to the public URL.

This fix falls back to the configured URL's domain (from the URL env variable) when the request hostname doesn't match, ensuring cookies are always set for the domain the browser sees.

### DIFF
--- a/shared/utils/domains.ts
+++ b/shared/utils/domains.ts
@@ -93,6 +93,14 @@ export function getCookieDomain(domain: string, isCloudHosted: boolean) {
     }
   }
 
+  // When behind a reverse proxy, the request hostname may differ from the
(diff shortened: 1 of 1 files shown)


tommoor:
You should update your proxy configuration to pass the correct public host to Outline, the change made in this PR breaks behavior for existing domains as noted by Copilot.

PR #12093, opened 2026-04-17
fix: Handle missing team in OAuth authorize page during initial load

eisber:
The Authorize component called useCurrentTeam() which throws an invariant error if the team hasn't loaded yet. During the MCP OAuth flow on first login, the MobX stores may not have hydrated, causing the page to crash with 'Invariant Violation: team required'.

Change useCurrentTeam() to useCurrentTeam({ rejectOnEmpty: false }) and show a loading indicator while the team is still loading, matching the existing pattern for waiting on the OAuth client response.

### DIFF
--- a/app/scenes/Login/OAuthAuthorize.tsx
+++ b/app/scenes/Login/OAuthAuthorize.tsx
@@ -76,7 +76,7 @@ function inputScopes(scope?: string): string[] {
  * and allows the user to either authorize or cancel the request.
  */
 function Authorize() {
-  const team = useCurrentTeam();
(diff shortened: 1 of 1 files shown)


app/scenes/Login/OAuthAuthorize.tsx
```diff
@@ -76,7 +76,7 @@ function inputScopes(scope?: string): string[] {
  * and allows the user to either authorize or cancel the request.
  */
 function Authorize() {
-  const team = useCurrentTeam();
```
tommoor, line 79: This is already guarded on L41

PR #12110, opened 2026-04-19
fix: prevent crash when printing with right sidebar open

huiseo:
## Summary
- Added `isPrinting` check using `useMediaQuery("print")` to conditionally unmount the aside sidebar's children during print, preventing a crash when the comments or history panel is open while printing
- When `window.print()` is triggered with the aside sidebar open, the browser's `matchMedia("print")` change event fires and React's microtask scheduler invalidates pending callbacks in child components (`ResizeObserver` in `Scrollable`, `useMeasure` in `ResizingHeightContainer`), resulting in an `ErrorBoundary` error: "The provided callback is no longer runnable"
- Also refactored `useMediaQuery` to defer state updates via `setTimeout(0)` to avoid microtask conflicts during print transitions — this fix is consistent with the existing `useMediaQuery("print")` pattern used in `PageScroll.tsx` and `useBuildTheme.ts`, and the `@media print { display: none }` CSS rule matches the left sidebar's print-hiding approach in `Sidebar.tsx`

## Test plan
- [x] Verified printing with comments sidebar open no longer crashes
- [x] Verified printing with history sidebar open no longer crashes
- [x] Verified normal printing (no sidebar open) still works
- [x] Confirmed responsive behavior unchanged (`useMobile` hook still works)
- [x] Confirmed no regressions in `useBuildTheme` print detection

### DIFF
--- a/app/components/Sidebar/Aside.tsx
+++ b/app/components/Sidebar/Aside.tsx
@@ -7,6 +7,7 @@ import { depths, s } from "@shared/styles";
 import ErrorBoundary from "~/components/ErrorBoundary";
 import Flex from "~/components/Flex";
 import ResizeBorder from "~/components/Sidebar/components/ResizeBorder";
+import useMediaQuery from "~/hooks/useMediaQuery";
 import useStores from "~/hooks/useStores";
 import useWindowScrollbarWidth from "~/hooks/useWindowScrollbarWidth";
 import { sidebarAppearDuration } from "~/styles/animations";
@@ -22,6 +23,7 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
 function Aside({ children, border, className, skipInitialAnimation }: Props) {
   const theme = useTheme();
   const { ui } = useStores();
+  const isPrinting = useMediaQuery("print");
   const [isResizing, setResizing] = React.useState(false);
   const maxWidth = theme.sidebarMaxWidth;
   const minWidth = theme.sidebarMinWidth + 16; // padding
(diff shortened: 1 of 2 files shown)


tommoor:
I can't reproduce the crash you're speaking of, I don't think unrendering the element like this would be the correct fix

huiseo:
Thanks for taking a look! A few notes that might help:

**Reproduction conditions** (consistent in our environment)
- Document with at least one comment thread loaded
- Right sidebar open (comments or history)
- Production build where the `es2015.*.js` polyfill bundle is loaded for the request UA
- The error caught by `ErrorBoundary` is:
  > Error: The provided callback is no longer runnable.
  > at h (es2015.*.js:1:2506)
  > at unstable_runWithPriority (vendor-react.*.js)

It seems to surface only when the core-js `queueMicrotask` polyfill is loaded — that's likely why it doesn't reproduce on your end with native `queueMicrotask`.

**On the approach** — you're right that unmounting children is heavy. I went down that path after CSS-only `@media print { display: none }` (matching `Sidebar.tsx`) wasn't enough on our end, but I'd much prefer your guidance here. Would you suggest:

1. Just adding the CSS rule (mirroring left sidebar) and treating any residual polyfill-edge-case errors as out of scope
2. A more targeted fix at the source (e.g., guarding `Scrollable`'s ResizeObserver or `ResizingHeightContainer`'s useMeasure against zero-dimension elements)
3. Closing this and reporting separately so you can triage

Happy to revise or close — whichever you'd prefer.

PR #12152, opened 2026-04-24
feat: add Meilisearch search engine plugin

reasonable0:


lublak:
can it get reopend? i thin Meilsearch is really useful for some people @tommoor

PR #12326, opened 2026-05-12
fix: Printing includes extra blank page

tommoor:
closes #12324


andymarden:
Nice one. That's was quick!

PR #12328, opened 2026-05-12
fix: Batch document deletes when emptying trash

tommoor:
## Summary

Empty-trash issued a single unbatched DELETE against the `documents` table, holding an exclusive lock long enough to queue concurrent web SELECTs behind it. `documentPermanentDeleter` now chunks both the trailing `parentDocumentId` clear and the `Document.destroy` into batches of 100, keeping each lock window short.

## Test plan

- [x] Existing `server/commands/documentPermanentDeleter.test.ts` suite passes
- [ ] Verify on staging by emptying a large trash and watching documents-table lock waits

### DIFF
--- a/server/commands/documentPermanentDeleter.ts
+++ b/server/commands/documentPermanentDeleter.ts
@@ -1,4 +1,4 @@
-import { uniq } from "es-toolkit/compat";
+import { chunk, uniq } from "es-toolkit/compat";
 import { Op, QueryTypes } from "sequelize";
 import Logger from "@server/logging/Logger";
 import { Document, Attachment } from "@server/models";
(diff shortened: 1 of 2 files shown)


server/commands/documentPermanentDeleter.ts
```diff
@@ -76,26 +76,52 @@ export default async function documentPermanentDeleter(documents: Document[]) {
...
-        },
-      },
-      paranoid: false,
-    }
-  );
 
-  return Document.scope("withDrafts").destroy({
+  // Re-check deletedAt in the database to exclude documents that were restored
+  // between the caller's query and now. Otherwise the parentDocumentId clear
+  // below would detach children of a restored parent, breaking the hierarchy.
```
tommoor, line 88: Unrelated fix while we're here

PR #12344, opened 2026-05-13
fix: pass expires into /auth/redirect's accessToken JWT

awais786:
## Summary

`/auth/redirect` mints the new `accessToken` JWT via `user.getJwtToken(undefined, service)`, omitting the `expiresAt` payload claim. The validator at `server/utils/jwt.ts` skips its expiry check when that claim is missing, so the token is valid until the next `jwtSecret` rotation regardless of the cookie's `Max-Age`. If the token ever leaves the cookie (log capture, screenshot, browser history copy, etc.), it can be replayed indefinitely.

The two other call sites that mint this cookie (`server/middlewares/authentication.ts`, `server/utils/authentication.ts`) already pass `expires` into both `getJwtToken` and the cookie option. This PR brings the third path in line.

## Changes

- Compute `expires` once and pass it into both `getJwtToken(expires, service)` and the cookie's `expires` option.
- One-line comment at the call site noting why the binding matters.

## Testing

New test in `server/routes/auth/index.test.ts`: hits `/auth/redirect` with a transfer token, extracts the resulting `accessToken` cookie from `Set-Cookie`, decodes the JWT payload (`base64url`), and asserts `expiresAt` is set and future-dated. Guards against future regression to `getJwtToken(undefined, ...)`.

Existing tests (`should redirect to home`, `should redirect to first collection`, `should prevent token extension by rejecting JWT tokens`) all continue to pass.

## Before / after

| Scenario | Before | After |
|---|---|---|
| Cookie + JWT both held in browser | Token works for cookie's `Max-Age` | Same |
| Raw JWT exfiltrated (logs, history, screenshot) | Replayable indefinitely until `jwtSecret` rotation | Bounded by `expiresAt` (= cookie's `Max-Age`) |

### DIFF
--- a/server/routes/auth/index.test.ts
+++ b/server/routes/auth/index.test.ts
@@ -43,4 +43,32 @@ describe("auth/redirect", () => {
 
     expect(res.status).toEqual(401);
   });
+
+  it("should mint an accessToken JWT carrying an expiresAt claim", async () => {
+    const user = await buildUser();
+    const before = Date.now();
+
+    const res = await server.get(
+      `/auth/redirect?token=${user.getTransferToken()}`,
+      {
+        redirect: "manual",
+      }
+    );
+
+    expect(res.status).toEqual(302);
+
+    // Pull the `accessToken` cookie out of the Set-Cookie header(s).
+    const setCookie = res.headers.get("set-cookie") || "";
+    const match = setCookie.match(/accessToken=([^;,]+)/);
+    expect(match).not.toBeNull();
+    const jwt = match![1];
+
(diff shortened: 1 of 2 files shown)


tommoor:
This makes me want to make a few other improvements here, but we can merge this ahead of that if you sign CLA.

Cheers!

tommoor:
@awais786 I would like to merge this, are you able to sign the CLA? Otherwise I will re-implement soon

tommoor:
superseded by https://github.com/outline/outline/pull/12371

PR #12352, opened 2026-05-14
Tune collaborative undo grouping

### LINKED ISSUE #12351: Inconsistencies in Undo and Redo functionality

### Current Behavior

When using the Undo (`Ctrl + Z`) and Redo (`Ctrl + Y`) functions while composing text, the results are inconsistent with the intended actions. Specifically, performing a Redo immediately after an Undo does not restore the document to its original state.

**1. Unpredictable Undo Scope**
In most browsers, the Undo function typically rolls back the most recent action or reverses text input word-by-word (delimited by spaces). However, in Outline Wiki, an Undo often causes an entire sentence to disappear or reverts both the text and its styling simultaneously in an unpredictable manner.

**2. Unpredictable Redo Scope**
When performing a Redo, it is expected to restore exactly what was just undone. Instead, the Redo process often requires two or three separate steps to return to the previous state, rather than a single direct action.

The attached video is a screen recording of the Undo and Redo functions being used while writing text.

https://github.com/user-attachments/assets/0e94025c-6310-474b-b9a3-1232e248aaab

These inconsistencies lead to a confusing and frustrating experience for users when creating or editing documents.

### Expected Behavior

- Undo should only roll back the most recently entered word or formatting change.

- Redo should restore the document to its exact state immediately before the Undo was performed.

### Steps To Reproduce

Input `ctrl + z` and `ctrl + y` when write a text in document.

### Environment

```markdown
- Outline: 1.17.1
- Browser: MS Edge
```

### Anything else?

_No response_

mzl2233:
Adjusts the Yjs undo manager capture timeout used by the collaborative editor so separate typing and formatting actions are grouped less aggressively. This keeps undo and redo steps closer to the most recent editing action and makes redo restore the immediately undone change more predictably.

Verification: ran `git diff --check`; local TypeScript verification was not run because dependencies were not installed in the checkout.

Fixes #12351

### DIFF
--- a/app/editor/extensions/Multiplayer.ts
+++ b/app/editor/extensions/Multiplayer.ts
@@ -121,7 +121,9 @@ export default class Multiplayer extends Extension<MultiplayerOptions> {
         awarenessStateFilter,
         selectionBuilder,
       }),
-      yUndoPlugin(),
+      yUndoPlugin({
+        captureTimeout: 500,
+      }),
       new Plugin({
         props: {
           handleScrollToSelection: (view) => isRemoteTransaction(view.state.tr),


tommoor:
This option does not exist on the plugin I'm afraid

PR #12366, opened 2026-05-17
fix: initialize text selection state

### LINKED ISSUE #12365: Word counter doesn't update on selection

### Current Behavior

Currently, the word counter is hidden inside the menu, and it doesn't update on selected text. The stats are only for the whole document. I am not certain if this is a bug or just missing feature.

### Expected Behavior

Users expect a easier way to get word count info especially when actively typing and on selected snippets, similar to below:

Image

Image

### Steps To Reproduce

_No response_

### Environment

```markdown
- Outline:1.7.1
- Browser:Safari
```

### Anything else?

_No response_

mturac:
## Summary

Fixes #12365.

The document insights word counter already supports selected text stats, but the selection hook started from an empty state and only updated after a later `selectionchange` event. When opening the insights menu after selecting text, focus can move into the menu before the hook captures the existing selection, so the selected text stats remain empty.

This initializes the hook state from the current browser selection and reuses the same helper for future selection changes.

## Test plan

- `yarn test app/hooks/useTextSelection.test.ts`
- `yarn lint:changed`
- `yarn prettier --check app/hooks/useTextSelection.ts app/hooks/useTextSelection.test.ts`
- `yarn tsc --noEmit --pretty false`

Note: `yarn format:check app/hooks/useTextSelection.ts app/hooks/useTextSelection.test.ts` currently reports an unrelated pre-existing formatting warning in `shared/editor/rules/mark.ts`; the changed files pass direct Prettier check.


app/hooks/useTextSelection.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, expect, it, vi } from "vitest";
```
tommoor: This file doesn't test anything at all except the mocks, should remove

tommoor (approved):

PR #12518, opened 2026-05-29
fix: Pipes in math and code blocks within table cells

### LINKED ISSUE #12510: API MARKDOWN ROUND-TRIP ISSUE FOR TABLES, INLINE CODE, AND LATEX

### Current Behavior

Short report:
> When a document contains special Markdown formatting such as inline code, LaTeX/math, and pipe-containing expressions inside table cells, `documents.info` and `documents.update` may not be fully round-trip safe.
Re-importing exported Markdown can change formatting, split table rows, leak content outside the table, or truncate the original content. The API export does not always appear to be a reliable re-import payload for the same document.

## OBSERVED BEHAVIOR
The problematic cases seem to appear when Markdown combines:

- table cells
- inline code formatting such as `` `...` ``
- LaTeX/math blocks such as `$$...$$`
- special characters such as `|`, `_`, `{`, `}`, and backticks

Examples of failure modes observed through the API:

1. A table cell containing expressions such as `` `|Clock_NTP_Camera1 - Clock_GPS_Camera1|` `` is not preserved exactly after API import/export.
2. Math content containing pipe characters may be exported in one Markdown shape and then re-imported into a different internal structure.
3. Re-importing Markdown through `documents.update` can split one logical table row into multiple fragments.
4. Content that should stay inside a table cell can appear outside the table after re-import.
5. Exact Markdown formatting is not preserved, even when the logical content is intended to remain unchanged.

This seems especially relevant for technical documentation using:

- formulas
- code-like expressions
- identifiers with underscores/braces
- pipe-delimited expressions
- tables combining prose and notation

### Expected Behavior

Ideally, the following would hold:

1. `documents.info` would return Markdown that can be sent back through `documents.update` without changing the document structure unexpectedly.
2. If Outline intentionally normalizes Markdown, that normalization would still preserve the exact document structure and content boundaries.
3. Special formatting inside table cells would remain stable and would not truncate content, split rows, or leak nodes outside the table.

### Steps To Reproduce

Use a document containing a Markdown table with a cell that mixes prose and special formatting.

Example input:

```md
| Condition | Facts |
| --- | --- |
| Absolute time difference | The system checks `|Clock_NTP_Camera1 - Clock_GPS_Camera1|` and `|Clock_NTP_Camera2 - Clock_GPS_Camera2|`. |
```

Then:
1. create or update the document through `documents.update`
2. fetch it with `documents.info`
3. send the fetched Markdown back again through `documents.update`
4. fetch it once more

Observed result:

- the exported Markdown may no longer match the original input
- replaying the exported Markdown may further alter the document
- the table cell may be truncated or structurally split

## IMPACT
This makes reliable external tooling much harder for clients that sync Markdown to Outline through the public API.

### Environment

```markdown
- Outline:
- Browser:
```

### Anything else?

_No response_

tommoor:
Alternative approach to https://github.com/outline/outline/pull/12516

closes https://github.com/outline/outline/issues/12510

### DIFF
--- a/server/models/helpers/ProseMirrorHelper.test.ts
+++ b/server/models/helpers/ProseMirrorHelper.test.ts
@@ -6,7 +6,7 @@ import * as Y from "yjs";
 import type { ProsemirrorData } from "@shared/types";
 import { MentionType } from "@shared/types";
 import { createContext } from "@server/context";
-import { schema } from "@server/editor";
+import { parser, schema, serializer } from "@server/editor";
 import { buildProseMirrorDoc, buildUser } from "@server/test/factories";
 import type { MentionAttrs } from "./ProsemirrorHelper";
 import { ProsemirrorHelper } from "./ProsemirrorHelper";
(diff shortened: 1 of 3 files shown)


beretienne:
Hello Tom, thanks for the fix so far.
My findings from local testing:

- Markdown table round-trip is still unsafe for inline LaTeX-style expressions in mixed-content cells, especially forms using delimiter commands such as `\left|...\right|`;
they may render as math on import but come back with mutated markdown.

   Example: `$\left|Clock_{NTP_{Camera1}} - Clock_{GPS_{Camera1}}\right|$<br>and compares it with \`camera1.ntp_offset\`.`

   comes back as

   `$\\left\|Clock_{NTP_{Camera1}} - Clock_{GPS_{Camera1}}\\right\|$<br>and compares it with\`camera1.ntp_offset\`.`

- Also, **display math** `$$...$$` inside markdown table cells is not supported; only the inline `$...$` case currently round-trips as a math node.

tommoor:
Honestly I'm not sure this will merge – the fix is too complex for the problem. Markdown is not supposed to be fully roundtrippable, it's a fundamentally lossy format unfortunately.

beretienne:
Thanks for your quick response.
Understand these are also "edge" cases. 
I'm working in the technical doc, up until now with the Myst-Parser (namely used in Jupyter Notebooks), which pushes even deeper the MD directives & roles.
I believe that as long as an outline "export" still looks right in a GitHub flavoured renderer this is ok.
Maybe documenting not supported edge cases or at least making a note where coverage is still not 100% ?
Also, I was testing with the full corrections brought to you code, after code scan.
Thanks a lot for your time and dedication.
BTW, I could help with doc. in French if necessary.

PR #12548, opened 2026-06-01
feat(s3): add optional CloudFront support for attachments with signed URLs

oliinykdm:
Added support for serving all attachments (document files, avatars, previews) through CloudFront using signed URLs instead of direct S3 presigned URLs.

- Keeps the S3 bucket fully private with Origin Access Control (OAC)
- Saves a ton of S3 egress costs
- Still uses /api/attachments.redirect for proper auth checks

What changed

- New env vars: `AWS_CLOUDFRONT_URL`, `AWS_CLOUDFRONT_KEY_PAIR_ID`, `AWS_CLOUDFRONT_PRIVATE_KEY` (and _BASE64 alternative)
- Uses @aws-sdk/cloudfront-signer with default 1-hour TTL
- Full fallback to original S3 presigned URLs if CloudFront vars are empty (nothing breaks)
- If `AWS_CLOUDFRONT_KEY_PAIR_ID` and `AWS_CLOUDFRONT_PRIVATE_KEY` are not provided - unsigned CloudFront URLs are used (static URLs without expiration - simple redirect through CF)

Everything was thoroughly tested on a real self-hosted instance: uploads, previews, downloads, avatar updates, public shares - all work perfectly with signed URLs and private bucket.

**Why AWS_S3_ACCELERATE_URL does NOT solve this**

`AWS_S3_ACCELERATE_URL` changes the endpoint for uploads (PUT) as well, sending them directly to the accelerate URL instead of through S3. Since CloudFront doesn't proxy PUT requests (it's meant for GET only), uploads fail with 403/404 errors (tested on a real bucket). Downloads via CloudFront work fine.

If these kind of improvements are welcomed - I can try to add a support of R2, B2 as well.


oliinykdm:
It works flawlessly for months, anybody would benefit from the implementation since 1TB free egress by Cloudfront is a huge, or even with flat-rate Cloudfront packages it's no brainer (50TB egress for $15 per month). 
What should I test or implement not help move forward with this? This is my main wiki for myself and I'm rely on it daily. But I'm afraid to share documents or files due to S3 egress costs, so CloudFront just solves the issue and basically could turn Outline to real file-server with no-limits (any without DoW attack possibility in case of flat-rate usage).

@tommoor 

Thanks!

PR #12558, opened 2026-06-02
Disable branding visibility in Document scene.

oliinykdm:


tommoor:
Hi @oliinykdm – removing branding is part of the licensed edition btw

oliinykdm:
> Hi @oliinykdm – removing branding is part of the licensed edition btw

Sure, it was a mistake.

PR #12597, opened 2026-06-05
only users with "manage" permissions can edit page permissions

orikad:
If I don't miss anything, there is not difference between users with 'edit' permissions to a collection or page or users with 'manage' permissions. 
In my mind, only users with manage permissions should be able to edit permissions for the collection/page. Right now even an editor can do so, even remove managers from the collection/page. 
This PR fixes that such that the permissions submenu does not appear for editor and only for managers.
Tests for that included, but the actual change is just in `document.ts`.
I use this change internally succesfully and have not found any issues with it.
Would appreciate feedback/merging this in.

Thank you


orikad:
Hey @tommoor I've fixed the previously failing test. Would appreciate your review/comments. 
Thank you

server/policies/document.ts
```diff
@@ -107,11 +107,13 @@ allow(User, "publish", Document, (actor, document) =>
 
 allow(User, "manageUsers", Document, (actor, document) =>
   and(
-    can(actor, "update", document),
+    !!document?.isActive,
```
tommoor: I think we should remove the active check here, as it could be useful to remove users from an archived document

tommoor (changes requested):
Looks good, overall I agree with one change

tommoor (approved):

PR #12611, opened 2026-06-06
feat: Inline editor menu

### LINKED ISSUE #10546: Skip intermediate menu on table row
Right now we need to show a "..." menu to trigger the context menu from, this is weird and annoying – clicking the table row grip should be able to open a context menu directly.

Image

This will also unlock a better UX for #10545 as we can move the table menu to the same context menu style, allowing more options to be displayed clearly than a toolbar would allow.

tommoor:
Thanks to https://github.com/outline/outline/pull/11084 which provided the first pass at this, pre-menu refactor, this did end up 1/3 smaller and once the interface for menu items is aligned we can remove even more code.

image

related #11084
closes #10546


app/editor/menus/tableRow.tsx
```diff
@@ -97,7 +133,7 @@ export default function tableRowMenuItems(
           {
             name: "toggleRowBackgroundAndCollapseSelection",
             label: t("None"),
-            icon: <DottedCircleIcon retainColor color="transparent" />,
```
tommoor, line 100: Separate bug, this placeholder icon never have retained color

PR #12656, opened 2026-06-10
feat(s3): add optional CloudFront support for signed attachment URLs

oliinykdm:
## Summary

- Adds optional CloudFront CDN support for **attachment downloads** (files, avatars, previews) while keeping **uploads on the direct S3 endpoint**
- Uses `@aws-sdk/cloudfront-signer` to generate short-lived signed CloudFront URLs with the same TTL as existing S3 presigned URLs
- Preserves the existing auth model: access checks still happen via `/api/attachments.redirect` before redirecting to storage
- Fully backward compatible: if CloudFront env vars are unset, behavior is unchanged (S3 presigned URLs only)

## Motivation

Self-hosted Outline instances with many attachments or public shares can incur significant **S3 egress costs**. CloudFront reduces egress cost (including AWS free tier and flat-rate plans) and improves download performance globally.

`AWS_S3_ACCELERATE_URL` does **not** solve this use case because it replaces the storage endpoint for **both** downloads and uploads. CloudFront does not proxy S3 presigned POST uploads the same way S3 does — pointing accelerate URL at a CloudFront domain breaks uploads (403/signature mismatch). This PR introduces a **separate** `AWS_CLOUDFRONT_URL` used only in `getSignedUrl` / `getUrlForKey`, leaving upload paths untouched.

## Related

- Supersedes #12548 and earlier attempts (#11853, #11814)
- Discussion: https://github.com/outline/outline/discussions/12655

## What changed

| File | Change |
|------|--------|
| `server/storage/files/S3Storage.ts` | CloudFront signed URLs for downloads; S3 fallback on missing/invalid signing config |
| `server/env.ts` | New env vars for CloudFront URL and signing credentials |
| `.env.sample` | Sample CloudFront configuration |
| `package.json` / `yarn.lock` | Add `@aws-sdk/cloudfront-signer` |
| `server/storage/files/S3Storage.cloudfront.test.ts` | Unit tests for signing and fallback behavior |

## Configuration

New environment variables (all optional):

```env
# CloudFront distribution domain (no trailing slash)
AWS_CLOUDFRONT_URL=https://d1234567890abcdef.cloudfront.net

# Required together for signed download URLs
AWS_CLOUDFRONT_KEY_PAIR_ID=K2XXXXXXXXXX
AWS_CLOUDFRONT_PRIVATE_KEY=          # PEM, multi-line OK in docker-compose
AWS_CLOUDFRONT_PRIVATE_KEY_BASE64=   # alternative to PEM; takes precedence
```

**Expected AWS setup (self-hosted):**

1. S3 bucket remains **private**
2. CloudFront distribution with **Origin Access Control (OAC)** to the bucket
3. CloudFront **trusted key group** + RSA key pair for signed URLs
4. Outline only needs the distribution URL and signing key — no bucket policy changes in app code

**Behavior matrix:**

| `AWS_CLOUDFRONT_URL` | Key pair + private key | Download URL |
|----------------------|------------------------|--------------|
| unset | — | S3 presigned (unchanged) |
| set | set | CloudFront signed URL |
| set | missing | S3 presigned + warning log |
| set | set but signing fails | S3 presigned + error log |

Uploads (`getPresignedPost`, `store`) always use the S3 endpoint — never CloudFront.

## Security notes

- CloudFront private keys are **not** exposed to the client (`@Public` decorator is not used on secrets)
- Unsigned CloudFront URLs are **not** returned as a fallback — if signing is unavailable, the code falls back to S3 presigned URLs instead of permanent public CDN links
- Auth checks before redirect are unchanged (`/api/attachments.redirect`)

## Differences from previous PR #12548

1. **Single clean commit** on branch `feat/cloudfront` (not `main` of fork with 37+ merge commits)
2. **No `@Public` on private keys** — previous version would leak signing secrets to the web client
3. **Safe fallback** — missing or broken signing config uses S3 presigned URLs, not unsigned CloudFront URLs
4. **Unit tests** added for CloudFront signing paths
5. Rebased on current `outline/main`

## Test plan

Manual (self-hosted with real AWS):

- [ ] Upload attachment via editor → file lands in S3 (not via CloudFront)
- [ ] Download attachment via `/api/attachments.redirect` → redirects to CloudFront signed URL
- [ ] Image preview in document renders correctly
- [ ] Avatar upload/update works
- [ ] Public document share with attachments works
- [ ] With `AWS_CLOUDFRONT_URL` unset → identical behavior to before (S3 presigned URLs)
- [ ] With `AWS_CLOUDFRONT_URL` set but keys missing → downloads still work via S3 fallback; warning in logs

Automated:

```bash
yarn test server/storage/files/S3Storage.cloudfront.test.ts
yarn test server/storage/files/S3Storage.test.ts
yarn lint
yarn tsc
```

## Notes for reviewers

- This is an **opt-in** infra feature for self-hosted deployments; cloud Outline is unaffected unless these env vars are set
- Happy to add hosting docs in a follow-up PR once approach is approved
- CI from fork PRs may need workflow approval from a maintainer

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -122,7 +122,7 @@ AWS_S3_ACL=private
 
 # Optional CloudFront CDN for attachment downloads (uploads stay on S3).
 # Requires a key pair for signed URLs; otherwise downloads fall back to S3.
-AWS_CLOUDFRONT_URL=https://d1234567890abcdef.cloudfront.net
+AWS_CLOUDFRONT_URL=
 AWS_CLOUDFRONT_KEY_PAIR_ID=K2XXXXXXXXXX
 AWS_CLOUDFRONT_PRIVATE_KEY=
 AWS_CLOUDFRONT_PRIVATE_KEY_BASE64=
--- a/.env.sample
+++ b/.env.sample
@@ -123,9 +123,8 @@ AWS_S3_ACL=private
 # Optional CloudFront CDN for attachment downloads (uploads stay on S3).
 # Requires a key pair for signed URLs; otherwise downloads fall back to S3.
 AWS_CLOUDFRONT_URL=
-AWS_CLOUDFRONT_KEY_PAIR_ID=K2XXXXXXXXXX
+AWS_CLOUDFRONT_KEY_PAIR_ID=
 AWS_CLOUDFRONT_PRIVATE_KEY=
-AWS_CLOUDFRONT_PRIVATE_KEY_BASE64=
 
 # ––––––––––––––––––––––––––––––––––––––
 # ––––––––––––––––  SSL  –––––––––––––––
--- a/.env.sample
+++ b/.env.sample
@@ -120,6 +120,12 @@ AWS_S3_UPLOAD_BUCKET_NAME=bucket_name_here
 AWS_S3_FORCE_PATH_STYLE=true
 AWS_S3_ACL=private
 
+# Optional CloudFront CDN for attachment downloads (uploads stay on S3).
+# Requires a key pair for signed URLs; otherwise downloads fall back to S3.
+AWS_CLOUDFRONT_URL=https://d1234567890abcdef.cloudfront.net
+AWS_CLOUDFRONT_KEY_PAIR_ID=K2XXXXXXXXXX
+AWS_CLOUDFRONT_PRIVATE_KEY=
+AWS_CLOUDFRONT_PRIVATE_KEY_BASE64=
 
 # ––––––––––––––––––––––––––––––––––––––
 # ––––––––––––––––  SSL  –––––––––––––––
--- a/server/env.ts
+++ b/server/env.ts
@@ -651,24 +651,19 @@ export class Environment {
    * private key when AWS_CLOUDFRONT_URL is set.
    */
   @IsOptional()
-  public AWS_CLOUDFRONT_KEY_PAIR_ID =
-    environment.AWS_CLOUDFRONT_KEY_PAIR_ID ?? "";
+  public AWS_CLOUDFRONT_KEY_PAIR_ID = this.toOptionalString(
+    environment.AWS_CLOUDFRONT_KEY_PAIR_ID
+  );
 
   /**
-   * PEM-encoded RSA private key for CloudFront signed URLs. Use a YAML block
-   * scalar in docker-compose for multi-line values.
+   * PEM-encoded RSA private key for CloudFront signed URLs, or a base64-encoded
+   * PEM string on a single line. Use a YAML block scalar in docker-compose for
+   * multi-line PEM values.
    */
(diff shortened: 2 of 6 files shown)


.env.sample
```diff
@@ -120,6 +120,12 @@ AWS_S3_UPLOAD_BUCKET_NAME=bucket_name_here
 AWS_S3_FORCE_PATH_STYLE=true
 AWS_S3_ACL=private
 
+# Optional CloudFront CDN for attachment downloads (uploads stay on S3).
+# Requires a key pair for signed URLs; otherwise downloads fall back to S3.
+AWS_CLOUDFRONT_URL=https://d1234567890abcdef.cloudfront.net
```
tommoor:
```suggestion
AWS_CLOUDFRONT_URL=
```

I'd remove this value, we try to keep the defaults here be values you could use.

server/env.ts
```diff
@@ -635,6 +635,41 @@ export class Environment {
...
+  /**
+   * Optional CloudFront distribution URL for serving attachment downloads.
+   * Uploads continue to use the S3 endpoint directly. When set together with
+   * AWS_CLOUDFRONT_KEY_PAIR_ID and a private key, signed CloudFront URLs are
+   * used for downloads. If signing credentials are missing, S3 presigned URLs
+   * are used instead.
+   * Example: https://d1a2b3c4d5e6f.cloudfront.net (no trailing slash)
+   */
+  @IsOptional()
+  public AWS_CLOUDFRONT_URL = environment.AWS_CLOUDFRONT_URL ?? "";
```
tommoor:
```suggestion
  public AWS_CLOUDFRONT_URL = this.toOptionalString(environment.AWS_CLOUDFRONT_URL);
```

Use this helper for these so they will be properly `undefined`

server/env.ts
```diff
@@ -635,6 +635,41 @@ export class Environment {
...
+  @IsOptional()
+  public AWS_CLOUDFRONT_PRIVATE_KEY =
+    environment.AWS_CLOUDFRONT_PRIVATE_KEY ?? "";
+
+  /**
+   * Base64-encoded CloudFront private key. Takes precedence over
+   * AWS_CLOUDFRONT_PRIVATE_KEY when set.
+   */
+  @IsOptional()
+  public AWS_CLOUDFRONT_PRIVATE_KEY_BASE64 =
```
tommoor: I'd probably prefer this is combined with `AWS_CLOUDFRONT_PRIVATE_KEY` - it's easy to differentiate base64 encoded content

tommoor (changes requested):

server/env.ts
oliinykdm: I agree, it makes it easier to handle during setup

tommoor (approved):

PR #12673, opened 2026-06-12
fix(mcp): honor legacy append:true / prepend:true on update_document

### LINKED ISSUE #12672: update_document MCP tool: append: true parameter is silently ignored, body is always replaced

### Current Behavior

The update_document tool exposes an append: true boolean parameter which should append the provided text to the existing document body. In actual behavior, the parameter is silently ignored: the document body is always replaced by the provided text, resulting in silent data loss.
Steps to reproduce (verified on Outline 1.8.0 self-hosted, built-in MCP):

### Expected Behavior

When update_document is called with append: true, the text parameter should be appended to the existing document body rather than replacing it. The resulting body should contain the original content followed by the newly supplied text, with appropriate whitespace separation between the two (typically a newline).
Concretely, given an initial document body:
## BLOCK A
Marker: AAA-001
and the call:
update_document(id, text: "## BLOCK B\nMarker: BBB-002", append: true)
the resulting document body should be:
## BLOCK A
Marker: AAA-001

## BLOCK B
Marker: BBB-002
This is the standard semantic of an append flag in any content API — it preserves existing content and adds the new payload at the end. The parameter name itself (append) signals this contract to the client.

### Steps To Reproduce

Create a document with initial body containing "BLOCK A — Marker AAA-001".
Call update_document(id, text: "## BLOCK B — Marker BBB-002", append: true).
Fetch the document.
Expected: body contains BLOCK A followed by BLOCK B.
Actual: body contains only BLOCK B. BLOCK A is silently gone.

Impact: High — AI assistants using the MCP receive a successful response, but users lose existing document content. The data loss is silent and only detected when users manually inspect the page.
Workaround: Use update_document without the append flag, providing the full body (fetch first, append locally, then call update with the complete new body).

### Environment

```markdown
- Outline: 1.8.0 selfhosted
- Browser: Latest Safari Browser on macOS (latest)
```

### Anything else?

_No response_

tsushanth:
## Summary

Closes #12672 — silent data loss when an MCP client calls `update_document` with `append: true`.

## Root cause

Two places parse update-document input, and they disagreed on the legacy shape:

- **REST API** (`server/routes/api/documents/schema.ts`) accepts the deprecated `append: true` boolean and runs a `.transform()` block that rewrites it into `editMode: "append"` before the handler ever sees it. This is intentional back-compat.
- **MCP tool** (`server/tools/documents.ts`) only documents `editMode`. The boolean shorthand isn't in the schema, so Zod silently strips it.

LLM clients reaching for the MCP tool often infer `append: true` from the REST API shape, general API convention, or the verbal phrasing of the tool description ("In append mode it is added to the end"). The boolean disappears at validation, `editMode` falls back to its default of `replace`, and the document body is overwritten instead of appended — only detected when a user later opens the doc and notices content is gone.

## Fix

Add the two deprecated aliases to the MCP tool's input schema and fold them into `editMode` inside the handler before forwarding to `documentUpdater`:

```ts
const { append, prepend, editMode: rawEditMode, ...rest } = input;
let editMode = rawEditMode;
if (editMode === undefined) {
  if (append) {
    editMode = TextEditMode.Append;
  } else if (prepend) {
    editMode = TextEditMode.Prepend;
  }
}
```

Both schema descriptions explicitly mark the new fields as deprecated and steer callers toward `editMode`, so an LLM that reads the tool definition picks the canonical name in new code:

```ts
append: z
  .boolean()
  .optional()
  .describe(
    'Deprecated alias for editMode="append". When true and editMode is not set, text is added to the end of the document instead of replacing it. Prefer editMode="append" in new code.'
  ),
```

**Precedence:** explicit `editMode` always wins. The alias is a fallback, not an override — so `append: true, editMode: "replace"` still replaces.

## Tests

Three new tests in `server/tools/documents.test.ts`:

- `appends text instead of replacing when the legacy append:true alias is set` — regression for #12672, builds a doc with `"Existing content"`, calls the tool with `append: true`, asserts both substrings are present.
- `prepends text instead of replacing when the legacy prepend:true alias is set` — same shape for the prepend alias, also asserts ordering.
- `prefers an explicit editMode over the legacy append:true alias` — guards the precedence rule.

## Test plan

- [x] `yarn tsc --noEmit` — clean
- [x] Test code mirrors the existing `update_document` test patterns (`buildOAuthUser` / `callMcpTool` / `Document.findByPk`).
- [ ] Local `yarn test server/tools/documents.test.ts` — could not run end-to-end on my machine due to a missing local Postgres `user` role (env issue, not test code). CI will run the full suite.

## Notes for review

- I considered adding the aliases to the canonical `documentUpdater` props instead of stripping them in the tool handler. Kept the translation localized to the MCP layer because the REST API already handles its own (slightly different) Zod transform there, and `documentUpdater` would otherwise grow back-compat semantics that have nothing to do with its model contract.
- The tool description's existing "always prefer editMode patch" guidance is left intact — this PR just makes the misread cases safe, it doesn't change which mode LLMs should reach for first.

### DIFF
--- a/server/tools/documents.test.ts
+++ b/server/tools/documents.test.ts
@@ -394,6 +394,90 @@ describe("update_document", () => {
     expect(res?.result?.isError).toBe(true);
   });
 
+  it("appends text instead of replacing when the legacy append:true alias is set", async () => {
+    // Regression for #12672. The REST API silently accepts `append: true` and
+    // translates it to editMode="append" — the MCP tool used to drop the flag,
+    // collapse to editMode="replace", and overwrite the document body.
+    const { user, accessToken } = await buildOAuthUser();
+    const collection = await buildCollection({
+      teamId: user.teamId,
+      userId: user.id,
+    });
+    const document = await buildDocument({
+      teamId: user.teamId,
+      userId: user.id,
+      collectionId: collection.id,
+      text: "Existing content",
+    });
+
+    const res = await callMcpTool(server, accessToken, "update_document", {
+      id: document.id,
+      text: "Added content",
+      append: true,
+    });
+
+    expect(res?.result?.isError).toBeUndefined();
+    const reloaded = await Document.findByPk(document.id);
+    expect(reloaded?.text).toContain("Existing content");
+    expect(reloaded?.text).toContain("Added content");
+  });
+
+  it("prepends text instead of replacing when the legacy prepend:true alias is set", async () => {
+    const { user, accessToken } = await buildOAuthUser();
+    const collection = await buildCollection({
+      teamId: user.teamId,
+      userId: user.id,
+    });
+    const document = await buildDocument({
+      teamId: user.teamId,
+      userId: user.id,
+      collectionId: collection.id,
+      text: "Existing content",
+    });
+
+    const res = await callMcpTool(server, accessToken, "update_document", {
+      id: document.id,
+      text: "Prepended content",
+      prepend: true,
+    });
+
+    expect(res?.result?.isError).toBeUndefined();
+    const reloaded = await Document.findByPk(document.id);
+    expect(reloaded?.text).toContain("Existing content");
+    expect(reloaded?.text).toContain("Prepended content");
+    expect(reloaded?.text?.indexOf("Prepended content")).toBeLessThan(
+      reloaded?.text?.indexOf("Existing content") ?? -1
+    );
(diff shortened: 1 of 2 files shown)


tommoor:
Thanks, but there is no reason to bring long deprecated parameters from the API to the new MCP server

PR #12694, opened 2026-06-15
feat: support running under a URL sub-path via the URL env var

hemker:
Derive a base path from the path component of the URL environment variable (e.g. URL=https://host/outline) and thread it through routing, assets, websockets, cookies, and auth so the app can run under a sub-path behind a prefix-preserving reverse proxy.

The base path is applied at runtime, not baked at build time: bundle-internal asset URLs resolve through a window.__assetUrl runtime hook (vite renderBuiltUrl), so a prebuilt image works under any sub-path set at container start without rebuilding. When served from the domain root the base path is an empty string and behavior is unchanged.

- server: prefix koa mounts (/api, /mcp, /auth, /oauth, routes), websocket and collaboration paths, OAuth discovery endpoints, opensearch, and redirects
- client: React Router basename, API client base, service worker scope, auth form actions and redirects
- expose BASE_PATH to the client via the public env

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -38,6 +38,7 @@ import type {
   WebsocketEntitiesEvent,
   WebsocketEntityDeletedEvent,
 } from "~/types";
+import env from "~/env";
 import { AuthorizationError, NotFoundError } from "~/utils/errors";
 import Logger from "~/utils/Logger";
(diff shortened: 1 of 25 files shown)


tommoor:
Thanks, but this has been proposed before and we I not wish to maintain this particular option.

You're welcome to keep this logic on your own fork of course.

PR #12742, opened 2026-06-19
feat: Add installation switcher to desktop login

tommoor:
Allows the desktop app to be used with non-cloud instances


Codinator007:
Hi Tom,

thank you for the great feature.
Updated & tried it but I think there is an issue regarding self hosted instances with SSO (EntraID) enabled.
"Weiter mit Microsoft" opens a tab with Login but instead of closing or saying "successfully authenticated", the session is on browser and the app does not recognize the login. Any idea?
App & Outline on latest versions (macOS).

NeptunedGmbH:
> Hi Tom,
> 
> thank you for the great feature. Updated & tried it but I think there is an issue regarding self hosted instances with SSO (EntraID) enabled. "Weiter mit Microsoft" opens a tab with Login but instead of closing or saying "successfully authenticated", the session is on browser and the app does not recognize the login. Any idea? App & Outline on latest versions (macOS).

yep same here

Brramble:
> Hi Tom,
> 
> thank you for the great feature. Updated & tried it but I think there is an issue regarding self hosted instances with SSO (EntraID) enabled. "Weiter mit Microsoft" opens a tab with Login but instead of closing or saying "successfully authenticated", the session is on browser and the app does not recognize the login. Any idea? App & Outline on latest versions (macOS).

I'm also having this issue with Passkeys and PocketID

PR #12749, opened 2026-06-19
feat: Allow Passkey logins on desktop

tommoor:
Bounce the user to browser to complete passkey ceremony and then return to desktop with a single-use transfer token


app/scenes/Login/components/PasskeyAuthenticationProvider.tsx
```diff
@@ -0,0 +1,177 @@
...
+        {fields &&
+          Object.entries(fields).map(([fieldName, value]) => (
+            <input
+              key={fieldName}
+              type="hidden"
+              name={fieldName}
+              value={value}
+              readOnly
+            />
+          ))}
```
tommoor, line 154: Refactor from the previous imperative approach

PR #12788, opened 2026-06-22
feat: Add toggle headings

tommoor:
Adds convenience items in the `/` menu for creating a toggle container with a heading within


shared/editor/plugins/PlaceholderPlugin.ts
```diff
@@ -46,14 +53,18 @@ export class PlaceholderPlugin extends Plugin {
...
+    nodeTypes: string[]
+  ) {
     const paras: Array<{
       node: Node;
       $start: ResolvedPos;
       parent: Node | null;
     }> = [];
     state.doc.descendants((node, pos, parent) => {
-      if (node.type.name === "paragraph") {
+      if (nodeTypes.includes(node.type.name)) {
```
tommoor, line 67: Enable placeholders to work in headings too

bx8361:
May I ask: will this remove the original heading folding? 
To get foldable headings, I now have to use toggle headings, which add a margin and make them appear indented.
Though it's expected, this makes it work more like a toggle block than foldable headings.

tommoor:
Yes, the old folding is removed in 1.9.0 (https://github.com/outline/outline/pull/12729). 

This had to happen for a couple of reasons. Namely we need the new toggle container as it is much more flexible and efficient than only being able to toggle headings. The behavior is also subtly different as the state of the new container _always starts closed_ (this is key) and the open state does not sync to other users - everyone has control over their own view.

With headings we could never start them closed and the behavior synced to other users which was very annoying in multiplayer sessions.

bx8361:
> Yes, the old folding is removed in 1.9.0 (#12729).
> 
> This had to happen for a couple of reasons. Namely we need the new toggle container as it is much more flexible and efficient than only being able to toggle headings. The behavior is also subtly different as the state of the new container _always starts closed_ (this is key) and the open state does not sync to other users - everyone has control over their own view.
> 
> With headings we could never start them closed and the behavior synced to other users which was very annoying in multiplayer sessions.

Okay, I rely on folding headings, though. I'll probably revert folding for headings on my side. 🎉

PR #12864, opened 2026-06-30
fix(mcp): return text content for empty list tool results

### LINKED ISSUE #12858: MCP `list_documents` empty content

### Current Behavior

When a document search matches nothing, the MCP `list_documents` tool still reports success but comes back empty — it returns an empty content list instead of any "no results" indication. An empty result is valid per the MCP spec, but it causes two real problems for clients: the calling model can't tell "nothing found" apart from "the tool broke," and some clients reject an empty result outright. Through **Anthropic's MCP connector**, for example, a no-match search doesn't come back as "no results" — it comes back as the error *"The MCP tool returned content in a format not currently supported by the Anthropic API."* So any search that happens to match nothing fails through that client and is effectively unusable. The same happens with the other list-style tools (`list_collections`, `list_collection_documents`) whenever they return nothing.

### Expected Behavior

A zero-match search should return something a client reads as "no results" — a non-empty `content` with an explicit no-results block — not an empty array. Options (decision yours):

1. *(preferred)* In the shared `success()` helper, return a single text block (e.g. `"No results"` / `JSON.stringify([])`) when the payload is empty. One change; covers every `list_*` tool.
2. Handle the empty case per-tool in `list_documents`.
3. Document the empty-array behavior and leave it to clients.

The empty array is the worst option for LLM clients: silently indistinguishable from failure, and it breaks at least one major connector.

### Steps To Reproduce

1. Call the `list_documents` MCP tool with a query that matches nothing:

```bash
curl -s https://<your-instance>/mcp \
  -H 'content-type: application/json' \
  -H 'accept: application/json, text/event-stream' \
  -H 'authorization: Bearer <api-token>' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"list_documents","arguments":{"query":"zzqqxx-no-such-doc"}}}'
```

2. Response is a success with an **empty content array**:

```
data: {"result":{"content":[]},"jsonrpc":"2.0","id":1}
```

(A query that *does* match returns `content: [{ "type": "text", ... }, ...]` normally.)

3. The same zero-match call via Anthropic's MCP connector (Messages API `mcp_servers`, beta `mcp-client-2025-11-20`) returns a `tool_result` with content = *"…content in a format not currently supported…"* — the call fails rather than returning "no results."

### Environment

```markdown
- **Outline**: 1.8.0-1 (self-hosted, official image `outlinewiki/outline:1.8.0-1`)
- **MCP transport**: Streamable HTTP at `https://<your-instance>/mcp`; protocol negotiated `2025-06-18` (`serverInfo: {name: "outline", version: "1.8.0-1"}`)
- **Affected client**: Anthropic Messages API MCP connector (beta `mcp-client-2025-11-20`)
- Reproducible at the protocol layer with the raw `tools/call` above — independent of client, reverse proxy, OS, or DB.
```

### Anything else?

Root cause (verified on `main`): `list_documents` returns `success(presented)` ([documents.ts#L200](https://github.com/outline/outline/blob/main/server/tools/documents.ts#L200), [#L238](https://github.com/outline/outline/blob/main/server/tools/documents.ts#L238)); on a no-match search `presented` is `[]`. The shared helper [`success()` in server/tools/util.ts#L74-L85](https://github.com/outline/outline/blob/main/server/tools/util.ts#L74-L85) maps the array to text blocks, so `success([])` → `{ content: [] }`. Affects any list-style tool that can return zero items (`list_documents`, `list_collections`, `list_collection_documents`).

For reference, the client side: Anthropic's [MCP connector](https://platform.claude.com/docs/en/docs/agents-and-tools/mcp-connector) only serializes text/image content into a `tool_result` and rejects an empty content array.

Happy to open a small PR adding the empty-case guard to `success()` if that approach sounds right.

syf2211:
## Summary

When MCP list tools return zero items, `success([])` previously produced `{ content: [] }`. Anthropic's MCP connector rejects that format and surfaces a misleading API error instead of "no results."

This change returns a single JSON text block (`[]`) for empty list payloads so clients can distinguish zero results from a broken tool response.

## Motivation

Fixes #12858. The issue author confirmed the preferred approach is a shared guard in `success()` rather than per-tool handling.

## Changes

- Guard empty arrays in `server/tools/util.ts` `success()` helper
- Add unit tests for empty, multi-item, and single-object responses
- Add `parseMcpListContent` test helper for the `[]` sentinel block
- Update integration tests that can receive zero list results

## Tests

- `yarn test:server server/tools/util.test.ts` — 13 passed (3 new `success` tests + existing unit tests; DB integration tests require PostgreSQL)
- `yarn lint server/tools/util.ts server/tools/util.test.ts` — no new issues in changed files

## Notes

- Applies to all list-style MCP tools (`list_documents`, `list_collections`, `list_collection_documents`, `list_users`, `list_templates`, etc.)
- Non-empty list responses are unchanged (one text block per item)
- Empty collection trees still return a single `[]` block, which is the machine-readable equivalent of zero items

### DIFF
--- a/server/test/McpHelper.ts
+++ b/server/test/McpHelper.ts
@@ -120,3 +120,25 @@ export async function callMcpTool(
       }
     | undefined;
 }
+
+/**
+ * Parses list-style MCP tool content blocks into an array of items.
+ * Zero-result list tools return a single `[]` sentinel block instead of
+ * an empty content array.
+ *
+ * @param content - the content blocks from an MCP tool result.
+ * @returns the parsed list items.
+ */
+export function parseMcpListContent<T = unknown>(
+  content: { text?: string }[] | undefined
+): T[] {
+  if (!content?.length) {
+    return [];
+  }
+
+  if (content.length === 1 && content[0]?.text === "[]") {
+    return [];
+  }
+
+  return content.map((c) => JSON.parse(c.text ?? "{}") as T);
+}
--- a/server/tools/collections.test.ts
+++ b/server/tools/collections.test.ts
@@ -13,16 +13,17 @@ describe("collection tools", () => {
     });
 
     const res = await callMcpTool(server, accessToken, "list_collections");
-    const data = parseMcpListContent<{ id: string }>(res?.result?.content);
+    const data = parseMcpListContent<{ id: string; url: string }>(
+      res?.result?.content
+    );
 
     expect(data.length).toBeGreaterThanOrEqual(1);
-    const ids = data.map((c: { id: string }) => c.id);
+    const ids = data.map((c) => c.id);
     expect(ids).toContain(collection.id);
 
-    const match = data.find((c: { id: string }) => c.id === collection.id) as {
-      url: string;
-    };
-    expect(match.url).toMatch(/^https?:\/\//);
+    const match = data.find((c) => c.id === collection.id);
+    expect(match).toBeDefined();
+    expect(match!.url).toMatch(/^https?:\/\//);
   });
 
   it("list_collections does not return collections from another team", async () => {
--- a/server/tools/collections.test.ts
+++ b/server/tools/collections.test.ts
@@ -1,6 +1,6 @@
 import { buildCollection, buildUser } from "@server/test/factories";
 import { getTestServer } from "@server/test/support";
-import { buildOAuthUser, callMcpTool } from "@server/test/McpHelper";
+import { buildOAuthUser, callMcpTool, parseMcpListContent } from "@server/test/McpHelper";
 
 const server = getTestServer();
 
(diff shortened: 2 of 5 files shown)


syf2211:
I have read the CLA Document and I hereby sign the CLA

syf2211:
recheck

server/test/McpHelper.ts
```diff
@@ -120,3 +120,25 @@ export async function callMcpTool(
...
+ * @returns the parsed list items.
+ */
+export function parseMcpListContent<T = unknown>(
+  content: { text?: string }[] | undefined
+): T[] {
+  if (!content?.length) {
+    return [];
+  }
+
+  if (content.length === 1 && content[0]?.text === "[]") {
```
tommoor, line 139: When would this clause be true?

syf2211:
I have read the CLA Document and I hereby sign the CLA

syf2211:
recheck

syf2211:
That branch handles the empty-list sentinel emitted by `success([])` in `server/tools/util.ts`.

When a list MCP tool finds zero items, we now return:

```json
{ "content": [{ "type": "text", "text": "[]" }] }
```

instead of `{ "content": [] }` (which Anthropic rejects). Tests that previously did `(content ?? []).map(...)` need to treat that single `"[]"` block as an empty list rather than a real item — hence the `content.length === 1 && content[0]?.text === "[]"` guard in `parseMcpListContent`.

It would not match a normal one-item list response, because those serialize the item object (e.g. `{"id":"..."}`), not the literal string `[]`.

tommoor (approved):

PR #12872, opened 2026-07-02
fix(mermaid): use opacity:0 render container to fix foreignObject sizing

### LINKED ISSUE #12812: Multiple Mermaid diagram types render with massively oversized SVG nodes in Outline

### Current Behavior

Multiple Mermaid diagram types render with massively oversized SVG dimensions in Outline.

The issue has been reproduced with:

- requirementDiagram
- classDiagram
- erDiagram

The issue persists after upgrading Outline to the latest version and clearing browser/site cache.

This is not caused by long labels, a large number of nodes, browser zoom, or self-hosting configuration. It can be reproduced with small Mermaid examples.

The generated SVG is already oversized before it is inserted into the document. The issue appears to occur during Mermaid rendering in Outline's offscreen render container.

For example, given the following diagram:

```
erDiagram
    CAR ||--o{ NAMED-DRIVER : allows
    CAR {
        string registrationNumber PK
        string make
        string model
        string[] parts
    }
    PERSON ||--o{ NAMED-DRIVER : is
    PERSON {
        string driversLicense PK "The license #"
        string(99) firstName "Only 99 characters are allowed"
        string lastName
        string phone UK
        int age
    }
    NAMED-DRIVER {
        string carRegistrationNumber PK, FK
        string driverLicence PK, FK
    }
    MANUFACTURER only one to zero or more CAR : makes
```
this is the result I get:

Image

but  the same diagram look on https://mermaid.live/

Image

### Expected Behavior

Mermaid diagrams should render at a normal size, with node dimensions based on their actual text content and diagram layout.

Small diagrams should not produce SVG files with multi-thousand-pixel node dimensions or extremely large viewBox values.

## Actual behavior

The generated SVG contains abnormally large node paths and a huge viewBox.

For example:

- A small `requirementDiagram` produces:
viewBox="0 0 2756 5209.984375"
style="max-width: 2756px; width: 100%;"
- A small `classDiagram` produces:
viewBox="0 0 7401.00048828125 5965.95361328125"
style="max-width: 7401px; width: 100%;"

In the affected SVG output, text elements are only around 20–160px wide, but their corresponding node paths are generated at approximately 1,000–2,700px wide/high.

For example, a Class03 label has a text width of approximately 66px, but its node path is approximately 1144 × 1180px:

<foreignObject width="65.68359375" height="23.984375">
  ...
  <p>Class03</p>
</foreignObject>

<path d="M-572 -590 L572 -590 L572 590 L-572 590">

This causes the entire diagram to become excessively large and difficult to view in Outline.

### Steps To Reproduce

1. Create a new Outline document.
1. Insert one of the Mermaid examples below.
1. Observe that the diagram is rendered much larger than expected.
1. Open browser DevTools and inspect the generated SVG.
1. Confirm that the SVG viewBox, max-width, and node paths are abnormally large.

## erDiagram

```
erDiagram
    CAR ||--o{ NAMED-DRIVER : allows
    CAR {
        string registrationNumber PK
        string make
        string model
        string[] parts
    }
    PERSON ||--o{ NAMED-DRIVER : is
    PERSON {
        string driversLicense PK "The license #"
        string(99) firstName "Only 99 characters are allowed"
        string lastName
        string phone UK
        int age
    }
    NAMED-DRIVER {
        string carRegistrationNumber PK, FK
        string driverLicence PK, FK
    }
    MANUFACTURER only one to zero or more CAR : makes
````

## requirementDiagram

```
requirementDiagram
requirement test_req {
  id: 1
  text: the test text.
  risk: high
  verifymethod: test
}

element test_entity {
  type: simulation
}

test_entity - satisfies -> test_req
```
## classDiagram

```
classDiagram
Class01 <|-- AveryLongClass : Cool
Class03 *-- Class04
Class05 o-- Class06
Class07 .. Class08
Class09 --> C2 : Where am i?
Class09 --* C3
Class09 --|> Class07
Class07 : equals()
Class07 : Object[] elementData
Class01 : size()
Class01 : int chimp
Class01 : int gorilla
Class08 <--> C2: Cool label

```

### Environment

```markdown
- Outline: v1.8.1
- Browser: edge,chrome
```

### Anything else?

Additional context

The generated SVG IDs use the offscreen-mermaid-* prefix, for example:

<svg id="offscreen-mermaid-..." ...>

This suggests the incorrect dimensions may be produced while Mermaid measures text and lays out nodes in Outline's offscreen rendering container.

The affected diagram types use structured nodes with multiple sections and foreignObject-based text rendering or measurement.

This is not related to self-hosting configuration. The issue is reproducible using small Mermaid examples and the generated SVG dimensions are already incorrect before the SVG is inserted into the document.

This may be related to the offscreen Mermaid rendering changes introduced for high-DPI / RDP display handling, but the issue also occurs after cache clearing and with the latest Outline version.

tsushanth:
Fixes #12812

## Root cause

`classDiagram`, `erDiagram`, and `requirementDiagram` render node labels using `<foreignObject>` elements. Browsers skip layout computation for `<foreignObject>` content inside a `visibility:hidden` SVG — when mermaid calls `getBBox()` on those nodes to drive the dagre/ELK layout, it gets zero-size boxes and the layout engine distributes elements across a canvas that is orders of magnitude too large.

The symptom is a viewBox like `0 0 7401 5966` for a three-class diagram that should fit in ~400×300.

## Fix

Replace `visibility: hidden` with `opacity: 0` + `pointer-events: none` on the render container.

`opacity: 0` keeps the element fully in the render tree — layout, paint, and `getBBox()` all run normally for both regular SVG shapes and `<foreignObject>` content. The user sees nothing because the element is transparent and behind everything (z-index: -1).

The previous switch from `position: absolute; left: -9999px` → `visibility: hidden` fixed Chromium's `getBBox()` bug for off-screen elements (mermaid-js/mermaid#6146). `opacity: 0` avoids both that bug and the `foreignObject` layout-skip issue.

The session-storage cache prefix is bumped from `v2` → `v3` so any incorrectly-sized diagrams cached by the previous render path are discarded and re-rendered with the fix.

### DIFF
--- a/shared/editor/extensions/Mermaid.ts
+++ b/shared/editor/extensions/Mermaid.ts
@@ -28,9 +28,9 @@ export type MermaidState = {
   editingId?: string;
 };
 
-// The `v2` namespace discards entries cached before the #11782 fix, so
+// The `v3` namespace discards entries cached before the foreignObject fix, so
 // previously mis-sized diagrams are re-rendered instead of served from cache.
-const STORAGE_PREFIX = "mermaid:v2:";
+const STORAGE_PREFIX = "mermaid:v3:";
 const MAX_STORAGE_ENTRIES = 20;
 
 class Cache {
@@ -160,16 +160,22 @@ class MermaidRenderer {
       return;
     }
 
-    // Create a temporary element for rendering. We use visibility:hidden instead of
-    // offscreen positioning so the browser computes correct bounding boxes for SVG
-    // elements — offscreen elements can produce incorrect getBBox() results, leading
-    // to wrong viewBox dimensions (see mermaid-js/mermaid#6146).
+    // Create a temporary element for rendering. We use opacity:0 instead of
+    // visibility:hidden because browsers skip layout of <foreignObject> content
+    // inside visibility:hidden SVGs, causing mermaid's layout engine to measure
+    // zero-size nodes and produce inflated viewBox dimensions for diagram types
+    // that use foreignObject-based text (classDiagram, erDiagram,
+    // requirementDiagram). opacity:0 keeps the element in the render tree and
+    // fully laid out without being visible to the user. We previously used
+    // offscreen positioning (left:-9999px) which broke getBBox() in Chromium
+    // (mermaid-js/mermaid#6146); opacity:0 avoids both problems.
     const renderElement = document.createElement("div");
     const tempId =
       "offscreen-mermaid-" + Math.random().toString(36).substr(2, 9);
     renderElement.id = tempId;
     renderElement.style.position = "fixed";
-    renderElement.style.visibility = "hidden";
+    renderElement.style.opacity = "0";
+    renderElement.style.pointerEvents = "none";
     renderElement.style.top = "0";
     renderElement.style.left = "0";
     const width = this.editor.view?.dom.clientWidth ?? window.innerWidth;


plugins/notion/server/notion.ts
```diff
@@ -207,24 +207,52 @@ export class NotionClient {
...
+    let restarts = 0;
 
     while (hasMore) {
-      const response = await this.fetchWithRetry(() =>
-        this.client.search({
-          start_cursor: cursor,
-          page_size: this.pageSize,
-        })
-      );
+      let response;
```
tommoor: Unrelated changes

tsushanth: Good catch — those notion.ts changes belonged to a separate PR (#12871). Force-pushed to drop them; this PR now only touches Mermaid.ts.

tsushanth:
I'll sign the CLA via the link — just a moment.

tommoor:
I have never been able to reproduce this issue so I'm not sure if this fixes it - but it's worth a try, doesn't appear to _break_ anything.

The verbose AI commentary I can live with here given how many rounds this fix has had

tommoor (approved):

PR #12873, opened 2026-07-02
Improve sidebar menu and navigation tree performance

Ali-ovo:
## Summary

This PR reduces unnecessary work during sidebar rendering and document navigation.

The main changes are:

- Lazily build context menu items only while the context menu is open.
- Avoid creating document menu actions for sidebar rows until their context menu is opened.
- Cache document child lookups in `Collection` and `NavigableModel` so `getChildrenForDocument` no longer recursively walks the full navigation tree on each call.
- Reuse `useMenuAction` root actions when the action array has not changed.
- Share `IntersectionObserver` instances in `useOnScreen`, batch observer callbacks with `requestAnimationFrame`, and skip redundant state updates.

## Motivation

Sidebar rows can mount many context menus and document actions at once. Previously, closed context menus still converted actions into menu items, and `getChildrenForDocument` recursively scanned the document tree each time it was called. This could add noticeable React/MobX work when opening documents, changing the active document, or rendering starred/sidebar rows.

These changes keep the same behavior while moving expensive work to the point where it is actually needed.

## Testing

- Added regression coverage for lazy context menu item creation.
- Added coverage for `useMenuAction` avoiding unnecessary root action recreation.
- Added coverage for cached child document lookup.
- Added coverage for shared `IntersectionObserver` usage in `useOnScreen`.


app/components/Menu/ContextMenu.test.tsx
```diff
@@ -0,0 +1,95 @@
+/* oxlint-disable */
```
tommoor: Lets remove this useless test

app/hooks/useMenuAction.test.tsx
```diff
@@ -0,0 +1,55 @@
+/* oxlint-disable */
```
tommoor: Same lets remove this test, its slop overkill

tommoor (changes requested):
Broadly good but there are too many disparate changes in this - can you remove all of the tests which are over-mocked and separate out the `useOnScreen` observer sharing which I'm really not confident about.

The rest looks pretty mergable.

Ali-ovo:
> Broadly good but there are too many disparate changes in this - can you remove all of the tests which are over-mocked and separate out the `useOnScreen` observer sharing which I'm really not confident about.
> 
> The rest looks pretty mergable.

Do you mean that the changes to `useOnScreen`  will be separated separately?

Ali-ovo:
I removed the `useOnScrren` change first，then I'll explain the reason and function of this modification. I think there will be quite a significant performance optimization, but it can be carried out later

Previously, each row using useOnScreen created its own IntersectionObserver. When a large sidebar tree is expanded, this can result in many observers. During scrolling, many visibility changes can then trigger many React state updates.

The change did three things:

1. Share a single IntersectionObserver for elements with the same root / rootMargin / threshold, reducing the number of observers.
2. Collect multiple IntersectionObserver callbacks and flush them with requestAnimationFrame, so visibility updates are batched into a frame instead of firing continuously during scroll.
3. Track the current visibility value with isIntersectingRef and skip setState when the value has not changed, avoiding unnecessary re-renders.

Ali-ovo:
@tommoor  Maybe this pr can be merged

app/components/Sidebar/components/DocumentLink.tsx
```diff
@@ -378,6 +380,7 @@ const DocumentLinkInner = observer(function DocumentLinkInner({
 
   const contextMenuAction = useDocumentMenuAction({
     documentId: node.id,
+    enabled: contextMenuOpen,
```
tommoor: Coming back to this again, all this `enabled` plumbing does not feel good. Why would we define the context menu action here and then plumb the open/close events through from the `ContextMenu` component in SidebarLink?

It would be much simpler to allow ContextMenu to take a factory method - so `useDocumentMenuAction` just returns a function but does no actual calculations until opened

Ali-ovo: That makes sense. I’ll refactor ContextMenu to accept an action factory and have useDocumentMenuAction return a memoized factory. This will remove the enabled flag and the open/close event plumbing through the sidebar components while keeping action construction lazy.

Ali-ovo:
@tommoor  Based on the Chrome Performance traces, the current changes provide some measurable performance improvements. However, the largest remaining bottleneck is still `useOnScreen`.


When many sidebar rows are mounted or their visibility is recalculated at once, each row calls `getBoundingClientRect` and creates its own `IntersectionObserver`, causing significant performance overhead. This optimization could, however, be discussed and addressed in a separate follow-up PR.

tommoor (approved):

PR #12876, opened 2026-07-02
fix: Serialize API and imported markdown hard breaks as double spaces

beretienne:
Fix to a finding highlighted during .md export/import testing. 

## Summary
- When exporting/pulling a document or importing a Markdown file, hard breaks currently serialize as an internal escaped backslash (`\n`) rather than a standard CommonMark hard break.
- This updates `DocumentHelper`, `DocumentConverter`, and `documentImporter` to serialize with `commonMark: true`, so hard breaks materialize as standard CommonMark trailing double spaces instead.

## Test plan
- [x] `yarn tsc --noEmit`
- [x] `yarn lint`
- [x] `yarn test server/commands/documentImporter.test.ts server/models/helpers/DocumentHelper.test.ts`

### DIFF
--- a/server/commands/documentImporter.test.ts
+++ b/server/commands/documentImporter.test.ts
@@ -214,6 +214,24 @@ describe("documentImporter", () => {
     expect(response.title).toEqual("Title");
   });
 
+  it("should preserve hard breaks as trailing double spaces in markdown imports", async () => {
+    const user = await buildUser();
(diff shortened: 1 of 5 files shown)


tommoor:
This is a backwards compatibility issue for us as it will change the serialized content in the db, probably not going to change this

PR #12877, opened 2026-07-03
feat: add MCP tools for importing documents from uploaded files

Ali-ovo:
## Summary

Adds MCP tools for creating documents from uploaded source files using Outline’s existing Import document pipeline.

This introduces:
- `prepare_document_file_upload` to create a temporary pre-signed upload target for a document source file.
- `create_document_from_uploaded_file` to create a document from that uploaded source file.
- MCP guidance that tells clients to use this flow for file-based document creation instead of manually converting supported files to markdown.

## Motivation

MCP clients and AI agents often generate rich documents as files, such as HTML or Word documents. Today, the MCP `create_document` tool only accepts markdown-like document content, so agents either need to inline converted text or perform their own HTML/Word-to-markdown conversion.

That conversion can be lossy, especially for documents containing images, media, tables, or other rich structure. Outline already has a built-in Import document flow for these source files, but that flow was not available through MCP.

This change exposes that existing import path to MCP clients so generated or local document files can be uploaded and imported through the same pipeline as the product UI.

### DIFF
--- a/server/tools/documents.ts
+++ b/server/tools/documents.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import { z } from "zod";
 import { type McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { type CallToolResult } from "@modelcontextprotocol/sdk/types.js";
@@ -8,15 +9,28 @@ import documentCreator, {
 import documentMover from "@server/commands/documentMover";
 import documentRestorer from "@server/commands/documentRestorer";
 import documentUpdater from "@server/commands/documentUpdater";
-import { Collection, Document, Template } from "@server/models";
+import { ValidationError } from "@server/errors";
+import {
+  Attachment,
+  Collection,
+  Document,
+  Team,
+  Template,
+} from "@server/models";
+import AttachmentHelper from "@server/models/helpers/AttachmentHelper";
 import { sequelize } from "@server/storage/database";
+import FileStorage from "@server/storage/files";
 import { authorize, can } from "@server/policies";
+import presentAttachment from "@server/presenters/attachment";
 import {
   presentDocument as presentDocumentBase,
   presentNavigationNode,
 } from "@server/presenters";
+import type { DocumentImportTaskResponse } from "@server/queues/tasks/DocumentImportTask";
+import DocumentImportTask from "@server/queues/tasks/DocumentImportTask";
 import AuthenticationHelper from "@shared/helpers/AuthenticationHelper";
 import { UrlHelper } from "@shared/utils/UrlHelper";
+import { bytesToHumanReadable } from "@shared/utils/files";
 import {
   error,
   success,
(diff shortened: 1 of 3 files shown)


tommoor:
> MCP clients and AI agents often generate rich documents as files, such as HTML or Word documents

Your statement doesn't really hold water, it's not possible for an agent to create a Word file. HTML I can understand – though it's probably better to just have a tool to create directly from HTML rather than messing around with file uploads which agents are bad at?

Ali-ovo:
> > MCP clients and AI agents often generate rich documents as files, such as HTML or Word documents
> 
> Your statement doesn't really hold water, it's not possible for an agent to create a Word file. HTML I can understand – though it's probably better to just have a tool to create directly from HTML rather than messing around with file uploads which agents are bad at?

image
The intent here is to expose the existing “Import document” capability to MCP clients

Maybe my description is a little off.

tommoor:
The aim of the MCP isn't to mirror the API 1-1 though, so there needs to be a good reason to expose new tools. What is your usecase? Would an HTML input not suffice?

Ali-ovo:
> The aim of the MCP isn't to mirror the API 1-1 though, so there needs to be a good reason to expose new tools. What is your usecase? Would an HTML input not suffice?

Sometimes AI generates HTML files, and when I try to create a document from that HTML using the MCP `create_document` tool, media assets in the HTML such as videos and images can be lost.

I tested the built-in `Import document` feature, and it handles this use case correctly, including preserving those media assets. That’s why I’d like to expose this capability as an MCP tool.

server/tools/documents.ts
```diff
@@ -291,6 +305,227 @@ export function documentTools(server: McpServer, scopes: string[]) {
...
+              key,
+              acl,
+              size,
+              expiresAt: AttachmentHelper.presetToExpiry(preset),
+              contentType,
+              teamId: user.teamId,
+              userId: user.id,
+            });
+
+            const presignedPost = await FileStorage.getPresignedPost(
```
sviat-barbutsa, line 373: Hi, nice addition. One compatibility detail I’d like to highlight: `prepare_document_file_upload` always builds a presigned POST form here, but the existing MCP `create_attachment` tool branches on `env.AWS_S3_UPLOAD_METHOD === "put"` and returns a PUT URL/headers shape for installations configured that way.

For those deployments, this new document-import upload flow can give agents a POST form even though normal MCP attachments use PUT uploads. Could you, please, mirror the upload-mode handling from `server/tools/attachments.ts`, or add a focused test with `AWS_S3_UPLOAD_METHOD=put` so this tool returns the same mode/shape as `create_attachment`?

sviat-barbutsa (commented):
Hi, everything is great but could you, please, check one small follow-up comment?

tommoor (changes requested):
Lets just support HTML as an input for create_document tool and see how far that gets us

Ali-ovo:
> Lets just support HTML as an input for create_document tool and see how far that gets us

Are these changes still needed? Should I close this PR？

tommoor:
I would close in favor of HTML tool input. Please file issues in future so we can discuss implementation before code is written

Ali-ovo:
Got it, I’ll close this PR and open an issue to discuss supporting HTML input for the MCP document creation flow instead.

PR #12893, opened 2026-07-05
fix(locale): translate missing pt-BR string

noadevbr:


tommoor:
Thanks, however translations must be provided through CrowdIn - edits in these files are overwritten

Docs - https://github.com/outline/outline/blob/main/docs/TRANSLATION.md

PR #12895, opened 2026-07-05
Add option to link to a heading when inserting a document

### LINKED ISSUE #11712: Add option to link to a heading when inserting a document link
Not sure on the exact UX yet, but after typing `@` and selecting a document perhaps you can press the right arrow to expose headings within the document that can be directly linked to.

salihudickson:
closes #11712 

Demo: 

https://github.com/user-attachments/assets/712dd48a-505c-4765-abfc-1d7389016428

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,4 @@
   "typescript.validate.enable": true,
   "typescript.format.enable": true,
   "editor.formatOnSave": true,
-  "editor.defaultFormatter": "oxc.oxc-vscode"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,4 +4,5 @@
   "typescript.validate.enable": true,
   "typescript.format.enable": true,
   "editor.formatOnSave": true,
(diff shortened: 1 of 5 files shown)


tommoor:
Yea, the UX here isn't great as now it's an extra awkward tap/click to mention a document which is the 90% action. Also, the document content is not guaranteed to be loaded so the headings will only show up sometimes.

This is one of the reasons I haven't tackled this - there are not many great options for how to achieve it. 

One quick win we could do is ensure that pasting a doc link with a heading correctly creates a mention, that isn't the case today.

salihudickson:
Yeah, I think the idea is really good but implementation still needs some workshopping, I thought putting together a quick PR might help push it a bit.

> One quick win we could do is ensure that pasting a doc link with a heading correctly creates a mention, that isn't the case today.

Sure! I'll create a separate PR for this.

salihudickson:
> Also, the document content is not guaranteed to be loaded so the headings will only show up sometimes

was going to ask about this cause I wasn't even sure, and didn't want to add any unnecessary overhead.

PR #12937, opened 2026-07-08
fix: Wrap inline code in comments

VdustR:
Follow-up to #12735. Inline code in comments can still overflow narrower comment bubbles because the shared editor decorates inline code words with `.code-word` spans, and that class uses `white-space: nowrap` at tablet widths. The existing max word length guard handles very long tokens, but shorter domain-like tokens can still exceed the comment bubble width.

- Scope the wrapping change to comment rendering only.
- Allow `code.inline` to wrap anywhere inside comments.
- Let `.code-word` inherit the comment inline code wrapping behavior without changing normal document editor styling.

Verification:
- `yarn oxlint --type-aware app/scenes/Document/components/Comments/CommentThreadItem.tsx`
- `yarn oxfmt --check app/scenes/Document/components/Comments/CommentThreadItem.tsx`
- `yarn tsc --noEmit --pretty false`
- Ran Outline locally with a seeded comment and verified the inline code wraps inside the comments panel (`codeOverflowThread=false`).

Screenshot:

Before
file-c5d9f6f3db5a1cb25e0f7e4dea445c63

After
file-fb86f5b6abe63082743e653c4021951d

### DIFF
--- a/app/scenes/Document/components/Comments/CommentThreadItem.tsx
+++ b/app/scenes/Document/components/Comments/CommentThreadItem.tsx
@@ -9,6 +9,7 @@ import { toast } from "sonner";
 import styled, { css } from "styled-components";
 import breakpoint from "styled-components-breakpoint";
 import EventBoundary from "@shared/components/EventBoundary";
+import { EditorStyleHelper } from "@shared/editor/styles/EditorStyleHelper";
 import { s, hover } from "@shared/styles";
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks, I went a different direction where the functionality is simply never included in the comment editor instead of being suppressed via css. (https://github.com/outline/outline/pull/12961)

Issue #12901, opened 2026-07-06
Support HTML input in MCP create_document tool


Ali-ovo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

About https://github.com/outline/outline/pull/12877

AI agents sometimes generate rich HTML content that includes media such as images or videos.

Currently, when that HTML is passed to the MCP `create_document` tool, it is treated as markdown/text content. This can cause media in the HTML to be lost or not converted into document attachments.

The built-in “Import document” flow handles this case well because it runs HTML through Outline’s existing import/conversion pipeline and preserves supported media assets.

### Expected Behavior

support HTML input in MCP create_document tool

### Environment

```markdown
- Outline:
- Browser:
```

PR #12945, opened 2026-07-09
MCP: omit document body from create/update tool results

BoySevenIII:
### What

`create_document` and `update_document` MCP tools present their result with `includeText: true`, so every write — even a one-line `append`/`patch` — returns the **entire document body** in the tool result.

This flips those two handlers to `includeText: false`, matching what the read/list tools already do (`search_documents`, `list_*` use `includeText: false`). `restore_document` is intentionally left unchanged (the caller didn't supply the content there).

```diff
 // create_document
   presentDocument(document, {
     includeData: false,
-    includeText: true,
+    includeText: false,
     includeUpdatedAt: true,
   }),
 // update_document
   presentDocument(updated, {
     includeData: false,
-    includeText: true,
+    includeText: false,
     includeUpdatedAt: true,
   }),
```

The handlers already destructure `const [{ text, ...attributes }, breadcrumb]` and place `text` separately, so with `includeText: false` it is simply `undefined` and drops out of the response — no other change needed. Title, url, id, revision, `updatedAt` and the breadcrumb are still returned for workspace awareness.

### Why

On a write, the caller **just supplied the text** — echoing the full body back is pure redundancy that inflates the agent's context. For agents that edit large/long-lived docs (runbooks, task queues), each append re-emits the whole document; across a fleet of agent sessions this adds up to a large, silent context/token tax (measured ~1.3M tokens/month in one deployment) and offers no new information the client doesn't already have. Read tools already made this choice; this just makes writes consistent.

### Note

If you'd rather preserve the current default, I'm happy to change this to an opt-in input parameter (e.g. `minimal` / `includeText`) on the write tools instead of flipping the default — just let me know your preference.

### DIFF
--- a/server/tools/documents.ts
+++ b/server/tools/documents.ts
@@ -393,7 +393,7 @@ export function documentTools(server: McpServer, scopes: string[]) {
           const [{ text, ...attributes }, breadcrumb] = await Promise.all([
             presentDocument(document, {
               includeData: false,
-              includeText: true,
+              includeText: false,
               includeUpdatedAt: true,
             }),
             getDocumentBreadcrumb(document, user),
@@ -660,7 +660,7 @@ export function documentTools(server: McpServer, scopes: string[]) {
           const [{ text, ...attributes }, breadcrumb] = await Promise.all([
             presentDocument(updated, {
               includeData: false,
-              includeText: true,
+              includeText: false,
               includeUpdatedAt: true,
             }),
             getDocumentBreadcrumb(updated, user),


tommoor:
I understand the concern of context usage, this change seems likely ok. One gotcha to resolve is the PR at #12901 which conflicts with this logic slightly – when passing HTML as the input we should still return the Markdown output

The failing tests also need fixing

tommoor:
Did you plan on taking this across the finish line?

PR #12947, opened 2026-07-09
Add `sidebarHidden` query option

tommoor:
Adds utility for reading query string booleans, shared usage.

Ref https://github.com/outline/outline/discussions/12941

### DIFF
--- a/app/components/Layout.tsx
+++ b/app/components/Layout.tsx
@@ -30,8 +30,9 @@ const Layout = React.forwardRef(function Layout_(
   ref: React.RefObject<HTMLDivElement>
 ) {
   const { ui } = useStores();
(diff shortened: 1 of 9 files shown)


pricard32:
Hi, can I ask one more thing.
The params do not hide the Sharing link button at top left.
All the top nav bar are still visible.
Would it be possible to have an option to only display the content ?

Many thanks

tommoor:
Could you provide a screenshot of what you're referring to?

pricard32:
image

I still can return to the previous page and subscribe to email.

tommoor:
This option should've been more like chromehidden then, those items aren't in the sidebar. It doesn't really make sense to include them.

PR #13044, opened 2026-07-18
feat: Split view

tommoor:
Adds the ability to open two routes side-by-side, completely optional - everything acts as normal if this functionality is never used. It's particularly useful in the desktop app

image


app/actions/definitions/documents.tsx
```diff
@@ -34,6 +34,7 @@ import {
   EditIcon,
   EmbedIcon,
   OpenIcon,
+  SidebarIcon,
```
tommoor: Update to `outline-icons@4.5.0` and use `SplitIcon` instead of `SidebarIcon` everywhere it's used in this PR

app/utils/splitView.ts
```diff
@@ -0,0 +1,248 @@
...
+  to: LocationDescriptor,
+  state?: LocationState
+): LocationDescriptorObject {
+  if (typeof to === "string") {
+    const location = parsePath(to);
+    return state === undefined ? location : { ...location, state };
+  }
+
+  return to;
+}
```
tommoor: Should be in utils/history?

app/utils/splitView.ts
```diff
@@ -0,0 +1,248 @@
...
+  focusListeners.forEach((listener) => listener(pane));
+}
+
+/**
+ * Observes changes to the focused split view pane.
+ *
+ * @param callback invoked with the newly focused pane on every change.
+ * @returns a function that removes the observer.
+ */
+export function observeFocusedSplitPane(
```
tommoor: Use an observable box instead of manually maintained listeners? Is there anywhere that's using as a pure callback that would prevent this

PR #13046, opened 2026-07-19
feat: Add onNotice callback to decouple shared editor from toast library

tommoor:
Shared editor code previously imported and called `sonner`'s toast directly to surface notices (upload errors, "Copied to clipboard", unsupported link messages, etc).

This introduces an `onNotice(message, type?)` callback on the editor props, defaulting to a new `toastNotice` handler in the app layer, so shared editor code no longer depends on any specific notification library. All direct `toast.*` calls in `shared/editor/` are routed through this callback.


tommoor:
Note, this change _also_ reduces memory usage by avoiding loading sonner on the server.

PR #13048, opened 2026-07-19
feat(mcp): Add available icon names as resource

tommoor:
Not ever MCP client can read resources, but as this is complementary data that's ok.

ref https://github.com/outline/outline/discussions/12951


marklynk:
Thank you for this!

PR #13054, opened 2026-07-20
fix: Serialize GitHub issue/PR/project mentions as links in Markdown

### LINKED ISSUE #13053: GitHub issue/PR/project mentions serialize to opaque mention:// URIs in Markdown export

### Current Behavior

When a document contains a GitHub issue, pull request, or project mention (via the GitHub integration), serializing that document to Markdown produces an internal URI instead of a link:

```
@[Epic 1: Control plane Helm chart](mention://0c440212-8b40-49fa-8a64-2548d6b60d59/issue/c85a0d80-3a89-4b25-a0cd-e7fc83f0d226)
```

The `mention://` URI only carries Outline-internal UUIDs; there is no github.com URL in it, so the reference can't be resolved by anything outside Outline. It shows up in the three places that share the same serializer:

1. Markdown export (single document and collection zip export)
2. The documents API Markdown output
3. The built-in MCP server document reads (`server/routes/mcp/index.ts`)

So any consumer of that Markdown, including an agent reading a document over MCP, gets an opaque UUID with no path back to the underlying issue.

### Expected Behavior

A plain Markdown link to the real URL, consistent with what the HTML render already produces, for example:

```
[Epic 1: Control plane Helm chart](https://github.com/acme/infra/issues/2)
```

The node already stores the URL in `attrs.href`. In `shared/editor/nodes/Mention.tsx`, `toDOM` uses it for the anchor `href`/`data-url` on Issue/PullRequest/Project, and the `Enter` handler navigates to it. Only `toMarkdown` drops it and falls through to the `mention://` branch.

Same class of problem as #5293, where Markdown export emitted non-portable internal references while HTML used the full URLs.

### Steps To Reproduce

1. Connect the GitHub integration, paste a github.com issue link into a document and choose the mention option.
2. Export the document as Markdown (or read it through the documents API / MCP).
3. The output contains a `mention://` URI instead of a link.
4. Repeat with a pull request and a project link.

### Environment

- Outline: `main` (reproduced against the current `shared/editor/nodes/Mention.tsx`)

### Anything else?

One thing I'd like to settle before opening a PR: is emitting a plain link acceptable as the default, or would you rather have it behind an export option so the mention round-trip is preserved on re-import? And should the `url` mention type get the same treatment, since it also carries an `href`?

I have a small patch that turns Issue/PullRequest/Project mentions into links from `href`, guarded so anything without an `href` (and user/group/date mentions) keeps the existing `mention://` output. Glad to adjust the scope based on what you prefer.

impuls42:
Closes #13053

### What

GitHub issue, pull request and project mentions serialize to an internal `mention://` URI in Markdown:

```
@[Epic 1: Control plane Helm chart](mention://0c440212-.../issue/c85a0d80-...)
```

That URI only carries Outline-internal UUIDs, so the reference can't be resolved outside Outline. It affects the three paths that share this serializer: Markdown export (single doc and collection zip), the documents API Markdown output, and the built-in MCP server document reads.

These three types already store the real URL in `attrs.href` — `toDOM` uses it for the anchor `href`/`data-url`, and the `Enter` handler navigates to it. Only `toMarkdown` dropped it and fell through to the `mention://` branch.

This changes `toMarkdown` in `shared/editor/nodes/Mention.tsx` to emit a plain link from `href` for Issue/PullRequest/Project:

```
[Epic 1: Control plane Helm chart](https://github.com/acme/infra/issues/2)
```

The branch is guarded on `href` being present, so anything without one — and user/group/date/url mentions — keeps the existing `mention://` output. The link goes through `sanitizeUrl`, matching `toDOM`. The parser is untouched, so existing `mention://` documents still parse.

Same class of problem as #5293 (Markdown export emitting non-portable internal references while HTML used full URLs).

### Points worth flagging

- **Round-trip changes for these types.** Exported Markdown becomes a plain link, so re-importing produces a regular link rather than a live mention until it is re-unfurled. #8610 suggests the mention round-trip is valued for document mentions, so I'm happy to gate this behind an export option / setting if you'd prefer `mention://` to stay the default.
- **`url` mentions left as-is.** They also carry `href` and could serialize to a plain link the same way, but they're a more general unfurl surface so I left them out of this change. Happy to include them if you want.

### Tests

- New `shared/editor/nodes/Mention.test.ts` covering the serializer:
  - Issue / PullRequest / Project with an `href` → plain `[label](href)` link
  - Issue without an `href` → `mention://` fallback
  - User / Group unchanged, and `url` (with `href`) still `mention://`
- Existing mention parser tests (`shared/editor/rules/mention.test.ts`, `shared/utils/parseMentionUrl.test.ts`) still pass unchanged.

`yarn lint`, `yarn tsc` and the affected test files are green.

### Follow-ups

- The external-URL type check is factored into a small `isExternalUrlMention` helper now shared by `toDOM` (data-url), the `Enter` handler and `toMarkdown`, so they can't drift on which mention types are real links.
- The write-direction counterpart — turning inbound `github.com` links back into live mentions on API/MCP document writes — is tracked separately in #13055.

### DIFF
--- a/shared/editor/nodes/Mention.test.ts
+++ b/shared/editor/nodes/Mention.test.ts
@@ -1,0 +1,119 @@
+import { extensionManager, schema } from "../../test/editor";
+
+const serializer = extensionManager.serializer();
+
+function serializeMention(attrs: Record<string, unknown>) {
+  const doc = schema.nodeFromJSON({
+    type: "doc",
+    content: [
+      {
+        type: "paragraph",
+        content: [{ type: "mention", attrs }],
+      },
+    ],
+  });
+  return serializer.serialize(doc, { commonMark: true }).trim();
+}
+
+describe("Mention serialization", () => {
+  const id = "0c440212-8b40-49fa-8a64-2548d6b60d59";
+  const modelId = "c85a0d80-3a89-4b25-a0cd-e7fc83f0d226";
+
+  describe("external integration mentions", () => {
+    it("serializes an issue mention with an href to a plain link", () => {
+      expect(
+        serializeMention({
+          type: "issue",
+          label: "Epic 1: Control plane Helm chart",
+          modelId,
+          id,
+          href: "https://github.com/acme/infra/issues/2",
+        })
+      ).toBe(
+        "[Epic 1: Control plane Helm chart](https://github.com/acme/infra/issues/2)"
+      );
+    });
+
+    it("serializes a pull request mention with an href to a plain link", () => {
+      expect(
+        serializeMention({
+          type: "pull_request",
+          label: "Add Helm chart",
+          modelId,
+          id,
+          href: "https://github.com/acme/infra/pull/42",
+        })
+      ).toBe("[Add Helm chart](https://github.com/acme/infra/pull/42)");
+    });
+
+    it("serializes a project mention with an href to a plain link", () => {
+      expect(
+        serializeMention({
+          type: "project",
+          label: "Q3 Roadmap",
+          modelId,
+          id,
+          href: "https://github.com/orgs/acme/projects/7",
+        })
+      ).toBe("[Q3 Roadmap](https://github.com/orgs/acme/projects/7)");
+    });
+
+    it("falls back to mention:// when an issue mention has no href", () => {
+      expect(
+        serializeMention({
+          type: "issue",
+          label: "Epic 1",
+          modelId,
+          id,
+        })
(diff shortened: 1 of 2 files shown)


tommoor (changes requested):
We have an existing serializer option `commonMark` – the intent of this option is to denote markdown that is leaving the platform (it probably needs renaming now).

It makes sense to serialize as a plain link in the case this option is true, but for e.g MCP server markdown we want to retain the existing UUID format.

tommoor:
Superseded by https://github.com/outline/outline/pull/13459

PR #12943, opened 2026-07-09
feat: Add HTML input support to MCP create_document

Ali-ovo:
close https://github.com/outline/outline/issues/12901

## Summary

Adds optional HTML input support to the MCP `create_document` tool.

- Adds a `format` option to `create_document`, defaulting to markdown.
- When `format: "html"` is provided, the tool reuses the existing document import pipeline to convert HTML into document state.
- Preserves supported embedded media, such as images, as attachments instead of leaving raw HTML/media sources in the document.
- Keeps markdown behavior unchanged when `format` is omitted.

PR #13093, opened 2026-07-22
fix(mcp): clarify create_document HTML media handling in tool schema

Ali-ovo:
## Summary
- Clarifies the `create_document` `text` parameter description for HTML input that includes images/videos.
- Documents that remote and base64 media are imported as attachments automatically when `format` is `"html"`.
- Explicitly steers agents away from two failure modes seen in practice: converting media-rich HTML to markdown, or uploading the HTML file itself as an attachment instead of creating a document.
## Context
After adding HTML support to `create_document` (#12943), agents sometimes still:
1. Flatten HTML with `` / `<video>` into markdown (dropping media), or
2. Upload the whole HTML file via `create_attachment` instead of creating a document
The server already handles remote URL and base64 media via `documentImporter` → `ProsemirrorHelper.replaceImagesWithAttachments`. This change only updates the MCP tool schema so models know to pass the HTML through unchanged.

### DIFF
--- a/server/routes/mcp/index.ts
+++ b/server/routes/mcp/index.ts
@@ -67,6 +67,8 @@ Document and collection markdown support @mentions using the syntax: @[Display N
 
 Read images and attachments with the "fetch" tool by setting resource to "attachment" and passing either the attachment ID or an /api/attachments.redirect?id=... URL; the tool will return a signed URL for download.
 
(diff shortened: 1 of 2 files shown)


tommoor:
I would honestly move this mostly to the default instructions here:

https://github.com/outline/outline/blob/main/server/routes/mcp/index.ts#L64-L70

We can detail there that base64 encoded images are supported in all formats

Ali-ovo:
> I would honestly move this mostly to the default instructions here:
> 
> https://github.com/outline/outline/blob/main/server/routes/mcp/index.ts#L64-L70
> 
> We can detail there that base64 encoded images are supported in all formats

It's really a better approach

tommoor (approved):

PR #13133, opened 2026-07-25
chore: Audit and move inline actions to definitions

tommoor:
- Refactor inline actions into `/definitions` for reuse
- Added common action factories for dialogs and rename
- Consistent naming of factories with "Factory" suffix
- Less code, even with the new common action definitions


app/hooks/useUserMenuActions.tsx
```diff
@@ -1,215 +1,45 @@
...
-    await users.delete(targetUser);
-  }, [users, targetUser]);
-
-  const resendInvitation = React.useCallback(async () => {
-    if (!targetUser) {
-      return;
-    }
-    try {
-      await users.resendInvite(targetUser);
-      toast.success(t(`Invite was resent to ${targetUser.name}`));
```
tommoor, line 110: We also caught a broken translation interpolation here

PR #13195, opened 2026-07-29
fix: googletagmanager not included in CSP unless enabled through env

### LINKED ISSUE #13191: Google Analytics integration is blocked by CSP on self-hosted instances

### Current Behavior

On a self-hosted Outline 1.9.2 installation, enabling Google Analytics through **Settings → Integrations → Google Analytics** causes the client to request:

```text
https://www.googletagmanager.com/gtag/js?id=G-TEST123456
```

The request is blocked by the Content Security Policy because the generated `script-src` does not include `https://www.googletagmanager.com`:

```text
script-src https://outline.example.com outline.example.com/static/ 'nonce-<dynamic>'
```

The CSP is generated by Outline and contains a per-response nonce. Adding a second CSP header in a reverse proxy does not extend the first policy, and replacing Outline's header with a static policy would lose the dynamic nonce.

There are two separate Google Analytics configuration paths in the code:

- `server/middlewares/csp.ts` adds `www.googletagmanager.com` only when the server-level `GOOGLE_ANALYTICS_ID` environment variable is set.
- `app/components/Analytics.tsx` also loads `gtag.js` when a database-backed `IntegrationService.GoogleAnalytics` integration is active.
- `server/routes/app.ts` dynamically adds script sources for analytics integrations only when `integration.settings.instanceUrl` exists. The Google Analytics integration stores a `measurementId`, but no `instanceUrl`, so it never registers the Google Tag Manager script source.

As a workaround, setting `GOOGLE_ANALYTICS_ID=G-TEST123456` and disabling the UI integration works because the environment-variable code path updates the CSP. Keeping both enabled is undesirable because the same measurement ID can be configured twice in the client.

### Expected Behavior

When the Google Analytics integration is enabled, Outline should add `https://www.googletagmanager.com` to `script-src` while preserving the generated nonce.

When the integration is not enabled, the CSP should remain unchanged. The fix should not add `'unsafe-inline'` for scripts.

### Steps To Reproduce

1. Run self-hosted Outline 1.9.2 without `GOOGLE_ANALYTICS_ID`.
2. Configure a GA4 measurement ID under **Settings → Integrations → Google Analytics**.
3. Reload Outline and inspect the browser network/console.
4. Observe that `https://www.googletagmanager.com/gtag/js?id=...` is blocked by CSP.
5. Inspect the response's `Content-Security-Policy` header and note that the Google Tag Manager source is missing.

### Environment

```markdown
- **Outline**: 1.9.2 (self-hosted)
- **Browser**: Chrome
- **Reverse proxy**: Nginx Proxy Manager, passing Outline's CSP through unchanged
```

### Anything else?

The official documentation currently instructs self-hosted users to configure Google Analytics only through **Settings → Integrations → Google Analytics**, but that documented path is broken by the CSP issue above:

https://docs.getoutline.com/s/guide/doc/google-analytics-ifqw7tzomJ

Relevant code:

- CSP generation and environment-variable allowlist:
  https://github.com/outline/outline/blob/f8ccdf143c1deab7ac1fc76753608bd9bf9f6aaf/server/middlewares/csp.ts#L47-L117
- Integration-specific CSP source injection:
  https://github.com/outline/outline/blob/f8ccdf143c1deab7ac1fc76753608bd9bf9f6aaf/server/routes/app.ts#L101-L114
- GA4 script loading for both environment and integration settings:
  https://github.com/outline/outline/blob/f8ccdf143c1deab7ac1fc76753608bd9bf9f6aaf/app/components/Analytics.tsx#L44-L83

A minimal fix would make the dynamic analytics CSP path add `https://www.googletagmanager.com` whenever `IntegrationService.GoogleAnalytics` is active, with regression coverage for both the enabled and disabled cases. The documentation should either mention the environment-variable workaround until this is fixed or be verified/updated together with the code change.

tommoor:
Snuck some overdue refactors in here too, moving logic out of routes and into the model and helper class.

closes #13191

### DIFF
--- a/server/models/helpers/IntegrationHelper.ts
+++ b/server/models/helpers/IntegrationHelper.ts
@@ -1,0 +1,51 @@
+import { uniq } from "es-toolkit/compat";
+import { IntegrationService, type IntegrationType } from "@shared/types";
+import env from "@server/env";
+import type Integration from "@server/models/Integration";
+
+/**
+ * Helper class for working with integrations.
+ */
+export default class IntegrationHelper {
+  /**
+   * Returns the script sources that must be allowed by the Content Security
+   * Policy for the given analytics integrations to load.
+   *
+   * @param integrations The analytics integrations.
+   * @returns A list of CSP script sources.
+   */
+  public static getAnalyticsScriptSrc(
+    integrations: Integration<IntegrationType.Analytics>[] = []
+  ): string[] {
+    // The cloud policy already allows the third-party services that are
+    // available to hosted teams.
+    if (env.isCloudHosted) {
+      return [];
+    }
+
+    return uniq(
+      integrations.flatMap((integration) =>
+        IntegrationHelper.scriptSrcForIntegration(integration)
+      )
+    );
+  }
+
+  private static scriptSrcForIntegration(
+    integration: Integration<IntegrationType.Analytics>
+  ): string[] {
+    if (integration.service === IntegrationService.GoogleAnalytics) {
+      return ["www.googletagmanager.com", "www.google-analytics.com"];
+    }
+
+    // Self-hosted analytics services are loaded from the configured instance.
+    if (integration.settings?.instanceUrl) {
+      try {
+        return [new URL(integration.settings.instanceUrl).host];
+      } catch {
+        return [];
+      }
+    }
+
+    return [];
+  }
+}
--- a/server/middlewares/csp.ts
+++ b/server/middlewares/csp.ts
@@ -51,6 +51,17 @@ export function allowScriptSrc(ctx: Context, sources: string[]) {
 }
 
 /**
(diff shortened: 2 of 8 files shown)


server/models/helpers/IntegrationHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+      )
+    );
+  }
+
+  private static scriptSrcForIntegration(
+    integration: Integration<IntegrationType.Analytics>
+  ): string[] {
+    if (integration.service === IntegrationService.GoogleAnalytics) {
+      return ["www.googletagmanager.com", "www.google-analytics.com"];
+    }
```
tommoor, line 38: This is the 🍖 , as this integration does not have an "instanceUrl" it did not contribute to the csp string

PR #13201, opened 2026-07-30
fix: Copy as Markdown does not include unsaved changes

### LINKED ISSUE #11491: When editing, and not yet saving "Copy as Markdown" is outdated

### Current Behavior

I just have edited a document and not yet clicked on "Done editing", then I tried to "Copy as Markdown" and noticed that the clipboard actually does not contain my latest delete operation. I then clicked on "Done editing" and then copied again => Now the clipboard is consistent with the document.

### Expected Behavior

Copy as Markdown is exactly the same as current document state.

### Steps To Reproduce

described above

### Environment

```markdown
- Outline:
- Browser:
```

### Anything else?

_No response_

UditDewan:
Fixes #11491

### Problem

`copyDocumentAsMarkdown` calls `documents.export`, which serializes the copy of the document held by the server. The collaboration server persists on a delay, so anything typed in the last few seconds is missing from the clipboard until the document is saved.

A previous attempt (#11773) moved the serialization to the client but dropped the signed attachment urls that #10821 added, which is the tradeoff described in the issue thread. The approach here is the one suggested there: build the Markdown locally, then sign the attachments in a second pass through a purpose-built endpoint.

### Changes

**`attachments.signUrls`** — new endpoint that takes attachment ids and returns a map of id to signed url, with an optional `expiresIn`. Public-bucket attachments return their canonical url. Scoped to the acting user's team, the same rule `attachments.redirect` applies, since attachments are owned by the workspace rather than by individual documents.

**`attachmentsToSignedUrls`** in `app/utils/files.ts` — the client-side counterpart of the server's `TextHelper.attachmentsToSignedUrls`. Finds attachment redirect urls in the text (relative or absolute), signs them in one request and substitutes them. No request is made when the document has no attachments.

**Reaching the editor from an action** — `DocumentContext` already holds the editor for the document in view, so it gains a `getEditorData(documentId)` accessor. `DocumentContextProvider` moves from `AuthenticatedLayout` up to sit above `ActionContextProvider` in `app/index.tsx`, which lets the action context expose `getEditorData` to every action, including from the command bar. The action falls back to `document.data` when the document is not open in an editor, for example when copying from a sidebar context menu.

### Notes

- The client serializer writes CommonMark, the same dialect the editor already uses when copying a selection to the clipboard. The server export path is untouched and still emits Outline-flavoured Markdown so that exports round-trip on import.
- The document title is prepended with its emoji icon exactly as `DocumentHelper.toMarkdown` does, so the copied output keeps the shape it had before.
- "Copy as text" reads `document.data` and has the same staleness; left alone here to keep the change focused.

### Tests

- `attachments.signUrls`: signs an attachment, omits attachments from another team, requires authentication, rejects an empty id list.
- `attachmentsToSignedUrls`: skips the request with no attachments, replaces relative and absolute urls, leaves urls untouched when the attachment is not returned.

I could not run the server suite locally as it needs Postgres, so those four are unverified outside CI — the app tests, typecheck and lint all pass.

### DIFF
--- a/app/actions/definitions/documents.tsx
+++ b/app/actions/definitions/documents.tsx
@@ -40,12 +40,12 @@ import { toast } from "sonner";
 import { errToString } from "@shared/utils/error";
 import Icon from "@shared/components/Icon";
 import type { NavigationNode } from "@shared/types";
-import { ExportContentType } from "@shared/types";
+import { ExportContentType, IconType } from "@shared/types";
 import { isMobile } from "@shared/utils/browser";
 import { getEventFiles } from "@shared/utils/files";
+import { determineIconType } from "@shared/utils/icon";
 import { Week } from "@shared/utils/time";
 import type UserMembership from "~/models/UserMembership";
-import { client } from "~/utils/ApiClient";
 import DocumentDelete from "~/scenes/DocumentDelete";
 import { ProsemirrorHelper } from "~/models/helpers/ProsemirrorHelper";
 import DocumentPermanentDelete from "~/scenes/DocumentPermanentDelete";
@@ -71,6 +71,7 @@ import {
   TrashSection,
 } from "~/actions/sections";
 import { setPersistedState } from "~/hooks/usePersistedState";
+import { attachmentsToSignedUrls } from "~/utils/files";
 import history from "~/utils/history";
 import {
   documentHistoryPath,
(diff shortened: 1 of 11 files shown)


app/actions/definitions/documents.tsx
```diff
@@ -802,16 +803,26 @@ export const copyDocumentAsMarkdown = createAction({
...
     if (document) {
-      const res = await client.post("/documents.export", {
-        id: document.id,
-        signedUrls: Week.seconds, // 7 days (AWS S3 max for presigned URLs)
-      });
-      copy(res.data);
+      // Serialize from the editor rather than asking the server for the
+      // document, as the collaboration server persists on a delay and edits
+      // made in the last few seconds would otherwise be missing.
+      const data = getEditorData(document.id) ?? document.data;
```
tommoor, line 814: I think we can just use `document.data` and avoid all this new editor plumbing

PR #13208, opened 2026-07-31
fix: Remove associated notifications when access is lost

### LINKED ISSUE #13207: Notifications return documents the recipient can no longer read
### Summary

`notifications.list` and the `notifications.create` / `notifications.update` websocket events return a fully presented document — including `title` and `text` — for a document the recipient may no longer be able to read.

### Details

`Notification`'s `@DefaultScope` eagerly includes the `document` association (`server/models/Notification.ts`), and `presentNotification` serializes it in full:

```ts
document: notification.document
  ? await presentDocument(ctx, notification.document)
  : undefined,
```

`server/routes/api/notifications/notifications.ts` filters notifications only by `teamId` / `userId` — there is no authorization check on the embedded document, and no `policies` are returned alongside it.

On the client, `Notification.document` is a `@Relation`, so the nested payload is inserted into `DocumentsStore` via the relation setter (`app/models/decorators/Relation.ts`). Home's list getters render everything in that store, so a leaked document also surfaces on Home.

### Impact

A user who was notified about a document (mention, comment, update) and later lost access to it — removed from the collection or group, document membership revoked, or the document moved into a private collection — still receives the current title and text of that document whenever their notifications are loaded.

### Proposed fix

Revoke the notification when the underlying permission changes, rather than filtering at read time. Discussed as the preferred approach over authorizing in the presenter, since it also stops the notification from lingering in the UI.

### Related

Split out of the work on outline/outline#13205, where the leaked document appearing in Home > Popular was traced to several separate vectors. This one was deliberately left out of that fix.

tommoor:
closes outline/outline#13207

### DIFF
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -715,6 +715,10 @@ function useNotificationHandlers() {
       }
     );
 
+    socket.on("notifications.delete", (event: WebsocketEntityDeletedEvent) => {
+      notifications.remove(event.modelId);
+    });
+
     socket.on(
       "subscriptions.create",
       (event: PartialExcept<Subscription, "id">) => {
--- a/app/components/WebsocketProvider.tsx
+++ b/app/components/WebsocketProvider.tsx
@@ -792,6 +792,10 @@ function useNotificationHandlers() {
       }
     );
 
+    socket.on("notifications.delete", (event: WebsocketEntityDeletedEvent) => {
+      notifications.remove(event.modelId);
+    });
+
     socket.on(
       "subscriptions.create",
       (event: PartialExcept<Subscription, "id">) => {
--- a/plugins/webhooks/server/tasks/DeliverWebhookTask.ts
+++ b/plugins/webhooks/server/tasks/DeliverWebhookTask.ts
@@ -169,6 +169,7 @@ export default class DeliverWebhookTask extends BaseTask<Props> {
       case "authenticationProviders.update":
       case "notifications.create":
       case "notifications.update":
+      case "notifications.delete":
       case "access_requests.create":
         // Ignored
         return;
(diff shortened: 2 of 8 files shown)


pacocartones:
Took a careful read through since this is a nice security hardening and it has been sitting a few weeks — sharing what I found in case it is useful.

The proactive-revocation approach is the right call — destroying the notification when access is lost beats read-time filtering, since there's nothing left to leak or escape later. The event-driven design (`NotificationsRevokeProcessor` → `RevokeUserNotificationsTask`) reads clean.

**One real bug though — I think Copilot was right about the drafts case, and it runs deeper than it looks.** In `RevokeUserNotificationsTask.ts`:

```ts
const documents = await Document.scope({ method: ["withMembership", user.id] }).findAll({ ... });
```

`Document.scope(...)` still layers on the model's `@DefaultScope` (`publishedAt != null`), and `withMembership` adds to it rather than overriding. So notifications pointing at **drafts** (`publishedAt === null`) never show up in `documents`, never get marked unreadable, and never get revoked.

The subtle part: the `"should revoke notifications for a draft the user can no longer read"` test passes green — but for the wrong reason. The draft gets revoked via the *collection* path (collection `permission: null`, user not a member), not the document path. If that same draft lived in a readable collection and access were lost through `documents.remove_user`, the notification would survive. So the test is green by accident, not by coverage.

Fix that matches the model's own convention (`findByPk`/`findByIds` already do this):

```ts
Document.unscoped().scope({ method: ["withMembership", user.id] })
// or: Document.scope(["withDrafts", { method: ["withMembership", user.id] }])
```

**Smaller thing:** `documents.move` isn't in `applicableEvents` on `NotificationsRevokeProcessor`. Moving a doc to a collection the recipient can't read is another access-loss path that could strand a notification with a readable title/text. Might be worth adding, or confirming it's covered elsewhere.

Adding `notifications.delete` to `DeliverWebhookTask`'s ignore list looks correct. Didn't run the suite locally, so worth a sanity check on the fix, but the drafts scope issue seemed worth flagging before merge. Thanks for tackling this one @tommoor 🙏

PR #13210, opened 2026-07-31
Document databases: typed properties and saved views over documents

Plinsboorg:
Adds **document databases**: typed properties on documents, and saved views — table, board, list and gallery — over the documents that carry them.

It is opened for discussion rather than as a finished proposal. It follows discussion #8476, where several people asked whether an implementation existed. A running instance is linked below so the design can be tried rather than read about.

> **Implemented by Claude.** Every line of this branch was written by Claude, working from a spec and iterating against review and hands-on testing by @Plinsboorg. Flagging it plainly so you can weigh the code accordingly.

## The design in two decisions

**A row is an ordinary document.** Not a record with a description field — a real Outline document with the same editor, revisions, comments, search indexing and permissions. Rows keep their `collectionId`, which is what makes the rest cheap: search, sharing and access control needed no special cases.

**A database is a model owned by a collection**, not a kind of collection. A collection can hold any number of databases alongside its ordinary documents. An earlier iteration made a database *be* a collection and it was rebuilt — it forced every collection to be either a folder or a table, and pushed teams into creating top-level collections just to hold small tables.

The feature is behind `TeamPreference.DocumentDatabases`, **off by default**, and enforced server-side on every route.

## What's here

- **Ten property types** — text, number, select, multi-select, checkbox, date, url, person, relation, rollup. Values are keyed by property id, never by name, so renaming a column cannot orphan data.
- **Four view layouts** with per-view filters, multi-level sorts, column order, visibility and grouping. A database may hold any number of views.
- **Filters and sorts compiled to SQL** against the JSON properties column rather than applied client-side, so pagination and totals stay correct on large databases.
- **Column summaries** (count, filled, empty, unique, sum, avg, min, max) computed across every row matching the view, not only the loaded page.
- **Bidirectional relations** — declaring an inverse creates and maintains the mirror property on the target database — and **rollups** across them, computed at read time and never stored.
- **Archive and restore**, modelled on `collections.archive`/`restore`: archiving hides a database and its rows while keeping every property value, and restore matches rows on the archive's timestamp so separately-archived rows stay archived.
- **Embedded databases** in a document via `/database`.
- **Frontmatter → typed properties** when importing markdown into a collection holding exactly one database.
- Permissions delegate to the parent collection; there is no second permission model.

## Deliberately not built

- **Formulas.** A formula language is a large surface — parser, type system, evaluation model, client-vs-SQL execution. Rollups cover the common aggregate case.
- **Filtering or sorting by a rollup.** Rollup values are computed at read time and never stored, so there is no column for `WHERE`/`ORDER BY` to address. The code raises a clear validation error rather than returning nothing quietly.
- **Calendar and timeline layouts.** Nothing structural blocks them; they were cut to keep this reviewable.
- **Star, subscribe or share on a database.** Those models hold literal `documentId`/`collectionId` columns with no polymorphic target. Rows support all three, since rows are documents.

## Try it

A demo instance is running the branch: **https://outline-db-test.plinsburg.tech**

To let anyone look at this without an invite, the instance runs one extra change: the email sign-in path was modified so that entering any address creates an account and sends a magic link. **That change is not part of this PR** — it exists only so the feature can be tried, and it is not proposed for upstream.

The Welcome collection there documents the feature using the feature: the rows of the "Feature tour" database *are* its capabilities, including the gaps listed above.

## Notes for review

- 28 commits, kept as a readable phase-by-phase series and rebased on current `main` rather than merged.
- Migrations are reversible and verified in both directions; the schema additions are additive.
- Tests are collocated and cover the routes, models, policies and the shared property/validation helpers.
- Several bugs in here were found by using the instance rather than by the suite — private-collection permission scoping, rows created without a `databaseId`, and a row lock combined with a user-scoped document load. Each has a regression test.
- No new markdown files, per `CLAUDE.md`.

Happy to restructure, split this into smaller PRs, or take it in a different direction entirely — including "not like this". Feedback on the two design decisions at the top is the most useful thing.


Plinsboorg:
if you build it from source to test, make sure to enable the database feature in the settings. 
otherwise you'll see no difference from the stock behavior of the app
image

tommoor:
I don't see a path to merging this volume of AI code, but you're welcome to maintain it on your own branch. 

This is somewhat duplicative of the [Data Attributes](https://docs.getoutline.com/s/guide/doc/data-attributes-8nPrT7b6Qa?q=data) feature we have on the enterprise licensed edition - I'm currently figuring out how to bring pieces of that across to the community edition, and that will be the path to take here rather than a very notion-like database feature.

PR #13223, opened 2026-08-01
Add nightly docker builds

tommoor:
- Add nightly tagged docker builds
- Add `actionlint` CI job
- Refactor build logic into shared workflow


.github/workflows/docker-nightly.yml
```diff
@@ -0,0 +1,21 @@
...
+    secrets: inherit
+    with:
+      ref: main
+      environment: dockerhub-nightly
+      flavor: |
+        latest=false
+      tags: |
+        type=raw,value=nightly
+        type=raw,value=nightly-{{date 'YYYYMMDD'}}
+        type=sha,prefix=nightly-,format=short
```
rlenferink, line 21: Trying to think along:

The `nightly` tag will simply be overwritten every night and unused layers will be overwritten which does not create dangling images. It simply provides a pointer to the latest commit of the day. Previous layer will be garbage collected by Docker Hub (if they are no longer referenced).

The `nightly-<date>` and `nightly-<sha>` tags will also be created nightly, but as far as I can tell, they won't be removed. Will these `nightly-<date>` and `nightly-<sha>` tags become a problem storage wise? I am not sure on which plan Outline is, but I have had similar issues in the past. Will people really use these - or a better question: should these really be kept - because the most interesting thing is to point to the latest commit.

tommoor: We don't pay for anything, it's open source. 365 images a year seems fine - only a couple of mb of layers changes each day

.github/workflows/docker-nightly.yml
```diff
@@ -0,0 +1,21 @@
+name: Publish nightly build
+
+on:
+  schedule:
+    # 03:00 UTC daily
+    - cron: "0 3 * * *"
```
rlenferink, line 6: What would be the reason to do a nightly build, instead of a build after something changed on the main branch?

This allows users (like me) who reported issues, to quickly verify changes.

The only thing I can think of is compute costs for e.g. GitHub runners, but that's something I don't know (I use self-hosted runners for my own projects, so am not too up-to-date on GitHub action costs).

Suggestion:

```suggestion
  push:
    branches:
      - 'main'
```

In that case you probably also want to use something like `dev` as tag, instead of `nightly`.

tommoor: Every merge to main would be vastly more images and compute costs than nightly. Nightly is good enough for occasional testing needs.

.github/workflows/docker-nightly.yml
rlenferink: Thanks for confirming, then it indeed won't be an issue.

.github/workflows/docker-nightly.yml
rlenferink: > be vastly more images

If only e.g. `dev` is pushed (not `dev-<sha>`), than the amount of images will only be one (which is overwritten on every push).

> and compute costs

This is indeed something to keep in mind, hence my question :)
If e.g. sponsored runners are used, then this also won't be an issue and it is just a matter of preference:

* Having `nightly` tags, which includes the date/sha checksum.
  * Pro: easier to test against a given checksum.
  * Con: more images in the container registry, but this is not too big an issue, since only a couple of MB changes per day.
* Having e.g. a `dev` tag.
  * Pro: almost immediately after a merge a user can test/verify changes. Otherwise the user must wait till the nightly timer has triggered.
  * Pro: only 1 tag, so no storage usage on the Docker Hub side.
  * Con: no checksum in image tags. But IMO this is also not wat a `dev` tag is for.
  * Con: added build time for each merge/commit to `main`.

> Nightly is good enough for occasional testing needs.

I agree that nightly is already better than nothing, so thanks for the quick action!

PR #13248, opened 2026-08-02
fix: exclude soft-deleted documents from userMemberships.list pagination

### LINKED ISSUE #13238: The "Shared with me" sidebar section disappears **entirely** — no error, no empty state — when the first page of a user's memberships contains only soft-deleted documents.

### Current Behavior

The "Shared with me" sidebar section disappears **entirely** — no error, no empty
state — when the first page of a user's memberships contains only soft-deleted
documents.

The user can still open the shared documents by URL and via notifications, so the
permissions themselves are intact. Only the sidebar section is missing.

### Expected Behavior

"Shared with me" lists the documents actually shared with the user.

### Root cause

Two behaviours combine:

1. Soft-deleting a document does **not** remove its `user_permissions` rows.
2. `userMemberships.list` does **not** filter memberships pointing to
   soft-deleted documents — it returns them, but `Document.findByIds` omits the
   documents themselves, so the response carries memberships without their docs.

The sidebar fetches only the first page (`Pagination.sidebarLimit = 10`), ordered
by `"user_permission"."index" collate "C"`. Newly created shares receive a
*lower* fractional index, so stale memberships from an earlier batch monopolise
page 1. That page then side-loads **zero** documents.

`user.documentMemberships` filters on `documents.get(m.documentId)` being
present and active, so it computes to an empty array. The guard in
`SharedWithMe.tsx`

```js
if (!user.documentMemberships.length && !user.groupsWithDocumentMemberships.length) {
  return null;
}
```

hides the section — and because the section is hidden, the "show more" control
that would load page 2 (where the valid shares live) is never rendered. Deadlock
until the trash is emptied; permanent deletion *does* cascade the rows.

### Observed on a real instance

- `userMemberships.list` → **200**, 17 memberships, **4** documents, 21 policies
- 13 of the 17 point to soft-deleted documents and occupy page 1
- The 4 valid shares sit at positions 14–17
- Emptying the trash removed all stale rows and the section reappeared

### Steps To Reproduce

1. Share ≥10 documents directly with user U (`documents.add_user`)
2. Soft-delete those documents (move to trash — do **not** empty it)
3. Share 1+ new document with U
4. Log in as U

### Environment

```markdown
- Outline **1.9.2**, self-hosted (Docker)
- Reproduced on two independent instances
```

### Anything else?

- #10499 — same pagination root cause; closed 2025-11-08, but the fix PRs
  #10539 and #10588 show `merged: null`, and 1.9.2 still exhibits the behaviour.
- #7478 — hiding deleted documents client-side is what makes the wasted slots
  *invisible*: before it, they showed up as broken entries rather than silently
  consuming the page.

okxint:
The `userMemberships.list` handler fetches memberships without filtering out those that point to soft-deleted documents. The default Sequelize `findAll` without an explicit `include` does not join the `document` table at all, so Sequelize's paranoid `WHERE deletedAt IS NULL` guard never fires — stale memberships from deleted documents consume pagination slots as if the documents still exist.

When a user accumulates ≥10 stale memberships (the default page size), every row on the first page is orphaned, no valid memberships are returned, and the "Shared with me" sidebar disappears entirely even though there are active shares on later pages.

Adding `include: [{ association: "document", required: true }]` turns the join into an INNER JOIN. Sequelize's paranoid filter then automatically appends `AND "document"."deletedAt" IS NULL`, excluding stale rows before `LIMIT`/`OFFSET` is applied.

Fixes #13238

### DIFF
--- a/app/components/Sidebar/components/DocumentLink.tsx
+++ b/app/components/Sidebar/components/DocumentLink.tsx
@@ -331,14 +331,6 @@ const DocumentLinkInner = observer(function DocumentLinkInner({
       if (!moveCollectionId) {
         return;
       }
-      if (expansion.isExpanded(node.id)) {
-        return {
-          documentId: item.id,
-          collectionId: moveCollectionId,
-          parentDocumentId: node.id,
-          index: 0,
-        };
-      }
       return {
         documentId: item.id,
         collectionId: moveCollectionId,
--- a/server/queues/processors/DocumentMovedProcessor.ts
+++ b/server/queues/processors/DocumentMovedProcessor.ts
@@ -38,8 +38,8 @@ export default class DocumentMovedProcessor extends BaseProcessor {
             : [],
         ]);
 
-      await this.destroyUserMemberships(document.id);
-      await this.destroyGroupMemberships(document.id);
+      await this.destroyUserMemberships(document, transaction);
+      await this.destroyGroupMemberships(document, transaction);
 
       await this.recalculateUserMemberships(
         parentDocumentUserMemberships,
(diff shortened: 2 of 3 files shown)


tommoor:
Lots of unrelated work in here again?

closed by https://github.com/outline/outline/pull/13255

PR #13298, opened 2026-08-04
fix: Reduce peak memory duplicating large document trees, stabilize useActionContext identity

### LINKED ISSUE #13296: Memory regression: duplicating a large document tree loads all documents' content into memory at once
## Summary

#13270 ("fix: Remap internal links when duplicating a document tree") changed document duplication to a two-pass design: first assign new ids/urlIds to every document in the tree being duplicated (needed so cross-document links/mentions can be remapped to the copies), then write the duplicates.

The planning pass (`buildChildItems` in `server/commands/documentDuplicator.ts`) fetches every descendant document via `original.findChildDocuments(...)`, which uses the model's default scope — this includes the full `content`/`text` fields. Because the *entire* tree is walked and fetched before any duplicate is written, every original document's content is held in memory simultaneously for the whole tree, for the whole duration of the operation.

Previously, duplication was a single interleaved pass (fetch child → create its duplicate → recurse), so memory was bounded by the current recursion path (ancestors + not-yet-processed siblings along that path), not the entire tree.

## Impact

Duplicating a large document tree (e.g. thousands of documents, a large wiki/collection export) now holds roughly the full tree's content in memory at once, where before it was bounded by tree depth. This raises OOM risk on the API/worker process for large trees, where none existed before.

## Location

- `server/commands/documentDuplicator.ts`, `buildChildItems`/`buildItem` (structure-discovery pass) and `duplicateItem` (write pass).

## Suggested fix

Fetch documents in the planning pass without `content`/`text`/`state` (only what's needed to assign ids and preserve tree order), and re-fetch each document's full content immediately before it is written in `duplicateItem`. This keeps peak memory bounded to roughly one document's content at a time instead of the whole tree's, while preserving the upfront id-assignment needed for link remapping.

---

Found via an automated review of PRs merged in the last 24 hours. A targeted fix is included in the linked PR.

### LINKED ISSUE #13297: Performance regression: DropdownMenu re-evaluates all action visibility on every render, even while closed
## Summary

#13292 ("Dismissable invite and import onboarding links in sidebar") added an `isEmpty` check to `DropdownMenu` (`app/components/Menu/DropdownMenu.tsx`) so the trigger renders nothing when none of its actions are visible (e.g. hiding "Empty trash" for non-admins):

```ts
const isEmpty = useComputed(() => {
  const resolvedAction = typeof action === "function" ? action() : action;
  return !hasVisibleActions(
    resolvedAction.children as ActionVariant[],
    actionContext
  );
}, [action, actionContext]);
```

The intent (per the code comment) is that this only runs "while the menu is closed" — i.e. it should be cheap and only recompute when its inputs genuinely change, deferring the more expensive per-item resolution (`menuItems`) to when the menu is actually opened.

However, `actionContext` (from `useActionContext({ isMenu: true })`) does not have a stable identity — `useActionContext` returns a freshly spread object (`{...contextValue, ...overrides}`) on every call when overrides are passed, which they always are here. Since `actionContext` is in `isEmpty`'s dependency array, the underlying `useMemo`/mobx `computed()` wrapper is invalidated and rebuilt on *every* render of `DropdownMenu`, so `hasVisibleActions` — which resolves the `visible()` callback of every action, touching stores/policies — reruns on every render, regardless of whether the menu is open or closed.

The sibling `menuItems` computed has the same identity-churn problem, but avoids the cost via an explicit `if (!open) return [];` short-circuit. `isEmpty` has no such guard, so it pays this cost unconditionally.

## Impact

`DropdownMenu` is used in ~28 places across the app (e.g. `DocumentMenu`, `TemplateMenu`, `CollectionMenu`), often once per row in a list. Any parent re-render (typing in search, hover state, an unrelated MobX store update) now re-evaluates the visibility of every action for every closed `DropdownMenu` on the page — work that did not happen at all before this menu had any items rendered/opened.

## Location

- `app/components/Menu/DropdownMenu.tsx` (`isEmpty`, added in #13292)
- `app/hooks/useActionContext.tsx` (`useActionContext`, root cause of the unstable reference)

## Suggested fix

Stabilize the object returned by `useActionContext` across renders when its shallow contents haven't changed, so consumers that memoize on it (`useComputed`/`useMemo`) get real caching back.

---

Found via an automated review of PRs merged in the last 24 hours. A targeted fix is included in the linked PR.

tommoor:
## Summary

Found via an automated review of PRs merged in the last 24 hours. Two medium-severity regressions, each fixed here:

- **Memory regression in document tree duplication** (introduced by #13270). `documentDuplicator`'s planning pass now assigns ids across the *entire* tree before writing any duplicates (needed to remap cross-document links to the copies), but it eagerly fetched every descendant's full `content`/`text` during that pass — holding the whole tree's content in memory simultaneously for large trees, where before duplication was interleaved (fetch → write → recurse) and memory was bounded by the current recursion path. Fixes #13296.

  Fix: fetch documents in the planning pass without `content`/`text`/`state` (only what's needed to assign ids and preserve order), and re-fetch each document's full content immediately before it's written in `duplicateItem`. Peak memory is now bounded to roughly one document's content at a time. The upfront id-assignment (and therefore link remapping) behavior is unchanged.

- **Performance regression in `DropdownMenu`** (introduced by #13292). The new `isEmpty` check (added so the trigger hides itself when none of its actions are visible) depends on `actionContext` from `useActionContext({ isMenu: true })`. That hook returns a freshly spread object on every render whenever overrides are passed, which they always are here — this breaks the `useComputed` memoization `isEmpty` relies on, so `hasVisibleActions` (which resolves every action's `visible()` callback against stores/policies) reruns on *every* render of *every* `DropdownMenu` on the page, even while closed. `DropdownMenu` is used in ~28 places, often once per list row, so this is paid on unrelated re-renders (typing in search, hover, any MobX store update). Fixes #13297.

  Fix: `useActionContext` now returns a referentially stable object across renders as long as its shallow contents are unchanged (`useStableValue`, backed by an exported, unit-tested `shallowEqual` helper), restoring memoization for `isEmpty` and the neighboring `menuItems` computed.

## Test plan

- [x] `yarn tsc --noEmit` — passes
- [x] `yarn lint` (oxlint, type-aware) on changed files — clean
- [x] `yarn test server/commands/documentDuplicator.test.ts` — 12/12 passing, including recursive children, drafts, link/mention remapping, and fullWidth propagation, confirming the lighter planning-pass query and lazy content re-fetch don't change duplication behavior
- [x] `yarn test server/queues/tasks/DuplicateCollectionDocumentsTask.test.ts` — 4/4 passing
- [x] `yarn test app/hooks/useActionContext.test.ts` — new tests for the extracted `shallowEqual` helper, 7/7 passing

---
_Generated by [Claude Code](https://claude.ai/code/session_017HG8YoPALKcHj56vvDhDm2)_


server/commands/documentDuplicator.ts
```diff
@@ -48,6 +48,14 @@ type Root = {
...
   /** The document being duplicated */
   original: Document;
+  /**
+   * Whether `original` has its content loaded. Child documents are fetched
+   * without content while identifiers are being assigned across the whole
+   * tree, so their content is re-fetched immediately before it is written –
+   * this keeps peak memory bounded to one document's content at a time
+   * rather than the entire tree's.
+   */
+  hasContent: boolean;
```
tommoor: Instead of piping this through just check if "content" is set

app/hooks/useActionContext.tsx
```diff
@@ -216,6 +222,60 @@ export const ActionContextProvider = observer(function ActionContextProvider_({
...
 
+/**
+ * Compares two objects' own enumerable properties with `Object.is`, one
+ * level deep.
+ *
+ * @param a the first object.
+ * @param b the second object.
+ * @returns true if `a` and `b` have the same keys with the same values.
+ */
+export function shallowEqual(
```
tommoor: Use shallow equal from es-compat?

server/commands/documentDuplicator.ts
```diff
@@ -165,6 +175,11 @@ async function duplicateRoots(
     original: Document,
     originalCollection: Collection | null
   ) {
+    // Identifiers are assigned across the whole tree before any content is
+    // written, so this pass only needs each document's structural fields –
+    // fetching content/text here as well would hold every document's full
+    // content in memory at once for the entire duration of a large tree
+    // duplication. It is re-fetched per document in `duplicateItem` instead.
```
tommoor: Reduce comment size

server/commands/documentDuplicator.ts
```diff
@@ -194,26 +214,36 @@ async function duplicateRoots(
     item: DuplicateItem,
     options: { parentDocumentId?: string; publish: boolean }
   ) {
+    // Content was deliberately left unfetched for child documents while
+    // building the tree above (see `buildChildItems`) – load it now, right
+    // before it's needed, so at most one document's content is held at a
+    // time instead of the whole tree's.
```
tommoor: Remove comment

app/hooks/useActionContext.tsx
```diff
@@ -216,6 +223,48 @@ export const ActionContextProvider = observer(function ActionContextProvider_({
...
```
tommoor: Remove shallowEqual and tests, just inline use of isEqualWith

PR #13299, opened 2026-08-04
fix(editor): re-measure code block toolbar when block resizes

### LINKED ISSUE #13282: Code block toolbar is vertically misaligned after navigating between pages

### Current Behavior

The code block toolbar (Copy, Wrap, and Language selector) is occasionally displayed in the vertical center of the code block after navigating between documents.

The issue occurs intermittently. The toolbar initially appears correctly at the top-right corner, but after switching between documents, it may move to the vertical center of the code block.

Refreshing the page usually restores the toolbar to the correct position.

Image

### Expected Behavior

The code block toolbar should always remain positioned at the top-right corner of the code block, regardless of document navigation.

### Steps To Reproduce

1. Open a document containing a long code block.
2. Navigate to another document using the Outline sidebar.
3. Navigate back to the original document.
4. Repeat navigation between documents several times.
5. Observe that the code block toolbar occasionally appears in the vertical center instead of the top-right corner.

### Environment

```markdown
- Outline: 1.9.2
- Deployment: Self-hosted
- Browser: Mozilla Firefox
- Operating system: Linux
```

### Anything else?

_No response_

okxint:
## Problem

The code block toolbar (Copy, Wrap, Language selector) drifts to the vertical center of the code block instead of the top-right corner after navigating between documents. Refreshing the page restores the correct position.

**Root cause:** `StickyBlockToolbar` stores the block's bounding rect and uses it to position the toolbar overlay. The measurement only re-ran when React triggered a re-render — on window resize or ProseMirror's `forceUpdate()`. When the block's DOM height changed after initial mount without a React re-render (e.g. a collapsed block auto-expanding asynchronously, or a custom font finishing loading and shifting line metrics), the stale rect became permanent until a full page reload.

## Fix

Add a `ResizeObserver` on the block element. Any dimension change fires `forceUpdate` (a `useReducer` dispatch with a stable identity), triggering a re-render and a fresh measurement via the existing `useLayoutEffect`. The observer is cleaned up when the element changes or the component unmounts.

```tsx
const [, forceUpdate] = React.useReducer((x: number) => x + 1, 0);

React.useEffect(() => {
  if (!element) return;
  const observer = new ResizeObserver(forceUpdate);
  observer.observe(element);
  return () => observer.disconnect();
}, [element, forceUpdate]);
```

Fixes #13282

### DIFF
--- a/app/editor/components/StickyBlockToolbar.tsx
+++ b/app/editor/components/StickyBlockToolbar.tsx
@@ -88,11 +88,23 @@ const StickyBlockToolbar = React.forwardRef(function StickyBlockToolbar_(
   const { view } = useEditor();
   const trackRef = ref || React.createRef<HTMLDivElement>();
   const [rect, setRect] = React.useState<TrackRect | null>(null);
+  const [, forceUpdate] = React.useReducer((x: number) => x + 1, 0);
 
   // Re-measure when the window resizes; scroll is handled by `position: sticky`.
   useWindowSize();
 
   const element = getBlockElement(view);
+
+  // Re-measure when the block's size changes (e.g. after an auto-expand that
+  // settles after the initial mount, or a font load that shifts layout).
+  React.useEffect(() => {
+    if (!element) {
+      return;
+    }
+    const observer = new ResizeObserver(forceUpdate);
+    observer.observe(element);
+    return () => observer.disconnect();
+  }, [element, forceUpdate]);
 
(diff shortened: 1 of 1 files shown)


tommoor (approved):

tommoor:
Waiting on CLA

okxint:
> Waiting on CLA

/check-cla

PR #13302, opened 2026-08-04
feat: PDF import support

tommoor:
Add support for basic import of PDF files (no images), and rejiggers the import dialog to continue to work with more file formats now that the list is getting longer.

---

getoutline outline dev_3000_doc_coated-paper-development-roadmap-1qJeqpjZDJ


server/converters/PdfConverter.ts
```diff
@@ -0,0 +1,101 @@
...
+   * @param markdown The markdown emitted by the parser.
+   * @returns The markdown with underline tags replaced.
+   */
+  private static replaceUnderlineTags(markdown: string): string {
+    return (
+      markdown
+        .replace(
+          // Underlines do not span lines, so a tag left open by the end of one
+          // is unpaired and falls through to be stripped below.
+          /<u>([^\n]*?)<\/u>/gi,
```
tommoor, line 64: There may be other HTML tags output, but this is the only one I could find

PR #13327, opened 2026-08-06
feat: Allow attributing documents.create to an existing user

nickmeinhold:
Adds an optional `createdById` to `documents.create` so an admin can attribute a new document to another user in their team.

Comes out of the discussion in https://github.com/outline/outline/discussions/12963, where @tommoor said the audit log must stay uncorruptible and later asked for the permission checks to go through the policies rather than hardcoded ones. Both of those are addressed below.

## Permissions

Both checks go through the policy layer as you suggested:

- `authorize(user, "update", user.team)` for the admin gate, which resolves to `isTeamAdmin`.
- `cannot(author, "read", collection)` for whether the target can be assigned as the creator.

I read `can(user, "read", ...)` as constraining the **prospective author** rather than the acting admin, and used the **collection** since no document exists yet at create time. If either reading is wrong, say so and I'll change it.

The admin gate fires on the mere presence of `createdById`, so supplying an author is admin-only even when it is the caller's own id. It sits above `authorizeDocumentCreate` so a non-admin is rejected before any user lookup, which keeps the endpoint from being used to probe which user ids exist. Author resolution sits below it because it needs the resolved collection.

The collection is re-loaded in the author's scope before the policy check. `allow(User, "read", Collection, ...)` consults memberships preloaded by `withMembership(userId)`, so testing the author against the acting admin's instance would evaluate the wrong user's memberships on a private collection. There's a test covering group-only access to confirm the reload populates `groupMemberships` too.

## Audit log

`ctx.state.auth` is never touched, so `Event.createFromContext` still records the real API caller in `actorId`. Only `document.createdById` moves. There's a test asserting the event actor is the admin and not the attributed author, so it can't regress silently.

## One thing worth flagging

I described this in the discussion as parity with the JSON importer's author-mapping, and having now read that code I don't think that was accurate. The importer sets `sourceMetadata.createdByName`, a display-only string surfaced by the presenter, while the document's real `createdById` stays the importer. This change assigns a real user id, which confers actual ownership rather than just a displayed name.

That's a stronger power than the precedent I cited, so it may deserve a stronger check. `read` answers "could this person see the result"; `createDocument` / `createChildDocument` would answer "could this person have created it here". I've implemented `read` because that's what you named, but I think the case for the stronger check is real and I'd rather raise it than quietly pick.

Related: `documentCreator` is also reached from the MCP `create_document` tool, the import task, and the duplicator. None of them pass `createdById` today, so there's no bypass, but the gate does live at the route rather than in the command.

## Fails closed

Rejects an unresolved id, a cross-team user, a suspended user, and an author who can't read the destination collection. No path falls back to the acting user.

## Tests

`documents.test.ts` 281/281 pass, `tsc` and lint clean. The negative cases assert no document was persisted, not just the status code.

### DIFF
--- a/server/routes/api/documents/documents.ts
+++ b/server/routes/api/documents/documents.ts
@@ -1719,16 +1719,65 @@ router.post(
       fullWidth,
       templateId,
       createdAt,
+      createdById,
     } = ctx.input.body;
     const editorVersion = ctx.headers["x-editor-version"] as string | undefined;
 
     const { transaction } = ctx.state;
     const { user } = ctx.state.auth;
 
+    // Attributing a document to another user is an admin-only capability. The
+    // gate is on the mere presence of createdById — supplying an author is
+    // admin-only, even when it is the acting user's own id — so the code
+    // enforces exactly the contract the schema documents. It runs before
+    // collection resolution so a non-admin is rejected without probing user
+    // existence (no oracle) or touching unrelated state.
+    if (createdById) {
+      authorize(user, "update", user.team);
+    }
+
     const { collection } = await authorizeDocumentCreate(ctx, {
       collectionId,
       parentDocumentId,
     });
+
+    // Resolve the author once the destination is known. The target must belong
+    // to the same team and must themselves be able to read the collection the
+    // document will live in, so a document cannot be attributed to someone who
+    // could not see it. This mirrors the JSON importer's author-mapping
+    // (#11879) but fails closed on an unresolved, cross-team or unauthorized
+    // target rather than silently falling back to the acting user, since an
+    // explicit API argument asserts intent.
+    if (createdById) {
+      const author = await User.findByPk(createdById, { transaction });
+      if (!author || author.teamId !== user.teamId) {
+        throw ValidationError(
+          "createdById must reference a user in the same team"
+        );
+      }
+      if (author.isSuspended) {
+        throw ValidationError("createdById must reference an active user");
+      }
+      // A draft with no collection has no collection ACL to consult, so only
+      // the team and active-user checks above apply. This is intentional: an
+      // admin may place a draft in another member's drafts. Do not add a
+      // soft fallback here — there is nothing to authorize against.
+      if (collection) {
+        // Re-load the collection in the author's scope. The collection read
+        // policy consults memberships preloaded by withMembership(userId), so
+        // testing the author against the acting admin's instance would
+        // evaluate the wrong user's memberships on a private collection.
+        const authorCollection = await Collection.findByPk(collection.id, {
+          userId: author.id,
+          transaction,
+        });
+        if (cannot(author, "read", authorCollection)) {
+          throw ValidationError(
+            "createdById must reference a user who can access the collection"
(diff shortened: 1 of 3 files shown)


server/routes/api/documents/documents.ts
```diff
@@ -1719,17 +1719,66 @@ router.post(
...
+    if (createdById) {
+      const author = await User.findByPk(createdById, { transaction });
+      if (!author || author.teamId !== user.teamId) {
+        throw ValidationError(
+          "createdById must reference a user in the same team"
+        );
+      }
+      if (author.isSuspended) {
+        throw ValidationError("createdById must reference an active user");
+      }
```
tommoor: Replace with authorize(user, "read", author);

server/routes/api/documents/documents.ts
```diff
@@ -1719,17 +1719,66 @@ router.post(
...
+      // A draft with no collection has no collection ACL to consult, so only
+      // the team and active-user checks above apply. This is intentional: an
+      // admin may place a draft in another member's drafts. Do not add a
+      // soft fallback here — there is nothing to authorize against.
+      if (collection) {
```
tommoor: We should ensure that createdById can only be passed with a collection or you could create a draft for another user

server/routes/api/documents/documents.ts
```diff
@@ -1719,17 +1719,66 @@ router.post(
...
     const { user } = ctx.state.auth;
 
+    // Attributing a document to another user is an admin-only capability. The
+    // gate is on the mere presence of createdById — supplying an author is
+    // admin-only, even when it is the acting user's own id — so the code
+    // enforces exactly the contract the schema documents. It runs before
+    // collection resolution so a non-admin is rejected without probing user
+    // existence (no oracle) or touching unrelated state.
+    if (createdById) {
+      authorize(user, "update", user.team);
```
tommoor: Move down inside the existing if block

server/routes/api/documents/documents.ts
```diff
@@ -1719,17 +1719,66 @@ router.post(
...
       templateId,
       createdAt,
+      createdById,
     } = ctx.input.body;
     const editorVersion = ctx.headers["x-editor-version"] as string | undefined;
 
     const { transaction } = ctx.state;
     const { user } = ctx.state.auth;
 
+    // Attributing a document to another user is an admin-only capability. The
```
tommoor: de slop these comments, most are not neccessary

tommoor (changes requested):

nickmeinhold:
All four addressed, plus one tightening.

- `authorize(user, "read", author)` replaces the hand-rolled team check, so those rejections are 403 now and the not-found case is covered.
- `createdById` requires `collectionId` and `publish`. I added the publish half because a collection alone still allows an attributed draft — happy to drop it back to collection only.
- Admin check moved inside the block, above the author lookup.
- Comments trimmed.

Also dropped a suspended-user check I'd added, and cut the tests from 13 to 7. The collection is re-loaded in the author's scope before the read check.

PR #13337, opened 2026-08-07
fix: Restore embeds when parsing markdown

impuls42:
### Summary

`Embed.toMarkdown` serialises an embed node as `[href](href)`, but nothing parses that back into an embed node, so the round-trip is lossy. `Embed.parseMarkdown()` already exists and expects a token of type `embed` — no markdown-it rule in `shared/editor/rules/` ever produces that token, so the method is currently unreachable.

Export a document and re-import it, or create one through the API, and every embed becomes an ordinary link. The API path is the more visible symptom: there is no way to create a document containing an embed programmatically.

Interestingly this is half-wired already — `documents.create` calls `convertBareUrlsToEmbedMarkdown`, whose docstring says it rewrites bare URLs to `[url](url)` *"so the markdown parser can recognize them as embeds"*. Without the rule below, that preprocessing has nothing to hand off to.

### Changes

A markdown-it core rule (`shared/editor/rules/embeds.ts`) that converts a paragraph containing exactly one self-linking, embeddable URL into an `embed` token, registered through a `rulePlugins` getter on the Embed node. `Embed.parseMarkdown()` already reads `token.attrGet("href")`, so it is unchanged.

Two details that may not be obvious from the diff:

**The match has to skip `matchOnInput: false` descriptors.** `getMatchingEmbed` returns a match for *every* `http(s)` URL, because the generic iframe embed is last in the list and matches `^https?://(.*)$`. Gating the rule on it directly would turn every self-linking URL in every existing document into an embed. `server/utils/embeds.ts` already had the right predicate for bare URLs; it is lifted to `shared/editor/lib/embeds.ts` as `getMatchingEmbedOnInput` and both callers now share it.

**The rule is pushed to the end of the core chain, not anchored after `attachments`.** `Embed` is registered before `Attachment` in `richExtensions`, so `md.core.ruler.after("attachments", …)` throws `Parser rule not found: attachments`. Pushing runs it last — still after the attachments rule, which inserts itself mid-chain after `breaks` — and keeps the ordering stable if the extension list is ever reordered.

The serializer writes `_` as `%5F`; the rule reverses that when setting the attribute, so hrefs containing underscores round-trip unchanged.

### Tests

`shared/editor/rules/embeds.test.ts`, matching the style of the existing rule tests:

- a paragraph holding a single self-linking URL becomes an `embed`
- link text differing from the href stays a link
- two links in one paragraph stay links
- a link inside a longer sentence stays a link
- a URL matching no descriptor stays a link
- a URL only the generic embed matches stays a link
- an href containing `_` survives the `%5F` escape
- a self-linking URL in a table cell converts, nested in the `td`
- full round-trip: build a doc containing an embed, serialise, parse, assert it is still an `embed` with the same href

The last one is the assertion that fails today.

One deviation worth flagging: table cells **do** convert. Outline's table rule wraps cell content in paragraphs, so the rule sees `paragraph_open` there, and the result is a valid document with the embed inside the `td`. Since `Embed.toMarkdown` has an `inTable` branch that produces exactly this markdown, parsing it back makes tables lossless too. Happy to suppress it if you would rather keep the change narrower.

Verified no regressions: `yarn test:shared` 1193 passed, `yarn test:app` 167 passed, `server/utils/embeds.test.ts` 32 passed. Removing only the `rulePlugins` wiring fails 10 of the 20 new tests, so they are not vacuous.

### DIFF
--- a/server/utils/embeds.ts
+++ b/server/utils/embeds.ts
@@ -1,5 +1,8 @@
 import type { EmbedDescriptor } from "@shared/editor/embeds";
-import { getMatchingEmbed } from "@shared/editor/lib/embeds";
+import {
+  getMatchingEmbed,
+  getMatchingEmbedOnInput,
+} from "@shared/editor/lib/embeds";
 import embeds from "@shared/editor/embeds";
 import fetch, { chromeUserAgent } from "./fetch";
 import { Second } from "@shared/utils/time";
@@ -188,15 +191,7 @@ export async function checkEmbeddability(
  * @returns True if the URL matches an embed pattern with `matchOnInput` enabled.
  */
 function isEmbedUrl(url: string, embedDescriptors: EmbedDescriptor[]): boolean {
-  for (const embed of embedDescriptors) {
-    if (!embed.matchOnInput) {
-      continue;
-    }
-    if (embed.matcher(url)) {
-      return true;
-    }
-  }
-  return false;
+  return !!getMatchingEmbedOnInput(embedDescriptors, url);
 }
 
 /**
--- a/shared/editor/lib/embeds.ts
+++ b/shared/editor/lib/embeds.ts
@@ -16,6 +16,28 @@ export function getMatchingEmbed(
   return undefined;
 }
 
+/**
+ * Find the embed that should claim a URL the user did not explicitly ask to
+ * embed — one pasted into the editor, or parsed out of markdown.
+ *
+ * Descriptors with `matchOnInput` disabled are skipped. The generic iframe
+ * embed is one of them and matches every http(s) URL, so including it would
+ * make any URL look like a match.
+ *
+ * @param embeds the descriptors to search.
+ * @param href the URL to match.
+ * @returns the matching descriptor and its regex matches, or undefined.
+ */
+export function getMatchingEmbedOnInput(
+  embeds: EmbedDescriptor[],
+  href: string
+): { embed: EmbedDescriptor; matches: RegExpMatchArray } | undefined {
+  return getMatchingEmbed(
+    embeds.filter((embed) => embed.matchOnInput),
+    href
+  );
+}
+
 export function transformListToEmbeds(listNode: Node, schema: Schema): Node[] {
   const nodes: Node[] = [];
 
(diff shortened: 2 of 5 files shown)


myi1:
Thanks for this. It fixes a real problem for us. We run a self-hosted 1.10.1 wiki where Claude edits pages over MCP, and every text update (`replace`, `append`, and even `update_document` with `editMode: "patch"` on an unrelated block) turned our YouTube embeds into plain links. That's 32 video pages.

We've applied this PR on top of `v1.10.1` (`git apply` is clean) and are running it in production. On a draft, `patch`, `replace` and `append` through MCP now all keep the embed.

One gap we hit: **autolinks with an escaped underscore stay links.** For documents created through `documents.import`, `documents.info` and MCP `fetch` return the embed line as an autolink:

```md
<https://www.youtube.com/watch?v=0oV%5F%5F5LsnsU>
```

So a client that fetches a page and sends it back with `replace` sends that form. markdown-it runs `normalizeLinkText` on autolink text, which percent-decodes it, but it leaves the href as `…0oV%5F%5F5LsnsU`. The `text.content !== href` check in `selfLinkingHref` then fails, and the paragraph stays a link. Any YouTube id containing `_` is affected; that's 3 of our 32 videos.

The change we're running, in `shared/editor/rules/embeds.ts`:

```ts
const href = open.attrGet("href");
if (!href) {
  return undefined;
}

// An autolink (<url>) has its text percent-decoded by markdown-it while the
// href keeps the %5F escapes, so accept a text matching either form.
const url = unescapeHref(href);
if (text.content !== href && text.content !== url) {
  return undefined;
}

return url;
```

A test case that would cover it:

```ts
it("converts an autolink whose href contains an escaped underscore", () => {
  const doc = parser.parse("<https://www.youtube.com/watch?v=0oV%5F%5F5LsnsU>");
  expect(doc.firstChild?.type.name).toBe("embed");
  expect(doc.firstChild?.attrs.href).toBe(
    "https://www.youtube.com/watch?v=0oV__5LsnsU"
  );
});
```

I checked this with the compiled server parser in a patched 1.10.1 image, not with the repo's test suite. Stock 1.10.1 turns all three forms into links (`[url](url)` with `%5F`, `<url>` with `%5F`, and `<url>` without). With the PR plus this change, all three become embeds, and a serialize → parse round trip keeps the embed. Links inside a sentence, named links (`[the video](url)`), and non-embeddable URLs alone on a line all stay links.

Happy to open a follow-up PR with the test if that's easier than folding it in here.

PR #13364, opened 2026-08-10
Normalize document export options

tommoor:
Removes the dialog, adds a submenu that includes "Print" (removes one top-level item).
Ensure that revisions have the same options as documents


gilleard:
I'll maybe open a separate ticket but wanted to check if this was intended (or something else recently) to remove the ability to Download as a zip **with child documents**?

I may be missing something but that seems to have been dropped in the last week or so.

It's something we were actively using, wanted to first check if that was intended or not?

Thanks

tommoor:
My mistake, fix going in here - https://github.com/outline/outline/pull/13465

PR #13396, opened 2026-08-12
Add container image badges to README

rlenferink:
I think it might be a good idea to list the latest available version in the README, so hereby my proposal to add two badges for it:
* One to point to the latest stable release
* One to point to the latest nightly snapshot

**NOTE**: I am no text-writer, so I tried to keep it as clear as possible ;)


rlenferink:
@tommoor do you have any feedback on this?

tommoor (approved):
Seems fine, I might move them elsewhere but no harm here

PR #13419, opened 2026-08-13
fix: Match base domain on a label boundary when parsing hosts

tommoor:
- Host parsing looked for the base domain anywhere in the string, so a host that merely contained the install's domain was classified as one of its subdomains rather than as a custom domain.
- The base domain is now matched against the end of the host, on a label boundary.
- The custom domain CNAME check reuses `parseDomain` instead of keeping its own comparison.

### DIFF
--- a/server/routes/api/urls/urls.ts
+++ b/server/routes/api/urls/urls.ts
@@ -3,7 +3,7 @@ import Router from "koa-router";
 import { traceFunction } from "@server/logging/tracing";
 import isUUID from "validator/lib/isUUID";
 import { MentionType, UnfurlResourceType } from "@shared/types";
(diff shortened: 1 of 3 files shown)


server/routes/api/urls/urls.ts
```diff
@@ -292,14 +292,13 @@ router.post(
...
+    if (!address) {
       throw ValidationError("No CNAME record found");
     }
 
-    const address = addresses[0];
-    const likelyValid = address.endsWith(getBaseDomain());
-
-    if (!likelyValid) {
+    // the record must point at the base domain, or a subdomain of it
+    if (parseDomain(address).custom) {
```
tommoor, line 301: Unused, but fixed while we're here as it should be used soon

PR #13425, opened 2026-08-14
fix(editor): preserve toggle fold state across remote and local document edits

### LINKED ISSUE #13424: Toggle blocks re-collapse on their own after remote collaborative updates
### Describe the bug

When reading or editing a document, an expanded toggle block sometimes collapses back on its own — typically right after another user edits the document (it can also happen after the user's own edit). The user has to keep re-expanding it. This also affects read-only viewers.

### To Reproduce

1. Open the same document in two browsers (A and B), with multiplayer active.
2. In browser A, expand a toggle block that was previously collapsed.
3. Make localStorage writes fail in A (any of: fill the ~5MB quota, block cookies/site data, or override `Storage.prototype.setItem` to throw in DevTools). This is the real-world trigger — see root cause.
4. In browser B, type any character in the document.
5. The toggle block in A collapses on its own. (Step 4 can be replaced by typing in A itself — a local edit triggers it too.)

### Expected behavior

A document edit — remote or local — should never change the fold state of toggle blocks the user has already interacted with. Fold state is per-user/local by design (it lives in localStorage, not in the document).

### Root cause

`shared/editor/nodes/ToggleBlock.ts`, fold plugin `apply`:

```ts
if (isRemoteTransaction(tr)) {
  const foldedIds = this.initFoldedIds(newState);
  ...
}
```

with:

```ts
private initFoldedIds(state: EditorState) {
  const pluginState = toggleFoldPluginKey.getState(state);
  const foldedIds = new Set<string>(pluginState?.foldedIds);
  ...
  const stored = Storage.get(toggleStorageKey(id));
  if (stored?.fold !== false) {
    foldedIds.add(id);
  }
```

Two things combine here:

1. **The entire fold set is re-derived from storage on every remote transaction.** The `new Set(pluginState?.foldedIds)` seed looks like it preserves existing state, but during the plugin's own `apply`, `toggleFoldPluginKey.getState(newState)` reads the plugin's own state slot before ProseMirror assigns it — it is always `undefined`, so the seed is effectively dead code and every block's fold state is rebuilt purely from localStorage on each collaborative update. The local doc-changing branch has the same disease: any id not currently in `foldedIds` is re-checked against storage and re-added whenever the stored value isn't an explicit `fold: false`.

2. **Storage writes can fail silently.** Unfolding persists via `Storage.set(...)`, but `Storage` (`shared/utils/Storage.ts`) swallows all errors, including `QuotaExceededError` and denied storage. When the write fails, a stale `{ fold: true }` (or no entry, which defaults to folded) remains, so the next edit transaction — the user's own or a collaborator's — re-folds the block the user just expanded.

An amplifier: `toggle:<uuid>` keys are written for every toggle block ever seen and never removed when blocks or documents are deleted, so localStorage gradually fills toward the ~5MB quota, making silent write failures increasingly likely over time.

### Proposed solution

Make in-memory fold state authoritative and only *initialize* fold state from storage for toggle blocks seen for the first time:

- Track a `knownIds` set alongside `foldedIds` in the plugin state.
- On `init`, collect both sets from the document.
- On remote and local doc-changing transactions, reconcile using the plugin's previous state (the second `apply` argument, which is reliable — unlike `getState(newState)`): for each toggle block in the new doc, if its id is already known, preserve its current fold decision verbatim; only unknown ids get initialized from storage (default folded) — this still handles new blocks arriving via sync.
- `FOLD`/`UNFOLD` handlers keep updating `foldedIds` and writing to storage, so storage becomes a cross-session cache rather than the source of truth.

Optionally, prune `toggle:<id>` keys for ids no longer present in any open document to bound storage growth.

### Note

There is also an intentional auto-fold of empty-body toggles (any doc-changing transaction folds a toggle whose body is empty) — that is separate, by-design behavior and not what this issue is about.

I have a patch implementing the above and can open a PR if the approach sounds good.

chenniannian90:
Fixes #13424

### Problem

Toggle blocks re-collapse on their own after a document edit — remote or local. See #13424 for full analysis. In short:

- The fold plugin re-derives fold state from localStorage on every remote collaborative transaction, and partially on local doc-changing ones.
- The `new Set(pluginState?.foldedIds)` seed in `initFoldedIds` never takes effect there, because `toggleFoldPluginKey.getState(newState)` reads the plugin's own state slot before ProseMirror assigns it — it is `undefined` during the plugin's own `apply`.
- `Storage.set` swallows all errors (quota exceeded, blocked storage), so a failed unfold write leaves a stale `{ fold: true }`, and the next edit transaction re-folds the block the user just expanded.
- Stored fold state is also never pruned: a key is written for every toggle block ever seen and never removed, so storage grows towards the quota and makes those silent failures more likely over time.

### Solution

Two layers, one per root cause:

**In-memory state is authoritative within a session.** The plugin tracks a `knownIds` set alongside `foldedIds`; on remote and local doc-changing transactions it reconciles from its previous state, preserving existing fold decisions verbatim and only initializing fold state for blocks seen for the first time. `FOLD`/`UNFOLD` still write to storage. This stops blocks collapsing mid-session — including when storage is healthy, since the dead `foldedIds` seed meant every remote transaction rebuilt fold state from storage.

**Storage is bounded and degrades gracefully:**

- Fold state lives in an `LRUCache` mirrored to local storage (`toggle` namespace, max 500 entries), so the number of stored keys is capped and the least recently used fold states are evicted first, instead of one key per block ever seen. The persisted key format (`toggle:<id>`) is unchanged, so fold state still syncs across tabs via storage events.
- `Storage` falls back to session storage when a write to the primary interface fails at runtime (e.g. quota exceeded): reads prefer the fallback copy while it is the most recent successful write, and a later successful write drops it again. A quota'd write now degrades to session persistence instead of being silently lost.

The empty-body auto-fold behavior is unchanged.

### Note

Fold state written before this change is not in the cache index, so on upgrade stored fold state resets once per browser (the same tradeoff the `mermaid:v3` namespace made); from then on it is bounded.


chenniannian90:
Both points from your comment are now in 8db1131:

- **Fallback** — `Storage` writes to sessionStorage when a write to localStorage fails at runtime, and reads prefer that copy while it is the most recent successful write, so persistence degrades to the session rather than being silently lost.
- **Pruning** — fold state goes through the existing `LRUCache` mirrored to localStorage (max 500), bounding the number of stored keys and evicting the least recently used first, instead of keeping a key per block ever seen.

The first commit's in-memory reconciliation is unchanged and independent of storage health: the `new Set(pluginState?.foldedIds)` seed was dead code (the plugin's own state slot reads as `undefined` during `apply`), so every remote transaction rebuilt fold state from storage — that is what collapsed expanded blocks even in the healthy case.

One behavior note: pre-existing entries are not in the cache index, so stored fold state resets once per browser on upgrade (same tradeoff as the `mermaid:v3` namespace bump).

shared/utils/LRUCache.ts
```diff
@@ -6,15 +6,15 @@ export interface LRUCacheOptions {
   max: number;
   /** Prefix under which entries are stored, used to namespace and version them. */
   namespace?: string;
-  /** Mirror entries to sessionStorage, requires a namespace. */
-  persistToSession?: boolean;
+  /** Mirror entries to the given web storage, requires a namespace. */
+  persistTo?: "local" | "session";
```
tommoor:
```suggestion
  storage: "memory" | "localStorage" | "sessionStorage";
```

this would be much less ambiguous

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -156,20 +169,24 @@ export default class ToggleBlock extends Node {
...
             decorations: this.createDecorations(state.doc, foldedIds),
           };
         },
 
         apply: (tr, pluginState, _oldState, newState) => {
+          // Remote (collaborative) document updates land here. Preserve the
+          // user's existing fold decisions verbatim and only initialize fold
+          // state for toggle blocks seen for the first time. Re-deriving fold
+          // state from storage on every remote update would collapse blocks
+          // the user just expanded whenever a collaborator edits: storage
```
tommoor: Lets calm this down, it's just not true - especially with the LRU

shared/editor/nodes/ToggleBlock.ts
```diff
@@ -65,6 +66,18 @@ export const toggleEventPluginKey = new PluginKey("toggleBlockEvent");
 /** Build the localStorage key used to persist a toggle block's fold state. */
 export const toggleStorageKey = (id: string) => `toggle:${id}`;
 
+/**
+ * A bounded cache of fold state per toggle block id, mirrored to local
+ * storage. Bounding the cache keeps the number of stored keys from growing
+ * without limit as documents are visited; evicted blocks simply fall back to
+ * the default folded state.
+ */
```
tommoor, line 73:
```suggestion
/**
 * A bounded cache of fold state per toggle block id, mirrored to local
 * storage. Evicted blocks fall back to the default folded state.
 */
```

tommoor (changes requested):
Make this one interface change and fix the CI.

CLAUDE.md
tommoor: Remove this file, can you review you're own work? I don't have time to point out obvious stuff - it's faster to just run claude myself.

chenniannian90:
Addressed in cc8c58d14:

- `CLAUDE.md` is back to the upstream symlink and out of the diff, along with the unrelated `scrollToNonce` changes (those belong to a separate change).
- Reviewing my own work I also found a real bug in the storage fallback: `clear()` was clearing **all** of session storage — and `ErrorBoundary` calls `clear()`. Fallback writes are now tracked per key and `clear()` removes only those (`Storage.test.ts` covers it).
- `fallbackInterface` is private now, and the constructor probe no longer uses the magic `"test"` key.

Local `yarn lint` and the `shared` test suite pass. The fork workflow runs will need an approve-and-run.

tommoor (approved):

PR #13429, opened 2026-08-14
fix: re-scroll when the same table of contents entry is clicked again

### LINKED ISSUE #13428: Clicking the same table of contents entry again does nothing after scrolling away
### Describe the bug

In the document Contents (TOC) sidebar, clicking an entry navigates to that heading and sets the URL hash. After manually scrolling elsewhere, clicking the **same** entry again does nothing — no scroll, no feedback. Clicking a *different* entry works fine, which makes the dead clicks feel intermittent.

### To Reproduce

1. Open a document with multiple headings (Contents sidebar visible).
2. Click a TOC entry → the view scrolls to that heading, URL hash becomes `#h-<heading>`.
3. Manually scroll away from that heading.
4. Click the same TOC entry again → nothing happens.

### Root cause

The TOC click handler (`app/scenes/Document/components/Contents.tsx`) pushes a location with only the hash changed:

```ts
history.push(patchLocation(history.location, { hash: `#${id}` }));
```

The editor receives `scrollTo` derived from the current hash (`app/scenes/Document/components/Editor.tsx`):

```tsx
scrollTo={decodeURIComponentSafe(window.location.hash)}
```

and `scrollToAnchor` is only re-invoked when the prop value changes (`app/editor/index.tsx`):

```ts
if (this.props.scrollTo && this.props.scrollTo !== prevProps.scrollTo) {
  void this.scrollToAnchor(this.props.scrollTo);
}
```

When the target heading is already the one in the hash, `scrollTo` is unchanged, the guard short-circuits, and the click is a no-op. The guard is otherwise sensible (it prevents unrelated re-renders from re-scrolling), it just also swallows legitimate repeated clicks.

I verified this with a jsdom simulation replicating the exact component chain (TOC click handler → hash-derived `scrollTo` prop → `componentDidUpdate` guard) against React 17 / react-router 5.3.4: the second click **does** re-render the scene, but `scrollToAnchor` is not called again.

### Expected behavior

Clicking a TOC entry should always scroll to that heading, even if it was already visited.

### Suggested fix

Include a monotonically changing value in the pushed location's state (e.g. `tocScrollNonce: Date.now()`), thread it through as a `scrollToNonce` prop next to `scrollTo`, and re-run `scrollToAnchor` when either `scrollTo` or `scrollToNonce` changes:

```ts
if (
  this.props.scrollTo &&
  (this.props.scrollTo !== prevProps.scrollTo ||
    this.props.scrollToNonce !== prevProps.scrollToNonce)
) {
  void this.scrollToAnchor(this.props.scrollTo);
}
```

This preserves the existing protection against unrelated re-renders while making repeated clicks re-trigger the scroll. I verified the same simulation with the fix applied: the re-click scrolls again.

Happy to open a PR with the patch if the approach sounds good.

chenniannian90:
Fixes #13428

### Problem

Clicking a TOC entry pushes a location with only the hash changed. The editor receives `scrollTo` derived from the current hash and only re-runs `scrollToAnchor` when that prop value changes — so once the hash points at a heading, clicking that same TOC entry again is a no-op (verified with a jsdom simulation replicating the exact component chain: the scene does re-render, but the guard in `componentDidUpdate` blocks the call).

### Solution

Include a monotonically changing `tocScrollNonce` in the pushed location's state, thread it through as a `scrollToNonce` prop alongside `scrollTo`, and re-run `scrollToAnchor` when either value changes:

- `Contents.tsx` — include `tocScrollNonce: Date.now()` in the pushed location state.
- `Editor.tsx` — read the hash via `useLocation` and pass `scrollToNonce` from the location state.
- `editor/index.tsx` — new optional `scrollToNonce` prop; the update guard now also fires when the nonce changes.
- `MultiplayerEditor.tsx` — pass `scrollToNonce` through to the cache editor.

Unrelated re-renders still never re-scroll (the hash and nonce are both unchanged for them); only explicit repeated clicks do.

### DIFF
--- a/app/editor/index.tsx
+++ b/app/editor/index.tsx
@@ -106,6 +106,12 @@ export type Props = {
   maxLength?: number;
   /** Heading id to scroll to when the editor has loaded */
   scrollTo?: string;
+  /**
+   * Monotonic value that changes on each navigation to the same `scrollTo`
+   * target, so repeated clicks re-trigger scrolling even when the hash is
+   * unchanged.
+   */
+  scrollToNonce?: number;
   /** Callback for handling uploaded images, should return the url of uploaded file */
   uploadFile?: (
     file: File | string,
@@ -309,7 +315,11 @@ export class Editor extends React.PureComponent<
       );
     }
 
-    if (this.props.scrollTo && this.props.scrollTo !== prevProps.scrollTo) {
+    if (
+      this.props.scrollTo &&
+      (this.props.scrollTo !== prevProps.scrollTo ||
+        this.props.scrollToNonce !== prevProps.scrollToNonce)
+    ) {
       void this.scrollToAnchor(this.props.scrollTo);
     }
 
--- a/app/scenes/Document/components/Contents.tsx
+++ b/app/scenes/Document/components/Contents.tsx
@@ -38,9 +38,20 @@ function Contents() {
         return;
       }
       // Navigate via history so the location state (active sidebar context) is
-      // retained rather than dropped by a native hash navigation.
+      // retained rather than dropped by a native hash navigation. The nonce
+      // forces a re-scroll when the same heading is clicked again after the
+      // hash is already set — the editor otherwise skips scrolling when the
+      // target is unchanged.
       event.preventDefault();
-      history.push(patchLocation(history.location, { hash: `#${id}` }));
+      const previousState =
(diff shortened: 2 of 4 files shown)


tommoor:
Superseded by simpler approach in https://github.com/outline/outline/pull/13458

PR #13434, opened 2026-08-14
fix: Make code fence markdown serialization round-trip stable

tommoor:
- Code fences now serialize with more backticks than the longest backtick run in their content, so a content line can no longer terminate the fence early when the markdown is parsed again
- The fence info string and `language` attribute are reduced to a single safe token on both parse and serialize, instead of storing arbitrary text verbatim
- Previously an affected document changed block structure on every export → import round trip, which could collapse most of a document into a single code block and made revision diffs extremely expensive
- Adds shared `parser`/`serializer` exports to the editor test helper and removes the duplicated construction from existing tests

### DIFF
--- a/shared/editor/nodes/CodeFence.ts
+++ b/shared/editor/nodes/CodeFence.ts
@@ -64,6 +64,20 @@ const COLLAPSE_HEIGHT_RATIO = 0.5;
 /** Approximate rendered line height of a code block, in pixels. */
 const CODE_LINE_HEIGHT = 20;
 
+/**
+ * Reduce a language attribute or fence info string to a single safe token, so
+ * it cannot break the fence line when written back to markdown.
+ *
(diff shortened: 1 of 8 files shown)


shared/editor/nodes/CodeFence.ts
```diff
@@ -717,10 +731,17 @@ export default class CodeFence extends Node<CodeFenceOptions> {
       ? escapeRawTableCell(node.textContent)
       : node.textContent;
 
-    state.write("```" + (node.attrs.language || "") + "\n");
+    // The fence must be longer than any backtick run in the content, or the
+    // content could terminate the fence early when the markdown is parsed.
+    const backticks = content.match(/`{3,}/g);
+    const fence = "`".repeat(
```
tommoor, line 737: TIL this is valid markdown

PR #13438, opened 2026-08-15
feat: Support OIDC back-channel logout

tommoor:
Adds `/auth/oidc.backchannel_logout`, which accepts a provider-initiated logout notification and ends the Outline sessions of the identified user.

- Mounted only when discovery supplies the key set and issuer, so there is no new configuration – the URL just needs registering with the provider.
- Logout tokens are validated against the spec: signature from the cached provider keys, issuer, audience, freshness, the logout event claim, absence of a nonce, and single use by `jti`.
- Provider signing keys are cached and refetched on an unknown key id, so key rotation needs no restart.
- Revocation rotates the user's JWT secret, so the sign-out covers all of their devices. A token scoped to a provider session alone is rejected, as individual sessions cannot be revoked.

### DIFF
--- a/plugins/oidc/server/JWKSCache.ts
+++ b/plugins/oidc/server/JWKSCache.ts
@@ -1,0 +1,129 @@
+import crypto, { type JsonWebKey } from "node:crypto";
+import { InternalError } from "@server/errors";
+import fetch from "@server/utils/fetch";
+import { Minute, Second } from "@shared/utils/time";
+
+interface JSONWebKey extends JsonWebKey {
+  kid?: string;
+  use?: string;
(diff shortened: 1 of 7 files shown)


ByTrist4n (commented):
Cool :)

forkrails:
> Revocation rotates the user's JWT secret, so the sign-out covers all of their devices. A token scoped to a provider session alone is rejected, as individual sessions cannot be revoked.

I realize it is a limitation of the stateless auth, but this is not the expected behavior when using backchannel logout. Most people will not want to sign out of all sessions when logging out of one.

PR #13459, opened 2026-08-16
fix: Serialize external mentions as links in exported markdown

### LINKED ISSUE #13053: GitHub issue/PR/project mentions serialize to opaque mention:// URIs in Markdown export

tommoor:
Issues, pull requests, and project mentions carry the real URL, but always serialized to a `mention://` reference that nothing outside of Outline can resolve.

- Write a plain markdown link for these mentions when the `commonMark` serializer option is set, falling back to the existing reference when there is no url
- Set that option on the paths where markdown leaves the platform – document and revision export, zip exports, and the public share markdown response
- Internal paths keep the existing format, so MCP reads, the API text field and stored document text are unchanged

Closes #13053
Supersedes #13054

### DIFF
--- a/shared/editor/nodes/Mention.tsx
+++ b/shared/editor/nodes/Mention.tsx
@@ -277,10 +277,10 @@ export default class Mention extends Node {
         ) {
           const mentionType = selection.node.attrs.type;
 
-          let link: string;
+          let link: string | undefined;
 
           if (isExternalMention(mentionType)) {
-            link = selection.node.attrs.href;
+            link = sanitizeUrl(selection.node.attrs.href);
           } else {
             const { modelId } = selection.node.attrs;
 
@@ -296,7 +296,9 @@ export default class Mention extends Node {
             }`;
           }
 
-          this.editor.props.onClickLink?.(link);
+          if (link) {
+            this.editor.props.onClickLink?.(link);
+          }
           return true;
         }
         return false;
--- a/shared/editor/nodes/Mention.tsx
+++ b/shared/editor/nodes/Mention.tsx
@@ -49,6 +49,22 @@ function dateMentionLabel(node: ProsemirrorNode): string {
     : node.attrs.label;
 }
 
+/**
+ * Whether a mention points at a resource outside of Outline, in which case the
+ * real URL is stored in `attrs.href` rather than addressed with a `mention://`
+ * reference.
+ *
+ * @param type the mention type.
+ * @returns true if the mention links to an external URL.
+ */
+function isExternalMention(type: MentionType): boolean {
+  return (
+    type === MentionType.Issue ||
+    type === MentionType.PullRequest ||
+    type === MentionType.Project
+  );
+}
+
 export default class Mention extends Node {
   get name() {
     return "mention";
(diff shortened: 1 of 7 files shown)


shared/editor/nodes/Mention.tsx
```diff
@@ -148,12 +164,9 @@ export default class Mention extends Node {
           "data-id": node.attrs.modelId,
           "data-actorid": node.attrs.actorId,
           "data-anchor-id": node.attrs.anchorId,
-          "data-url":
-            node.attrs.type === MentionType.PullRequest ||
-            node.attrs.type === MentionType.Issue ||
-            node.attrs.type === MentionType.Project
-              ? sanitizeUrl(node.attrs.href)
-              : `mention://${node.attrs.id}/${node.attrs.type}/${node.attrs.modelId}`,
+          "data-url": isExternalMention(node.attrs.type)
```
tommoor, line 167: Needed this refactor anyway

PR #13499, opened 2026-08-20
fix: Update GitLab token handling for compatibility with older instances

hant121:
- Made `expires_in` and `refresh_token` optional in the AccessTokenResponseSchema to accommodate older self-hosted GitLab versions( Gitlab version < 15 ) that may not provide these fields.
- Adjusted token handling in the API to use fallback values for `refreshToken` when not provided.
- Updated IntegrationAuthentication to handle optional `expires_in` when calculating `expiresAt` for tokens.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,7 +3,7 @@ Business Source License 1.1
 Parameters
 
 Licensor:             General Outline, Inc.
-Licensed Work:        Outline 1.9.1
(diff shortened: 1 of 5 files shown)


tommoor:
I'm not sure it makes sense for us to support such old GitLab instances for auth, especially given this clearly compromises the security

PR #13516, opened 2026-08-21
fix(i18n): repair corrupted interpolation placeholders in locale files

### LINKED ISSUE #13514: Spanish welcome email renders literal "{ appName }" instead of "Outline"

### Current Behavior

When an invited user receives the welcome email in Spanish (`es_ES`), the
application name placeholder is not interpolated.

The email renders:

> ¡Bienvenido a { appName }!

### Expected Behavior

The heading should render:

> ¡Bienvenido a Outline!

### Steps To Reproduce

1. Run a self-hosted Outline instance using version 1.9.2.
2. Use the Spanish locale (`es_ES`).
3. Invite a new non-guest user.
4. Let the user accept the invitation or sign in.
5. Inspect the welcome email.

### Environment

```markdown
- Outline: 1.9.2
- Deployment: Self-hosted Docker
- Locale: es_ES
```

### Anything else?

The Spanish translation currently contains a single-brace placeholder:

```json
"Welcome to {{ appName }}!": "¡Bienvenido a { appName }!"

It should use the same double-brace interpolation syntax as the source string:

"Welcome to {{ appName }}!": "¡Bienvenido a {{ appName }}!"

The affected translation is located here:
https://github.com/outline/outline/blob/v1.9.2/shared/i18n/locales/es_ES/translation.json#L1757
The issue is also present in the current main branch.

yzxcj797:
## Summary

Fixes #13514 — the Spanish welcome email rendered a literal `{ appName }` because the translation used a single-brace placeholder that i18next never interpolates.

While fixing that one string I audited **every shipped locale against `en_US`** for interpolation-placeholder integrity and found five more corruptions of the same class, all of which leak literal placeholder text into the UI or emails today:

| Locale | Key | Corruption |
|---|---|---|
| es_ES | `Welcome to {{ appName }}!` | `{ appName }` — single brace (the reported bug) |
| he_IL | `{{ minutes }}m read` | `{{ דקות }}` — placeholder *name* translated, never interpolates |
| uk_UA | `{{userName}} requested` | `{{username}}` — case mismatch, interpolates to empty |
| uk_UA | `No results for "{{ term }}"` | `{{ terms }}` — case mismatch |
| vi_VN | `{{ count }} member(_plural)` | `{{$count}}` — dollar prefix, never interpolates |
| pt_PT | `{{ groupsCount }} groups with access(_plural)` | `{{ usersCount }}` — a variable `MembershipPreview` never passes to this string |
| fa_IR | `{{ releasesBehind }} versions behind` | placeholder dropped, renders "یک نسخه عقب است" regardless of count |

## Prevention

Adds `app/utils/i18nPlaceholders.test.ts` — one assertion per locale that fails when a translation contains:

- a single-brace `{ x }` form of a variable the source interpolates,
- a `{{$x}}` dollar-prefixed form,
- a non-identifier placeholder name (e.g. a translated name), or
- a case-renamed variable (`{{username}}` vs `{{userName}}`).

This way a future Crowdin sync that reintroduces this corruption class fails CI instead of shipping.

## Test plan

- [x] `yarn vitest run app/utils/i18nPlaceholders.test.ts app/utils/i18n.test.ts` — 39/39 pass (27 new locale assertions + existing)
- [x] Differential check: reverting the data fixes makes the new test fail with exactly the corrupted strings listed above; with the fixes it is green
- [x] All six JSON files re-parse cleanly and `git diff` touches only the listed strings

Note: the Crowdin memory for these strings will need the same correction applied by a maintainer so the next "New Crowdin updates" sync keeps the fixed values; the new test guards exactly that sync.


tommoor:
Sorry, whilst these fixes look correct, they can't be made directly in the code base – they have to be made on CrowdIn, the translation platform or they will simply be overwritten.

I updated these so the translations will be pulled into next release

PR #13531, opened 2026-08-23
fix: Contents highlight lost when scrolling deep headings

tommoor:
The contents lists headings up to level four, but tracked the active heading across every heading in the document. Scrolling into a level five or six section selected a heading with no entry in the list, so the highlight disappeared until the next listed heading came into view.

- Run the level filter once, up front, so the active heading tracking, indentation base, empty check and render all read the same list
- A reader inside a deep section now keeps the nearest listed heading highlighted, and the list keeps auto-scrolling to follow them
- A document made only of level five and six headings no longer renders an empty "Contents" panel
- Prefix labels are still computed over every heading, so the numbering keeps matching the document

### DIFF
--- a/app/scenes/Document/components/Contents.tsx
+++ b/app/scenes/Document/components/Contents.tsx
@@ -18,6 +18,9 @@ import { decodeURIComponentSafe } from "~/utils/urls";
 
 const HEADING_OFFSET = 20;
 
+/** Headings deeper than this level are not listed in the contents. */
+const MAX_HEADING_LEVEL = 4;
+
 function Contents() {
   const history = useHistory();
   const [scrolledSlug, setScrolledSlug] = useState<string>();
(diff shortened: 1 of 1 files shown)


UmbraCi:
Thanks for fixing the deep-heading highlight behavior. Could you share the rationale for intentionally limiting authoring and the table of contents to H1–H4 while retaining H5/H6 support for imported content? Most modern documentation platforms expose all six HTML heading levels. Is the four-level limit a product or UX decision (for example, outline readability), or a historical implementation constraint? Understanding this would help downstream integrations decide whether they should preserve the same boundary.

tommoor:
It is a long standing UX decision and purposeful limitation. There are quite a few places that we have made decisions like this to help guide folks into writing better documents, and this is one of those.

Headings beyond L4 would all render the same size too.

PR #13557, opened 2026-08-24
perf: Delay loading emoji data index

tommoor:
`shared/utils/emoji.ts` imported the full (large) emoji index statically, and that module is reached from the share page via `Reaction.tsx` and the editor’s `EmojiMenu`.

The dataset is only needed for emoji search/picker UIs, not for rendering a document.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -79,6 +79,16 @@
           {
             "paths": [
               {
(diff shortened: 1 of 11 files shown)


tommoor:
Decided this complicates too much for not enough gain

PR #13568, opened 2026-08-26
fix(History): scroll to first change when opening a revision

### LINKED ISSUE #13567: History: opening a revision does not scroll to the first change
## Describe the bug

When a revision is opened from the **History** sidebar, the revision content renders with diff highlights, but the viewport stays at the top of the document. In long documents the changed content is often far below the fold, so nothing on screen indicates that anything changed — the user has to scroll manually or reach for the next/prev change arrows in the header to find the highlighted difference.

## To reproduce

1. Open a long document with several revisions where a change is below the fold.
2. Open **History** and click a revision.
3. The revision opens scrolled to the top of the document; the diff highlight is not in view.

### Expected behavior

The viewport should land on the first change of the diff, the same way the existing next/prev change navigation in the header does (`scrollIntoView` with `scrollMode: "if-needed"`), so the user immediately sees where the revision differs.

## Details

- Outline version: 1.9.1
- Browser: any

chenniannian90:
## Description

Opening a revision from the History sidebar renders the diff highlights but always leaves the viewport at the top of the document. In long documents the changed content is often below the fold, so nothing on screen indicates where the revision differs.

The root cause is that the Diff extension initializes with no change selected (`currentChangeIndex = -1`), and only the next/prev controls in the header ever trigger navigation — nothing moves the viewport when a revision is first shown.

This PR triggers `nextChange` once from `RevisionViewer` after the diff editor has mounted with its decorations in place. That reuses the existing navigation behavior:

- scrolls to the first change only when it is not already in view (`scrollMode: "if-needed"`, centered),
- highlights it as the current change,
- and updates the header counter to "1 of N changes" so the arrows continue from there.

The effect is keyed on the same `editorKey` that remounts the editor whenever the diff configuration changes, so it also re-positions the viewport when switching between revisions in the list, and is a no-op when change highlighting is disabled.

Fixes #13567

## Testing

Automated: `oxlint` and `oxfmt` clean; existing `Diff` extension tests pass.

Manual test plan: open a long document with revisions whose change is below the fold and click a revision in History — the viewport should land centered on the first change with the "1 of N changes" counter in the header; revisions with the first change already visible stay put; with "Highlight changes" off there should be no scrolling or counter change; switching revisions in the list should re-position to the new first change.

### DIFF
--- a/app/scenes/Document/components/RevisionViewer.tsx
+++ b/app/scenes/Document/components/RevisionViewer.tsx
@@ -1,5 +1,6 @@
 import { observer } from "mobx-react";
 import * as React from "react";
+import { mergeRefs } from "react-merge-refs";
 import { colorPalette } from "@shared/constants";
 import type Document from "~/models/Document";
 import type Revision from "~/models/Revision";
@@ -43,6 +44,7 @@ function RevisionViewer(props: Props, ref: React.Ref<TEditor>) {
   const { document, children, revision } = props;
   const { revisions } = useStores();
   const query = useQuery();
+  const localRef = React.useRef<TEditor | null>(null);
   const showChanges = props.showChanges ?? query.has("changes");
   const compareToParam = query.get("compareTo");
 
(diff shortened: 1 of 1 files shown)


tommoor:
I'm not moving forward with this one for the reasoning described here https://github.com/outline/outline/issues/13567#issuecomment-5423125409

PR #13569, opened 2026-08-26
feat: Provider-agnostic email delivery (Microsoft Graph, Gmail, Amazon SES)

cmfrazier340B:
Implements the provider-agnostic email delivery discussed in #13019, where SMTP AUTH being disabled on Microsoft 365 tenants leaves no way to send Outline email. Naming and env-only configuration per your comments there.

## Design

- `BaseEmailProvider` implements nodemailer's `Transport`; each provider implements a single `sendMessage(mail)` method and can serialize the fully assembled RFC 5322 message via `getMimeMessage()`. Because providers receive the assembled message rather than mapping onto each vendor's JSON schema, threading headers, `List-Unsubscribe`, and inline attachments behave identically across providers with no per-provider field mapping to maintain.
- Selected with a new `EMAIL_PROVIDER` env var (default `smtp`, existing behavior unchanged) resolved through a new `Hook.EmailProvider` — mirroring `SEARCH_PROVIDER` / `BaseSearchProvider`.
- `EMAIL_PROVIDER` is validated against registered providers at startup, so a typo or missing provider credentials fails fast at boot instead of surfacing as per-email queue failures. `SMTP_FROM_EMAIL` is required whenever a provider is configured.

## Providers

| Provider | Auth | Why |
|---|---|---|
| `msgraph` (plugins/azure) | OAuth client credentials → `users/{mailbox}/sendMail` (MIME) | M365 tenants with SMTP AUTH disabled; works with Conditional Access |
| `gmail` (plugins/google) | Service account + domain-wide delegation → `messages/send` | Google Workspace without app passwords |
| `ses` (plugins/ses) | Standard AWS credential chain → SES v2 `SendEmail` (raw) | Instance roles/IRSA — no SMTP credentials to store or rotate |

Details handled: per-provider message tagging (SES tags preserved between SMTP and API paths), Bcc semantics per transport (submission endpoints keep the header, SES strips it and delivers via the envelope), OAuth token caching with coalesced concurrent refresh, sovereign-cloud endpoint overrides for Graph, and base64/escaped-PEM key handling.

## Testing

63 tests across the new code drive real nodemailer through mocked provider endpoints (msw) and assert on the decoded wire format — including cryptographic verification of the Gmail JWT assertion against a generated keypair. Full server suite passes. Not yet exercised against live M365/Workspace/SES tenants.

Docs updated in `.env.sample` and `app.json`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -220,6 +220,28 @@ SMTP_USERNAME=
 SMTP_PASSWORD=
 SMTP_FROM_EMAIL=
 
+# Email can instead be delivered over a provider's API, which avoids needing
+# mailbox credentials at all. SMTP_FROM_EMAIL remains the sender address.
+# EMAIL_PROVIDER=smtp
+
+# Microsoft Graph (EMAIL_PROVIDER=msgraph), for tenants where SMTP AUTH is
+# disabled. Needs an Entra ID app registration with the Mail.Send application
+# permission granted. These are separate from any sign-in credentials above –
+# set them explicitly even when reusing the same app registration.
+# AZURE_MAIL_CLIENT_ID=
+# AZURE_MAIL_CLIENT_SECRET=
+# AZURE_MAIL_TENANT_ID=
+
+# Gmail (EMAIL_PROVIDER=gmail). Needs a service account granted domain-wide
+# delegation for the https://www.googleapis.com/auth/gmail.send scope.
+# GOOGLE_MAIL_CLIENT_EMAIL=
+# GOOGLE_MAIL_PRIVATE_KEY=
+
+# Amazon SES (EMAIL_PROVIDER=ses). Credentials are resolved from the standard
+# AWS credential chain, so an instance role can be used in place of keys.
+# AWS_SES_REGION=
+# AWS_SES_CONFIGURATION_SET=
+
 
 # ––––––––––––––––––––––––––––––––––––––
 # ––––––––––  RATE LIMITER  ––––––––––––
--- a/app.json
+++ b/app.json
@@ -181,6 +181,10 @@
       "value": "26214400",
       "required": false
     },
+    "EMAIL_PROVIDER": {
+      "description": "How email is delivered: 'smtp' (default), 'msgraph', 'gmail' or 'ses' (optional)",
+      "required": false
+    },
     "SMTP_HOST": {
       "description": "smtp.example.com (optional)",
       "required": false
@@ -212,6 +216,34 @@
     "SMTP_SECURE": {
       "value": "true",
       "description": "Use a secure SMTP connection (optional)",
+      "required": false
+    },
+    "AZURE_MAIL_CLIENT_ID": {
+      "description": "Entra ID application with the Mail.Send permission, for EMAIL_PROVIDER=msgraph (optional)",
+      "required": false
+    },
+    "AZURE_MAIL_CLIENT_SECRET": {
+      "description": "(optional)",
+      "required": false
+    },
+    "AZURE_MAIL_TENANT_ID": {
+      "description": "(optional)",
+      "required": false
+    },
+    "GOOGLE_MAIL_CLIENT_EMAIL": {
+      "description": "Service account with domain-wide delegation, for EMAIL_PROVIDER=gmail (optional)",
+      "required": false
+    },
+    "GOOGLE_MAIL_PRIVATE_KEY": {
+      "description": "PEM private key for the service account, base64 encoded (optional)",
+      "required": false
+    },
+    "AWS_SES_REGION": {
+      "description": "The region SES is configured in, for EMAIL_PROVIDER=ses (optional)",
+      "required": false
+    },
+    "AWS_SES_CONFIGURATION_SET": {
+      "description": "SES configuration set used to publish sending events (optional)",
       "required": false
     },
     "SMTP_DISABLE_STARTTLS": {
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   ],
   "dependencies": {
     "@aws-sdk/client-s3": "^3.1115.0",
+    "@aws-sdk/client-sesv2": "^3.1115.0",
     "@aws-sdk/cloudfront-signer": "^3.1111.0",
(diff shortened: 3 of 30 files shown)


cmfrazier340B:
For the [hosting docs SMTP page](https://docs.getoutline.com/s/hosting/doc/smtp-cqCJyZGMIB), here's proposed updated content to apply if/when this merges. The existing content is preserved verbatim under "Deliver over SMTP" (only `SMTP_FROM_EMAIL`/`SMTP_REPLY_EMAIL` move up, since they now apply to every delivery method); the new material is the intro and the "Deliver over a provider API" section. Suggest keeping the page title/URL to avoid breaking links, since SMTP remains the default. The `:::info` callout uses Outline's own syntax, ready to paste.

---

To enable outgoing transactional emails—such as "document updated" or "you've been invited"—you must configure email delivery. This is also essential for email "magic link" sign-in functionality.

Email can be delivered in two ways: over **SMTP** (the default) or over a **provider API**. The API option exists for environments where SMTP authentication is unavailable or undesirable — for example Microsoft 365 tenants that disable SMTP AUTH, Google Workspace accounts without app passwords, or AWS deployments that prefer IAM roles over stored SMTP credentials.

In both cases, set:

**SMTP_FROM_EMAIL** – Sender address; supports mailbox format: `Outline <noreply@mycompany.com>`. Required for all delivery methods.

**SMTP_REPLY_EMAIL** – Optional reply address; supports mailbox format

## Deliver over SMTP

Email configuration offers two approaches based on your service provider. For "known" services, the first method is recommended for simplest setup.

### With Known Service

**SMTP_SERVICE** – A supported email service from the list below

**SMTP_USERNAME** – Mail server username

**SMTP_PASSWORD** – Mail server password

#### Known Services

1und1, AOL, DebugMail.io, DynectEmail, FastMail, GandiMail, Gmail, Godaddy, GodaddyAsia, GodaddyEurope, hot.ee, Hotmail, iCloud, mail.ee, Mail.ru, Mailgun, Mailjet, Mandrill, Naver, Postmark, QQ, QQex, SendCloud, SendGrid, SES, Sparkpost, Yahoo, Yandex, Zoho

### Custom Mail Server

For providers not listed above or self-hosted solutions, use these settings:

**SMTP_HOST** – Mail server hostname (e.g., mail.mycompany.com)

**SMTP_PORT** – Mail server port, typically 465

**SMTP_USERNAME** – Authentication username

**SMTP_PASSWORD** – Optional password for secured servers

**SMTP_TLS_CIPHERS** – Optional TLS ciphers configuration

**SMTP_SECURE** – Optional; disables TLS when false (defaults to true)

**SMTP_NAME** – Optional hostname parameter for server communication

## Deliver over a provider API

Set **EMAIL_PROVIDER** to choose how email is delivered. The value is case-insensitive and is validated when the server starts — if the named provider is missing its configuration, the server logs the reason and exits rather than failing silently at send time.

**EMAIL_PROVIDER** – One of `smtp` (default), `msgraph`, `gmail`, or `ses`

No SMTP variables other than `SMTP_FROM_EMAIL` and `SMTP_REPLY_EMAIL` are needed when using a provider API.

### Microsoft Graph

For Microsoft 365 tenants, including those with SMTP AUTH disabled. Sends via the Graph `sendMail` endpoint using OAuth client credentials, so it is compatible with Conditional Access and MFA policies.

```
EMAIL_PROVIDER=msgraph
```

**AZURE_MAIL_CLIENT_ID** – Application (client) ID of an Entra ID app registration

**AZURE_MAIL_CLIENT_SECRET** – A client secret for that app registration

**AZURE_MAIL_TENANT_ID** – Directory (tenant) ID

**AZURE_MAIL_FROM_USER_ID** – Optional; the mailbox to send from, as a user principal name or object ID. Defaults to the address in `SMTP_FROM_EMAIL`. A shared mailbox works and needs no license.

**AZURE_MAIL_AUTHORITY_URL** / **AZURE_MAIL_GRAPH_BASE_URL** – Optional; only for sovereign clouds such as US Government or China. Defaults target the global cloud.

Setup in Entra ID: create an app registration (no redirect URI needed), add the **application** permission `Mail.Send` under Microsoft Graph, and grant admin consent. These variables are separate from any `AZURE_*` sign-in configuration — set them explicitly even when reusing the same app registration, and note that the sign-in app does not normally have `Mail.Send`.

:::info
Application-level `Mail.Send` allows sending as **any** mailbox in the tenant. Scope it to the sending mailbox with Exchange Online's [role based access control for applications](https://learn.microsoft.com/en-us/exchange/permissions-exo/application-rbac). Sent messages are saved to the mailbox's Sent Items folder.
:::

### Gmail

For Google Workspace organizations. Sends via the Gmail API using a service account with domain-wide delegation — this cannot be used with a personal Gmail account.

```
EMAIL_PROVIDER=gmail
```

**GOOGLE_MAIL_CLIENT_EMAIL** – The service account's email address

**GOOGLE_MAIL_PRIVATE_KEY** – The service account's PEM private key; either the PEM itself or base64-encoded on a single line

**GOOGLE_MAIL_FROM_USER_ID** – Optional; the mailbox to impersonate. Defaults to the address in `SMTP_FROM_EMAIL`. If set to a different mailbox, Gmail rewrites the From header to this address unless the `SMTP_FROM_EMAIL` address is configured as a send-as alias of it.

Setup: create a service account in Google Cloud, then in the Workspace admin console grant it domain-wide delegation for the scope `https://www.googleapis.com/auth/gmail.send`.

### Amazon SES

Sends via the SES API using the standard AWS credential chain, so an EC2 instance profile, ECS task role, or EKS IRSA works with no email credentials stored in Outline's configuration.

```
EMAIL_PROVIDER=ses
```

**AWS_SES_REGION** – The region SES is configured in; defaults to `AWS_REGION`

**AWS_SES_CONFIGURATION_SET** – Optional; an SES configuration set for publishing sending events

Requirements: the `SMTP_FROM_EMAIL` address (or its domain) must be a verified identity in that region, the credentials need the `ses:SendEmail` IAM action, and the account must be out of the SES sandbox to email unverified recipients.

PR #13574, opened 2026-08-27
fix: Keep OIDC provider identifier in sync with configured domain

### LINKED ISSUE #13572: Generic OIDC provider ID remains on old domain after changing provider domain

### Current Behavior

After changing the domain of an existing Generic OIDC identity provider, Outline continues to show the old provider domain in the UI.

All OIDC environment variables were updated to the new domain and the Outline container was recreated. Authentication through the new OIDC domain works correctly.

However, the old domain remained visible in Outline.

Checking the PostgreSQL database showed that the old domain was still persisted in:

authentication_providers.providerId

For example:

providerId = id.old-domain.example

while the active OIDC configuration was already using:

https://id.new-domain.example

Manually updating authentication_providers.providerId to the new domain and restarting Outline immediately fixed the value shown in the UI. Authentication continued to work normally.

### Expected Behavior

When the domain of the configured Generic OIDC provider changes, Outline should keep the persisted authentication provider information consistent with the active OIDC configuration.

Ideally, Outline should either update authentication_providers.providerId automatically or provide a supported way to migrate/update this value.

Changing an OIDC provider domain should not require a manual modification of the Outline PostgreSQL database.

### Steps To Reproduce

1. Run a self-hosted Outline instance using Generic OIDC.

2. Configure an OIDC provider using a domain such as:

   https://id.old-domain.example

3. Log in successfully so that the authentication provider is created in Outline.

4. Move the same OIDC provider to a new domain, for example:

   https://id.new-domain.example

5. Update all Outline OIDC environment variables to use the new domain.

6. Recreate/restart the Outline container.

7. Log in through the new OIDC domain and verify that authentication works correctly.

8. Open the authentication/provider settings in Outline.

9. Observe that the old provider domain is still displayed.

10. Query the database:

    SELECT id, "providerId"
    FROM authentication_providers;

    The stored providerId still contains:

    id.old-domain.example

11. Update the value manually:

    UPDATE authentication_providers
    SET "providerId" = 'id.new-domain.example'
    WHERE "providerId" = 'id.old-domain.example';

12. Restart Outline.

13. The new provider domain is now displayed correctly and OIDC authentication continues to work.

### Environment

```markdown
- Outline: 2.14.0
- Deployment: Docker
- Database: PostgreSQL 17
- Authentication: Generic OIDC
- Identity provider: Pocket ID
- Reverse proxy: Caddy
- Browser: Microsoft Edge
```

### Anything else?

The OIDC authentication itself works correctly after the domain migration. This appears to be a persistence/consistency issue rather than an OIDC configuration or authentication failure.

I found a few related reports, but none seem to describe this exact Generic OIDC provider domain migration case:

- Discussion #6130 also involves an incorrect or stale authentication_providers.providerId and database-level changes as a workaround, but covers a different authentication-provider scenario.

- Issue #6858 concerns authentication mappings during a domain change for Google authentication and appears conceptually related, but does not cover Generic OIDC.

- Discussion #8250 contains related cases involving persisted OIDC authentication-provider records, but describes a different scenario.

The workaround in this case was to manually update the persisted provider ID:

UPDATE authentication_providers
SET "providerId" = 'id.new-domain.example'
WHERE "providerId" = 'id.old-domain.example';

After restarting Outline, the stale domain disappeared from the UI and OIDC authentication continued to work correctly.

It would be helpful if Outline either synchronized this persisted value when the Generic OIDC provider changes or provided/documented a supported migration procedure.

tommoor:
On each OIDC sign in the router reuses the identifier stored on the team's existing provider record. That value is only ever derived from the configured authorization URL at first sign in, so moving the identity provider to a new domain left the old hostname persisted and displayed in settings, with a manual database edit as the only remedy.

- Update the stored provider identifier when it no longer matches the configured OIDC location
- Guard the update so a conflicting record cannot interrupt sign in

closes #13572

### DIFF
--- a/plugins/oidc/server/auth/oidcRouter.ts
+++ b/plugins/oidc/server/auth/oidcRouter.ts
@@ -164,8 +164,24 @@ export function createOIDCRouter(
 
             // Derive a providerId from the OIDC location if there is no existing provider.
             const oidcURL = new URL(endpoints.authorizationURL);
-            const providerId =
+            let providerId =
               authenticationProvider?.providerId ?? oidcURL.hostname;
+
+            // The provider can move to a different domain, in which case the
+            // stored identifier must follow the configured location.
+            if (authenticationProvider && providerId !== oidcURL.hostname) {
+              try {
+                await authenticationProvider.update({
+                  providerId: oidcURL.hostname,
+                });
+                providerId = oidcURL.hostname;
+              } catch (error) {
+                Logger.warn(
+                  "Could not update OIDC authentication provider identifier",
+                  { providerId, hostname: oidcURL.hostname, error }
+                );
+              }
+            }
 
             if (!domain) {
               throw OIDCMalformedUserInfoError();


IamTaoChen:
it doesn't work.  

image
image

The OIDC issuer is `https://auth.xxx.com/xxx` and my email is `xxx@outlook.com`

`@139.com` is the firsr user's eamil domain

PR #13648, opened 2026-09-04
fix: Mermaid diagrams laid out with nodes thousands of pixels apart

### LINKED ISSUE #13638: MermaidJS renderer produces unusable results

## Steps to reproduce:

Insert a basic MermaidJS diagram. Example:

```
graph TD
    A[Start] --> B{Decision}
    B -->|Yes| C[Do this]
    B -->|No| D[Do that]
```

View this in Outline.

### Expected Behavior

Something like this (rendered via https://mermaid.live )

Image

### Steps To Reproduce

The rendered result in Outline is:

Image

### Environment

```markdown
Outline: SaaS
Browser: Tested in latest Chrome and Firefox
```

### Anything else?

A vaguely related issue in #12812

tommoor:
Mermaid sizes each node for layout from the node group's bounding box, which includes the HTML label \`foreignObject\` that starts at 2000×2000 px before being shrunk. On some Windows and Linux machines the browser keeps that size, so nodes are laid out thousands of pixels apart and edges collapse to short stubs.

- Render diagrams with SVG text labels instead of HTML labels, which are measured reliably. Mindmaps keep HTML labels because Mermaid mis-centers their labels otherwise.
- Remove the offscreen render container, which was never passed to Mermaid anyway 🙈 
- Bump the session cache namespace so stale diagrams re-render.

closes #13638

### DIFF
--- a/shared/editor/extensions/Mermaid.ts
+++ b/shared/editor/extensions/Mermaid.ts
@@ -28,11 +28,12 @@ export type MermaidState = {
   editingId?: string;
 };
 
-// The `v3` namespace discards entries cached before the foreignObject fix, so
-// previously mis-sized diagrams are re-rendered instead of served from cache.
+// The `v4` namespace discards entries cached before the switch to SVG text
+// labels, so previously mis-sized diagrams are re-rendered instead of served
+// from cache.
 const cache = new LRUCache<string>({
   max: 20,
-  namespace: "mermaid:v3",
+  namespace: "mermaid:v4",
   storage: "sessionStorage",
 });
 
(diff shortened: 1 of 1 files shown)


pradhankukiran:
@tommoor I noticed while reading through since htmlLabels is set globally per initialize() but the renders fire off without awaiting, wouldnt a page with a flowchart and a mindmap end up rendering both with whichever setting ran last? So the flowchart slips back into the broken mode.
I see Copilot flagged the same thing above, but it cant tell you if it actually bites at runtime. I traced the queue order and it does.

PR #13675, opened 2026-09-06
fix: re-render revision viewer when late-loaded content arrives

### LINKED ISSUE #13674: Revision view does not re-render when late-loaded revision content arrives (diff highlights missing)
**Describe the bug**

When a revision (or a comparison) is opened whose content hasn't been loaded yet, the revision view does not re-render once the content arrives:

1. Opening a revision with `?changes=true` shortly after a page load frequently shows the revision **without any diff highlights** — sometimes with blank/stale content — even though the fetch completed in the background.
2. The same for `?compareTo=…` / `compareTo=latest` views: the comparison revision is fetched correctly now that #13669 routes synthetic `latest` ids, but when its content lands nothing re-renders, so the `Diff` extension is never built — no highlights, no change count.
3. In the history sidebar, the "Current version" entry depends on the latest revision's content being compared against the document; when that content arrives late the check does not re-evaluate.

In all cases the view only corrects itself after an unrelated re-render (toggling the sidebar, switching revisions, etc.).

**To Reproduce**

1. Edit a document a few times so it has revisions containing changes.
2. **Reload the page**, then open History and click a revision — or navigate directly to `/doc/<slug>/history/<id>?changes=true`.
3. The revision opens, but the diff highlights / "N changes" indicator are missing. Toggling the sidebar (any unrelated re-render) makes them suddenly appear.

**Root cause**

`app/scenes/Document/components/RevisionViewer.tsx` fetches late content fire-and-forget:

```tsx
React.useEffect(() => {
  if (showChanges && beforeRevisionId) {
    void revisions.fetch(beforeRevisionId);
  }
}, [showChanges, beforeRevisionId, revisions]);
```

Revisions are listed without their content, so it arrives **after** the component has rendered. Outline models only annotate a field once it is first written, so reads of not-yet-loaded `data` in the render above are not tracked by `observer` — when the content lands there is no re-render. The `editorKey` remount mechanism from #13140 / #13656 (which swaps `comparisonData`/`revision.data` between `pending` and `loaded`) can only take effect on a render, so without one the editor is never rebuilt with the completed diff configuration. The same applies to the revision's own content when linking directly to a revision, and to the `compareTo` revision fetched by the document data loader.

`app/scenes/Document/components/History/History.tsx` has the same shape for the latest revision added in #13669 — `void revisions.fetch(latestRevision.id)` — relying on observer tracking that doesn't apply to never-written fields, so the `isDocUpdated` comparison below it doesn't re-evaluate when the content lands.

**Fix**

Bump a local version once each fetch resolves, so the component re-renders — and the editor remounts with the completed diff configuration — when late content lands:

```tsx
const [, bumpContentLoaded] = React.useReducer((v: number) => v + 1, 0);

React.useEffect(() => {
  if (revision.data) {
    return;
  }
  revisions.fetch(revision.id).then(
    () => bumpContentLoaded(),
    () => {}
  );
}, [revision.id, revision.data, revisions]);
```

…and the equivalent `.then(() => bumpContentLoaded(), …)` bumps for the `before` and `compareTo` revision fetches in `RevisionViewer`, and for the latest-revision fetch in `History`.

Verified on a self-hosted instance across: direct links to revisions, history navigation, `?changes=true`, `compareTo=latest`, and the "Current version" entry — content, highlights and change counts now appear on the first render without needing an unrelated re-render.

**Environment**

Self-hosted, main @ `4ba91fcb8`.

chenniannian90:
## Description

Fixes #13674.

Revisions are listed without their content, so the content of the viewed revision, of the `before` revision and of the `compareTo` revision is fetched separately and usually arrives **after** the revision viewer has rendered. Because models only annotate a field once it is first written, reads of not-yet-loaded content in the render are not tracked by `observer` — when the content lands, nothing re-renders, so the `editorKey` remount (from #13140 / #13656) never re-evaluates and the editor is not rebuilt with the completed diff configuration. The view stays blank or without highlights until an unrelated re-render happens.

- `RevisionViewer`: bump a local version (`useReducer`) once each of the three fetches resolves, so the component re-renders and the editor remounts with the diff when late content lands. The revision's own content is also fetched when missing (direct links to a revision rely on the document DataLoader, which may resolve after this component has mounted).
- `History`: the same bump for the latest-revision fetch added in #13669, so the `isDocUpdated` "Current version" comparison re-evaluates when its content arrives.

## Test plan

Verified manually on a self-hosted instance running essentially this code:

- Reloaded the page, then opened History and a revision with `?changes=true`: content and diff highlights now appear on the first render without toggling anything.
- Direct navigation to `/doc/<slug>/history/<id>?changes=true` for revisions whose content was not yet in the store: the revision renders with highlights once the fetch resolves.
- `?compareTo=latest`: comparison content arriving late now builds the diff and the change count.
- The History sidebar "Current version" entry appears/disappears correctly when the latest revision's content arrives late.
- Regression-checked: switching between revisions, toggling highlights off/on, and returning to the live document all behave as before.

### DIFF
--- a/app/scenes/Document/components/History/History.tsx
+++ b/app/scenes/Document/components/History/History.tsx
@@ -185,18 +185,27 @@ function History() {
     ? revisions.get(latestRevisionEvent.id)
     : undefined;
 
+  // Revision content arrives via a separate fetch, and models only annotate a
+  // field once it is first written, so reads of not-yet-loaded content are not
+  // tracked by `observer` — bump a local version when it lands so the
+  // `isDocUpdated` check below re-evaluates.
+  const [, bumpContentLoaded] = React.useReducer((v: number) => v + 1, 0);
+
   React.useEffect(() => {
     if (latestRevision && !latestRevision.data) {
-      void revisions.fetch(latestRevision.id);
+      revisions.fetch(latestRevision.id).then(
+        () => bumpContentLoaded(),
+        () => {}
+      );
     }
   }, [revisions, latestRevision]);
 
   // Whether the current document has unsaved changes beyond its latest
   // revision, in which case a "Current version" entry is shown. Computed in the
-  // render body (rather than the memo below) so the observer re-evaluates it
-  // once the lazily-loaded revision content arrives. The content-aware check is
-  // deferred until the content has loaded to avoid showing an entry that would
-  // then vanish.
+  // render body (rather than the memo below) so it re-evaluates once the
+  // lazily-loaded revision content arrives (the version bump above). The
+  // content-aware check is deferred until the content has loaded to avoid
+  // showing an entry that would then vanish.
   const isDocUpdated =
     !!latestRevision &&
     !!document &&
--- a/app/scenes/Document/components/RevisionViewer.tsx
+++ b/app/scenes/Document/components/RevisionViewer.tsx
@@ -72,11 +72,47 @@ function RevisionViewer(props: Props, ref: React.Ref<TEditor>) {
   const beforeRevisionId = compareToRevisionId
     ? undefined
     : revision.before?.id;
+
+  // Revision content is fetched separately and arrives after this component
+  // has mounted. Outline models only annotate a field once it is first
+  // written, so reads of not-yet-loaded content are not tracked by `observer`
+  // and its arrival alone will not re-render this component. Bump a local
+  // version once the fetch resolves so the editor, and its diff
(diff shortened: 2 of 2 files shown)


tommoor:
I dislike this fix, pretty last-resort to bump state like this just to trigger a render - will see if there is an alternative way

tommoor:
Fixed at the root cause in https://github.com/outline/outline/pull/13683

PR #13676, opened 2026-09-06
fix: make revision change count indicator and current-of-N navigation reliable

### LINKED ISSUE #13670: History: changes indicator missing on load and never shows "current of N" when navigating
**Describe the bug**

In document history, when a revision is opened with change highlighting, the "N changes" indicator with prev/next navigation in the header is unreliable:

1. On a fresh load the indicator is usually **missing**, even though the revision clearly has changes highlighted in the document.
2. Once it does appear, clicking the prev/next change buttons moves the highlight in the document, but the label **never switches to "current of N"** — it stays at "N changes" forever.

**To Reproduce**

1. Edit a document collaboratively so it has at least two revisions, then make another edit so a revision has multiple changes (e.g. three separate edits in one session).
2. Open the history panel and click a revision that has changes (the `?changes=true` view).
3. Observe the header: the "N changes" pill often does not appear. It only shows up after an unrelated re-render (toggling the sidebar, switching revisions, etc.).
4. Once visible, click the next/previous change buttons: the `.current-diff` highlight moves through the document, but the label stays "N changes" and never shows which change you are on.

**Root cause**

`app/scenes/Document/components/ChangesNavigation.tsx` (since #10824) reads the editor instance through a plain React ref during render:

```tsx
const diffExtension = editorRef.current?.extensions.extensions.find(
  (ext) => ext instanceof Diff
) as Diff | undefined;
const currentChangeIndex = diffExtension?.getCurrentChangeIndex() ?? -1;
const totalChanges = diffExtension?.getTotalChangesCount() ?? 0;
```

Two things break here:

- **Writing `editorRef.current` does not re-render the component.** A ref mutation is invisible to React, so whatever was read last render is reused.
- **The editor mounts asynchronously.** `app/components/Editor.tsx` renders the editor through `lazyWithRetry` inside `React.Suspense`, so at the moment `ChangesNavigation` first renders, `editorRef.current` is still `null`. Nothing notifies the component when the editor (and its `Diff` extension with the computed changeset) finally arrives.

Because the component never reads a MobX observable during render, the `observer` wrapper has nothing to track: even though `Diff.currentChangeIndex` is `@observable`, mutations to it cannot re-render this component. The revision viewer was recently taught to register its editor with `DocumentContext` (#13657), but `DocumentContext.editor` is a plain field, not `@observable`, so reading it in render would not notify either.

**Fix**

Make the arrival of the editor observable, and publish the change count through the same context:

1. Mark `DocumentContext.editor` as `@observable.ref` (it is already written by `setEditor` from the revision viewer). **It must be given an initializer**:

   ```ts
   @observable.ref
   editor: Editor | undefined = undefined;
   ```

   A bare `editor?: Editor` compiles to no property assignment in the constructor, so when `makeObservable(this)` runs the property does not exist yet. For decorator-based annotations mobx then *silently skips* the annotation (`ObservableObjectAdministration.make_` returns early with "will be annotated by subclass constructor" once `storedAnnotations` contains the key — it assumes a subclass constructor will apply it later), and the field stays a plain property: no error, no atom, no re-renders. Every other field in the class works because it has an initializer. (`Model.initialize()` in `app/models/base/Model.ts` works around this same mobx behavior for models by `Object.defineProperty`-ing each declared field before `makeObservable`.)
2. Add an observable `totalChanges` to `DocumentContext`, written by `RevisionViewer` once the changeset settles (the count is known before the editor mounts, so the indicator can render immediately):

```tsx
const { extensions, totalChanges } = React.useMemo(() => {
  const changeset = ChangesetHelper.getChangeset(revision.data, comparisonData);
  ...
  totalChanges: showChanges ? Diff.countChanges(changeset?.changes ?? null) : 0,
}, [revision.data, comparisonData, showChanges]);

React.useEffect(() => {
  setTotalChanges(totalChanges);
  return () => setTotalChanges(0);
}, [totalChanges, setTotalChanges]);
```

3. Have `ChangesNavigation` read `editor` and `totalChanges` from `useDocumentContext()` during render. Now the observer tracks (a) the editor arriving from its lazy chunk, (b) `totalChanges` settling, and (c) `currentChangeIndex` changes (read transitively through the observable `editor` reference), so the label updates to "current of N" as you navigate.

Verified on a self-hosted instance across a full matrix: cold load of five revisions covering insertions, deletions, node insertions, attribute modifications (checkbox toggle) and `compareTo=latest`; the "N changes" count matched the changeset in every case, and stripping the rendered `diff-insertion` / `diff-deletion` marks from the DOM reproduced the before/after revision texts fetched from the API exactly. The label now switches to "current of N" on prev/next (with wrap-around), survives toggling the highlight off and on, updates when switching revisions, and disappears when returning to the live document.

**Environment**

Self-hosted, main @ `c11c91071`.

chenniannian90:
## Description

Fixes #13670.

`ChangesNavigation` reads the editor through a plain React ref during render (`editorRef.current`), but a ref mutation never re-renders a component and the editor mounts asynchronously from a lazy chunk — so on first render the ref is still `null` and nothing notifies the component when the editor (and its `Diff` extension with the computed changeset) arrives. The `observer` wrapper has nothing tracked, so even though `Diff.currentChangeIndex` is observable, mutations to it cannot re-render this component.

Three changes:

- Mark `DocumentContext.editor` as `@observable.ref` **with an initializer**. The initializer matters: without an own property at the time `makeObservable` runs, mobx silently skips the annotation (assuming a subclass constructor will apply it later) and the field stays plain — no error, no atom, no re-renders. Every other field in the class works because it has an initializer.
- Publish the change count through the same context (`totalChanges`, written by `RevisionViewer` once the changeset settles) so the indicator can render before the editor chunk has even loaded.
- Have `ChangesNavigation` read `editor` and `totalChanges` from `useDocumentContext()` during render. Now the observer tracks the editor arriving, the count settling, and `currentChangeIndex` changes (read transitively through the observable editor reference), so the label updates to "current of N" while navigating.

Also extracts `Diff.countChanges()` as a static so the count can be computed from the changeset before the editor exists.

Touches the same file as #13675 (different hunks; should merge independently).

## Test plan

Verified manually on a self-hosted instance across a full matrix:

- Cold load of five revisions covering insertions, deletions, node insertions, attribute modifications (checkbox toggle) and `compareTo=latest`: the "N changes" count matched the changeset in every case, and stripping the rendered `diff-insertion` / `diff-deletion` marks from the DOM reproduced the before/after revision texts fetched from the API exactly.
- The indicator now appears on first load, without needing an unrelated re-render.
- Clicking prev/next switches the label to "current of N" and wraps around in both directions; `Diff.currentChangeIndex` changes re-render the header.
- Survives toggling the highlight off and on, updates when switching revisions, and disappears when returning to the live document.

### DIFF
--- a/app/components/DocumentContext.tsx
+++ b/app/components/DocumentContext.tsx
@@ -28,6 +28,14 @@ class DocumentContext {
   @observable
   headings: Heading[] = [];
 
+  /**
+   * The total number of changes highlighted in the revision viewer. Written
+   * after the revision editor is (re)built, because the header's changes
+   * navigation cannot observe the editor instance through a ref alone.
+   */
+  @observable
+  totalChanges: number = 0;
+
   constructor() {
     makeObservable(this);
   }
@@ -70,6 +78,16 @@ class DocumentContext {
   @action
   setFocusedCommentId = (commentId: string | null) => {
     this.focusedCommentId = commentId;
+  };
+
+  /**
+   * Set the total number of changes highlighted in the revision viewer.
+   *
+   * @param total - The number of changes, or 0 when none are highlighted.
+   */
+  @action
+  setTotalChanges = (total: number) => {
+    this.totalChanges = total;
   };
 
   @action
--- a/app/components/DocumentContext.tsx
+++ b/app/components/DocumentContext.tsx
@@ -13,7 +13,12 @@ class DocumentContext {
   /** The current document */
   document?: Document;
 
-  /** The editor instance for this document */
+  /**
+   * The editor instance for this document. Observable so that components
+   * reading it re-render when the (lazily loaded) editor finally mounts —
+   * a plain ref cannot notify them.
+   */
+  @observable.ref
   editor?: Editor;
 
   /** The ID of the currently focused comment, or null if no comment is focused */
--- a/app/components/DocumentContext.tsx
+++ b/app/components/DocumentContext.tsx
@@ -17,9 +17,13 @@ class DocumentContext {
    * The editor instance for this document. Observable so that components
    * reading it re-render when the (lazily loaded) editor finally mounts —
    * a plain ref cannot notify them.
+   *
+   * The initializer is required: without an own property at the time
+   * `makeObservable` runs, mobx silently skips the annotation (assuming a
+   * subclass constructor will apply it later) and the field stays plain.
    */
   @observable.ref
-  editor?: Editor;
+  editor: Editor | undefined = undefined;
 
   /** The ID of the currently focused comment, or null if no comment is focused */
   @observable
--- a/app/scenes/Document/components/ChangesNavigation.tsx
+++ b/app/scenes/Document/components/ChangesNavigation.tsx
@@ -23,13 +23,16 @@ export const ChangesNavigation = observer(function ChangesNavigation_({
   const { t } = useTranslation();
   const query = useQuery();
   const showChanges = query.get("changes");
-  const { totalChanges } = useDocumentContext();
+  // The editor mounts from a lazy chunk, and a ref update alone cannot
+  // re-render this component — read the instance through the document
+  // context so its arrival and any change of the current index are tracked.
+  const { editor, totalChanges } = useDocumentContext();
 
   if (!showChanges) {
     return null;
   }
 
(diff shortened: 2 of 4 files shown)


tommoor:
Dupe of #13673

PR #13708, opened 2026-09-08
fix: pinned document cards ignore clicks and trackpad taps (delay activation)

### LINKED ISSUE #4268: Click on pinned document doesn't open it
Clicking on a pinned document doesn't open it.

**To Reproduce**
1. Pin a document
2. Go to a different collection and back
3. Click on the pinned document, no effect
4. Ctrl + click or middle click on the pinned document, it opens on a new tab.

**Expected behavior**
Clicking on a pinned document opens it.

**Screenshots**
imagen

- Install: getoutline.com
- Version: ?

 - OS: Windowsa
 - Chrome 105

misha-sunborn:
Fixes #4268.

Clicks and especially trackpad taps on Home/collection pinned cards frequently do nothing. Cause: the `PointerSensor` in `PinnedDocuments` uses `activationConstraint: { delay: 100, tolerance: 5 }`. Any stationary press held 100 ms or longer activates a drag, `DocumentCard` then sets `pointer-events: none` while dragging, and the pointerup/click lands behind the card, so the `Link` never navigates. A firm mouse click is ~60-90 ms and usually works; a trackpad tap is held ~180 ms by libinput/macOS tap-and-drag detection, so it always fails. Measured with synthesized pointer sequences: holds of 70/90 ms navigate, 95/100/110/130/180 ms do not.

Fix: use a distance constraint instead (`{ distance: 8 }`). A stationary press never activates, so clicks and taps of any duration navigate; drag-to-reorder still starts after 8 px of movement. #13498 (1.10.0) added the context menu but left the delay constraint in place, so the bug is still present there.

Tested on 1.10.0 (self-hosted) in Firefox and Chromium with synthesized holds and real trackpad taps.

Side note, separate from this change: `build/app/sw.js` is served with `Cache-Control: max-age=31536000, immutable`. Firefox does not revalidate immutable responses even for service-worker update checks, so a changed `sw.js` never installs there without a URL change. Serving `sw.js` as `no-cache` would avoid that; happy to open a separate PR if wanted.

### DIFF
--- a/app/components/PinnedDocuments.tsx
+++ b/app/components/PinnedDocuments.tsx
@@ -82,8 +82,7 @@ function PinnedDocuments({
   const sensors = useSensors(
     useSensor(PointerSensor, {
       activationConstraint: {
-        delay: 100,
-        tolerance: 5,
+        distance: 8,
       },
     }),
     useSensor(KeyboardSensor, {


tommoor:
Seems probably fine? But I haven't experienced any issues with this control to be honest

misha-sunborn:
It's OS-dependent, consistent for me on Linux with trackpad tap-to-click, on Brave and Firefox. Mouse clicks work, taps don't.

The delay constraint starts a drag on any press held 100 ms without moving. Linux tap-to-click holds the button ~180 ms (libinput's tap-and-drag window), so every tap starts a drag, the card tilts, and the release lands on pointer-events:none. Click is lost. Same issue as [4268](https://github.com/outline/outline/issues/4268).

With this distance activation fix, every click works and reorder still starts after 8 px. Fixes it reliably for me in both browsers.

tommoor (approved):

Issue #13746, opened 2026-09-15
Markdown ZIP import: attachment records created with name/key resolving to literal "null" and size=0 (files never written)


Taavi95:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### This is not related to configuring Outline

- [x] The issue is not related to self-hosting config

### Current Behavior

When importing a Markdown ZIP archive (containing a nested folder structure of `.md` files plus a flat `attachments/` folder) via **Settings → Import → Markdown**, a subset of the created `Attachment` records end up broken:

- `key` resolves to `uploads/{userId}/{attachmentId}/null` (literal string `"null"` instead of the real filename)
- `size = 0`
- `contentType = application/octet-stream` (the mime-type fallback, not the real detected type)

The referencing document's `text`/content is otherwise completely correct — the page structure, hierarchy, and `/api/attachments.redirect?id=...` placeholders are all generated correctly, it's specifically the underlying `Attachment` row and the uploaded file bytes that are missing/corrupt.

The bug is deterministic and reproducible on the *same* input every time, but only affects a *subset* of the imported attachments per run (not all of them) — the exact subset seems to vary between otherwise-identical runs of the same ZIP.

## Expected Behavior

Every attachment referenced in the imported Markdown should have a corresponding `Attachment` row with a correct `key` (matching the real uploaded filename), correct `size`, and correct `contentType`, and the underlying file should actually exist in storage — regardless of how many other attachments are being imported in the same batch.

## Steps To Reproduce

1. Prepare a Markdown ZIP with the following structure (matching what [`[confluence-markdown-exporter](https://github.com/Spenhouet/confluence-markdown-exporter)`](https://github.com/Spenhouet/confluence-markdown-exporter) produces, using `export.attachment_href=relative`):
   ```
   my-space/
     page-a.md
     page-a/
       child-page.md
   attachments/
     123456.jpg
     123457.png
     ...
   ```
2. Reference the attachments from the `.md` files using relative Markdown image syntax, e.g. `![](../../attachments/123456.jpg)`.
3. Import the ZIP via **Settings → Import → Markdown** (this matters — importing via a Collection's own "Import" button produces a *flattened* hierarchy and is a separate, likely related issue; only the Settings-level import correctly preserves hierarchy).
4. Wait for the import to complete ("Import completed" in the worker logs).
5. Inspect the `attachments` table:
   ```sql
   SELECT id, key, size, "contentType"
   FROM attachments
[...]

PR #13750, opened 2026-09-15
fix: Imported attachment keys end in "null" when no name is set

mturac:
Attachments from an API import end up with a key like `uploads/<userId>/<uuid>/null`.

`uploadAttachments` reads the name off `alt` for images and `title` for video/attachment. Both come through as `null` when the markup doesn't set one, and `String(null)` is `"null"` — 4 chars, so the `name.length !== 0` check passes and the `node.type.name` fallback never runs. `getKey` then sticks `"null"` on the end of the key.

Fall back to the node type before the `String()`, same shape as `node.attrs.alt || "image"` in Image.tsx. Left the length check in, it still catches `title="   "`.

Only hits importers on the default `shouldUploadAttachmentsPerPage()` — markdown import opts out of this path.

Test checks the last key segment for all three node types, fails on main with `[ 'null', 'null', 'attachment' ]`. It mocks around Sequelize since `uploadAttachments` is private — say the word if you'd rather it sat somewhere else.

Might be related to #13746. That one blames the markdown zip import, which doesn't reach this code, but the key shape is the same.


tommoor:
Seems reasonable, please delete the slop test though

tommoor (approved):

PR #13769, opened 2026-09-17
feat: Show a toast with actions when a file download completes in the desktop app

tommoor:
The desktop app now tells the user when a file finishes downloading. The toast shows the file name with buttons to open the file or reveal it in the file manager of the operating system.

- Add a `DownloadToast` component that listens for the `onDownloadComplete` bridge event and shows a toast with "Open" and "Show in Finder" / "Show in Explorer" / "Show in folder" actions, labelled by platform
- Guard the new bridge calls so older desktop app versions without them do not throw
- Add `Desktop.protocol`, which resolves to `outline-dev://` in development so the desktop app can be tested against a local server without clashing with the production protocol handler

### DIFF
--- a/app/components/DesktopEventHandler.tsx
+++ b/app/components/DesktopEventHandler.tsx
@@ -3,6 +3,7 @@ import { useTranslation } from "react-i18next";
 import { useHistory } from "react-router-dom";
 import { toast } from "sonner";
 import KeyboardShortcuts from "~/scenes/KeyboardShortcuts";
+import { showDownloadToast } from "~/components/DownloadToast";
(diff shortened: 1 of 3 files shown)


tommoor:
The bridge side of this lives in outline/desktop#83 — `onDownloadComplete`, `openDownload` and `showDownloadInFolder` are implemented there, so this PR is inert until a desktop build ships them. All three calls are guarded with `?.()`, so an older desktop app is unaffected.

On the automated review's point about accumulating callbacks: the bridge calls `ipcRenderer.removeAllListeners("download-completed")` before subscribing (the same pattern as `redirect`, `focus`, `blur` and `updateDownloaded`), so re-registration replaces the listener rather than stacking.

PR #13784, opened 2026-09-17
feat: Add archive and deletion reasons

tommoor:
Allow users to explain why a document was archived or deleted, or why a collection was archived.

- Add an editable plain text reason to the notice banner
- Store `deprecatedReason` on documents and collections through the APIs, and MCP tools.
- Clear reasons when items are restored or unarchived, including restored child documents.


tommoor:
Final feedback resolved in #13793

